PR #4, opened 2025-04-08
Add KeyCloak config demo script

Alan-Cha:
Add script for configuring KeyCloak in the KeyCloak token exchange demo.


pdettori (approved):
/lgtm

PR #26, opened 2025-06-13
Update install documentation

mrsabath:
Some information requires minor updates. 
The required versions don't match the most recent install. We probably need to remove the max values. My podman has a different versioning than docker.


docs/cn-demos.md
```diff
@@ -28,7 +29,7 @@ git clone https://github.com/kagenti/kagenti.git
 cd kagenti
 ```
 
-Setup your env variables as follows:
+Setup your env variables as follow:
```
esnible:
```suggestion
Setup your env variables:
```

pdettori (approved):

PR #27, opened 2025-06-13
Doc fixes

esnible:
Small fixes that I needed to understand the demo and get it working.

The most important fix is the range for `kind`, because the stated range does not encompass the current version.


pdettori:
Thanks for the PR @esnible - could you pls. rebase and resolve conflicts ?

mrsabath:
Fixed! @pdettori @esnible

PR #44, opened 2025-06-27
Add SPIRE install

mrsabath:


pdettori (approved):

pdettori (commented):
Should perhaps rebase from upstream main? Lots of commits that seems to be previously merged.

maia-iyer (approved):
Tested and it works straight out of the box!

pdettori (approved):

PR #49, opened 2025-07-04
Make kagenti-installer work with podman

donaldh:
The 'docker' python library does not work with podman so replace all usage with subprocess commands.

This has been tested with podman on macOS and Fedora 42. Not that it has not been tested with docker.


pdettori (approved):
/lgtm - thanks @donaldh !

PR #67, opened 2025-07-17
Update cn-demos.md

msteinder:
It seems that the system requires agents to be in kagenti namespace. Need to debug this.

## Related issue(s)

Fixes #


pdettori:
Namespaces that are enabled are defined in .env - currently default and team1

PR #68, opened 2025-07-17
Add kagenti-client-registration

Alan-Cha:
Add `kagenti-client-registration`, which enables Kagenti to automatically register clients (agents and tools) with Keycloak


mrsabath (approved):
I tested the import weather agent and the agent was correctly registered using my namespace and agent name. E.g. `mariusz/acp-weather-service`.
LGTM!

PR #80, opened 2025-07-25
:sparkles: add configuration for running ollama on linux

### LINKED ISSUE #79: feature: ollama environment does not work on linux
### Feature Description

The ollama environment is defined in : https://github.com/kagenti/kagenti/blob/e2097116008e88ccfd4614e1bce6fde319964ed9/kagenti/installer/app/resources/environments.yaml#L8 and uses the "magic" host.docker.internal , which does not work on Linux by default. As a result, agents using ollama environment do not work by default. 

### Proposed Solution

There's two ways to get around this:

#### Add a host alias to the pod:
```yaml
spec:
   hostAliases:
      - ip: "172.18.0.1"
        hostnames:
        - "host.docker.internal"
```
iirc while this works well on linux, it may not work the same on mac

#### Add a headless service and use the service for the llm api base:
```bash
cat <<EOF | kubectl apply -f -
apiVersion: discovery.k8s.io/v1
kind: EndpointSlice
metadata:
  name: dockerhost
  labels:
    kubernetes.io/service-name: dockerhost
addressType: IPv4
endpoints:
- addresses:
  - $(docker network inspect kind | jq .[0].IPAM.Config[0].Gateway | sed s/\"//g)
  conditions:
    ready: true
ports:
- name: all-ports
  protocol: TCP
---
apiVersion: v1
kind: Service
metadata:
  name: dockerhost
spec:
  clusterIP: None
EOF
```
In this case, LLM_API_BASE would be http://dockerhost:11434/v1

In either scenario, the installer would need to determine if the user's OS is linux and then do the appropriate changes. 

I think scenario 2 (headless service) is probably the easier option, but am not wedded to that solution.

### Want to contribute?

### Additional Context

_No response_

psschwei:
## Summary

This PR allows Linux users to use kagenti with ollama. 

It does so by using a headless service to reach ollama on the host system, mapping the service to the kind docker network and then updating the ollama hostname in environment configmap.

The linux configuration is done via a script as that seemed like the easiest approach, but happy to try another way if folks think that makes more sense.

Used the weather agent demo to validate that this works on Fedora 41. 

## Related issue(s)

Fixes #79

### DIFF
--- a/kagenti/installer/app/components/agents.py
+++ b/kagenti/installer/app/components/agents.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import os
+import platform
 import typer
 from kubernetes import client, config as kube_config
 
@@ -75,6 +76,15 @@ def install():
                     ns,
                 ],
                 f"Creating 'openai-secret' in '{ns}'",
+            )
+        # if user operating system is linux, do some special config to enable ollama
+        if platform.system() == "Linux":
+            run_command(
+                [
+                    "sh",
+                    "linux/ollama-config.sh"
+                ],
+                "Customizing ollama environment for Linux",
             )
 
         run_command(
--- a/kagenti/installer/app/linux/ollama-config.sh
+++ b/kagenti/installer/app/linux/ollama-config.sh
@@ -1,0 +1,34 @@
+#! /bin/sh
+set -ex
+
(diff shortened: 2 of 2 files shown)


pdettori (commented):
LGTM

pdettori (approved):

PR #81, opened 2025-07-28
Weather agent fix

Alan-Cha:
* Make `KEYCLOAK_URL` optional for client registration (simply does not attempt registration with Keycloak)
* Add client name during Keycloak registration
* Ensure logging is working (print shows up with `kubectl logs... -c kagenti-client-registration` but not logging module)

### DIFF
--- a/kagenti/examples/identity/kagenti_client_registration/client_registration.py
+++ b/kagenti/examples/identity/kagenti_client_registration/client_registration.py
@@ -1,6 +1,4 @@
 import os
-import logging
-
 from keycloak import KeycloakAdmin, KeycloakPostError
 
 KEYCLOAK_URL = os.environ.get('KEYCLOAK_URL')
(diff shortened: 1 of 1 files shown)


kagenti/examples/identity/kagenti_client_registration/client_registration.py
```diff
@@ -1,42 +1,67 @@
...
-            username=KEYCLOAK_ADMIN_USERNAME,
-            password=KEYCLOAK_ADMIN_PASSWORD,
-            realm_name=KEYCLOAK_REALM,
-            user_realm_name='master')
+    # Create client
+    try:
+        llama_stack_client_id = keycloak_admin.create_client(
+            {
+                "name": client_name,
+                "clientId": clientId,
```
Alan-Cha, line 48: To confirm, `name` is just the {name}, whereas `clientId` is `{namespace}/{name}`. Or do we want them to be the same? But if they're the same, why do we want both?

mrsabath (approved):
LGTM!

PR #90, opened 2025-07-30
🐛 Modify deployment imagePullPolicy to use Always

### LINKED ISSUE #89: bug: Fix imagePullPolicy for Agent/Tool Deployments to use Always instead of IfNotPresent
The current Deployment manifests generated in the UI for agents and tools use `imagePullPolicy: IfNotPresent`, which prevents proper image updates when using internal container registry within Kagenti Kubernetes cluster. This should be changed to `imagePullPolicy: Always` to ensure the latest images are pulled.

1. Modify constants.py to define the image pull policy.
2. Modify build_utils.py to use defined image pull policy

cwiklik:
## Summary
Updates the container imagePullPolicy from `IfNotPresent` to `Always` to ensure the latest image versions are pulled during deployments of Agents and Tools.

**Motivation**
The previous IfNotPresent policy could cause issues where:

- Cached images with the same tag but different content were used
- Development/staging environments didn't reflect the most recent image builds

Using Always ensures that Kubernetes will pull the latest version of the image on every deployment, providing more predictable and up-to-date deployments.

## Related issue(s)

Fixes #89

### DIFF
--- a/kagenti/ui/lib/build_utils.py
+++ b/kagenti/ui/lib/build_utils.py
@@ -342,7 +342,7 @@ def _construct_agent_resource_body(
                         "image": image_name,
                         "imageTag": image_tag,
                         "imageRegistry": image_registry_prefix,
-                        "imagePullPolicy": "IfNotPresent",
+                        "imagePullPolicy": constants.DEFAULT_IMAGE_POLICY,
                     },
                     "containerPorts": [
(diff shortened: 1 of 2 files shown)


mrsabath (approved):
Nice! Thank you for making a variable! LGTM

PR #95, opened 2025-08-04
❇️ Add Docker image deployment support to Kegenti UI

### LINKED ISSUE #70: Add Support for Deploying Agents and Tools from Existing Docker Images
**Description**

Enhance the Kegenti UI deployment flow to support deploying Agents and Tools from pre-built Docker images in addition to the current source code building process. This will reduce deployment time for users who already have containerized applications.

**Problem Statement**

Currently, the Kegenti UI only supports building images from source code, which can be time-consuming. Users with existing Docker images hosted in repositories should be able to deploy directly from these images for faster deployment.

cwiklik:
## Summary
This PR enhances Kegenti UI deployment flow to support deploying Agents and Tools from pre-built Docker images in addition to the current source code building process. This improvement reduces deployment time for users who already have containerized applications.

## Related issue(s)

Fixes #70

### DIFF
--- a/kagenti/installer/app/components/agents.py
+++ b/kagenti/installer/app/components/agents.py
@@ -63,6 +63,23 @@ def install():
                 ],
                 f"Creating 'github-token-secret' in '{ns}'",
             )
+        # Create docker-registry secret so that we can pull images from private repos
+        if not secret_exists(v1_api, "ghcr-secret", ns):
+            run_command(
+                [
+                    "kubectl",
+                    "create",
+                    "secret",
+                    "docker-registry",
+                    "ghcr-secret",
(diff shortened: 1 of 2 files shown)


kagenti/installer/app/components/agents.py
```diff
@@ -63,6 +63,23 @@ def install():
                 ],
                 f"Creating 'github-token-secret' in '{ns}'",
             )
+        # Create docker-registry secret so that we can pull images from private repos
```
mrsabath, line 66:
```suggestion
        # Create docker-registry secret in user-specified namespace, so that we can pull images from private repos
```

Issue #118, opened 2025-08-13
bug: "Waiting for Postgres rollout" does not complete


esnible:
### Describe the bug

Kagenti installation does not complete.

### Steps To Reproduce

1. Run `uv run kagenti-installer`

### Expected Behavior

Kagenti is installed.

### Additional Context

The failure occurs even if `uv run kagenti-installer --preload-images` is done.
postgres-0 shows ImagePullBackOff.
oc -n istio-system describe pod prometheus-... shows
```
Failed to pull image "prom/prometheus:v3.1.0": failed to pull and unpack image "docker.io/prom/prometheus:v3.1.0": failed to copy: httpReadSeeker: failed open: unexpected status code https://registry-1.docker.io/v2/prom/prometheus/manifests/sha256:6559acbd5d770b15bb3c954629ce190ac3cbbdb2b7f1c30f0385c4e05104e218: 429 Too Many Requests - Server message: toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit
```

I have a PR.

PR #119, opened 2025-08-13
:bug: Don't insist on repulling Postgres container image

esnible:
## Summary
Sometimes Kubernetes isn't able to verify if a container image is the latest.

## Related issue(s)
#118

Fixes #
#118


kellyaa (approved):
This makes sense, especially given that postgres image isn't expected to frequently change.

PR #121, opened 2025-08-13
Keycloak login

Alan-Cha:


kagenti/ui/Home.py
```diff
@@ -95,4 +97,46 @@
...
+# AUTH_ENDPOINT="http://localhost:8080/realms/master/protocol/openid-connect/auth"
+# TOKEN_ENDPOINT="http://localhost:8080/realms/master/protocol/openid-connect/token"
+# REDIRECT_URI="http://localhost:8502/oauth2/callback"
+# SCOPE="openid profile email"
+CLIENT_ID = os.environ.get('CLIENT_ID')
+CLIENT_SECRET = os.environ.get('CLIENT_SECRET')
+AUTH_ENDPOINT = os.environ.get('AUTH_ENDPOINT')
+TOKEN_ENDPOINT = os.environ.get('TOKEN_ENDPOINT')
+REDIRECT_URI = os.environ.get('REDIRECT_URI')
+SCOPE = os.environ.get('SCOPE')
```
esnible: Add comments or default values so that developers know the format of these values, and that they relate to OIDC/OAuth.  Consider making the default values something that works for Kagenti developers, e.g. http://keycloak.localtest.me:8080/realms/master/protocol/openid-connect/auth

kagenti/ui/Home.py
```diff
@@ -95,4 +97,46 @@
...
+
+oauth2 = OAuth2Component(CLIENT_ID, CLIENT_SECRET, AUTH_ENDPOINT, TOKEN_ENDPOINT)
+
+# Check if token exists in session state
+if TOKEN_STRING in st.session_state:
+    # User is not logged in
+    st.info("User is logged in")
```
esnible: Include the userid and possibly name or email address in this information.

kagenti/examples/identity/kagenti_auth_redirect/README.md
```diff
@@ -0,0 +1,127 @@
...
+           ✓ Sharing gateway access for UI done.                                      utils.py:88
+           ✗ Waiting for kagenti-ui rollout failed.                                   utils.py:93
+           Error: error: deployment "kagenti-ui" exceeded its progress deadline       utils.py:96
+
+Installation aborted.
+```
+
+### Create `kagenti` client in Keycloak
+
+Go to Keycloak at [http://keycloak.localtest.me:8080](http://keycloak.localtest.me:8080).
```
esnible:
```suggestion
Go to Keycloak at [http://keycloak.localtest.me:8080/admin/master/console/#/master/clients](http://keycloak.localtest.me:8080/admin/master/console/#/master/clients).
```

kagenti/examples/identity/kagenti_auth_redirect/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+Login with username `admin` and password `admin`.
+
+Create a new client 
+  * General settings
+    * Set Client ID to `kagenti`
+  * Capatibility confg
```
esnible:
```suggestion
  * Capatibility config
```

(Throughout this PR)

kagenti/examples/identity/kagenti_auth_redirect/README.md
```diff
@@ -0,0 +1,127 @@
...
+    * Enable Client Authentication
+  * Login settings
+    * Set Root URL to `http://kagenti-ui.localtest.me:8080/`
+
+After creating the client, go to Credentials tab and get the client secret.
```
esnible:
```suggestion
After creating the client, go to the Credentials tab and get the client secret.
```

kagenti/examples/identity/kagenti_auth_redirect/README.md
```diff
@@ -0,0 +1,127 @@
...
+cd kagenti/installer
+uv run kagenti-installer
+```
+
+### Change the `kagenti-ui` image
+
+Build the `ui-auth` image.
+
+```sh
+cd ui
```
esnible:
```suggestion
cd ../ui
```

Alan-Cha: I understand your reasoning but this is inconsistent with what I've been recommended for other documentation I've written.

For example: https://github.com/kagenti/kagenti/blob/main/kagenti/examples/identity/kagenti_client_registration/README.md#usage

```
Build the image

cd kagenti/examples/identity/kagenti-client-registration
docker build -t kagenti-client-registration .

Install Kagenti

cd kagenti/installer
uv run kagenti-installer
```

I understand that this is a nice flow and I've tried to do something similar in my previous documentation but because demos in Kagenti often require us to start multiple terminals and jump back and forth between different terminal instances, I've been recommended not to jump around different paths but instead, assume when changing paths that we start from the kagenti root directory.

Any thoughts?

esnible: Fair enough.  Then change it to `cd kagenti/ui`.

kagenti/examples/identity/kagenti_auth_redirect/README.md
```diff
@@ -0,0 +1,151 @@
...
+uv lock
+docker build -t ui-auth .
```
esnible: Do we want special instructions for building?  Or can we make a single binary that can run with and without auth?

maia-iyer: +1 to not having to build - when i ran this it gave me error because the base image was arm64 instead of amd64, and I had to manually change this. Better to not support more docker images

pdettori: +1 for single binary - we could use flags or env vars to enable/disable auth

deployments/ui/kagenti-ui.yaml
```diff
@@ -25,6 +25,42 @@ spec:
...
+            - name: "REDIRECT_URI"
+              valueFrom:
+                secretKeyRef:
+                  name: "auth" 
+                  key: "REDIRECT_URI"
+            - name: "SCOPE"
+              valueFrom:
+                secretKeyRef:
+                  name: "auth" 
+                  key: "SCOPE"
```
Alan-Cha, line 63: I'm not sure if it makes sense to put/get all these values from the `auth` secret.

I wonder if some of these values should come from the `environments` configmap.

I think the `auth` secret may only need to contain `CLIENT_SECRET`.

However, recall that `auth` and `kagenti-ui` are both in the `kagenti-system` namespace whereas the `environments` configmap is in `kagenti` namespace.

kagenti/examples/identity/auth_secret/auth_secret.py
```diff
@@ -0,0 +1,141 @@
...
+        print(f'Created Keycloak client "{client_id}": {internal_client_id}')
+    except KeycloakPostError as e:
+        print(f'Could not create Keycloak client "{client_id}": {e}')
+
+        error_json = json.loads(e.error_message)
+        if error_json['errorMessage'] == f'Client {client_id} already exists':
+            internal_client_id = keycloak_admin.get_client_id(client_id)
+            print(f'Obtained internal client ID of existing client "{client_id}": {internal_client_id}')
+
+    return internal_client_id
```
Alan-Cha, line 61: This code is modified from `kagenti/examples/identity/kagenti_client_registration/client_registration.py`.

I haven't replicated these changes in that files yet however I'm wondering if this code should exist elsewhere in the repo, not within the example directory. Maybe in https://github.com/kagenti/kagenti/blob/5952ee41cce01fbca1a0485dd32c083882977269/kagenti/ui/lib/build_utils.py#L41

kagenti/examples/identity/auth_secret/auth_secret.py
```diff
@@ -0,0 +1,141 @@
...
+client_secret = secrets.get("value", "") if secrets else ""
+data = {
+    "ENABLE_AUTH": "true", # string true
+    "CLIENT_SECRET": client_secret,
+    "CLIENT_ID": CLIENT_ID,
+    "AUTH_ENDPOINT": "http://keycloak.localtest.me:8080/realms/master/protocol/openid-connect/auth",
+    "TOKEN_ENDPOINT": "http://keycloak.keycloak.svc.cluster.local:8080/realms/master/protocol/openid-connect/token",
+    "REDIRECT_URI": "http://kagenti-ui.localtest.me:8080/oauth2/callback",
+    "SCOPE": "openid profile email"
+}
```
Alan-Cha, line 110: This is the data contained in the `auth` secret.

There are some hard coded values here. I'm not sure where and how I can obtain the `endpoints`. It might be possible to fetch or dynamically generate them.

Some other values, I'm not sure if they should be contained in the `environments` configmap, like `ENABLE_AUTH`.

kagenti/examples/identity/kagenti_client_registration/client_registration.yaml
```diff
@@ -1,39 +0,0 @@
...
-            secretKeyRef:
-              name: keycloak-secret
-              key: KEYCLOAK_ADMIN_PASSWORD
-        - name: CLIENT_NAME
-          valueFrom:
-            secretKeyRef:
-              name: keycloak-secret
-              key: CLIENT_NAME
-      restartPolicy: Never
-  backoffLimit: 1
```
Alan-Cha, line 39: This file isn't being used anywhere. I kept it here because I thought it could be a helpful reference.

Again, this goes back to the concern of I'm not sure where to keep this resources I'm creating. I'm just putting it into the `examples/identity` for now.

kagenti/installer/app/components/keycloak.py
```diff
@@ -171,6 +171,23 @@ def install():
...
+            "patch",
+            "statefulset",
+            "keycloak",
+            "-n",
+            "keycloak",
+            "--type=json",
+            "-p",
+            '[{"op": "replace", "path": "/spec/template/spec/containers/0/args", "value":["start-dev"]}]',
+        ],
+        "Updating Keycloak statefulset to enable OAuth",
```
Alan-Cha, line 188: This is where we automatically configure Keycloak to use `start-dev`.

However, eventually, we don't want to use developer mode for production? So I guess we need to document to remove this?

kagenti/installer/app/components/ui.py
```diff
@@ -21,6 +21,27 @@
...
+        [
+            "kubectl",
+            "wait",
+            "--for=condition=complete",
+            "job/auth-secret",
+            "-n", "kagenti",
+            "--timeout=300s",
+        ],
+        "Waiting for auth secret job to complete",
+    )
```
Alan-Cha, line 44: This starts a job and also creates a service account, a role, and a role binding.

The purpose of the job is to create a `kagenti` client within Keycloak, extract its client secret, and put the client secret into a Kubernetes secret named `auth.

The service account, role, and role binding are required because the job is running in the `kagenti` namespace so it can access `environments` configmap but needs to create the `auth` secret in the `kagenti-system` namespace.

kagenti/installer/app/resources/auth-secret.yaml
```diff
@@ -0,0 +1,74 @@
...
+              name: environments
+              key: KEYCLOAK_ADMIN_PASSWORD
+        - name: NAMESPACE
+          value: "kagenti-system"
+        - name: CLIENT_ID
+          value: "kagenti"
+        - name: ROOT_URL
+          value: "http://kagenti-ui.localtest.me:8080"
+        - name: SECRET_NAME
+          value: "auth"
```
Alan-Cha, line 42: There are some hard coded values.

`NAMESPACE`, `CLIENT_ID`, and `SECRET_NAME` should always be "kagenti-system", "kagenti", and "auth".

I'm not sure where to get `ROOT_URL` from.

I'm not sure if these values should also be in the `environments` configmap.

kagenti/ui/lib/agent_details_page.py
```diff
@@ -221,6 +220,9 @@ def render_agent_details_content(agent_k8s_name: str):
 
     agent_chat_name_for_sdk = agent_k8s_name
 
+    # Show access token
+    st.session_state[constants.TOKEN_STRING][constants.ACCESS_TOKEN_STRING]
```
Alan-Cha, line 224: Here, I'm trying to show the access token before we call `_handle_chat_interaction()`

kagenti/examples/identity/auth_secret/README.md
```diff
@@ -0,0 +1,46 @@
...
+```sh
+kind load docker-image ui-auth --name agent-platform
+```
+
+Change `kagenti-ui` deployment so it uses `ui-auth` image.
+
+```sh
+kubectl set image deployment/kagenti-ui \
+  kagenti-ui-container=ui-auth \
+  -n kagenti-system
```
esnible, line 35: We should get the image from the same place the other images come from.  The user should not need to build and apply it on top of the previous no-auth image.

kagenti/installer/app/components/keycloak.py
esnible: We use `start-dev` because that allows HTTP in addition to HTTPS and relaxes host name checking.  Probably the Kagenti installer needs an option for this that is used when configuring Keycloak.  On OpenShift it would create Routes to terminate TLS, on Kind it would either use `start-dev` or install an NGINX to terminate TLS.

kagenti/installer/app/resources/auth-secret.yaml
```diff
@@ -0,0 +1,74 @@
...
+        - name: KEYCLOAK_ADMIN_USERNAME
+          valueFrom:
+            configMapKeyRef:
+              name: environments
+              key: KEYCLOAK_ADMIN_USERNAME
+        - name: KEYCLOAK_ADMIN_PASSWORD
+          valueFrom:
+            configMapKeyRef:
+              name: environments
+              key: KEYCLOAK_ADMIN_PASSWORD
```
esnible, line 34: The Keycloak admin password should be in a Secret, not a ConfigMap.  Probably the admin username as well.

kagenti/installer/app/resources/auth-secret.yaml
esnible: The root-url should always be that for _kind_, but might be a Kagenti installer parameter for OpenShift.

kagenti/ui/Home.py
```diff
@@ -94,5 +96,52 @@
...
+    TOKEN_ENDPOINT = os.environ.get('TOKEN_ENDPOINT')
+    REDIRECT_URI = os.environ.get('REDIRECT_URI')
+    SCOPE = os.environ.get('SCOPE', 'openid profile email')
+
+    oauth2 = OAuth2Component(CLIENT_ID, CLIENT_SECRET, AUTH_ENDPOINT, TOKEN_ENDPOINT)
+
+    # Check if token exists in session state
+    if TOKEN_STRING in st.session_state:
+        # User is not logged in
+        st.info("User is logged in")
```
esnible, line 127: Please show something from the decoded token, such as user name or id, in this message.

kagenti/examples/identity/kagenti_auth_redirect/README.md
Alan-Cha: This is just instructions for building a local image of `kagenti-ui` just for you guys to test with.

I didn't want to push a `kagenti-ui` yet because it'll break Kagenti. This is not meant to be a permanent addition.

Alan-Cha: Yes, there is the `ENABLE_AUTH` env var for disabling auth.

kagenti/installer/app/components/keycloak.py
Alan-Cha: So should we make this optional somehow? Configurable with an env var?

I was just wondering if we should document this as an issue. If we leave it in, it'd be a security vulnerability?

kagenti/installer/app/resources/auth-secret.yaml
Alan-Cha: So I can move this into the `environments` configmap but my question is if we should try to dynamically compose the name or if this URL can also be derived from some manifest that already exists in the cluster

kagenti/examples/identity/auth_secret/README.md
Alan-Cha: This is just a temporary step. The intention is to push this `ghcr.io/kagenti/kagenti/ui:latest`. I cannot do it now because it'll break Kagenti until the rest of this is reviewed.

kagenti/installer/app/resources/auth-secret.yaml
Alan-Cha: I agree that the password should not be in the environments configmap. I'm not sure where to put it because this is statically defined.

I guess the `KEYCLOAK_ADMIN_USERNAME` and `KEYCLOAK_ADMIN_PASSWORD` should be passed in as options or environment variables for the installer? @pdettori

kagenti/installer/app/components/keycloak.py
esnible: The `uv run kagenti-installer` is responsible for installing Keycloak.  It should also be responsible for configuring TLS.  Instead of trying to handle security in this PR, just get it working.  Create a new Issue for someone to create installer options to terminate TLS.  The issue should point out that once TLS is terminated properly we can use `start` instead of `start-dev` and TLS endpoints for Keycloak.

kagenti/installer/app/resources/auth-secret.yaml
esnible: It won't know it.  So, for example, imagine the K8s cluster is IBM's internal https://c115-e.us-south.containers.cloud.ibm.com, and the host name for keycloak is alan-keycloak.draco.res.ibm.com.  The hostname comes from things like Routes and Ingress, and you can't get them from something on the cluster.

kagenti/examples/identity/auth_secret/README.md
esnible: If you divide this PR into a piece for the UI code, and make ENABLE_AUTH's default `false`, then everything would be ready for a second PR that used `ghcr.io/kagenti/kagenti/ui:latest` and didn't break Kagenti.

kagenti/installer/app/components/keycloak.py
pdettori: +1

kagenti/installer/app/resources/auth-secret.yaml
esnible: People are often OK with the initial install having a hard-coded password.  The issue is that in a cluster with security we want it to be possible to set it up so that only the admin can `kubectl get` the password.  The normal way to do that is by letting only the admin view Secrets.

kagenti/examples/identity/auth_secret/README.md
pdettori: +1

@Alan-Cha - just a reminder - we should only use the GH actions CI to build and push to  ghcr.io/kagenti/kagenti/ui:latest  - a manual push with GH PAT would mess up permissions in a way that the CI could no longer push (happened in the past with another project)

kagenti/installer/app/resources/auth-secret.yaml
pdettori: We could use the `installer/app/.env` to setup sensitive keys - that is already done for other things. We could even set default user/passwords (that user could change before running the installer) in `installer/app/.env_template`

PR #138, opened 2025-08-22
Keycloak redirect UI changes

Alan-Cha:
Add the UI changes required for Keycloak redirect
`ENABLE_AUTH` is set to `false` so there shouldn't be able any visual or functional changes as before


Alan-Cha:
# Instructions for testing

### Install Kagenti

```sh
cd kagenti/installer
uv run kagenti-installer
```

### Build and replace `kagenti-ui` image

Build the `kagenti-ui-auth` image.

```sh
cd kagenti/ui
uv lock
docker build -t kagenti-ui-auth .
```

Inject the `kagenti-ui-auth` image into kind cluster.

```sh
kind load docker-image kagenti-ui-auth --name agent-platform
```

Change `kagenti-ui` deployment so it uses `kagenti-ui-auth` image.

```sh
kubectl set image deployment/kagenti-ui \
  kagenti-ui-container=kagenti-ui-auth \
  -n kagenti-system
```

### Check UI

Go to Kagenti: [http://kagenti-ui.localtest.me:8080/](http://kagenti-ui.localtest.me:8080/)

Ensure that all pages look and behave the same as before.

## Manually perform the work of `auth-secret` job

`auth-secret` job, which is not included in this PR, talks to Keycloak, creates a `kagenti` client, gets the client secret, and puts it in the `auth` K8s secret.

So in order to test this functionality, you'll have to do it manually.

### Create a `kagenti` client in Keycloak

Go to Keycloak: [http://keycloak.localtest.me:8080/admin/master/console/#/master/clients](http://keycloak.localtest.me:8080/admin/master/console/#/master/clients)

Log in with username `admin` and password `admin`

Create a new client
  * General settings
    * Set Client ID to `kagenti`
  * Capatibility config
    * Enable Client Authentication
  * Login settings
    * Set Root URL to `http://kagenti-ui.localtest.me:8080/`

Get the client secret

### Create the `auth` K8s secret

```sh
kubectl create secret generic auth \
  --namespace kagenti-system \
  --from-literal=CLIENT_ID="kagenti" \
  --from-literal=CLIENT_SECRET="..." \
  --from-literal=AUTH_ENDPOINT="http://keycloak.localtest.me:8080/realms/master/protocol/openid-connect/auth" \
  --from-literal=TOKEN_ENDPOINT="http://keycloak.keycloak.svc.cluster.local:8080/realms/master/protocol/openid-connect/token" \
  --from-literal=REDIRECT_URI="http://kagenti-ui.localtest.me:8080/oauth2/callback" \
  --from-literal=SCOPE="openid profile email"
```

### Enable authentication

`kagenti-ui` has authentication disabled by default.

```sh
kubectl patch deployment kagenti-ui \
  -n kagenti-system \
  --type='json' \
  -p='[{"op": "replace", "path": "/spec/template/spec/containers/0/env/0/value", "value": "true"}]'
```

Alan-Cha:
Then after, we can create a new image and push to ghcr

kagenti/ui/Home.py
```diff
@@ -94,5 +97,51 @@
         """
     )
 
+# --- Login Redirect ---
+st.session_state[ENABLE_AUTH_STRING] = False
+ENABLE_AUTH = os.environ.get('ENABLE_AUTH')
+if ENABLE_AUTH == "true":
```
esnible: Make this check case-insensitive, and consider logging the enable auth setting at debug or info level.

deployments/ui/kagenti-ui.yaml
```diff
@@ -25,6 +25,45 @@ spec:
         - name: kagenti-ui-container
           image: ghcr.io/kagenti/kagenti/ui:latest
           imagePullPolicy: IfNotPresent
+          env:
+            - name: "ENABLE_AUTH"
+              value: "false"
+            - name: "CLIENT_ID"
+              valueFrom:
+                secretKeyRef:
+                  name: "auth"
```
esnible: "auth" is not a good name for a secret.  Maybe kagenti-ui-oauth-secret would be better.

This Pod depends on the secret being present.  Check in the secret as part of this PR so that the pod can start.  The values in the secret can be TODOs.  Ideally if the user enables auth it will fail with messages like "Unknown OIDC client ID: TODO update CLIENT_ID in kagenti-ui-oauth-secret".

Alan-Cha: > This Pod depends on the secret being present.

I set the env variables that depend on `auth` to optional so the pod can start without the secret.

Do you still think it's necessary to check in the auth secret?

esnible: `optional` is good enough; these parameters truly are optional because they are not needed if ENABLE_AUTH is not `true`.

esnible (approved):

pdettori (approved):
/lgtm

PR #141, opened 2025-08-25
Add Keycloak oauth secret job

Alan-Cha:
Enable authentication for Kagenti by default
Add job that creates the `kagenti-ui-oauth-secret` secret that is used by Kagenti authentication

Related to #121 #138

### DIFF
--- a/kagenti/installer/app/components/ui.py
+++ b/kagenti/installer/app/components/ui.py
@@ -20,7 +20,7 @@ from ..utils import console, run_command
 
 
 def install():
-    """Installs the Kagent UI from its deployment YAML."""
+    """Installs the Kagenti UI from its deployment YAML."""
     run_command(
         [
             "kubectl",


deployments/ui/kagenti-ui.yaml
```diff
@@ -27,7 +27,7 @@ spec:
           imagePullPolicy: IfNotPresent
           env:
             - name: "ENABLE_AUTH"
-              value: "false"
+              value: "true"
```
Alan-Cha, line 30: Enable authentication by default

kagenti/auth/client_registration/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+Go to `Clients` tab on the sidebar.
+
+After a while, a new client `kagenti/acp-ollama-researcher` should appear.
+
+# Publish image
+
+```sh
+docker -D build  --no-cache -t ghcr.io/kagenti/kagenti-ui-oauth-secret:latest -f Dockerfile .
+docker push ghcr.io/kagenti/kagenti-ui-oauth-secret:latest
```
pdettori: We should not do this manually - we should do this as part of the CI because 1) it can be traced back to a specific tag 2) CI builds are Multiplatform - see https://github.com/kagenti/kagenti/blob/4e20be727b9d839b0ffb6879d64c6aff1bd6e2fe/.github/workflows/build.yaml#L54-L63 - we can add more builds like that there

kagenti/auth/client_registration/client_registration.py
```diff
@@ -0,0 +1,67 @@
...
+        print(f'Could not create Keycloak client "{clientId}": {e}')
+
+register_client(
+    KEYCLOAK_URL,
+    KEYCLOAK_REALM,
+    KEYCLOAK_ADMIN_USERNAME,
+    KEYCLOAK_ADMIN_PASSWORD,
+    CLIENT_NAME,
+    NAMESPACE
+)
```
pdettori: A lot of this code seem to be moved out of `https://github.com/kagenti/kagenti/tree/main/kagenti/examples/identity/kagenti_client_registration` - if that is the case we should also remove the code from under `examples/identity/kagenti_client_registration`

kagenti/auth/client_registration/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+Select the `acp/aco_ollama_researcher`.
+
+Select `Build Agent`.
+
+### Verify client registration
+
+```sh
+kubectl port-forward service/keycloak -n keycloak 8081:8080
+```
```
pdettori: This should not be necessary if following the standard Kagenti installation - we are using the ingress gateway so you can just access the keycloak console at http://keycloak.localtest.me:8080/admin/master/console/

kagenti/auth/client_registration/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+Open Keycloak at [http://localhost:8081](http://localhost:8081).
+
```
pdettori: See above comment - can use http://keycloak.localtest.me:8080/admin/master/console/

kagenti/auth/ui-oauth-secret/README.md
```diff
@@ -0,0 +1,270 @@
...
+
+### Install Kagenti
+
+```sh
+cd kagenti/installer
+uv run kagenti-installer
+```
+
+The installer will fail because `kagenti-ui` requires a secret which is not in the cluster yet.
+
```
pdettori: Is this the case because ENABLE_AUTH=true ?

kagenti/auth/ui-oauth-secret/README.md
```diff
@@ -0,0 +1,270 @@
...
+
+
+
+# Publish image
+
+```sh
+docker -D build  --no-cache -t ghcr.io/kagenti/kagenti-ui-oauth-secret:latest -f Dockerfile .
+docker push ghcr.io/kagenti/kagenti-ui-oauth-secret:latest 
+```
```
pdettori: See previous comments about making this part of the CI build

pdettori: We should not manually push with personal tokens to `ghcr.io/kagenti/` but use the CI action to do that

kagenti/installer/app/components/ui.py
```diff
@@ -21,6 +21,27 @@
 
 def install():
     """Installs the Kagent UI from its deployment YAML."""
```
pdettori: should be Kagenti

esnible (changes requested):
This fails install on fresh _kind_ cluster:

```
╭───────────────╮
│ Installing Ui │
╰───────────────╯
           ✗ Creating OAuth secret failed.                                                                   utils.py:93
           Error: Error from server (NotFound): error when creating                                          utils.py:96
           "/Users/snible/src/kagenti/kagenti/installer/app/resources/ui-oauth-secret.yaml": namespaces                 
           "kagenti" not found   
```

kagenti/auth/client_registration/README.md
Alan-Cha: Ok, I will add more builds. But should this be documented? I recall @mrsabath once ask me to document how to we build and publish these images. For that reason, he also started working on #88

kagenti/auth/client_registration/client_registration.py
Alan-Cha: I believe that's been done. The `example/identity/kagenti_client_registration` directory should be gone.

kagenti/auth/client_registration/README.md
Alan-Cha: I can change this. This is old documentation that I moved from `kagenti/examples/kagenti_client_registration` and back then, I was not aware of this endpoint.

kagenti/auth/ui-oauth-secret/README.md
Alan-Cha: This is no longer true as all the environment variables for `kagenti-ui` that come from the secret have been made optional.

However, I believe you were reviewing and old snapshot of this PR. I've removed most of the documentation here.

kagenti/auth/client_registration/README.md
pdettori: for development purpose we should document the commands to build and load on the local cluster. But if we allow anybody with token access to do `push ghcr.io/kagenti/kagenti-ui-oauth-secret:latest ` then it would be hard to know what code has been pushed there - better use the CI that uses tagging to identify the tag that goes into that image

kagenti/auth/ui-oauth-secret/auth_secret.py
```diff
@@ -0,0 +1,141 @@
...
+    "ENABLE_AUTH": "true", # string true
+    "CLIENT_SECRET": client_secret,
+    "CLIENT_ID": CLIENT_ID,
+    "AUTH_ENDPOINT": "http://keycloak.localtest.me:8080/realms/master/protocol/openid-connect/auth",
+    "TOKEN_ENDPOINT": "http://keycloak.keycloak.svc.cluster.local:8080/realms/master/protocol/openid-connect/token",
+    "REDIRECT_URI": "http://kagenti-ui.localtest.me:8080/oauth2/callback",
+    "SCOPE": "openid profile email"
+}
+
+config.load_incluster_config()
```
pdettori: You can use `lib.is_running_in_cluster()` to determine the right call here

kagenti/installer/app/components/keycloak.py
```diff
@@ -171,7 +172,23 @@ def install():
...
+        [
+            "kubectl",
+            "patch",
+            "statefulset",
+            "keycloak",
+            "-n",
+            "keycloak",
+            "--type=json",
+            "-p",
+            '[{"op": "replace", "path": "/spec/template/spec/containers/0/args", "value":["start-dev"]}]',
```
maia-iyer: after some testing, I learned we don't even need the `start-dev` - start is fine, so I think we can remove this

.github/workflows/build.yaml
```diff
@@ -39,25 +39,68 @@ jobs:
...
+
+      - name: Build and push ui-oauth-secret Docker image
+        uses: docker/build-push-action@v6
+        with:
+          context: ./kagenti/auth/ui-oauth-secret
+          file: ./kagenti/auth/ui-oauth-secret/Dockerfile
+          push: true
+          platforms: linux/amd64,linux/arm64
+          tags: ${{ steps.ui-oauth-secret-meta.outputs.tags }}
+          labels: ${{ steps.ui-oauth-secret-meta.outputs.labels }}
```
Alan-Cha, line 84: I added a section for the `ui-oauth-secret` image

Previously, the image was `ghcr.io/kagenti/kagenti-ui-oauth-secret` but now it's `ghcr.io/kagenti/kagenti/ui-oauth-secret` to match the `ui` image.

@pdettori Do we need to do something to trigger the flow so that we can test the PR?

.github/workflows/build.yaml
```diff
@@ -39,25 +39,68 @@ jobs:
...
+
+      # - name: Build and push client-registration Docker image
+      #   uses: docker/build-push-action@v6
+      #   with:
+      #     context: ./kagenti/auth/client-registration
+      #     file: ./kagenti/auth/client-registration/Dockerfile
+      #     push: true
+      #     platforms: linux/amd64,linux/arm64
+      #     tags: ${{ steps.client-registration-meta.outputs.tags }}
+      #     labels: ${{ steps.client-registration-meta.outputs.labels }}
```
Alan-Cha, line 106: I added a section for the `client-registration` image as well.

Previously, the image was `ghcr.io/kagenti/kagenti-client-registration` but now it's `ghcr.io/kagenti/kagenti/client-registration`. However, I'm commenting this out for now because I also need to make changes to https://github.com/kagenti/kagenti-operator/ 

I want to make smaller changes so I will do this in a follow up PR

kagenti/auth/ui-oauth-secret/auth_secret.py
```diff
@@ -0,0 +1,146 @@
...
+# TODO: use the function from kagenti/ui/lib/kube.py
+def is_running_in_cluster() -> bool:
+    return bool(os.getenv("KUBERNETES_SERVICE_HOST"))
+
+# TODO: refactor this function so kagenti-client-registration image can use it
+def register_client(
+    keycloak_admin: KeycloakAdmin,
+    client_id: str,
+    client_payload
+):
```
Alan-Cha, line 49: `is_running_in_cluster()` is copied from `kagenti/ui/lib/kube.py`

This is not ideal. The issue is multifaceted. `kagenti/auth/ui-oauth-secret/` shares a sibling relationship with `kagenti/ui/lib/. The former uses a `requirements.txt` and the latter uses `pyproject.toml`. I think there needs to be some reorganization of files. I'm looking for suggestions but I'd like to that in a future PR.

Similarly, it would be nice if we can pull out `register_client()` and keep it in `kagenti/ui/lib/build_utils.py` so I can reuse it for the `client-registration` image as well.

kagenti/installer/app/components/ui.py
```diff
@@ -21,6 +21,39 @@
...
+        [
+            "kubectl",
+            "apply",
+            "-n",
+            "kagenti-system",
+            "-f",
+            str(config.RESOURCES_DIR / "global-environments.yaml"),
+        ],
+        f"Applying global-environments configmap in 'kagenti-system'",
+    )
```
Alan-Cha, line 34: Adding the new `global-environments` configmap. Later on, we can also split this up into a configmap and a secret.

kagenti/installer/app/resources/ui-oauth-secret.yaml
```diff
@@ -0,0 +1,74 @@
...
+metadata:
+  name: kagenti-ui-oauth-job
+  namespace: kagenti-system
+spec:
+  template:
+    spec:
+      serviceAccountName: secret-writer
+      containers:
+      - name: client
+        image: ghcr.io/kagenti/kagenti/ui-oauth-secret:latest
```
Alan-Cha, line 12: Pointing to the image that the workflow will create.

kagenti/installer/app/components/keycloak.py
Alan-Cha: I removed it.

pdettori (approved):
/lgtm

Issue #115, opened 2025-08-07
bug/question: Keycloak connection recovery (long-lived KIND cluster?)


evaline-ju:
### Describe the bug

I’ve left a kagenti installation for a couple days each time but it will puzzlingly stop working. The containers all stay up, but it is unclear to me how to ‘reset’ the connections where necessary.

In the `keycloak` pod in the `keycloak` namespace I can see logs such as 
```
2025-08-07 17:13:47,207 WARN  [io.agroal.pool] (agroal-11) Datasource '<default>': The connection attempt failed.
2025-08-07 17:13:47,208 WARN  [org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator] (JPA Startup Thread) HHH000342: Could not obtain connection to query metadata: org.hibernate.exception.JDBCConnectionException: unable to obtain isolated JDBC connection [The connection attempt failed.] [n/a]
	at org.hibernate.exception.internal.SQLStateConversionDelegate.convert(SQLStateConversionDelegate.java:100)
	at org.hibernate.exception.internal.StandardSQLExceptionConverter.convert(StandardSQLExceptionConverter.java:58)
	at org.hibernate.engine.jdbc.spi.SqlExceptionHelper.convert(SqlExceptionHelper.java:108)
	at org.hibernate.engine.jdbc.spi.SqlExceptionHelper.convert(SqlExceptionHelper.java:94)
	at org.hibernate.resource.transaction.backend.jta.internal.JtaIsolationDelegate.doTheWork(JtaIsolationDelegate.java:206)
	at org.hibernate.resource.transaction.backend.jta.internal.JtaIsolationDelegate.lambda$delegateWork$3(JtaIsolationDelegate.java:91)
	at org.hibernate.resource.transaction.backend.jta.internal.JtaIsolationDelegate.doInSuspendedTransaction(JtaIsolationDelegate.java:125)
	at org.hibernate.resource.transaction.backend.jta.internal.JtaIsolationDelegate.delegateWork(JtaIsolationDelegate.java:88)
	at org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator.getJdbcEnvironmentUsingJdbcMetadata(JdbcEnvironmentInitiator.java:336)
	at org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator.initiateService(JdbcEnvironmentInitiator.java:129)
[...]

PR #151, opened 2025-08-27
🐛 Ensure keycloak in dev mode has SSL disabled for Postgres

pdettori:
## Summary

Seems that after some time Keycloak forgets that it is not using a SSL connection to the Postgres DB and tries to open new connection using SSL mode and the pod crash loop. This should address #115


pdettori:
closing because of wrong rebase

PR #177, opened 2025-09-04
❇️ Replace hardcoded operator version with dynamic GitHub tag detection

### LINKED ISSUE #171: Kagenti installer uses hardcoded operator version
### Problem Description
The Kagenti installer currently uses a hardcoded version of the operator in kagenti/installer/app/config.py. Each time the operator version changes, the installer must be manually updated to reference the new version.

Below is a snippet from config.py showing current configuration:
```
# --- Cluster & Operator Configuration ---
CLUSTER_NAME = "agent-platform"
OPERATOR_NAMESPACE = "kagenti-system"
TEKTON_VERSION = "v0.66.0"
**LATEST_TAG = "0.2.0-alpha.4"**
```

### Expected Behavior

- Installer should automatically detect and use the latest operator version
- No manual updates required when new operator versions are released

### Proposed Solution
Replace hardcoded version with dynamic version detection using python code below:

```
def get_latest_operator_version():
  try:
      result = subprocess.run(
           ["git", "describe", "--tags", "--abbrev=0"],
           capture_output=True, text=True, check=True
      )
      return result.stdout.strip().lstrip('v')
....

```

cwiklik:
## Summary
Fixes Kagenti installer which used a hardcoded operator version. The installer now automatically detects and uses the latest operator release from the kagenti-operator repository.

## Problem

- Installer contained hardcoded operator version (ex. 0.2.0-alpha.4)

- Required manual updates every time a new operator version was released

- Risk of deploying outdated operator versions in new installations

- Created operational overhead for maintenance

## Solution
Implemented dynamic version detection that:

- Queries the kagenti-operator repository for the latest semantic version tag

- Uses `git ls-remote` to avoid pull limit error

## Changes Made
Added `get_latest_operator_version()` function

- Uses `git ls-remote --tags --sort=-version:refname` to query remote tags

- Parses semantic versions using the packaging library

- Returns the highest semantic version found

Fixes #171

### DIFF
--- a/kagenti/installer/app/config.py
+++ b/kagenti/installer/app/config.py
@@ -35,7 +35,7 @@ REQ_VERSIONS = {
     "docker": {"min": "5.0.0", "max": "29.0.0"},
     "kubectl": {"min": "1.29.0", "max": "1.35.0"},
     "helm": {"min": "3.14.0", "max": "3.19.0"},
-    "git": {"min": "2.99.0", "max": "3.0.0"},
+    "git": {"min": "2.30.0", "max": "3.0.0"},
 }
 
 # --- Images to preload in the kind cluster ---


kagenti/installer/app/components/operator.py
```diff
@@ -13,12 +13,54 @@
...
+    Args:
+        fallback (str): The fallback version to return if fetching fails.
+
+    Returns:
+        str: The latest version tag or the fallback version.
+    """
+    try:
+        result = subprocess.run(
+            [
+                "git", "ls-remote", "--tags", "--sort=-version:refname",
```
pdettori, line 34: Looks like we added git to the list of pre-reqs - I think we need to add that to https://github.com/kagenti/kagenti/blob/f4f34da78095318fa43ac03f4e8e8830d3262802/kagenti/installer/app/config.py#L34

cwiklik: Looks like the minimum version of Git that supports `git ls-remote --sort` is **Git 2.18**. What should be used for the max? The latest git (2.51.0)?

pdettori: Yes or even better we can also use a very large number since we are more interested in the minimum version (e.g. 2.99.0 )

pdettori (approved):
/lgtm

PR #178, opened 2025-09-04
:seedling::recycle: Create or patch secrets and leverage Kube API

### LINKED ISSUE #167: feature: Update secrets in agent namespaces when re-running installer
### Feature Description

There are a number of secrets we create in each namespace enabled for agents and tools:

```
$ kubectl get secrets
NAME                     TYPE                             DATA   AGE
ghcr-secret              kubernetes.io/dockerconfigjson   1      3d16h
github-token-secret      Opaque                           2      3d16h
keycloak-client-secret   Opaque                           1      46m
openai-secret            Opaque                           1      3d16h
slack-secret             Opaque                           1      3d16h
```

However, if we update any of the env variables in the .env file and re-run the installer, such secrets are not updated with the new tokens/credentials

### Proposed Solution

Update the secret creation logic to patch the secret when we run the installer instead then skipping when secret exists as the installer currently does.

### Want to contribute?

### Additional Context

_No response_

evaline-ju:
## Summary
Leverage kube API instead of shell to create and update secrets - this is limited to those affected by env vars from .env currently. The rationale was to use the typed Python objects as opposed to the string lists, which could be more typo-prone, avoid shell injection risks and avoid spawning new processes. The secret update part is additional - prior to this change, secrets would only be created if non-existent, otherwise not changed on subsequent installation runs.

## Related issue(s)

Fixes #167

### DIFF
--- a/kagenti/installer/app/components/agents.py
+++ b/kagenti/installer/app/components/agents.py
@@ -77,7 +77,7 @@ def install():
         ghcr_secret = client.V1Secret(
             api_version="v1",
             kind="Secret",
-            metadata=client.V1ObjectMeta(name="ghcr-secret"),
+            metadata=client.V1ObjectMeta(name="ghcr-secret", namespace=ns),
             data={
                 ".dockerconfigjson": base64.b64encode(
                     json.dumps(docker_config).encode("utf-8")


kagenti/installer/app/components/agents.py
```diff
@@ -50,68 +51,65 @@ def install():
...
+                    "username": github_user,
+                    "password": github_token,
+                    "auth": base64.b64encode(github_auth.encode("utf-8")).decode("utf-8")
+                }
+            }
+        }
+        ghcr_secret = client.V1Secret(
+            api_version="v1",
+            kind="Secret",
+            metadata=client.V1ObjectMeta(name="ghcr-secret"),
```
pdettori: Missing the namespace (this would create only in `default`) - should have `namespace=ns`

evaline-ju: ah yes in testing this, this was avoided with the `create_namespaced_secret` and `patch_namespaced_secret` which contains the namespace, but agree it would be better to be explicit on the secret itself

PR #184, opened 2025-09-05
🌱 bump max version of required kubectl

pdettori:
## Summary

bump max version of required kubectl - installation on a new system fails because of too new kubectl


huang195 (approved):
/lgtm

PR #186, opened 2025-09-05
✨ Update installer to allow for docker or podman as container engine options

### LINKED ISSUE #179: feature: Support using podman without symbolic link to docker
### Feature Description

It would be helpful to allow users to choose whether or not they want to use docker or podman as their container engine of choice without the need for using a symbolic link. 

### Proposed Solution

Update `kagenti/installer/app/config.py` to include a `CONTAINER_ENGINE` variable that can be used to determine if podman or docker is used throughout the installation. 

### Want to contribute?

### Additional Context

_No response_

MichaelClifford:
## Summary
This PR allows users to choose whether they want to use docker or podman as their container engine during kagenti installation without the need for using any symbolic links as is suggested in the [docs](https://github.com/kagenti/kagenti/blob/a293d21664dfd8fa701df4d5ebfd7ba6530d07f3/docs/demos.md?plain=1#L14).

This PR adds a `CONTAINER_ENGINE` variable to `kagenti/installer/app/config.py`. By default this variable is set to `"docker"`, leaving the default behavior unchanged. It also adds an Enum class `ContainerEngine()` to ensure that only the strings `"podman"` or `"docker"` can be  passed to the relevant `subprocess.run()`.  

During the dependency check in `check_dependencies()` only the selected container engine's version will be evaluated.

This has been tested on a Mac M2 with podman. 

## Related issue(s)

Fixes #179

### DIFF
--- a/kagenti/installer/app/checker.py
+++ b/kagenti/installer/app/checker.py
@@ -23,6 +23,7 @@ from rich.panel import Panel
 from rich.text import Text
 
 from . import config
+from .config import ContainerEngine
 from .utils import console, get_command_version
 
 
@@ -32,7 +33,16 @@ def check_dependencies():
         Panel(Text("1. Checking Dependencies", justify="center", style="bold yellow"))
     )
     all_ok = True
+    try:
+        container_engine = ContainerEngine(config.CONTAINER_ENGINE)
+    except ValueError:
+        console.log(f"[bold red]✗ Container engine must be either 'docker' or 'podman'[/bold red]")
+        raise typer.Exit(1)
     for tool, versions in config.REQ_VERSIONS.items():
+        if tool == "docker" and tool != container_engine.value:
+            continue
+        if tool == "podman" and tool != container_engine.value:
+            continue
         with console.status(f"[cyan]Checking for {tool}..."):
             time.sleep(0.5)
             version = get_command_version(tool)
--- a/kagenti/installer/app/cluster.py
+++ b/kagenti/installer/app/cluster.py
@@ -24,6 +24,7 @@ from rich.panel import Panel
 from rich.text import Text
 
 from . import config
+from .config import ContainerEngine
 from .utils import console, run_command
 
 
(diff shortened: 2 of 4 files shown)


kagenti/installer/app/components/registry.py
```diff
@@ -43,13 +44,19 @@ def install():
             registry_ip = service.spec.cluster_ip
 
             container = f"{config.CLUSTER_NAME}-control-plane"
+            container_engine = ContainerEngine(config.CONTAINER_ENGINE)
+
             subprocess.run(
-                ["docker", "exec", container, "sh", "-c",
+                [container_engine, "exec", container, "sh", "-c",
```
pdettori: ContainerEngine is an enum object (`container_engine = ContainerEngine(config.CONTAINER_ENGINE)`), but here it is used as a string - it would be better to use `str(container_engine)` or `container_engine.value`

kagenti/installer/app/cluster.py
```diff
@@ -46,15 +47,19 @@ def kind_cluster_exists():
 def kind_cluster_running():
     """Checks if the Kind cluster is already running."""
     try:
+        container_engine = ContainerEngine(config.CONTAINER_ENGINE)
         result = subprocess.run(
-            ["docker", "ps", "--format", "{{.Names}}"],
+            [container_engine, "ps", "--format", "{{.Names}}"],
```
pdettori: See previous my comment about rather using `str(container_engine)` or `container_engine.value`

kagenti/installer/app/cluster.py
```diff
@@ -145,14 +150,20 @@ def create_kind_cluster(install_registry: bool, silent: bool):
...
     )
+    try:
+        container_engine = config.ContainerEngine(config.CONTAINER_ENGINE)
+    except ValueError as e:
+        console.log(f"[bold red]✗ Container engine must be set to either 'docker' or 'podman' in config.py[/bold red]")
+        raise typer.Exit(1)
+
     for image in images:
-        run_command(["docker", "pull", image], f"Pulling image {image}")
+        run_command([container_engine, "pull", image], f"Pulling image {image}")
```
pdettori: See previous my comment about rather using `str(container_engine)` or `container_engine.value`

kagenti/installer/app/checker.py
```diff
@@ -33,6 +33,10 @@ def check_dependencies():
     )
     all_ok = True
     for tool, versions in config.REQ_VERSIONS.items():
+        if tool == "docker" and tool != config.CONTAINER_ENGINE:
+            continue
+        if tool == "podman" and tool !=config.CONTAINER_ENGINE:
+            continue
```
pdettori, line 45: If CONTAINER_ENGINE is set to neither "docker" nor "podman", both checks will skip, and neither will be verified. Could be helpful to raise an error if an unsupported value is set. I think at this time we do need one or the other for the installation to work.

kagenti/installer/app/components/registry.py
MichaelClifford: Thanks for the review @pdettori! 😃 

The Enum class is defined as `class ContainerEngine(str, Enum)`. When defined this way it works as expected in this scenario. But, if you'd like it to be more explicit I can go ahead and update it to use `container_engine.value` instead.

kagenti/installer/app/checker.py
MichaelClifford: Good point. I will raise an error if neither option is selected.

kagenti/installer/app/components/registry.py
pdettori: Yes, I'd prefer `container_engine.value` thanks!

pdettori (approved):
/lgtm - thank you @MichaelClifford !

PR #193, opened 2025-09-10
Replacing wasm-based Gateway with ext-proc based Gateway

huang195:
Pulling in MCP Gateway from kagenti/mcp-gateway to replace WASM-based MCP Gateway. Again, the gateway components will be dormant until tool and agent developers follow our instructions (another PR coming up) to make use of the Gateway.


kagenti/installer/app/config.py
```diff
@@ -60,8 +60,7 @@ class ContainerEngine(str, Enum):
     "ghcr.io/spiffe/oidc-discovery-provider:1.12.4",
     "docker.io/nginxinc/nginx-unprivileged:1.29.0-alpine",
     "docker.io/nginx/nginx-prometheus-exporter:1.4.2",
-    "ghcr.io/kagenti/mcp:latest",
-    "ghcr.io/kagenti/mcp-gateway-wasm:latest",
+    "ghcr.io/kagenti/mcp-gateway:latest",
```
pdettori: Should this be consistent with `ghcr.io/kagenti/mcp-gateway:v0.1` used later on in `gateway-deployment.yaml` ?

kagenti/installer/app/resources/gateway-deployment.yaml
```diff
@@ -0,0 +1,237 @@
...
+        component: broker-router
+    spec:
+      serviceAccountName: mcp-broker-router
+      volumes:
+        - name: config-volume
+          configMap:
+            name: mcp-gateway-config
+      containers:
+        - name: mcp-broker-router
+          image: ghcr.io/kagenti/mcp-gateway:v0.1
```
pdettori, line 126: See comment in `config.py` for this image -  if the tag keeps on changing should we use `latest` or parametrize this?

kagenti/installer/app/components/mcp_gateway.py
```diff
@@ -19,60 +19,57 @@
...
+        ["kubectl", "apply", "-f", str(config.RESOURCES_DIR / "gateway-deployment.yaml")],
+        "Deploying MCPGateway Broker, Router, and Controller",
     )
+
+    #"""Init MCPServer CR."""
+    # This command creates an empty MCPServer resource
+    #run_command(
+    #    ["kubectl", "apply", "-f", str(config.RESOURCES_DIR / "gateway-mcpserver.yaml")],
+    #    "Initing MCPServer CR",
+    #)
```
pdettori, line 75: Is this commented because it is "dormant" for now?

kagenti/installer/app/resources/gateway.yaml
```diff
@@ -1,31 +1,26 @@
...
-      namespaces:
-        from: All
+    - name: mcp # going to the Broker, and no hostname field
+      port: 8080
+      protocol: HTTP
+      allowedRoutes:
+        namespaces:
+          from: All
+    - name: mcps 
+      hostname: '*.mcp.test.com' #public accessibly MCP host (doesn't need dns as it is for rewritten hosts)
```
pdettori, line 21: is this required ? is it for test?

kagenti/installer/app/resources/gateway-deployment.yaml
```diff
@@ -0,0 +1,237 @@
...
+    metadata:
+      labels:
+        app: mcp-broker-router
+        component: broker-router
+    spec:
+      serviceAccountName: mcp-broker-router
+      volumes:
+        - name: config-volume
+          configMap:
+            name: mcp-gateway-config
```
pdettori, line 123: Where this ConfigMap is defined ? There is no `mcp-gateway-config` ConfigMap in this PR.

kagenti/installer/app/config.py
huang195: good catch! will fix

kagenti/installer/app/resources/gateway-deployment.yaml
huang195: we should not use `latest` as that keeps on changing, and the gateways CRDs are getting churned, so we don't want all of that to impact us. The `v0.1` is a stable and tested version that works with Kagenti, so let's use this image tag.

kagenti/installer/app/components/mcp_gateway.py
huang195: Correct. To "activate" it, agent and tool developers will need to create certain resources. The doc PR is coming up soon.

kagenti/installer/app/resources/gateway.yaml
huang195: Yes, this is required. This is required as it's used by the Gateway Router components internally to redirect to the right MCP server backend for `tools/call`.

kagenti/installer/app/resources/gateway-deployment.yaml
huang195: This configmap will be created by the Gateway controller dynamically at startup time, and consumed by the Gateway broker/router pod.

PR #196, opened 2025-09-10
update gateway doc

huang195:
Documentation that describe how tools are registered with the Gateway and how agents can access these tools via the Gateway.


mrsabath (approved):
LGTM

PR #198, opened 2025-09-11
❇️ Loads global env vars from configMap

### LINKED ISSUE #189: feature: Kubernetes ConfigMap Integration for Environment Variable Editor
### Summary
Add the ability to load global environment variables directly from a Kubernetes ConfigMap into the existing Streamlit environment variable editor with one-click import functionality.

### Problem Statement
When using the UI to create new Agents/Tools, users have no visibility into what global environment variables are already defined in the kagenti system. This creates several issues:

- Discovery Problem: Users don't know what environment variables are available (API keys, service URLs, configuration values)
- Duplication Risk: Users may recreate variables that already exist globally
- Poor Developer Experience: Users must run kubectl get cm environments -n kagenti -o yaml manually to view ConfigMap contents, which is inconvenient and breaks the UI workflow
- Configuration Inconsistency: Without seeing global variables, users can't ensure their Agent/Tool configurations align with the system defaults

### Proposed Solution

- One-click ConfigMap loading: Single button press to import all environment variables from a named configMap in a named namespace

- Automatic parsing: Handle multiple ConfigMap data formats:

1. Direct key-value pairs 
 ```
  KEYCLOAK_ADMIN_PASSWORD: password
  KEYCLOAK_ADMIN_USERNAME: user
  KEYCLOAK_REALM: master
  KEYCLOAK_URL: http://keycloak.keycloak:8080
```
1. JSON arrays with environment configurations
```
  ollama: |
    [
      {"name": "LLM_API_BASE", "value": "http://host.docker.internal:11434/v1"},
      {"name": "LLM_API_KEY", "value": "dummy"},
      {"name": "LLM_MODEL", "value": "llama3.2:3b-instruct-fp16"}
    ]
```
1. Kubernetes secret references (valueFrom.secretKeyRef)
```
  mcp-slack-config: |
    [
      {
        "name": "SLACK_BOT_TOKEN",
        "valueFrom": {"secretKeyRef": {"name": "slack-secret", "key": "bot-token"}}
      }
    ]

```
- Visual distinction: Clear separation between ConfigMap and custom variables

### Want to contribute?

### Additional Context

_No response_

### PR DESCRIPTION
**Pull request by cwiklik**
## Summary
This PR adds the ability to load all environment variables from a k8s ConfigMap directly into the Kagenti UI environment editor via one-click import button. 

### Changes Made
**New UI Components**
image

- "Load Global Environment Vars" button: one-click loading of all ConfigMap variables
- "Reload Global Environment Vars" button: restores original ConfigMap values (eg. accidental var delete )
- Variable Type Bages: Displays `🗂️ ConfigMap`, `🔒 ConfigMap Secret`, `✏️ Custom` to identify variable types

image

**Secret Support**

- Secret detection: identifies vars with `valueFrom.secretKeyRef` 
- Secret formatting: displays secrets as `<secret-name:key>`
image

**Enhanced Variable Display**

- In-memory editing: All ConfigMap vars are editable 
- Deletable: ConfigMap vars can be removed from memory (doesn't effect actual ConfigMap in k8s)
- Visual separation: Clear distinction between global and custom variables

## Related issue(s)

Fixes #189

### DIFF
--- a/kagenti/ui/lib/build_utils.py
+++ b/kagenti/ui/lib/build_utils.py
@@ -1098,7 +1098,7 @@ def render_import_form(
                 st_object.warning(f"No configMap data available to load in namespace '{build_namespace_to_use}'")
                 return
             # First, remove any previously loaded configmap vars
-            user_custom_vars = [var for var in st.session_state[custom_env_key] if not var.get('configmap_origin', False)]
+            user_custom_vars = [var for var in st.session_state.get(custom_env_key,[]) if not var.get('configmap_origin', False)]
             # Then, add all configmap vars
             config_map_data = get_config_map_data(
                 core_v1_api, build_namespace_to_use, constants.ENV_CONFIG_MAP_NAME)
@@ -1644,15 +1644,15 @@ def validate_registry_config(registry_config, st_object):
 
     return True
 
-def parse_configmap_data_to_env_vars(env_options: Dict[str, Any]) -> List[Dict[str, str]]:
+def parse_configmap_data_to_env_vars(env_options: Dict[str, Any]) -> List[Dict[str, Any]]:
     """
     Parses ConfigMap data and return all environment variables as a flat list.
 
     Args:
-        configmap_data (Dict[str, Any]): The data from the ConfigMap.
+        env_options(Dict[str, Any]): The data from the ConfigMap.
 
     Returns:
-        List[Dict[str, str]]: A list of environment variable dictionaries.
+        List[Dict[str, Any]]: A list of environment variable dictionaries.
     """
 
     env_vars = []
--- a/deployments/ui/kagenti-ui.yaml
+++ b/deployments/ui/kagenti-ui.yaml
@@ -23,8 +23,9 @@ spec:
       serviceAccountName: kagenti-ui-service-account
       containers:
         - name: kagenti-ui-container
-          image: ghcr.io/kagenti/kagenti/ui:latest
-          imagePullPolicy: IfNotPresent
+#          image: ghcr.io/kagenti/kagenti/ui:latest
+          image: kagenti-demo-ui:latest
+          imagePullPolicy: Never
           env:
             - name: UV_CACHE_DIR
               value: "/app/.cache/uv"

### CODE REVIEW

File: kagenti/ui/lib/build_utils.py [hunk 1]
```diff
@@ -1088,6 +1093,65 @@ def remove_env_var(index):
...
+                    "value": var["value"],
+                    'configmap_origin': True,
+                    'configmap_section': var.get('section', ''),
+                    'configmap_type': var.get('type', 'configmap')
+                    })
+                
+            st.session_state[custom_env_key] = current_custom_vars + configmap_as_custom
+            st.session_state[configmap_loaded_key] = True
+
+        def reload_all_configmap_vars():
```
pdettori: This function is very similar to load_all_configmap_vars() - perhaps some common utility function could be extracted or shared by both, or could be one function parametrized with a flag

kagenti/ui/lib/build_utils.py
```diff
@@ -22,6 +22,7 @@
 import re
 import logging
 import json
+import yaml
```
pdettori: seems like this import has been added but not used later, can it be removed?

cwiklik: yup this should be removed

kagenti/ui/lib/build_utils.py
cwiklik: I'll try to combine these two. Thanks

kagenti/ui/lib/build_utils.py
```diff
@@ -1568,3 +1643,119 @@ def validate_registry_config(registry_config, st_object):
             return False
 
     return True
+
+def parse_configmap_data_to_env_vars(env_options: Dict[str, Any]) -> List[Dict[str, str]]:
```
pdettori: the type annotation for return value should be updated to reflect possible non-string types in the dict values. Best to use `List[Dict[str, Any]` if there are possibilities that the dict may contain lists or other dicts

kagenti/ui/lib/build_utils.py
```diff
@@ -1088,6 +1092,33 @@ def remove_env_var(index):
             if 0 <= index < len(st.session_state[custom_env_key]):
                 st.session_state[custom_env_key].pop(index)
 
+        def load_all_configmap_vars():
+            """load original configmap vars, keeping user-added custom vars"""
+            if not env_options:
+                st_object.warning(f"No configMap data available to load in namespace '{build_namespace_to_use}'")
+                return
+            # First, remove any previously loaded configmap vars
+            user_custom_vars = [var for var in st.session_state[custom_env_key] if not var.get('configmap_origin', False)]
```
pdettori: safer using .get() for st.session_state access to avoid possible KeyErrors.
```suggestion
            user_custom_vars = [
    var for var in st.session_state.get(custom_env_key, []) 
    if not var.get('configmap_origin', False)
]
```

kagenti/ui/lib/build_utils.py
```diff
@@ -1088,6 +1092,33 @@ def remove_env_var(index):
...
+                configmap = []
+                for var in all_configmap_vars:
+                    configmap.append({
+                        "name": var["name"], 
+                        "value": var["value"],
+                        'configmap_origin': True,
+                        'configmap_section': var.get('section', ''),
+                        'configmap_type': var.get('type', 'configmap')
+                        })
+                st.session_state[custom_env_key] = configmap + user_custom_vars
```
pdettori, line 1118: could overwriting `custom_env_key` erase other user changes? consider if merge would be safer here

kagenti/ui/lib/build_utils.py
```diff
@@ -1568,3 +1643,119 @@ def validate_registry_config(registry_config, st_object):
             return False
 
     return True
+
+def parse_configmap_data_to_env_vars(env_options: Dict[str, Any]) -> List[Dict[str, str]]:
+    """
+    Parses ConfigMap data and return all environment variables as a flat list.
+
+    Args:
+        configmap_data (Dict[str, Any]): The data from the ConfigMap.
```
pdettori: There is a mismatch - in the function signature it's `env_options`

kagenti/ui/lib/build_utils.py
cwiklik: Not in this case IMO. All user vars are in 'user_custom_vars'. If a user does 'Reload' the intention is to start fresh but preserve all custom vars previously defined and I think the code does this correctly.

kagenti/ui/lib/build_utils.py
cwiklik: Done

kagenti/ui/lib/build_utils.py
cwiklik: Done

kagenti/ui/lib/build_utils.py
cwiklik: Done

Issue #153, opened 2025-08-27
epic: Run Kagenti on OpenShift


Shreyanand:
## Overview
This issue documents the problems encountered when attempting to run Kagenti on OpenShift, including architecture compatibility, container management, security context, and registry configuration issues.

### Problem
Current container images are built for ARM architecture, but OpenShift environments may require AMD64/x86_64 images. It can become the default option as well. 

### Affected Components
- `kagenti-ui` container
- `kagenti-client-registration` container

### Note
- The `kagenti-client-registration` image is hardcoded in the [operator](https://github.com/kagenti/kagenti-operator/blob/118d96a98f9ece4c8d9e99480a3a7186f4ba6d83/platform-operator/internal/deployer/kubernetes/kubernetes.go#L239)
- As a workaround, deployments must be manually patched to update the image

### Impact
- Containers fail to start on AMD64-based OpenShift clusters

---

### Problem
Dockerfiles use `uv` package management which creates virtual environments within containers, causing compatibility issues.

### Affected Components
- `kagenti-ui` container
- Agent example images

### Proposed Solutions
1. **Immediate**: Switch to standard `pip install` for all containers
2. **Alternative**: Research proper `uv` usage in containerized environments and evaluate if `uv` provides benefits that justify the complexity

---

### Problem
PostgreSQL and kagenti-ui containers fail because they attempt to create files in root directories, which OpenShift's security policies prohibit.

### Root Cause
OpenShift's Security Context Constraints (SCC) prevent containers from writing to root filesystem locations.

### Affected Components
- PostgreSQL database container
- kagenti-ui container

### Proposed Solutions
1. **Immediate**: Use `/tmp` directory for file creation, but this is not a production-ready solution.
2. **Fix**:  Create and use PVC for storage 

---

### Problem
Agent/tool containers cannot pull from the `cr-system` internal registry because OpenShift's kubelet expects secure (HTTPS) registries, but the default registry is unsecured.

### Technical Details
- OpenShift enforces HTTPS for container registry communication
- Default `cr-system` registry runs on HTTP
- Securing the internal registry is complex and not straightforward

### Impact
- Agents and tools cannot be deployed from internal registry
- Requires external registry configuration

### Proposed Solution
Implement user-configurable registry support:
- Allow users to specify their own registry (e.g., Quay.io)
- Provide credential management for external registries
---

## **5. Other Minor Issues**
- Increase resource request for 'agentic-platform-controller-manager' to prevent 'OOMKilled' error
[...]

PR #199, opened 2025-09-12
📖 document how to install kagenti components in openshift

pdettori:
## Summary

This PR starts documenting how several kagenti prereqs can be installed in openshift, and provides templates that
help and automate the installation. This is just a first step in starting to add openshift support in the kagenti installer.

## Related issue(s)

Epic #153


kellyaa (approved):
I was able to perform these steps successfully, with some modifications. I will include my proposed changes in a separate PR. For now I think we can merge.

PR #211, opened 2025-09-16
✨ Start preparing installer for OCP detection and enablement

pdettori:
## Summary

This PR adds to the support for existing cluster, by detecting if connected to an OCP cluster and preparing for conditional logic in each component installer where for example an OpenShift operator can be used for that component.


pdettori:
This PR is going to be replaced by #229

PR #214, opened 2025-09-16
:seedling::sparkles: Leverage github release fetching for latest UI tag

### LINKED ISSUE #205: feature: use actual tag for UI image instead of `latest`
### Feature Description

It is always very hard to understand what version of an the UI image is running when using the "latest" tag. 

### Proposed Solution

Have the installer autodetect the latest tag and deploy using it

### Want to contribute?

### Additional Context

_No response_

evaline-ju:
## Summary

- Reusing function to grab operator version from #177 so moved to utils, with prints replaced with console.log...`
- Put constant info in installer config (seemed more appropriate than just the UI constants since the operator also uses the same logic)
- The "updated" yaml is written out temporarily to make this easy to apply with the current logic of applying multiple resources at one time [Alternatives considered were rewriting to the original file, but yamls may be reordered, with whitespace changes, and overall changes hard to see. Each resource could've also been applied individually with just the deployment created and/or updated with/without the Kube API - this may result in more code but could be considered in a future refactor. The rest of the UI updates are still leveraging the `run_command` so I decided to not update this in isolation to the Kube API.] 

## Related issue(s)

Fixes #205

### DIFF
--- a/kagenti/installer/app/utils.py
+++ b/kagenti/installer/app/utils.py
@@ -56,10 +56,14 @@ def get_latest_tagged_version(github_repo, fallback_version) -> str:
                 if '^{}' not in tag:  # Exclude annotated tags
                     return tag
 
-        console.log("[yellow]Could not find tag name in the response. Using fallback version.[/yellow]")
+        console.log(
+            "[yellow]Could not find tag name in the response. Using fallback version.[/yellow]"
+        )
         return fallback_version
     except subprocess.CalledProcessError as e:
-        console.log(f"[bold red]Error fetching latest version: {e}. Using fallback version.[/bold red]")
+        console.log(
+            f"[bold red]Error fetching latest version: {e}. Using fallback version.[/bold red]"
+        )
         return fallback_version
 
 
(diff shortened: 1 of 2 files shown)


kagenti/installer/app/utils.py
```diff
@@ -26,6 +26,42 @@
...
+            if line and 'refs/tags/' in line:
+                # Extract tag name
+                tag = line.split('refs/tags/')[-1]
+                if '^{}' not in tag:  # Exclude annotated tags
+                    return tag
+
+        print("Could not find tag_name in the response. Using fallback version.")
+        return fallback_version
+    except subprocess.CalledProcessError as e:
+        print(f"Error fetching latest version: {e}. Using fallback version.")
```
pdettori: I suggest using console.log() instead of print - e.g. `console.log(f"[bold red]Error fetching latest version: {e}. Using fallback version.[/bold red]")`

pdettori (commented):
@evaline-ju overall looks good - just minor comment

pdettori (approved):
LGTM

PR #216, opened 2025-09-17
:sparkles: Add pre commit for linting and formatting

evaline-ju:
## Summary

I noticed that the linting via `make lint` started to raise errors, so this is a proposal to leverage `pre-commit` to make this more automatic and consistent. For now, `make lint `only runs on some UI files. This is kept for now but can be expanded to the other files in this repo [but may require many updates for other files] and/or changed to leverage another linter like ruff. This mainly adds ruff-format which will format whitespace and other pep-8 violations.

Most of the changes to the `build_utils.py` are aesthetic - making spacing more consistent and standardizing on double quotes

Pre-commit only runs on the staged commits 

To install: `pip install pre-commit && pre-commit install` to set up Git hooks
To run manually: `pre-commit run --all-files`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,14 @@
+repos:
+- repo: local
+  hooks:
+    # Run the linter
+    - id: make-lint
+      name: Run make lint
+      entry: make lint
+      language: system
+      pass_filenames: false
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  rev: v0.4.7
+  hooks:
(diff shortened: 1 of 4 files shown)


pdettori (commented):
Thanks @evaline-ju this sounds like a good way to keep consistency. Could you please also add a section in the dev guide ( https://github.com/kagenti/kagenti/blob/main/docs/dev-guide.md ) to describe how to set it up (new contributors should not have to look back at the PR description to know how to set it up)

PR #221, opened 2025-09-19
:bug: Add missing CLIENT_SECRET to Weather Agent

mrsabath:
## Summary

Weather Agent does not set CLIENT_SECRET that is needed to obtain the Keycloak secret

## Related issue(s)

Fixes #

### DIFF
--- a/kagenti/installer/app/resources/environments.yaml
+++ b/kagenti/installer/app/resources/environments.yaml
@@ -24,7 +24,11 @@ data:
     ]
   mcp-weather: |
     [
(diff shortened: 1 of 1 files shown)


mrsabath:
> Actually I feel like the weather agent should be edited to remove the need to do this. Something may have gotten reverted

@maia-iyer if that is the case, I will simplify the `weather_agent` and remove the dependencies of Keycloak and the need to get the token. This way, the weather agent and the tool would be our easiest use-case for a demo. This would be a new PR in `agent-examples`. Once approved, we can close this PR.

maia-iyer:
Should be taken care of by PR 60 in agent examples repo now

PR #227, opened 2025-09-24
:sparkles: Expose Gateway following mcp-gateway pattern

esnible:
## Summary

Exposes the MCP server outside the Kind cluster at MCP=http://mcp.127-0-0-1.sslip.io:8888/mcp

This allows Kagenti to use the same endpoint address as the MCP-Gateway's dev environment.

Config based on https://github.com/kagenti/mcp-gateway/blob/377e7e8ebbdf6c5d904bf72b6ad48ea9f7a58001/config/istio/gateway/gateway.yaml#L11

### DIFF
--- a/kagenti/installer/app/resources/gateway.yaml
+++ b/kagenti/installer/app/resources/gateway.yaml
@@ -24,3 +24,10 @@ spec:
       allowedRoutes:
         namespaces:
           from: All
+    - name: mcp-mcp # gateway listener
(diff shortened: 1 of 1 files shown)


kagenti/installer/app/resources/gateway.yaml
```diff
@@ -24,3 +24,10 @@ spec:
       allowedRoutes:
         namespaces:
           from: All
+    - name: mcp-mcp # gateway listener
```
huang195, line 27: This adds a listener to the broker, but the 1st listener (unname) should route the traffic to the broker already

PR #228, opened 2025-09-24
Use SPIFFE SVID for client registration

Alan-Cha:


kagenti/auth/auth_demo/set_up_demo.py
```diff
@@ -36,7 +38,7 @@ def assign_realm_role_to_client_scope(admin: KeycloakAdmin, scope_id: str, role_
 slack_full_access_string = "slack-full-access"
 slack_partial_access_user_string = f"{slack_partial_access_string}-user"
 slack_full_access_user_string = f"{slack_full_access_string}-user"
-slack_client_id = "spiffe://localtest.me/sa/slack-tool"
+slack_client_id = f"spiffe://localtest.me/ns/{NAMESPACE}/sa/slack-tool"
```
Alan-Cha, line 41: As discussed, the Slack client ID of the auth demo now needs the namespace in order to match the one provided by SPIFFE

kagenti/auth/client-registration/client_registration.py
```diff
@@ -1,72 +1,93 @@
...
-    if keycloak_admin_username is None:
-        raise Exception('Expected environment variable "KEYCLOAK_ADMIN_USERNAME"')
-    if keycloak_admin_password is None:
-        raise Exception('Expected environment variable "KEYCLOAK_ADMIN_PASSWORD"')
-    if client_name is None:
-        raise Exception('Expected environment variable "CLIENT_NAME"')
-    if client_id is None:
-        raise Exception('Expected environment variable "CLIENT_ID"')
-    if namespace is None:
-        raise Exception('Expected environment variable "NAMESPACE"')
```
Alan-Cha: I did some refactoring to match the changes I made here previously: https://github.com/kagenti/kagenti/blob/693bd36b3c906ab3ebee93d2696f75f8712c8c2f/kagenti/auth/ui-oauth-secret/auth_secret.py#L45

Eventually, I want to refactor all these components so they utilize common libraries

kagenti/auth/client-registration/client_registration.py
```diff
@@ -1,72 +1,93 @@
...
+except Exception as e:
+    print(f"An error occurred: {e}")
+    
+if content is None or content.strip() == "":
+    raise Exception(f'No content read from SVID JWT.')
+
+decoded = jwt.decode(content, options={"verify_signature": False})
+if 'sub' not in decoded:
+    raise Exception('SVID JWT does not contain a "sub" claim.')
+client_id = decoded['sub']
```
Alan-Cha: This is the new section, the rest is pretty much the same.

Here, we are extracting the SPIFFE SVID that would be provided by `spiffe-helper`.

kagenti/installer/app/components/agents.py
```diff
@@ -131,6 +131,17 @@ def install(**kwargs):
...
+            [
+                "kubectl",
+                "apply",
+                "-n",
+                ns,
+                "-f",
+                str(config.RESOURCES_DIR / "spiffe_helper_config.yaml"),
+            ],
+            f"Applying spiffe_helper_config configmap in '{ns}'",
+        )
```
Alan-Cha, line 145: `spiffe-helper` requires a helper config in order to run. Because `spiffe-helper` will be running in all agent namespaces, I thought it made sense to add the helper config here, where it gets added to all agent namespaces like the environments config

kagenti/installer/app/components/keycloak.py
```diff
@@ -252,16 +252,16 @@ def install(use_existing_cluster: bool = False, **kwargs):
...
-                        "kagenti_keycloak_client_secret",
+                        "kagenti-keycloak-client-secret",
                         "--type=merge",
                         "-p",
                         patch_string,
                         "-n",
                         ns,
                     ],
-                    f"🔄 Patching 'kagenti_keycloak_client_secret' in namespace '{ns}'",
+                    f"🔄 Patching 'kagenti-keycloak-client-secret' in namespace '{ns}'",
```
Alan-Cha: This seems to be a bug I ran into.

Alan-Cha:
To test this PR

### Create a new image for client registration
Should we create a new tag to trigger a build? Otherwise, build locally:

```sh
cd kagenti/auth/client-registration
docker build -t kagenti-client-registration-test .
```

### Install Kagenti
```sh
cd kagenti/installer
uv run kagenti-installer
```

### Inject new image
Only do this if you created a local image.

```sh
kind load kagenti-client-registration-test --name agent-platform
```

### Create a new Kagenti operator
Should we trigger a build?

Otherwise, change the [client registration image](https://github.com/Alan-Cha/kagenti-operator/blob/fd1dccc088fcd8156f3cb87185f567382b232f93/platform-operator/internal/deployer/kubernetes/kubernetes.go#L308)

```diff
				{
					Name:            "kagenti-client-registration",
+					Image:           "kagenti-client-registration-test",
-					Image:           "ghcr.io/kagenti/kagenti/client-registration:latest",
					ImagePullPolicy: corev1.PullPolicy(imagePullPolicy),
					// Wait until /opt/jwt_svid.token appears, then exec
					Command: []string{
```

Create a new Kagenti operator image and install it
sh
```
./scripts/cleanup.sh 
make ko-local-build 
make install-local-chart
```

### Try it out
Creating a new agent or tool should result in the same Keycloak client name as before.

Try out the Slack auth demo as well as the auth script has also been updated.

kagenti/installer/app/resources/spiffe_helper_config.yaml
```diff
@@ -0,0 +1,14 @@
...
+  name: spiffe-helper-config
+data:
+  helper.conf: |
+    agent_address = "/spiffe-workload-api/spire-agent.sock"
+    cmd = ""
+    cmd_args = ""
+    svid_file_name = "/opt/svid.pem"
+    svid_key_file_name = "/opt/svid_key.pem"
+    svid_bundle_file_name = "/opt/svid_bundle.pem"
+    jwt_svids = [{jwt_audience="kagenti", jwt_svid_file_name="/opt/jwt_svid.token"}]
```
maia-iyer, line 13: The jwt_audience will probably need to be the Keycloak issuer name because it seems we'll be using the SVIDs primarily to talk to keycloak. In our case it's `http://keycloak.localtest.me:8080/realms/master`. Should be fine to hardcode for now, but may change depending on the deployment

kagenti/installer/app/components/agents.py
maia-iyer: SGTM!

maia-iyer (commented):
Initial comments - I'm going to test now

kagenti/auth/auth_demo/set_up_demo.py
maia-iyer: I think need to update the README for the slack demo to add that namespace variable; but that can be done in subsequent PR too

PR #241, opened 2025-09-29
🐛 fix for broken postgres image and other image related issues

pdettori:
## Summary

fix for broken postgres image and other image related issues


mrsabath (approved):
LGTM!

PR #242, opened 2025-09-29
🐛  add volumes and volume mounts to agents as well

pdettori:
## Summary
add volumes and volume mounts to agents as well. Previous PR only added to tools template.


mrsabath (approved):
LGTM!

PR #243, opened 2025-09-30
:sparkles: Retrieve the secret for a client

mrsabath:
## Summary
This PR allows the containerInit to register the client and retrieve it's secret so it can be passed to the main container for further interactions with Keycloak 

## Related issue(s)

Fixes #

### DIFF
--- a/kagenti/auth/client-registration/client_registration.py
+++ b/kagenti/auth/client-registration/client_registration.py
@@ -55,7 +55,7 @@ def get_secret(
     keycloak_admin: KeycloakAdmin,
     internal_client_id: str,
     client_name: str,
-    secret_file: str = "secret.txt",
+    secret_file_path: str = "secret.txt",
 ) -> None:
     """
     Retrieve the secret for a Keycloak client and write it to a file.
--- a/kagenti/auth/client-registration/client_registration.py
+++ b/kagenti/auth/client-registration/client_registration.py
@@ -65,16 +65,10 @@ def register_client(
 
     # Always try to get the secret
     try:
-        info = keycloak_admin.get_server_info()
-        # print(f'Server info: {info}')
-        clients = keycloak_admin.get_clients()
-        client_ids = [client['clientId'] for client in clients]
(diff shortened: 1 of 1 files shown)


kagenti/auth/client-registration/client_registration.py
```diff
@@ -61,6 +63,30 @@ def register_client(
...
+        print(f"Could not retrieve secret for client '{client_name}': {e}")
+        return
+
+    # Write secret to file
+    try:
+        with open(secret_file, "w") as f:
+            f.write(secret)
+        print(f'Secret written to file: "{secret_file}"')
+    except OSError as ioe:
+        print(f'Error writing secret to file: {ioe}')
```
Alan-Cha: Because we want to reuse this function, I don't think it makes sense to always try to get and write the secret. The function is `register_client`, not `register_client_and_fetch_secret`.

I think it would better if `register_client` returned the `internal_client_id`. Then after we call `register_client`, we can try getting the secret.

kagenti/auth/client-registration/client_registration.py
```diff
@@ -1,72 +1,92 @@
...
-        raise Exception('Expected environment variable "NAMESPACE"')
 
-    keycloak_admin = KeycloakAdmin(
-        server_url=keycloak_url,
-        username=keycloak_admin_username,
-        password=keycloak_admin_password,
-        realm_name=keycloak_realm,
-        user_realm_name='master'
-    )
+def get_secret(
```
Alan-Cha: Maybe `write_secret`?

kagenti/auth/client-registration/client_registration.py
```diff
@@ -1,72 +1,92 @@
...
+    keycloak_admin: KeycloakAdmin,
+    internal_client_id: str,
+    client_name: str,
+    secret_file: str = "secret.txt",
```
Alan-Cha:
```suggestion
    secret_file_path: str = "secret.txt",
```

PR #256, opened 2025-10-01
🐛  fix for allowing to correctly tag for tool image import

### LINKED ISSUE #231: bug: Deploy from Existing Image does not use supplied Image tag
### Describe the bug

If the user attempts to Import_New_Tool "Deploy from Existing Image" using the image ghcr.io/kagenti/mcp-gateway/test-server2:latest , the Kagenti UI creates a Pod with the image ghcr.io/kagenti/mcp-gateway/test-server2:v0.0.1 .

When deploying from an existing image, the tag should not be changed.

Related: #230 

### Steps To Reproduce

Import_New_Tool using "Deploy from Existing Image" using the image ghcr.io/kagenti/mcp-gateway/test-server2:latest 

The Kagenti UI creates a Deployment with the image ghcr.io/kagenti/mcp-gateway/test-server2:v0.0.1 , whose Pods get an ImagePullBackOff.

### Expected Behavior

The Deployment should use EXACTLY the same image as the user specified, including the tag.
Or the UI should let the user know that the tag will be ignored.

### Additional Context

This problem is especially difficult because the repo might not have the image with the `:v0.0.1` tag.  To get my image started I needed to `oc -n team1 edit deployment test-server2` and set `imagePullPolicy: IfNotPresent` to get the tool running.

_No response_

pdettori:
## Summary

fix for allowing to correctly tag for tool image import

Fixes #231

### DIFF
--- a/kagenti/ui/lib/build_utils.py
+++ b/kagenti/ui/lib/build_utils.py
@@ -206,6 +206,8 @@ def _construct_tool_resource_body(
             image_registry_prefix = "registry.cr-system.svc.cluster.local:5000"
     else:
         image_registry_prefix, image_name, _tag = parse_image_url(repo_url)
+        if _tag:
+            image_tag = _tag
 
     client_secret_for_env = _get_keycloak_client_secret(
         st_object, f"{k8s_resource_name}-client"


esnible (approved):
Works for me.
Thanks for reminding me about _./scripts/ui-dev-build.sh_.
The ui-dev-build.sh script has a flaw, it must be run in the directory above _scripts_.  If it is run from the scripts directly itself the Dockerfile can't be found.  Is that worthy of a fix?

PR #262, opened 2025-10-03
🐛 fix setting up mcp inspector proxy URL

### LINKED ISSUE #72: bug: MCP Inspector not connecting to weather tool
### Describe the bug

After importing weather tool and selecting "Connect with MCP Inspector", the inspector window opens. When I click "Connect" this error show up: Error Connecting to MCP Inspector Proxy - Check Console logs. There are no errors in mcp-inspector pod logs.

### Steps To Reproduce

As above.

### Expected Behavior

Inspector should be connecting and giving me the page to list tools.

### Additional Context

_No response_

pdettori:
## Summary

 fix setting up mcp inspector proxy URL. Accounts for MCP_PROXY_FULL_ADDRESS being part of the [configuration](https://github.com/modelcontextprotocol/inspector?tab=readme-ov-file#configuration) and not an environment variable so it can only be set in the UI or passed as query parameter.

Fixes #72

### DIFF
--- a/kagenti/ui/lib/tool_details_page.py
+++ b/kagenti/ui/lib/tool_details_page.py
@@ -79,6 +79,7 @@ def render_mcp_tool_details_content(tool_k8s_name: str):
     st.header(f"MCP Tool: {tool_k8s_name}")
 
     mcp_inspector_url = os.environ.get('MCP_INSPECTOR_URL', constants.MCP_INSPECTOR_URL)
+    mcp_proxy_url = os.environ.get('MCP_PROXY_FULL_ADDRESS', constants.MCP_PROXY_FULL_ADDRESS)
 
     custom_obj_api = get_custom_objects_api()
     if not custom_obj_api:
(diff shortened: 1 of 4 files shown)


kagenti/ui/lib/tool_details_page.py
```diff
@@ -156,11 +157,11 @@ def render_mcp_tool_details_content(tool_k8s_name: str):
     st.link_button(
     "Connect with MCP Inspector",
     # TODO - transport should be a property of MCP server
-    url=f"{mcp_inspector_url}?serverUrl={mcp_server_url}&transport=streamable_http",
+    url=f"{mcp_inspector_url}?serverUrl={mcp_server_url}&transport=streamable_http&MCP_PROXY_FULL_ADDRESS={mcp_proxy_url}",
```
esnible: Perhaps use `urllib.parse.quote()` in case the server or proxy URL was configured with a special character.  For security we don't want `<` unescaped, for correctness we should only allow URL characters and quote() will take care of that.

pdettori: done!

PR #268, opened 2025-10-06
🌱 Split Kagenti chart in kagenti-deps chart and kagenti-chart

pdettori:
## Summary

Since software dependencies change less frequently and could be also pre-installed on an existing cluster it makes more sense to separate the chart that installs the dependencies from the one that installs the actual Kagenti components and does the configurations of Keycloak etc.

### DIFF
--- a/docs/ocp/openshift-install.md
+++ b/docs/ocp/openshift-install.md
@@ -38,8 +38,8 @@ To start, ensure your `kubectl` or `oc` is configured to point to your OpenShift
 3. **Kagenti Dependencies Helm Chart Installation:**
    This chart includes all the OpenShift software components required by Kagenti.
    ```shell
-   # For example, if the latest tag is 0.0.4-alpha.18
-   LATEST_TAG=0.0.4-alpha.18
+   # For example, if the latest tag is 0.1.0-alpha.2
+   LATEST_TAG=0.1.0-alpha.2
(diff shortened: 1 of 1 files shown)


charts/gateway-api/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+name: gateway-api
+description: Helm Chart to install the Gateway API
+type: application
+version: 1.3.0
+appVersion: "1.3.0"
```
mrsabath, line 6: at some point we should tie this appVersion with the Gateway release tag value, but for now this is good.

docs/ocp/openshift-install.md
```diff
@@ -35,10 +35,16 @@ To start, ensure your `kubectl` or `oc` is configured to point to your OpenShift
    - Copy the [.secrets_template.yaml](https://github.com/kagenti/kagenti/blob/main/charts/kagenti/.secrets_template.yaml) to a local `.secrets.yaml` file.
    - Edit the `.secrets.yaml` to provide the necessary keys as per the comments within the file.
 
-3. **Run Helm Installation:**
+3. **Kagenti Dependencies Helm Chart Installation:**
+   This chart includes all the OpenShift software components required by Kagenti.
    ```shell
    # For example, if the latest tag is 0.0.4-alpha.18
    LATEST_TAG=0.0.4-alpha.18
```
mrsabath: Do we want to use the latest release value here?

mrsabath (approved):
LGTM

Issue #248, opened 2025-10-01
feature:  Kagenti installation chart should support vanilla Kubernetes cluster


pdettori:
### Feature Description

Currently the chart is mainly focused on deploying on OpenShift, we need to support also vanilla K8s.

### Proposed Solution

Work on templates for the various dependencies to ensure we can use the chart to install on K8s.

### Want to contribute?

- [ ] I would like to work on this issue.

d0w:
Hey I'm new to this project and would like to contribute to this issue! Please let me know if I could be assigned

d0w:
Hi @pdettori. Thanks for the heads up. I was looking at using the _helpers.tpl file to define a `kagenti.isOpenShift` value and then use that in the template files like so

`{{- if and .Values.components.certManager.enabled (include "kagenti.isOpenShift" .) }}`

Using tags seems more reasonable though and I'll look into using that instead.

PR #279, opened 2025-10-08
✨ Add kagenti-deps chart support for vanilla Kubernetes clusters for Kagenti installation 

d0w:
### To test
1. Install operator-sdk olm
```
operator-sdk olm install

# alternate method but was buggy on my side
curl -sL https://github.com/operator-framework/operator-lifecycle-manager/releases/download/v0.36.0/install.sh | bash -s v0.36.0
```

1.1 Optionally install Spire
```
helm upgrade --install -n zero-trust-workload-identity-manager spire-crds spire-crds --repo https://spiffe.github.io/helm-charts-hardened/ --create-namespace

helm upgrade --install -n zero-trust-workload-identity-manager spire spire --repo https://spiffe.github.io/helm-charts-hardened/
```

2. Install cert-manager
```
helm install \
  cert-manager oci://quay.io/jetstack/charts/cert-manager \
  --version v1.19.1 \
  --namespace cert-manager-operator \
  --create-namespace \
  --set crds.enabled=true
```

2. Install prometheus
```
helm install kagenti-prometheus kube-prometheus-stack --repo=https://prometheus-community.github.io/helm-charts --create-namespace --namespace=monitoring
```

4. Install kagenti-deps with proper values set for non-openshift clusters
```
cd charts/kagenti-deps

helm dependency build # may need

helm install -n kagenti-system kagenti-deps ./ -f ./values.yaml --create-namespace \
  --set openshift=false \
  --set tags.kubernetes=true \
  --set components.ingressGateway.enabled=true \
  --set components.gatewayApi.enabled=true
## --set components.[components].enabled=true if want to install other components like spire from earlier
```

5. To uninstall (this does not remove all CRDs or created CRs)
```
helm uninstall kagenti-deps -n kagenti-system

helm uninstall -n zero-trust-workload-identity-manager spire
helm uninstall -n zero-trust-workload-identity-manager spire-crds

helm uninstall -n monitoring kagenti-prometheus

helm uninstall -n cert-manager-operator cert-manager

operator-sdk olm uninstall
```

### Changes
This mainly targets `charts/kagenti-deps`

This seeks to add vanilla k8s cluster support when using the helm charts to install Kagenti. Currently, the charts only support OpenShift clusters. To do this, some components use subchart dependencies are added into the `kagenti-deps` chart to replace the Operator subscription counterpart and tagged with `kubernetes` so that the dependencies are installed when setting `tags.kubernetes=true` and `tags.openshift=false`. Other components use operator-sdk's OLM which allows the helm charts to create a Subscription CR to add in the operator, mimicking OpenShift operator installations.

Some notable differences between vanilla K8s and OpenShift that were a factor in these changes include:
- HTTPRoutes and the Gateway API are used instead of Routes.
- operator-sdk olm must be installed prior to helm installing
- Subcharts were added to `kagenti-deps` via an online repository or a separate helm chart within `kagenti-deps/charts` if operator-sdk's operater hub did not have a supported version (aka cert-manager).

## Related issue(s)
Issue: https://github.com/kagenti/kagenti/issues/248

Fixes #

### DIFF
--- a/charts/kagenti-deps/Chart.yaml
+++ b/charts/kagenti-deps/Chart.yaml
@@ -7,6 +7,8 @@ appVersion: "1.0"
 
 dependencies:
   - name: gateway-api
+    # Version constraint changed from exact '1.3.0' to '~1.3.0' to allow patch updates (e.g., 1.3.1, 1.3.2).
+    # Please ensure gateway-api maintains backward compatibility across patch versions, or review upgrade notes for details.
     version: "~1.3.0"
     repository: "file://../gateway-api"
     condition: components.gatewayApi.enabled


pdettori:
Question - do we need all `--set openshift=false --set tags.openshift=false --set tags.kubernetes=true` or can we just use combinations  of `--set openshift=false  --set tags.kubernetes=true`

jennydaman:
Hi, just a mysterious stranger passing by...

FYI it is possible to automatically detect whether k8s is openshift using Helm.

```gotmpl
{{ .Capabilities.APIVersions.Has "security.openshift.io/v1" }}
```

https://github.com/bitnami/charts/blob/0a101909ce511465df195263fda978dfae973ebc/bitnami/common/templates/_compatibility.tpl#L14

A practice set forth by bitnami is to have the `openshift` value be an enum, one of:

- `auto` (default): auto-detect OpenShift
- `force`: always apply OpenShift-specific configuration
- `disabled`: do not apply OpenShift-specific configuration

When using [helm-unittest](https://github.com/helm-unittest/helm-unittest), OpenShift can be mocked like this: 
https://github.com/FNNDSC/charts/blob/ce1aa52cd96ea49447e79d6da2b2b7b94e908b74/charts/pfcon/tests/set_security_context_in_openshift_test.yaml

charts/prometheus/values.yaml
jennydaman, line 1: A chart with an empty `templates/` and empty `values.yaml` is fishy. IMO it is an unnecessary layer of indirection. Having `kagenti/kagenti-deps` depend on `kube-prometheus-stack` would be preferable for the following reasons:

- Give credit where due: the name "prometheus-community/kube-prometheus-stack" credits "prometheus-community," whereas "kagenti/prometheus" might mislead people into believing that kagenti requires a custom configuration of prometheus created by kagenti.
- Lower maintenance burden: kube-prometheus-stack can be upgraded by cluster admins independently of kagenti/kagenti-deps.
  - P.S. As someone who has been running kube-prometheus-stack on vanilla k8s for 3 years, I can say that kube-prometheus-stack often requires manual intervention during upgrades. You do not want to have to copy-paste these manual steps to kagenti's documentation, nor do you want kagenti's reputation to be harmed when a breaking kube-prometheus-stack upgrade happens.
    - Steps for manual upgrades are documented here: https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/UPGRADE.md.
      - Gotcha! Those steps don't actually work, as reported in a 2022 comment: https://github.com/prometheus-community/helm-charts/issues/1668#issuecomment-1091717444

@d0w mentioned to me over email that it will be necessary to set different default values for kube-prometheus-stack to configure it to work with kagenti. It should be possible to achieve this without using an empty subchart.

Instead of: kagenti-deps depending on kagenti/prometheus depending on prometheus-community/kube-prometheus-stack,
I suggest: kagenti-deps directly depending on prometheus-community/kube-prometheus-stack

The `values.yaml` for kagenti-deps would contain:

```yaml
kube-prometheus-stack:
  enabled: true
  namespaceOverride: monitoring
  customRules:
    whateverIs: neededByKagenti
```

pdettori:
@jennydaman thank you for your insights! Do you know if we can use the auto detection for OpenShift on a helm chart condition ? The problem we found is that for some components we need to use dependency sub-charts for k8s and operators for OpenShift, could not find something better than combinations of tags and variables in the value files to control the inclusion of dependencies.

pdettori:
@d0w could you pls. rebase ?

jennydaman:
> Do you know if we can use the auto detection for OpenShift on a helm chart condition ? 

@pdettori No, subchart dependency condition must be a literal value not template code.

<details>
<summary>But you could use the forbidden knowledge...</summary>

You can run arbitrary code using [Helm hooks](https://helm.sh/docs/topics/charts_hooks/), so nothing is impossible. More specifically, run `kubectl` and `helm` inside a Pod during a pre-install hook, to install extra dependencies. Needless to say, as a project maintainer it is your call whether the complexity is worth it just to support a specific use case.

IMO, `cert-manager`, `gateway-api`, and `prom-stack` are somewhat ubiquitous even in vanilla k8s—any unintended/unexpected problem in trying to automate their installation as dependencies could cause severe trouble, e.g. on vanilla k8s where the user already has prom-stack, you might end up with a duplicate prom-stack, overwriting the existing prom-stack, or upgrading CRDs.

</details>

pdettori:
I started testing on OpenShift and found something I was not expecting - the dependency chart is installed even if we set `tags.kubernetes: false` in yaml file. Some search on this topic found the following:


Why Isn't tags.kubernetes: false in values.yaml Working?
This is a common source of confusion! The reason is:

Helm only uses the parent chart’s values.yaml to set dependency tags if the dependency is packaged as a subchart.

If you are using an OCI dependency (as in your example), or any external dependency, the tags and condition resolution is performed before your values.yaml is merged in.

Therefore:

The dependency is enabled before Helm reads your parent chart’s values.yaml.
Only values passed on the command line (--set or -f) are considered during dependency resolution.
References:

[Helm Docs: Chart Dependencies](https://helm.sh/docs/topics/charts/#chart-dependencies)
[Helm Docs: Tags and Condition fields](https://helm.sh/docs/chart_best_practices/dependencies/#tags-and-condition-fields)

pdettori:
There is some interesting questions here - what is the right balance between complexity in the chart (by trying to incorporate all in one) and complexity in user commands (by asking to run several helm install commands).  Based on @jennydaman 's comments I am considering that for this first pass we should keep this simpler and just remove the cert-manager chart from the dependencies - at least for OpenShift this does not affect the current behavior, and then can just add documentation on how to install it same as spire, prometheus etc.

pdettori:
Also, we should not have empty templates/ and empty values.yaml for spire and prometheus as previously  noted by @jennydaman  - I see they are still in the PR

PR #285, opened 2025-10-13
✨ update docs for namespace and kiali fix

cooktheryan:
## Summary
Allows for assets to appear within Kiali and the error to go away. Directions on how to see the namespace in the UI.

## Related issue(s)

Fixes #


pdettori:
After more investigation I found a fix by adding a service for the ztunnel pods and a servicemonitor to scrape the metrics for prometheus (tried also directly a podmonitor but did not work). See https://github.com/kagenti/kagenti/pull/312

Issue #267, opened 2025-10-06
feature: Create a Env. Variable to replace hard-coded Client Secret File path


mrsabath:
### Feature Description

Currently the `secret_file_path` it's hardcoded here: https://github.com/kagenti/kagenti/blob/7927591bf29ca90996db957612e9adea349c7111/kagenti/auth/client-registration/client_registration.py#L97

### Proposed Solution

We should replace this hardcoded value with a env. variable that can be modified. 

### Want to contribute?

- [ ] I would like to work on this issue.

maia-iyer:
We need this implemented in Authbridge in the case where SPIRE is not enabled.

PR #289, opened 2025-10-14
✨ made secret file path configurable via env

yenayuu:
## Related issue(s)

Fixes #
#267

### DIFF
--- a/kagenti/auth/client-registration/client_registration.py
+++ b/kagenti/auth/client-registration/client_registration.py
@@ -134,7 +134,7 @@ internal_client_id = register_client(
 print(
     f'Writing secret for client ID: "{client_id}", internal client ID: "{internal_client_id}"'
 )
-secret_file_path = get_env_var("SECRET_FILE_PATH")  
+secret_file_path = get_env_var("SECRET_FILE_PATH")
 write_client_secret(
     keycloak_admin,
     internal_client_id,


kagenti/installer/app/config.py
```diff
@@ -27,6 +27,7 @@
 OPERATOR_NAMESPACE = "kagenti-system"
 TEKTON_VERSION = "v0.66.0"
 KEYCLOAK_URL = "http://keycloak.localtest.me:8080/realms/master"
+DEFAULT_SECRET_FILE_PATH = "/shared/secret.txt"
```
maia-iyer: where is this variable used?

yenayuu: I assumed a shared volume mounted at "/shared/" as a default and used the SECRET_FILE_PATH constant from agents.py. Is there other preferred mount/path?

maia-iyer: I think the value is right, but I'm confused where `DEFAULT_SECRET_FILE_PATH` is used?

yenayuu: Thank you for the clarification. I just removed the unused variable and double checked with the new env var set up.

mrsabath (approved):
Thank you for the update. 
LGTM

Alan-Cha:
@maia-iyer @kellyaa This also needs to coincide with some changes to the agents, right? We've changed where the secret is written to but whatever is reading the secret also needs to know where the secret is now written to, right?

Issue #266, opened 2025-10-06
feature: Improve UI functionality by showing final Environment Variables for each component.


mrsabath:
### Feature Description

Currently we have several places where the environment variables can be for each component. They are done on the kagenti level, on the env. files passed when importing the component (agent or tool) or hardcoded to the deployment file. The only way to see the final value of the variable is to exec into the container and list it.

### Proposed Solution

Add the environment variables view on the Detailed View panel for Agents and Tools

### Want to contribute?

- [ ] I would like to work on this issue.

EthanL3:
Hi, can I be assigned to this issue if no one else is planning to take it?

PR #294, opened 2025-10-14
✨Display final value of env vars for agents/tools in UI

EthanL3:
## Summary
Currently the only way to view the final value of env vars for agents and tools is to `docker exec` into the containers and list them there, this addresses the problem by displaying them in the UI in the detail pages for agents and tools
## Related issue(s)
Issue https://github.com/kagenti/kagenti/issues/266

### DIFF
--- a/kagenti/ui/lib/kube.py
+++ b/kagenti/ui/lib/kube.py
@@ -486,7 +486,7 @@ def _find_pods_for_resource(core_v1_api, resource_name, namespace):
             continue
 
     error_message = (
-        f"No pods found with labels app.kubernetes.io/name={resource_name} "
+        f"No pods found with label app.kubernetes.io/name={resource_name} "
         f"or app={resource_name} in namespace {namespace}"
     )
     return None, error_message


pdettori:
tested the PR but it only works [locally ]( https://github.com/kagenti/kagenti/blob/main/docs/dev-guide.md#running-locally) - did not work when running the [image in a pod](https://github.com/kagenti/kagenti/blob/main/docs/dev-guide.md#running-your-image-in-kubernetes) - this is because it lacks the needed RBAC authorization to list and read pods.

Please update the `ClusterRole` in `deployments/ui/kagenti-ui.yaml` to add a rule for:

```
  - apiGroups: [""]
    resources: ["pods"]
    verbs: ["get", "list"]
 ```

File: kagenti/ui/lib/common_ui.py [hunk 1]
```diff
@@ -321,6 +321,92 @@ def display_resource_metadata(st_object, resource_details: dict):
     return tags
 
 
+def display_environment_variables(st_object, env_vars: dict) -> None:
```
pdettori, line 324: There is a data mismatch — display_environment_variables expects a different data shape than get_pod_environment_variables returns

common_ui.display_environment_variables — creation starts around common_ui.py [323,324]+ and iterates expecting env entries to be dictionaries with a nested "valueFrom" key and Kubernetes-style JSON keys like "configMapKeyRef" / "secretKeyRef": e.g. common_ui.py lines [323,360..373], [323,376..387] (see checks for "valueFrom", "configMapKeyRef", "secretKeyRef")


get_pod_environment_variables (new) — kube.py [467,530]+ returns flattened dicts when source is configmap/secret: e.g. it appends {"name": env_var.name, "source_name": source.config_map_key_ref.name, "source_key": source.config_map_key_ref.key} (kube.py lines [467,501..508], [467,509..516]).


Therefore display_environment_variables DOES NOT show configmap/secret/fieldref/resourcefield entries because it looks for a nested "valueFrom" structure that get_pod_environment_variables does not produce. As a result ConfigMap/Secret entries will not be displayed (or worse, code branches that assume keys exist may error) - this matches my test, where I cannot see the env vars for secrets.

pdettori: I suggest to normalize `get_pod_environment_variables` to produce the same structure `display_environment_variables expects` (i.e., return the original env dict or build a `{"valueFrom": {"configMapKeyRef": {...}}}-style structure)`. Example: for a configmap key ref, append:
`{"name": env_var.name, "valueFrom": {"configMapKeyRef": {"name": cm_name, "key": cm_key}}}`

pdettori (approved):
/lgtm - thank you @EthanL3 !

PR #313, opened 2025-10-17
use mcp gateway + kuadrant helm charts

huang195:
We have been cherrypicking from the mcp gateway repo on various artifacts to do installation, which is error prone. This PR switches to helm charts for installation. Additionally, this installs Kuadrant for authz related work.


kagenti/installer/app/components/mcp_gateway.py
```diff
@@ -18,66 +18,66 @@
...
+        "Installing Kuadrant operator",
     )
 
-    # """Init MCPServer CR."""
-    # This command creates an empty MCPServer resource
-    # run_command(
-    #    ["kubectl", "apply", "-f", str(config.RESOURCES_DIR / "gateway-mcpserver.yaml")],
-    #    "Initing MCPServer CR",
-    # )
+    """Installs Kudrant"""
```
pdettori: Typo "Kudrant" -> "Kuadrant"

kagenti/installer/app/components/mcp_gateway.py
```diff
@@ -18,66 +18,66 @@
...
         [
-            "kubectl",
-            "apply",
-            "-f",
-            str(config.RESOURCES_DIR / "gateway-namespaces.yaml"),
+            "helm",
+            "repo",
+            "add",
+            "kuadrant",
+            "https://kuadrant.io/helm-charts",
```
pdettori, line 69: Is Kuadrant now a requirement to use the MCP Gateway ?

pdettori (commented):
Don't we still need to create a Gateway (as in gateway.yaml) ? it does not look the mcp-gateway chart includes it...

kagenti/installer/app/components/mcp_gateway.py
huang195: will fix the typo

kagenti/installer/app/components/mcp_gateway.py
huang195: it's not a requirement, but for more involved use cases needing authz, we need to use Kuadrant. MCP-gateway is currently an optional component of Kagenti, and can be skipped during installation. In the future, we could make Kuadrant an optional component of MCP-gateway if we find the pods running in the `kuadrant-system` ns using too much resources.

kagenti/installer/app/components/mcp_gateway.py
```diff
@@ -18,66 +18,87 @@
...
+            "helm",
+            "upgrade",
+            "--install",
+            "kuadrant-operator",
+            "kuadrant/kuadrant-operator",
+            "--create-namespace",
+            "--wait",
+            "--timeout=600s",
+            "--namespace",
+            "kuadrant-system",
```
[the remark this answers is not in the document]
huang195: # Namespace is already created as part o the helm chart, so not using --create-namespace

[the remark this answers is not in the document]
huang195: https://github.com/kagenti/mcp-gateway/issues/290

huang195:
@pdettori rebased, and passed all the CI

PR #337, opened 2025-10-23
feat: Enable UI authentication with Keycloak on OpenShift

### LINKED ISSUE #317: feature: Enable UI Authentication & Token Management on OpenShift
According to the OCP installation guide UI authentication and token management are currently disabled in OpenShift installs. This limits access control and user experience in production. It's worth fixing this to ensure that we are operating from an even baseline with vanilla kubernetes.

We need to re-enable this functionality and ensure it works with OpenShift OAuth or with Keycloak, depending on the deployment.

-  Investigate how the UI currently disables or bypasses auth/token flow on OpenShift
-  Re-enable auth with either:
    - OpenShift OAuth (via OAuthClient + route annotations)
    - Or Keycloak integration, if preferred
-  Ensure UI token refresh and session logout behavior works under OpenShift
-  Test login/logout flow using Route-based access
-  Document steps in openshift-install.md

Shreyanand:
# Enable UI Authentication with Keycloak on OpenShift

## Summary
Enables Keycloak-based authentication for the Kagenti UI on OpenShift, resolving SSL certificate verification issues with self-signed certificates. Fixes #317

## Changes
- **SSL Fix**: ~~Monkey-patched `httpx.Client` and `httpx.AsyncClient` to disable SSL verification for OpenShift self-signed certificates~~ **Uses `SSL_CERT_FILE` environment variable with Kubernetes CA bundle** to properly verify OpenShift route certificates
- ~~**Enable Auth**: Set `ui.auth.enabled: true` by default in `values.yaml`~~ **Auth remains disabled by default** - users can enable via values.yaml
- **Documentation**: Updated OpenShift install guide with prerequisites and authentication setup instructions

## Prerequisites
Users must create `kagenti-ui-oauth-secret` in `kagenti-system` namespace with Keycloak OAuth configuration ~~before accessing the UI~~. **The secret must include `SSL_CERT_FILE` key pointing to the CA bundle path.** See documentation for details.

## Files Changed
- ~~`kagenti/ui/Home.py` - SSL bypass implementation~~ **No code changes**
- ~~`charts/kagenti/values.yaml` - Enable auth flag~~ **No changes to values.yaml**
- `docs/ocp/openshift-install.md` - Setup instructions with Keycloak client configuration
- `charts/kagenti/templates/ui.yaml` - Added `SSL_CERT_FILE` env var (optional on vanilla k8s, required on OpenShift)

## Testing
- Tested patching the env var on OpenShift Kagenti UI
- Login/logout flow verified  
- [To do] test the helm install and check if changes don't break Kind deployment  

---

### DIFF
--- a/docs/ocp/openshift-install.md
+++ b/docs/ocp/openshift-install.md
@@ -6,7 +6,6 @@
 
 These limitations will be addressed in successive PRs.
 
-- UI Auth and token management is disabled
 - Only [quay.io](https://quay.io) registry has been tested in build from source
 - Ollama models not tested - OpenAI key required for now
 
@@ -100,6 +99,72 @@ To start, ensure your `kubectl` or `oc` is configured to point to your OpenShift
    helm upgrade --install kagenti ./charts/kagenti/ -n kagenti-system --create-namespace -f ./charts/kagenti/.secrets.yaml
    ```
 
+## Authentication Configuration
+
+Kagenti UI now supports Keycloak authentication by default.
+
+### Prerequisites
+
+**IMPORTANT**: Before accessing the UI, you must create a Kubernetes secret named `kagenti-ui-oauth-secret` in the `kagenti-system` namespace with the following keys:
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: kagenti-ui-oauth-secret
+  namespace: kagenti-system
+type: Opaque
+stringData:
+  ENABLE_AUTH: "true"
+  CLIENT_ID: "kagenti"
+  CLIENT_SECRET: "<your-keycloak-client-secret>"
+  AUTH_ENDPOINT: "https://<keycloak-route>/realms/master/protocol/openid-connect/auth"
+  TOKEN_ENDPOINT: "https://<keycloak-route>/realms/master/protocol/openid-connect/token"
+  REDIRECT_URI: "https://<ui-route>/oauth2/callback"
+  SCOPE: "openid profile email"
+  SSL_CERT_FILE: "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+```
+
+Replace the placeholders:
+
+**1. Get Keycloak client secret:**
+
+First, find the Keycloak admin username and password in the `keycloak` namespace.
+
+Then, log in to the Keycloak admin console:
+1. Navigate to the Keycloak route URL
+2. Log in with admin credentials
+3. Go to **Clients** in the left sidebar
+4. Click on the `kagenti` client
+5. Go to the **Credentials** tab
+6. Copy the **Client Secret** value
+
+**2. Get route hostnames:**
+```shell
+# Get Keycloak route
+KEYCLOAK_ROUTE=$(oc get route keycloak -n keycloak -o jsonpath='{.spec.host}')
+
+# Get UI route  
+UI_ROUTE=$(oc get route kagenti-ui -n kagenti-system -o jsonpath='{.spec.host}')
+
+echo "Keycloak route: $KEYCLOAK_ROUTE"
+echo "UI route: $UI_ROUTE"
+```
+
+**3. Update the secret:**
+
+Replace in the YAML above:
+- `<your-keycloak-client-secret>`: The client secret from Keycloak (step 1)
+- `<keycloak-route>`: Value of `$KEYCLOAK_ROUTE`
+- `<ui-route>`: Value of `$UI_ROUTE`
+
+### Authentication Features
+
+- **Keycloak Integration**: Uses the deployed Keycloak instance for authentication
+- **Automatic Token Management**: Handles token refresh and session management
+- **User Session Management**: Supports login/logout functionality
+- **SSL Certificate Handling**: Automatically handles OpenShift self-signed certificates
+
 ## Access the UI
 
 After the chart is installed, follow the instructions in the release notes to access the UI. To print the UI URL, run:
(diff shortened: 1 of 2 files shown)


kagenti/ui/Home.py
```diff
@@ -18,11 +18,32 @@
...
+    kwargs['verify'] = False
+    return original_client_init(self, *args, **kwargs)
+
+def patched_async_client_init(self, *args, **kwargs):
+    kwargs['verify'] = False
+    return original_async_client_init(self, *args, **kwargs)
+
+httpx.Client.__init__ = patched_client_init
+httpx.AsyncClient.__init__ = patched_async_client_init
+
```
pdettori: This code disables TLS verification globally and monkey-patches httpx constructors:
it disables certificate validation for the entire process (all libs and requests), exposing tokens and credentials to MITM attacks. I think we should try not to have a global SSL-disable and the httpx monkey-patch.

Suggestions to try out:

- Remove the global ssl change and monkey-patch.

- If self-signed certs are the problem, mount the cluster/Keycloak CA into the container and pass verify="/path/to/ca.pem" (or set SSL_CERT_FILE) for the specific httpx client(s) used for OAuth.

- If that does not work, can still use the monkey-patching but gate it behind an explicit env var (e.g., ENABLE_INSECURE_SSL) and document a warning — this should be used only for dev/testing in clusters with self signed certs.

Minimal replacement pattern:

Provide a helper that constructs clients with controlled verify behavior: 

```
def make_httpx_client(insecure: bool = False, ca_path: Optional[str] = None): verify = False if insecure else (ca_path or True) return httpx.Client(verify=verify)
```

charts/kagenti/values.yaml
```diff
@@ -47,7 +47,7 @@ agentNamespaces:
 ui:
   configmap: kagenti-ui-config
   auth:
-    enabled: false
+    enabled: true
```
pdettori: Let's keep the default to `false` and use `--set ui.auth.enabled=true` until we get the other automation for KeyCloak setup in place

docs/ocp/openshift-install.md
```diff
@@ -83,15 +82,71 @@ To start, ensure your `kubectl` or `oc` is configured to point to your OpenShift
...
+  CLIENT_ID: "kagenti"
+  CLIENT_SECRET: "<your-keycloak-client-secret>"
+  AUTH_ENDPOINT: "https://<keycloak-route>/realms/master/protocol/openid-connect/auth"
+  TOKEN_ENDPOINT: "https://<keycloak-route>/realms/master/protocol/openid-connect/token"
+  REDIRECT_URI: "https://<ui-route>/oauth2/callback"
+  SCOPE: "openid profile email"
+```
+
+Replace:
+- `<your-keycloak-client-secret>`: The client secret from your Keycloak client
```
pdettori: some explanation or pointers on how to get the keycloak client secret would be helpful (until we have the automation in place)

docs/ocp/openshift-install.md
```diff
@@ -83,15 +82,71 @@ To start, ensure your `kubectl` or `oc` is configured to point to your OpenShift
...
+oc get route kagenti-ui -n kagenti-system -o jsonpath='{.spec.host}'
 ```
 
+### Authentication Features
+
+- **Keycloak Integration**: Uses the deployed Keycloak instance for authentication
+- **Automatic Token Management**: Handles token refresh and session management
+- **User Session Management**: Supports login/logout functionality
+- **SSL Certificate Handling**: Automatically handles OpenShift self-signed certificates
+
```
pdettori, line 167: Please add a note about mounting the Keycloak CA cert or setting SSL_CERT_FILE if Keycloak uses self-signed certs (to avoid insecure TLS workarounds).

I suggest also to document what happens if `kagenti-ui-oauth-secret` is missing (install will succeed but login will fail, or install will error).

kagenti/ui/Home.py
```diff
@@ -18,11 +18,32 @@
...
 
+# Disable SSL verification globally for self-signed certificates (OpenShift uses self-signed certs)
+ssl._create_default_https_context = ssl._create_unverified_context
+
+# Monkey-patch httpx to disable SSL certificate verification for OAuth flows
+# This is required because httpx_oauth uses httpx.AsyncClient internally
+original_client_init = httpx.Client.__init__
+original_async_client_init = httpx.AsyncClient.__init__
+
+def patched_client_init(self, *args, **kwargs):
```
pdettori: Pylint warns about missing function docstrings (C0116). If you keep small helper functions, add concise docstrings explaining purpose, arguments, and why they exist. That said, prefer removing the helpers by removing the monkey-patch if possible (see comments below) . If helpers remain, add explicit typing and docstrings.

kagenti/ui/Home.py
```diff
@@ -18,11 +18,32 @@
...
+    kwargs['verify'] = False
+    return original_client_init(self, *args, **kwargs)
+
+def patched_async_client_init(self, *args, **kwargs):
```
pdettori: see comment above about `patched_client_init`

pdettori:
@Alan-Cha @mrsabath pls. have a look as you have been working on this part for the dev / kind env.

kagenti/ui/Home.py
Shreyanand: sounds better, will add that instead.

Shreyanand:
@pdettori turns out adding `SSL_CERT_FILE` to the UI env vars solves the issue. No need for changing `home.py`. I could take a look at automatic `kagenti-ui-oauth-secret` secret creation on OpenShift similar to what exists for Kind as a part of this PR if that makes sense?

pdettori:
@Shreyanand could you please rebase? Thanks!

pdettori (approved):
/lgtm

PR #344, opened 2025-10-27
🐛  fix defaults for mcp inspector transport

### LINKED ISSUE #343: bug: MCP Inspector does not get correct streaming http option by default
### Describe the bug

When opening MCP inspector from UI, if this is the first time for a user, MCP inspector panel on left shows STDIO as default transport. It should instead start with correct setting (streamable http). It should also start with connection type set to proxy.

### Steps To Reproduce

1. Open a private window on your browser (to avoid picking up previous settings in the cache)
2. Open Kagenti UI
3. Go to observability page
4. Open MCP inspector page

### Expected Behavior

MCP Inspector should start with streamable http option set

### Additional Context

_No response_

pdettori:
## Summary

 Fix defaults for mcp inspector transport so that MCP inspector starts with streamable HTTP transport.

## Related issue(s)

Fixes #343

### DIFF
--- a/kagenti/ui/lib/tool_details_page.py
+++ b/kagenti/ui/lib/tool_details_page.py
@@ -166,7 +166,7 @@ def render_mcp_tool_details_content(tool_k8s_name: str):
     console_url = (
         f"{mcp_inspector_url}?"
         f"serverUrl={encoded_server_url}&"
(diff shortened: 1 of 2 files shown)


mrsabath (approved):
LGTM

PR #346, opened 2025-10-27
🐛 fix handling of ENABLE_AUTH when running locally

### LINKED ISSUE #345: bug:  AttributeError: 'NoneType' object has no attribute 'lower' when running UI in local mode
### Describe the bug

When testing with UI in local mode for local development starting with "uv run streamlit run Home.py " produces this error:

```
37 │   ENABLE_AUTH = os.environ.get("ENABLE_AUTH")                                 
     38 │   if ENABLE_AUTH_STRING not in st.session_state:                              
  ❱  39 │   │   if ENABLE_AUTH.lower() == "true":                                       
     40 │   │   │   logger.info("Authentication is enabled")                            
     41 │   │   else:                                                                   
     42 │   │   │   logger.info("Authentication is disabled")                           
────────────────────────────────────────────────────────────────────────────────────────
AttributeError: 'NoneType' object has no attribute 'lower'
```

### Steps To Reproduce

1. Run UI in local mode with `uv run streamlit run Home.py`
2. Open home page 
3. Observe the error above

### Expected Behavior

Should not throw an error when lacking that env var

### Additional Context

_No response_

pdettori:
## Summary

This PR fixes a bug in the authentication handling logic when running locally. The issue occurred when the ENABLE_AUTH environment variable was not set, causing a NoneType error when calling .lower() on None.

## Key Changes:

Added a default value of "false" for the ENABLE_AUTH environment variable to prevent None errors
Normalized the authentication check into a single boolean variable to avoid repeated string operations

## Related issue(s)

Fixes #345

### DIFF
--- a/kagenti/ui/Home.py
+++ b/kagenti/ui/Home.py
@@ -34,15 +34,19 @@ def render_login():
     Render the login page.
     """
 
-    ENABLE_AUTH = os.environ.get("ENABLE_AUTH")
+    # Default to 'false' when the env var is missing so calling .lower() is safe
+    ENABLE_AUTH = os.environ.get("ENABLE_AUTH", "false")
+    # Normalize once into a boolean to avoid repeated .lower() calls and None issues
+    enable_auth_enabled = str(ENABLE_AUTH).lower() == "true"
+
     if ENABLE_AUTH_STRING not in st.session_state:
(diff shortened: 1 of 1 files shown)


mrsabath (approved):
This would be very useful! 
LGTM

PR #371, opened 2025-10-31
❇️ : Add image pull secret input field for deploying from existing images

### LINKED ISSUE #370: bug: When Deploying Agent from Image the Form Does Not Ask For Remote Repo Secret
When trying to deploy an agent/mcp server from an image which is in a remote private repo, the form should ask for a secret name. Currently the form only has an image URL input field. 

Image

cwiklik:
## Summary

This PR adds a new input field to the Kagenti UI that allows users to specify a custom image pull secret name when deploying Agents or MCP Servers from existing container images. The specified secret name is then added to the Component CR's `imagePullSecrets` field.

**Example Component CR output:**

```yaml
spec:
  deployer:
    kubernetes:
      imageSpec:
        image: myrepo/myimage
        imageTag: v1.0.0
        imageRegistry: docker.io
        imagePullSecrets:
          - name: my-custom-secret
```

### Motivation

Previously, users could only specify image pull secrets when building from source via `registry_config`. This PR extends that capability to deployments from existing images, allowing users to:
- Deploy from private Docker registries (Quay, Docker Hub, GHCR , etc.)
- Use custom secret names instead of hardcoded values
- Deploy both public and private images with the same workflow
- 
## Related issue(s)

Depends on: https://github.com/kagenti/kagenti-operator/pull/124
Fixes #370

### DIFF
--- a/kagenti/ui/lib/build_utils.py
+++ b/kagenti/ui/lib/build_utils.py
@@ -1904,16 +1904,16 @@ def render_import_form(
         # *** The Kagenti installer will only copy the necessary configuration (like ConfigMaps and Secrets) for those specific
         #     namespaces.
         st_object.write(
-            f"Provide Docker image details to deploy a new {resource_type.lower()}."
+            f"Provide container image details to deploy a new {resource_type.lower()}."
         )
         docker_image_url = st_object.text_input(
-            "Docker Image (e.g., myrepo/myimage:tag)",
+            "Container Image (e.g., myrepo/myimage:tag)",
             key=f"{resource_type.lower()}_docker_image",
         )
         repo_secret_name = st_object.text_input(
             "Image Pull Secret Name (optional, leave empty for public images)",
             key=f"{resource_type.lower()}_repo_secret",
-            help="Name of the Kubernetes secret containing credentials for private Docker registries",
+            help="Name of the Kubernetes secret containing credentials for private container registries",
         )
         selected_protocol = default_protocol
         if protocol_options:


kagenti/ui/lib/build_utils.py
```diff
@@ -722,6 +734,30 @@ def _construct_agent_resource_body(
     return body
 
 
+def _get_image_pull_secrets(build_from_source, registry_config, image_pull_secret):
+    """
+    Determine imagePullSecrets based on deployment method.
+
+    Args:
+        build_from_source: Boolean indicating if building from source
+        registry_config: Registry configuration dict (used when build_from_source=True)
```
esnible, line 743: If the user is deploying a container image (no build_from_source), they may also need to credentials secret.  For example, if the image is `us.icr.io/research3/snible/mcp-everything:v0.0.1` I need to use a secret.

kagenti/ui/lib/build_utils.py
```diff
@@ -1871,7 +1910,11 @@ def load_all_configmap_vars():
             "Docker Image (e.g., myrepo/myimage:tag)",
             key=f"{resource_type.lower()}_docker_image",
         )
-
+        repo_secret_name = st_object.text_input(
+            "Image Pull Secret Name (optional, leave empty for public images)",
+            key=f"{resource_type.lower()}_repo_secret",
+            help="Name of the Kubernetes secret containing credentials for private Docker registries",
```
esnible:
```suggestion
            help="Name of the Kubernetes secret containing credentials for private container registries",
```

kagenti/ui/lib/build_utils.py
cwiklik: Indeed and this PR fixes it. Here is the screenshot:

image

kagenti/ui/lib/build_utils.py
cwiklik: This PR depends on changes in the platform-operator Component CR API. There is a related PR https://github.com/kagenti/kagenti-operator/pull/124 which should be merged first.

cwiklik: I've addressed the "Docker" issues mentioned above in a new commit.

Issue #380, opened 2025-11-04
feature: Workload Identity Abstraction


usize:
### Feature Description

  Kagenti currently hardcodes SPIRE/SPIFFE for workload identity, particularly in:
  - Client registration init containers (kagenti/auth/client-registration/) that read SPIRE JWT tokens
  - Infrastructure deployment via the installer that assumes SPIRE is available
  - Agent/Tool manifests that hardcode SPIFFE volumes and CSI drivers

This prevents deployment on platforms like older versions of OpenShift where SPIRE components may be unavailable. We need an abstraction that allows identity providers to be swapped without changing agent/tool code.``

### Proposed Solution

  Create a WorkloadIdentity abstraction that:

  1. Provides a consistent interface for accessing workload identity regardless of provider
  2. Supports SPIRE/SPIFFE as the default implementation
  3. Supports Kubernetes ServiceAccounts as a minimalistic fallback
  4. Can be configured at installation time

Interfaces might look like:

Python (for agents, tools, client-registration)

```
  # kagenti/kagenti/identity/provider.py

  from abc import ABC, abstractmethod
  from typing import Dict, Any

  class WorkloadIdentity(ABC):
      """Represents a workload's identity"""

      @abstractmethod
      def get_subject(self) -> str:
          """Returns the identity subject (e.g., SPIFFE ID or SA name)"""
          pass

      @abstractmethod
      def get_token(self) -> str:
          """Returns a token representing this identity"""
          pass

      @abstractmethod
      def get_claims(self) -> Dict[str, Any]:
          """Returns identity claims as a dictionary"""
          pass


  class IdentityProvider(ABC):
      """Manages workload identity provisioning"""

      @abstractmethod
      def get_name(self) -> str:
          """Returns provider name (e.g., 'spire', 'serviceaccount')"""
          pass

      @abstractmethod
      def get_current_identity(self) -> WorkloadIdentity:
          """Retrieves identity for the current workload"""
          pass

      @abstractmethod
      def get_manifest_patches(self, component_name: str) -> Dict[str, Any]:
          """Returns Kubernetes manifest patches to inject identity into a workload"""
          pass
```

  Go (for operators)

Note: need to talk to @cwiklik since he is working on major changes regarding the platform-operator where this would currently be applicable.

```
  // pkg/identity/provider.go

  type WorkloadIdentity interface {
      GetSubject() string
      GetToken(ctx context.Context) (string, error)
[...]

PR #395, opened 2025-11-14
feat: Add identity provider abstraction for SPIRE and ServiceAccount

Shreyanand:
## Identity Provider Abstraction

This PR introduces a pluggable identity provider abstraction layer that supports multiple workload identity providers for Kagenti components. Implements #380.

### Changes

- Added identity provider abstraction module (`kagenti/identity/`) with support for:
  - SPIRE/SPIFFE provider (default)
  - Kubernetes ServiceAccount provider
- Updated client registration to use the abstraction instead of hardcoded SPIRE logic
- Refactored installer to conditionally install SPIRE based on provider selection
- Updated Helm charts to configure identity provider via `identity.provider` value
- Removed auto-detection logic - provider must be explicitly configured

### Breaking Changes

- `KAGENTI_IDENTITY_PROVIDER` environment variable is now required
- Valid values: `spire` or `serviceaccount`
- Default Helm value is `spire`

### Configuration

Set identity provider via Helm values:
```yaml
identity:
  provider: "spire"  # or "serviceaccount"
```

Or via environment variable:
```bash
export KAGENTI_IDENTITY_PROVIDER=spire
```
### Tests
- [X] Test service account client registration on Kind
- [X] Test Spire client registration on Kind
- [X] Test service account client registration on OpenShift
- [ ] Test Spire client registration on OpenShift

### DIFF
--- a/kagenti/installer/app/cli.py
+++ b/kagenti/installer/app/cli.py
@@ -51,7 +51,7 @@ INSTALLERS = {
     InstallableComponent.CERT_MANAGER: cert_manager.install,
     InstallableComponent.OPERATOR: operator.install,
     InstallableComponent.ISTIO: istio.install,
-    InstallableComponent.SPIRE: identity.install,
+    InstallableComponent.IDENTITY: identity.install,
     InstallableComponent.MCP_GATEWAY: mcp_gateway.install,
     InstallableComponent.ADDONS: addons.install,
     InstallableComponent.UI: ui.install,
@@ -203,7 +203,9 @@ def main(
             deploy_component(
                 InstallableComponent.KEYCLOAK, skip_install, **install_params
             )
-            deploy_component(InstallableComponent.SPIRE, skip_install, **install_params)
+            deploy_component(
+                InstallableComponent.IDENTITY, skip_install, **install_params
+            )
             deploy_component(
                 InstallableComponent.MCP_GATEWAY, skip_install, **install_params
             )
--- a/charts/kagenti/values.yaml
+++ b/charts/kagenti/values.yaml
@@ -94,14 +94,6 @@ kagenti-platform-operator-chart:
           memory: 512Mi
 
 # ------------------------------------------------------------------
-#  Identity Provider Configuration
-# ------------------------------------------------------------------
-identity:
-  provider: "spire"  # spire or serviceaccount (required)
-  # spire: Use SPIRE provider (requires SPIRE to be installed) - DEFAULT
-  # serviceaccount: Use Kubernetes ServiceAccount provider (cluster-local only)
-
-# ------------------------------------------------------------------
 #  MCP Gateway Configuration
 # ------------------------------------------------------------------
 mcpGateway:
--- a/deployments/ansible/README.md
+++ b/deployments/ansible/README.md
@@ -54,6 +54,48 @@ Important variables you can override (via `-e` / `--extra-vars`):
 - `kind_cluster_name`, `kind_images_preload`, `container_engine`, `kind_config`,
    `kind_config_registry`, `preload_images_file` - Kind-related knobs (see `default_values.yaml`).
 
+## Identity Provider Configuration
+
+Kagenti supports multiple workload identity providers for authentication and authorization.
+The identity provider is configured via Helm chart values under `identity.provider`:
+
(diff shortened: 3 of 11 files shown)


kagenti/installer/app/cli.py
```diff
@@ -51,7 +51,7 @@
     InstallableComponent.CERT_MANAGER: cert_manager.install,
     InstallableComponent.OPERATOR: operator.install,
     InstallableComponent.ISTIO: istio.install,
-    InstallableComponent.SPIRE: spire.install,
+    InstallableComponent.SPIRE: identity.install,
```
usize: I think it's a little confusing that the InstallableComponent.SPIRE is now set to `identity.install` can we denote that either the installable component is simply IDENTITY _or_ can we make something like identity.spire.install ?

charts/kagenti-deps/templates/_helpers.tpl
```diff
@@ -72,3 +72,12 @@ It will be enabled if:
 {{- tpl "{{ and .Values.components.istio.enabled (not .Values.openshift) }}" . | toString -}}
 {{- end -}}
 
+{{/*
+Determines if SPIRE should be installed based on identity provider configuration.
+SPIRE will be installed if:
+  - Identity provider is explicitly set to "spire"
+*/}}
+{{- define "kagenti.identity.shouldInstallSpire" -}}
+{{- eq .Values.identity.provider "spire" }}
```
usize, line 81: I think we need to update `deployments/ansible/default_values.yaml` as well

with something like

```yaml
identity:
  - provider: "spire"
 ```

File: charts/kagenti-deps/templates/spire-operator.yaml [hunk 3]
```diff
@@ -1,4 +1,4 @@
-{{- if and .Values.components.spire.enabled .Values.openshift }}
+{{- if and (include "kagenti.identity.shouldInstallSpire" .) .Values.components.spire.enabled .Values.openshift }}
```
usize, line 1: There's also a spire installer for Ansible now at `deployments/ansible/roles/kagenti_installer/tasks/main.yml` that I think needs to be changed.

usize (changes requested):
This looks really super. The main thing I think we're missing is integration with the Ansible installer, since I think that's becoming the new source of truth.

We should update the Ansible installer readme, make sure we make it easy to support the various options in its defaults and also the environment specific vars files it has at: `deployments/envs` 

Then finally, it would be pretty awesome to just see a quick run through of a kind deployment with the serviceaccount provider.

Shreyanand:
> This looks really super. The main thing I think we're missing is integration with the Ansible installer, since I think that's becoming the new source of truth.
> 
> We should update the Ansible installer readme, make sure we make it easy to support the various options in its defaults and also the environment specific vars files it has at: `deployments/envs`
> 
> Then finally, it would be pretty awesome to just see a quick run through of a kind deployment with the serviceaccount provider.

@usize I have tested this with the original installer script and I can add a short demo video showing the installation and the running agent with serviceaccount provider. Is that what you meant? 
ftr, I have added the identity values for the latest Ansible files but I have not tested this installer.

charts/kagenti-deps/templates/spire-operator.yaml
Shreyanand: Ah, you're right, this will have to be changed from charts.spire to charts.identity.spire or something. Would it make sense to add it as a follow up PR with tests for the Ansible installer script?

usize:
> > This looks really super. The main thing I think we're missing is integration with the Ansible installer, since I think that's becoming the new source of truth.
> > We should update the Ansible installer readme, make sure we make it easy to support the various options in its defaults and also the environment specific vars files it has at: `deployments/envs`
> > Then finally, it would be pretty awesome to just see a quick run through of a kind deployment with the serviceaccount provider.
> 
> @usize I have tested this with the original installer script and I can add a short demo video showing the installation and the running agent with serviceaccount provider. Is that what you meant? ftr, I have added the identity values for the latest Ansible files but I have not tested this installer.

No need for a video. For posterity's sake, we could maybe copy the console output to a gist and leave it in a comment here. That may be useful in case we come across some unexpected behavior later.

Let's make sure the Ansible installer works, ensure that client registration is working then we could look at the copilot suggestions and run it by one more reviewer before we land.

PR #399, opened 2025-11-17
✨ Keycloak feature flags for installation and configuration

### LINKED ISSUE #335: feature: Identity - Enable Feature Flag
### Feature Description

Currently we check if "KEYCLOAK_URL" environment is set to enable or skip the Keycloak interaction. 

### Proposed Solution

This should be replaced by series of Feature Flags. 
e.g.: 
INSTALL_KEYCLOAK = true
REGISTER_CLIENT = true
ENABLE_TOKEN_EXCHANGE = true 
etc

### Want to contribute?

### Additional Context

_No response_

d0w:
## Summary
Seeks to add flags for keycloak that can be set when creating a new agent or tool. There are two new flags that are predefined in the `environments` ConfigMap that should be imported by the client-registration container to determine keycloak configuration
- `KEYCLOAK_TOKEN_EXCHANGE_ENABLED`
- `KEYCLOAK_CLIENT_REGISTRATION_ENABLED`

The client-registration image checks if `KEYCLOAK_CLIENT_REGISTRATION_ENABLED` is false before creating a client for a tool/agent. It does the same for `KEYCLOAK_TOKEN_EXCHANGE_ENABLED:
```python
  "attributes": {
      "standard.token.exchange.enabled": KEYCLOAK_TOKEN_EXCHANGE_ENABLED,  # Enable token exchange
  },

Both of these env variables are defaulted to true.
```

### Extra notes
This change requires changes in the kagenti [platform-operator](https://github.com/kagenti/kagenti-operator/blob/437f07ab28f285b9b2bd751835e2789541026471/platform-operator/internal/deployer/kubernetes/kubernetes.go#L352C6-L352C27). Namely we would need to add the flags defined above as environment variables.

This PR does not add a flag for determining whether or not to install Keycloak since the installer already has this functionality.
## Related issue(s)

Fixes #335 
Related to: https://github.com/kagenti/kagenti-operator/pull/136


mrsabath:
Thank you @d0w for this submission. I think it would be better to reverse the logic and have the variable as:
`KEYCLOAK_TOKEN_EXCHANGE_ENABLED` instead of `KEYCLOAK_DISABLE_TOKEN_EXCHANGE`

d0w:
> Thank you @d0w for this submission. I think it would be better to reverse the logic and have the variable as: `KEYCLOAK_TOKEN_EXCHANGE_ENABLED` instead of `KEYCLOAK_DISABLE_TOKEN_EXCHANGE`

@mrsabath I initially put in the disabled variables in order to keep backwards compatability with the existing operator. Since having the "*_ENABLED" flags implies that if the vars are not present to disable (we could also make the default value in the python file `True`, this would make it so deployments with the existing operator wont allow keycloak to work. 

I'll change the variables but please let me know if you want to make the default value "True" in the client registration script and then change the operator later or first merge the changes into the operator then merge this PR.

d0w:
@mrsabath There shouldn't be any linting errors though I'm not sure why the PR deployment is failing. Looking at the logs, its failing when installing the istio helm charts since there is an unrecognized "--all" flag. I'm not sure if these changes affected that but please advise if they did.

PR #442, opened 2025-12-05
🐛 fix broken helm chart

pdettori:
## Summary

The Kagenti Helm Chart is declaring as dependencies BOTH the platform operator and the agent operator, which have conflicting objects resulting in install errors.


ingpaolodettori-dev (approved):
/lgtm

PR #443, opened 2025-12-05
change from mcp.test.com to mcp.local to match the default of gateway…

huang195:
To address: https://github.com/kagenti/mcp-gateway/issues/418, so we have consistency between script install and ansible install of mcp gateway helmchart


rubambiza (approved):
LGTM

PR #448, opened 2025-12-05
🐛 allow to set correct value for spiffe trustDomain

pdettori:
## Summary

 allow to set correct value for spiffe trustDomain

## Related issue(s)

Fixes #


mrsabath (approved):
LGTM

PR #469, opened 2025-12-12
Add barebones installer environment for minimal kagenti-operator experiments.

usize:
Here I've added a barebones option to the Ansible installer. Why?

We've had a number of users ask for "light weight" deployments, with only an Operator. 

These users would like the basic discovery capabilities and simple lifecycle management.

This install pattern enables that.

It's also useful for Kagenti developers because it allows us to uncover hidden dependencies.

It's also a good opportunity to explore creating an API server with its own AuthProvider interface, so that the UI can work even if fundamental components like KeyCloak are disabled or swapped with new components.

### DIFF
--- a/deployments/envs/barebones_operator_values.yaml
+++ b/deployments/envs/barebones_operator_values.yaml
@@ -1,0 +1,79 @@
+# Barebones Kagenti installation - Operator (CRDs + Controller) only
+# Minimal dependencies for kagenti-operator with Agent and AgentCard CRDs and controllers
+# No build pipelines - use pre-built images with Agent resources directly
+
+####################################################################
+# Minimal kubectl-applied dependencies
+####################################################################
+gatewayApi:
+  enabled: false
+
+# cert-manager is required for operator webhook certificates
+certManager:
+  enabled: true
+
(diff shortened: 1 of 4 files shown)


deployments/envs/barebones_operator_values.yaml
```diff
@@ -0,0 +1,79 @@
...
+  ####################################################################
+  spire:
+    enabled: false
+
+####################################################################
+# Kind cluster settings
+####################################################################
+kind_images_preload: false
+# docker or podman
+container_engine: docker
```
pdettori, line 79: The idea was to set all false in https://github.com/kagenti/kagenti/blob/main/deployments/ansible/default_values.yaml and just opt-in what we want to install in the different configs - I see now that this convention was not followed for tool hive in any case I suggest to turn it off there and then only show what is enabled here (need also to turn on tool hive if needed on the other envs using it)

pdettori:
@usize I think it's a useful install profile to have - we should test this after the epic in #523 is completed.

mrsabath:
@usize what should we do with this request?

PR #476, opened 2025-12-15
feat: Make container engine configurable via environment variable

rhuss:
## Summary
This PR makes the container engine (Docker or Podman) configurable via the `CONTAINER_ENGINE` environment variable in the `.env` file, instead of hardcoding it in `config.py`.

## Changes
- Updated `installer/app/config.py` to read `CONTAINER_ENGINE` from environment with "docker" as the default fallback
- Added `CONTAINER_ENGINE=docker` to `.env_template` for documentation

## Benefits
- Users can easily switch between Docker and Podman by just changing the `.env` file
- No need to modify source code to change container engine preference
- Maintains backward compatibility with default to "docker"

## Testing
Tested locally by setting `CONTAINER_ENGINE=podman` in `.env` file.

### DIFF
--- a/kagenti/installer/app/config.py
+++ b/kagenti/installer/app/config.py
@@ -17,11 +17,17 @@ from enum import Enum
 from pathlib import Path
 import os
 
+from dotenv import load_dotenv
+
 # --- Core Paths ---
 SCRIPT_DIR = Path(__file__).parent.resolve()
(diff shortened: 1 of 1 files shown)


pdettori:
@rhuss thank you for your PR. Could you please amend you commit to pass the DCO ?

pdettori (approved):
/lgtm

PR #502, opened 2025-12-22
🌱 adjust resource limits for kagenti helm charts

pdettori:
## Summary

adjust resource limits for the kagenti helm charts

This PR addresses issues of scheduler not being able to place Kagenti components when running on a dev cluster with limited resources.


mrsabath (approved):
LGTM

PR #507, opened 2025-12-30
:book: Add AuthBridge Component details

mrsabath:
## Summary
Add AuthBridge component overview, installation, and other details. 
## Related issue(s)

Fixes #

### DIFF
--- a/docs/identity-guide.md
+++ b/docs/identity-guide.md
@@ -630,7 +630,7 @@ The [AuthBridge Component](https://github.com/kagenti/kagenti-extensions/tree/ma
 │  3. Caller gets token from Keycloak (audience: "caller's SPIFFE ID")            │
 │  4. Caller sends request to auth-target with token                              │
 │  5. Envoy intercepts,Go Processor validates Token Signature and Isusuer,        │
-│    then it exchanges token (audience: "auth-target")                           │
+│    then it exchanges token (audience: "auth-target")                            │
 │  6. Auth Target validates token and returns "authorized"                        │
 └─────────────────────────────────────────────────────────────────────────────────┘
 


docs/components.md
```diff
@@ -258,8 +258,57 @@ kubectl get route kagenti-ui -n kagenti-system -o jsonpath='{.status.ingress[0].
...
+**Auth Bridge** solves a critical challenge in microservices and agentic architectures: **how can workloads authenticate and communicate securely without pre-provisioned static credentials?**
+
+### Auth Bridge Components
+
+| Component | Purpose | Repository |
+|-----------|---------|------------|
+| **[Client Registration](https://github.com/kagenti/kagenti-extensions/tree/main/AuthBridge/client-registration)** | Automatic OAuth2/OIDC client provisioning using SPIFFE ID | `AuthBridge/client-registration` |
+| **[AuthProxy](https://github.com/kagenti/kagenti-extensions/tree/main/AuthBridge/AuthProxy)** | JWT validation and transparent token exchange | `AuthBridge/AuthProxy` |
+| **SPIRE** | Workload identity and attestation | External |
+| **Keycloak** | Identity provider and access management | External |
```
evaline-ju: I think it may be helpful to have links for these as resources even if the repos are not owned through kagenti

docs/components.md
```diff
@@ -279,8 +328,8 @@ Kagenti provides a unified framework for identity and authorization in agentic s
 | Feature | Description |
 |---------|-------------|
 | **User Management** | Create and manage Kagenti users |
-| **Client Registration** | OAuth clients for agents and UI |
-| **Token Exchange** | Exchange SPIRE tokens for Keycloak tokens |
+| **Client Registration** | OAuth clients for agents and UI (automated via Client Registration) |
```
evaline-ju: The addition seems confusing on its own, with "client registration...automated via client registration" - perhaps a link would help for the second reference.

docs/components.md
```diff
@@ -304,7 +353,28 @@ The Agent and Tool Authorization Pattern replaces static credentials with dynami
...
+
+```
+┌─────────────────────────────────────────────────────────────────────────────────┐
+│  1. SPIFFE Helper obtains SVID from SPIRE Agent                                 │
+│  2. Client Registration registers workload with Keycloak (using SPIFFE ID)      │
+│  3. Caller gets token from Keycloak (audience: "authproxy")                     │
+│  4. Caller sends request to target with token                                   │
+│  5. Envoy+GoPro AuthProxy (Go Processor) intercepts, exchanges token            │
+│     (audience: "auth-target")                                                   │
+└─────────────────────────────────────────────────────────────────────────────────┘
```
evaline-ju: q - is this the same as in the "AuthBridge Architecture" of the identity guide but without the last step? Mainly concerned if partial duplication can make it more confusing / difficult to maintain in two parts

evaline-ju (approved):
LGTM with a few suggestions - disclaimer that I haven't tried out the commands themselves :)

docs/components.md
mrsabath: Good point. I removed it.

docs/components.md
mrsabath: They are just below, line 315 and 323 but I added them here too. Thank you.

PR #521, opened 2026-01-14
🐛 fix tagged build in CI

pdettori:
## Summary

Remove legacy UI build from CI which was causing a build failure for tagged builds.


mrsabath (approved):
LGTM

PR #531, opened 2026-01-20
:sparkles: Design proposal for compositional agent platform architecture

### LINKED ISSUE #560: Design composition APIs for TokenExchange, AgentCard and AgentTrace
Design composition APIs for:
- TokenExchange (Identity & Delegated Authorization)
- AgentCard (Discoverability & Connectivity)
- AgentTrace (Observability & Tracing)

cwiklik:
### Summary
This PR introduces the design proposal for refactoring Kagenti from an inheritance-based architecture to a composition-based architecture. The proposal replaces the monolithic Agent Custom Resource with a two-layer system consisting of automatic identity infrastructure injection via mutating webhook and optional configuration through three independent pillar CRs.
### Motivation
The current Kagenti architecture uses the inheritance pattern (embedding PodSpec in a custom Agent CR), which has proven problematic:

- Tight coupling to upstream Kubernetes changes requiring constant rebasing
- Limited workload type support (only Deployments, no StatefulSets/Jobs/CronJobs)
- CRD size issues with PodSpec validation (~700KB)
- Migration overhead for users with existing Deployments
- Poor compatibility with standard Kubernetes tooling

Historical analysis shows that higher-level Kubernetes abstractions that replace standard workload types consistently fail (Knative, OAM, Appsody, DeploymentConfig), while composition-based approaches succeed (KEDA, Flagger, Prometheus Operator, cert-manager).

### Proposed Changes
Two-Layer Architecture:

Layer 1: Automatic Identity Infrastructure (Webhook)

- Mutating webhook injects identity sidecars on workloads with kagenti.io/inject: enabled label
- Provides SPIFFE identity, IdP registration, inbound auth validation, outbound token exchange
- Works immediately with sensible defaults—no CRs required

Layer 2: Optional Configuration & Discovery (Pillar CRs)

- TokenExchange - Configure identity/authorization settings
- AgentTrace - Configure telemetry and observability
- AgentCard - Enable agent capability discovery

Key Benefits:

- Users keep standard Kubernetes workloads (Deployment, StatefulSet, Job, CronJob)
- Each pillar can be adopted independently
- Configuration changes without pod restarts
- Easy removal—workloads continue running if Kagenti is uninstalled
- Full compatibility with existing tools (kubectl, GitOps, Helm)

## Related issue(s): 
https://github.com/kagenti/kagenti-extensions/issues/97

Fixes #560

### DIFF
--- a/docs/design-proposals/compositional-agent-platform-design.md
+++ b/docs/design-proposals/compositional-agent-platform-design.md
@@ -429,57 +429,92 @@ status:
 - Provide discovery API for other agents
 
 **API Structure**:
-
 ```yaml
 apiVersion: kagenti.io/v1alpha1
 kind: AgentCard
 metadata:
   name: weather-agent-card
   namespace: default
+  labels:
+    app.kubernetes.io/name: weather-agent-card
+    app.kubernetes.io/component: discovery
 spec:
+  syncPeriod: "30s"
   selector:
     matchLabels:
       app: weather-agent
-  
-  endpoint:
-    path: "/.well-known/agent.json"
-    port: 8081
-    scheme: http
-  
-  syncPeriod: 30s
-  
-  metadata:
-    name: "Weather Intelligence Agent"
-    description: "Production weather forecasting service"
-    tags:
-    - weather
-    - forecast
-    - production
-
+      kagenti.io/type: agent
 status:
-  phase: Active
-  lastSyncTime: "2026-01-13T16:10:00Z"
-  discoveredPods: 3
-  syncErrors: 0
-  cards:
-  - podName: weather-agent-7d5f8b9c4d-abc12
-    podIP: 10.244.0.10
-    url: "http://10.244.0.10:8081"
-    lastFetchTime: "2026-01-13T16:10:00Z"
-    fetchStatus: Success
-    card:
-      name: "Weather Intelligence Agent"
-      version: "2.1.0"
-      url: "http://weather-agent.default.svc.cluster.local"
-      capabilities:
-        streaming: true
-        batchProcessing: true
-      skills:
-      - name: "get_forecast"
-        description: "Get weather forecast for location"
-      - name: "get_climate_data"
-        description: "Get historical climate data"
+  protocol: "a2a"
+  lastSyncTime: "2026-01-21T10:30:00Z"
+  conditions:
+  - type: Synced
+    status: "True"
+    lastTransitionTime: "2026-01-21T10:30:00Z"
+    reason: SyncSuccess
+    message: "Agent card successfully fetched"
+  card:
+    name: "Weather Intelligence Agent"
+    description: "Provides weather forecasts and current conditions for locations worldwide"
+    version: "2.1.0"
+    url: "http://weather-agent.default.svc.cluster.local:8080"
+    capabilities:
+      streaming: true
+      pushNotifications: false
+    defaultInputModes:
+      - "application/json"
+      - "text/plain"
+    defaultOutputModes:
+      - "application/json"
+    skills:
+    - name: "get_forecast"
+      description: "Get weather forecast for a location"
+      inputModes:
+        - "application/json"
+      outputModes:
+        - "application/json"
+      parameters:
+      - name: "location"
+        type: "string"
+        description: "City name or coordinates (lat,lon)"
+        required: true
+      - name: "days"
+        type: "number"
+        description: "Number of days to forecast (1-14)"
+        required: false
+        default: "7"
+    - name: "get_current_conditions"
+      description: "Get current weather conditions for a location"
+      inputModes:
+        - "application/json"
+      outputModes:
+        - "application/json"
+      parameters:
+      - name: "location"
+        type: "string"
+        description: "City name or coordinates (lat,lon)"
+        required: true
+    - name: "get_climate_data"
+      description: "Get historical climate data for analysis"
+      inputModes:
+        - "application/json"
+      outputModes:
+        - "application/json"
+      parameters:
+      - name: "location"
+        type: "string"
+        description: "City name or coordinates"
+        required: true
+      - name: "start_date"
+        type: "string"
+        description: "Start date (YYYY-MM-DD)"
+        required: true
+      - name: "end_date"
+        type: "string"
+        description: "End date (YYYY-MM-DD)"
+        required: true
 ```
+
 
 **Adaptation Needed**: Change selector from Agent-specific to generic workload selector to work with any workload type via label matching.
 


docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+
+1. **Tight coupling to upstream Kubernetes changes** - Requires constant rebasing on K8s releases without substantial benefits
+2. **Responsibility for pass-through definitions** - Even when passing through PodSpec, Kagenti becomes responsible when something goes wrong in those definitions
+3. **Limited workload type support** - Only works with Deployments; cannot handle StatefulSets (essential for stateful agents with long-term memory), Jobs (ideal for one-shot agent runs), or CronJobs (scheduled agent executions)
+4. **CRD size issues** - Teams have found PodSpec validation can blow CRDs to ~700KB, hitting client-side apply limits
+5. **Migration overhead** - Users must convert existing Deployments to new CRD format
+6. **Missing integration** - Standard K8s tools/controllers may not recognize the new CRD
+
+Red Hat's existing portfolio already addresses many needs of agentic workloads: service meshes handle inter-service communication, secret managers protect credentials, and observability stacks capture metrics and traces. Introducing a monolithic agent framework would duplicate this existing functionality, increase coupling between components, and add conceptual overhead.
```
pdettori, line 79: I would generalize this - e.g. "CNCF existing ecosystem already addresses ..."

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+
+### Impact and Desired Outcome
+
+The compositional refactoring will:
+
+1. **Reduce adoption friction** - Users keep their existing Deployments, StatefulSets, and Jobs
+2. **Enable incremental adoption** - Each pillar can be adopted independently
+3. **Improve tool compatibility** - kubectl, dashboards, and GitOps continue working normally
+4. **Support multiple workload types** - StatefulSets for stateful agents, Jobs for one-shot runs, CronJobs for scheduled executions
+5. **Simplify removal** - If disabled/uninstalled, core workloads continue running
```
pdettori, line 89: Maybe we could say "minimal impact from removal on workloads"

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+---
+
+## Goals
+
+1. **Compose with existing Kubernetes workload types** - Never require users to abandon Deployment, StatefulSet, or Job
+2. **Avoid introducing new mandatory workload abstractions** - Factor out agent-specific concerns into separate, optional components
+3. **Keep surface area small** - Focus on integration points, not comprehensive workload management
+4. **Provide non-intrusive injection** - Automatic identity infrastructure via opt-in webhook
+5. **Enable independent pillar adoption** - Each CR can be used separately
+6. **Support dynamic reconfiguration** - Configuration changes without pod restarts
```
pdettori, line 119: I am not sure we can remove/update containers without a pod restart - that was one of the key issues that motivated the introduction of Istio Ambient, where the biggest advantage is that it can be added or removed or upgraded without any workload restart. The current design with sidecars cannot claim this goal.

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+7. **Maintain backward compatibility** - Provide migration path from Agent CR
+
+---
+
+## Non-Goals
+
+1. **Building another workload orchestrator** - Users keep their existing orchestration tools
+2. **Introducing mandatory cross-dependencies between pillars** - Optional integration is acceptable, but no pillar requires another
+3. **Replacing established deployment tools** - Helm, GitOps, OLM, and Terraform remain valid choices
+4. **Duplicating existing Red Hat portfolio functionality** - Secret managers, service meshes, and observability stacks continue to be used
```
pdettori, line 129: Again, I would make this more general "Estabilshed CNCF ecosystem projects functionality..."

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+└─────────────────────────────────────────────────────────┘
+                           ↓
+┌─────────────────────────────────────────────────────────┐
+│ LAYER 2: Optional Configuration & Discovery            │
+│─────────────────────────────────────────────────────────│
+│ • TokenExchange: Configure identity/auth sidecars       │
+│ • AgentTrace: Configure agent telemetry                 │
+│ • AgentCard: Discover agent capabilities                │
+│ • Controllers create ConfigMaps or cache data           │
+│ • No pod restarts needed                                │
```
pdettori, line 170: See previous point regarding pod restarts

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+- Injects identity and auth sidecars with sensible defaults
+- User gets working agent immediately
+- No CRDs required for basic functionality
+
+**Layer 2: Optional Configuration & Discovery**
+- `TokenExchange`: Configure identity/auth sidecars
+- `AgentTrace`: Configure agent telemetry
+- `AgentCard`: Discover agent capabilities
+- Controllers create ConfigMaps or cache data
+- No pod restarts needed for configuration changes
```
pdettori, line 185: This may require exploring a design based on or along the lines of envoy + XDS for dynamic config

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+| Identity | Delegated Authorization | `TokenExchange` | Provide workload identity and enable delegated authorization via RFC 8693 token exchange |
+| Discovery | Agent Registry | `AgentCard` | Enable rich agent discovery—knowing which agents are running, what they can do, and how to connect |
+| Observability | Tracing | `AgentTrace` | Make agent behavior observable end-to-end—tool call graphs, inter-agent communication, LLM I/O |
+
+### Success Metrics
+
+1. **Adoption Rate**: Percentage of agent workloads using composition pattern vs. legacy Agent CR
+2. **Time to First Agent**: Time from deployment to working agent with identity infrastructure
+3. **Configuration Change Latency**: Time from CR update to sidecar reconfiguration (target: <10s)
+4. **Removal Impact**: Zero workload disruption when Kagenti is removed
```
pdettori, line 200: 4. may be problematic to achieve with a sidecar-based architecture.

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+  timeoutSeconds: 10
+  failurePolicy: Fail
+  reinvocationPolicy: Never
+```
+
+**Key Design Decisions**:
+
+- **Workload-level targeting**: Webhook watches Deployment/StatefulSet/Job, not Pods, enabling deterministic ConfigMap naming based on workload name
+- **Explicit opt-in**: Each workload must have `kagenti.io/inject: enabled` and `kagenti.io/spiffe: enabled` labels to prevent accidental injection
+- **Fail-closed policy**: If webhook unavailable, workload creation blocks, ensuring agents never run without required identity infrastructure
```
pdettori, line 251: Not sure how to achieve this if the policy enforcement happens in the webhook itself ...

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+   • envoy-proxy: exchanges outbound tokens
+   • Agent container: ready to handle requests
+```
+
+#### Default Configuration
+
+All identity infrastructure components start with sensible defaults:
+
+**SPIFFE Helper Defaults**:
+- Trust domain: `cluster.local`
```
pdettori, line 291: this depends on the particular cluster configuration and should be configurable - for example on OCP we use 

```
export DOMAIN=apps.$(kubectl get dns cluster -o jsonpath='{ .spec.baseDomain }')
```

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+- Audience: `kagenti-agents`
+
+**Client Registration Defaults**:
+- Provider: Keycloak
+- IdP URL: `http://keycloak.kagenti-system.svc:8080`
+- Realm: `default`
+- Token exchange enabled: `true`
+
+**Auth Proxy Defaults**:
+- Inbound port: `8080`
```
pdettori, line 303: This will also depend on the agent config... should be configurable, initially at platform level then at agent level

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+**API Structure**:
+
+```yaml
+apiVersion: kagenti.io/v1alpha1
+kind: TokenExchange
+metadata:
+  name: weather-agent-auth
+  namespace: default
+spec:
+  targetRef:
```
pdettori, line 336: Can we apply the same config to more than one agent ? `targetRef` assumes 1:1 mapping, but we could re-use this config on multiple agents, thus a label selector may be a better choice here - e.g.:

```yaml
 selector:
    matchLabels:
      app: my-app
```

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+    inboundProxyReady: true
+    outboundProxyReady: true
+```
+
+**Controller Behavior**:
+1. Watches TokenExchange CRs for create/update/delete
+2. Resolves targetRef to find workload
+3. Creates ConfigMap: `{workload-name}-token-exchange`
+4. ConfigMap contains configuration for all identity sidecars
+5. Identity sidecars watch ConfigMap via fsnotify
```
pdettori, line 412: I see - this is a different approach than envoy + XDS - consider also that as design option as it is well established in this space.

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+kind: AgentCard
+metadata:
+  name: weather-agent-card
+  namespace: default
+  labels:
+    app.kubernetes.io/name: weather-agent-card
+    app.kubernetes.io/component: discovery
+spec:
+  syncPeriod: "30s"
+  selector:
```
pdettori, line 443: If AgentCard maps 1:1 with an agent this is a case where we want to use a `targetRef` instead.

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+        - "downstream:access"
+      destinationRules:
+      - match:
+          host: "premium-api.external.com"
+        target:
+          audience: "premium-api"
+          scopes:
+          - "weather:premium"
+          - "weather:historical"
+
```
pdettori, line 394: Scopes and audiences make a lot of sense here - but I am wondering if low level infrastructure details (which hopefully are the same for all agents) should really be here - e.g. `tokenUrl`, `keycloakUrl` etc.

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+**API Structure**:
+
+```yaml
+apiVersion: kagenti.io/v1alpha1
+kind: AgentTrace
+metadata:
+  name: weather-agent-trace
+  namespace: default
+spec:
+  targetRef:
```
pdettori, line 543: Maybe a label selector here too ? I suppose these configs can be shared across multiple workloads

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+- CPU: <200m
+- Single pod for simplified operations
+
+#### Duck-Typing Pattern for Workload References
+
+Following Knative's pkg pattern, use flexible workload targeting via targetRef:
+
+```yaml
+spec:
+  targetRef:
```
pdettori, line 643: where needed, we should be able to use Duck-Typing together with a label selector

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+- Works with Deployments, StatefulSets, Jobs (addressing current Kagenti limitation)
+- No PodSpec duplication
+- Minimal schema dependency
+- Easy adoption on existing workloads
+
+### Agent Code Requirements
+
+#### Telemetry Instrumentation
+
+Agent code must instrument itself with OpenTelemetry SDK.
```
pdettori, line 659: Not sure how this can be accomplished - normally a developer would instrument the telemetry in the agent code, but in any case the agent cannot instrument itself - it's either injected somehow by the platform or by the agent developer.

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+| Learning curve | Steep | Gentle |
+| Workload type support | Usually single type | Multiple types |
+| Tool compatibility | Limited | Full |
+| Removal impact | Workload disappears | Workload continues |
+| Maintenance burden | High (track K8s changes) | Low (stable references) |
+
+### Open Questions
+
+1. **Namespace-scoped vs. Cluster-scoped CRs**: Should TokenExchange be namespace-scoped (current proposal) or cluster-scoped for multi-tenant scenarios?
+2. **Default identity provider**: Should we assume Keycloak or support pluggable providers from day one?
```
pdettori, line 825: it would be great to abstract Keycloak as a IAM provider that implements OpenID Connect (OIDC) and OAuth 2.0 provider

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+## Risks and Mitigations
+
+### Risk 1: Webhook Availability
+
+**Risk**: If the mutating webhook is unavailable, agent workloads fail to create.
+
+**Mitigation**: 
+- Deploy webhook with multiple replicas
+- Use PodDisruptionBudgets
+- Fail-closed is intentional (security-first approach)
```
pdettori, line 856: Mentioned before, but does it mean that if the webhook is configured but it's down the deployment would fail?

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+- Consider webhook auto-recovery mechanisms
+
+### Risk 2: Configuration Drift
+
+**Risk**: ConfigMap changes may not propagate to all sidecar instances.
+
+**Mitigation**:
+- Sidecars use fsnotify for real-time ConfigMap watching
```
pdettori, line 864: Should compare this to XDS for Envoy-based sidecars

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+**E2E Tests**:
+- Complete agent deployment workflow
+- Migration from Agent CR to composition
+- Removal of Kagenti with workload continuity
+- Cross-namespace agent communication
+
+### Update/Rollback Compatibility
+
+**Forward Compatibility**:
+- New pillar CRs are additive; existing workloads with `kagenti.io/inject` continue working
```
pdettori, line 969: see previous comments for sidecar-based approaches

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+
+### Implementation Phases/History
+
+**Phase 1: Extract TokenExchange** (Q1 2026)
+- Create standalone TokenExchange CR with targetRef pattern
+- Extract current SPIFFE/SVID integration from Agent controller
+- Validate works with Deployment, StatefulSet, Job
+- Write migration documentation
+
+**Phase 2: Adapt AgentCard** (Q1 2026)
```
pdettori, line 997: I think this should be first

pdettori: we can always do the injection with platform wide config as per https://github.com/kagenti/kagenti-extensions/pull/75

docs/design-proposals/compositional-agent-platform-design.md
cwiklik: Yeah, poor use of words. What was meant is exactly as you say. It's developers job to instrument telemetry in an agent's code. Perhaps: A developer must instrument the telemetry in an agent's code usig OpenTelemetry SDK.

docs/design-proposals/compositional-agent-platform-design.md
cwiklik: What happens in this scenario depends on the `MutatingWebhookConfiguration` configuration for the webhook in question. The `failurePolicy` field in the webhook configuration determines what happens when the webhook is unavailable

docs/design-proposals/compositional-agent-platform-design.md
cwiklik: I was thinking to use fsnotify in identity sidecars to watch ConfigMap and reconfigure but granted not sure about complexity involved. Perhaps pod restart will be required.

docs/design-proposals/compositional-agent-platform-design.md
cwiklik: Worth considering, although I understood Roland Huss suggested 1:1 targeting with Duck-Typing Pattern:
```
targetRef:
apiVersion: apps/v1
kind: Deployment # or StatefulSet, Job, CronJob, etc.
name: my-agent
```

docs/design-proposals/compositional-agent-platform-design.md
pdettori: Reconfiguration only might be accomplished without restart - this is what envoy + xDS do today. What cannot be accomplished without restart is 1) adding or removing the sidecars 2) upgrading the image version of the sidecars

docs/design-proposals/compositional-agent-platform-design.md
pdettori: @rhuss ^^^

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+  • Auth proxy (inbound validation)
+  • Envoy proxy (outbound token exchange)
+        ↓
+Agent Pod Running with Defaults
+  • Identity and auth configured
+  • Agent code handles telemetry
+  • Agent code exposes capabilities
+        ↓
+Optional: User Creates Pillar CRs
+  • TokenExchange → customize identity/auth
```
maia-iyer, line 49: So the webhook-injected identity infrastructure has to do with token exchange as well. I am wondering if there's a reason to need both or should we be able to consolidate here?

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+  • No pod restarts required
+```
+
+---
+
+## Background
+
+### Motivation and Problem Space
+
+An AI Agents PLatform needs to define a coherent set of capabilities for users operating fleets of AI agents. When exploring existing AgentOps frameworks, a key insight emerged: framing the problem as a single, general-purpose "Kubernetes AgentOps framework" is likely the wrong abstraction.
```
maia-iyer, line 66:
```suggestion
An AI Agents Platform needs to define a coherent set of capabilities for users operating fleets of AI agents. When exploring existing AgentOps frameworks, a key insight emerged: framing the problem as a single, general-purpose "Kubernetes AgentOps framework" is likely the wrong abstraction.
```

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+│ LAYER 2: Optional Configuration & Discovery            │
+│─────────────────────────────────────────────────────────│
+│ • TokenExchange: Configure identity/auth sidecars       │
+│ • AgentTrace: Configure agent telemetry                 │
+│ • AgentCard: Discover agent capabilities                │
+│ • Controllers create ConfigMaps or cache data           │
+│ • No pod restarts needed                                │
+└─────────────────────────────────────────────────────────┘
+```
+
```
maia-iyer, line 173: I wonder if layer 1 can be generalized into "default platform-wide behavior"? if pod restarts are unavoidable as @pdettori says, this may be desired? 
This will also fix the earlier comment I had where I think the functionality of webhook-injected identity and the `TokenExchange` CRD have too much overlap that we should combine them

docs/design-proposals/compositional-agent-platform-design.md
cwiklik: @maia-iyer, I agree that we should strive for symmetry and consistence by appling platform-wide defaults to all pillars. Perhaps a webhook can inject defaults via a configMap, Annotations, and Labels. 

The latest webhook implementation in the kagenti-extensions is meant to inject both identity and token exchange sidecars.

docs/design-proposals/compositional-agent-platform-design.md
cwiklik: @maia-iyer @mrsabath I am waiting for domain experts to chime in on what makes sense in this API. Which fields are fixed and which can be modified.

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+  • Envoy proxy (outbound token exchange)
+        ↓
+Agent Pod Running with Defaults
+  • Identity and auth configured
+  • Agent code handles telemetry
+  • Agent code exposes capabilities
+        ↓
+Optional: User Creates Pillar CRs
+  • TokenExchange → customize identity/auth
+  • AgentTrace → configure telemetry
```
rubambiza, line 50: Note to self: look at the definition of AgentTrace and its potential relationship to the [Observability proposal](https://github.com/kubernetes-sigs/kube-agentic-networking/pull/33) in the kube-agentic-networking community.

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+- [ ] Documentation updates
+- [ ] Integration tests
+- [ ] E2E tests
+- [ ] Performance benchmarks
+
+---
+
+## Summary/Abstract
+
+This design proposal describes the refactoring of Kagenti from an inheritance-based architecture to a composition-based architecture for managing AI agent workloads on Kubernetes. The current monolithic `Agent` Custom Resource (CR) will be replaced with a two-layer system consisting of automatic identity infrastructure injection via mutating webhook and optional configuration through three independent pillar CRs: `TokenExchange`, `AgentTrace`, and `AgentCard`.
```
rubambiza, line 26: A link to the current definition of the `Agent Custom Resource` might be helpful context for noobs like me.

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+
+An AI Agents PLatform needs to define a coherent set of capabilities for users operating fleets of AI agents. When exploring existing AgentOps frameworks, a key insight emerged: framing the problem as a single, general-purpose "Kubernetes AgentOps framework" is likely the wrong abstraction.
+
+The Kubernetes ecosystem tells a consistent story: attempts to create higher-level workload abstractions fail, while composition-based add-ons succeed. This is not bad engineering or poor timing—it reflects something fundamental about how platform engineers and developers relate to their tools.
+
+**Current Kagenti Problems (Inheritance Pattern)**:
+
+1. **Tight coupling to upstream Kubernetes changes** - Requires constant rebasing on K8s releases without substantial benefits
+2. **Responsibility for pass-through definitions** - Even when passing through PodSpec, Kagenti becomes responsible when something goes wrong in those definitions
+3. **Limited workload type support** - Only works with Deployments; cannot handle StatefulSets (essential for stateful agents with long-term memory), Jobs (ideal for one-shot agent runs), or CronJobs (scheduled agent executions)
```
rubambiza, line 74: Would the event-driven agents being pursued by Alek count as another example of CronJobs?

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+### Impact and Desired Outcome
+
+The compositional refactoring will:
+
+1. **Reduce adoption friction** - Users keep their existing Deployments, StatefulSets, and Jobs
+2. **Enable incremental adoption** - Each pillar can be adopted independently
+3. **Improve tool compatibility** - kubectl, dashboards, and GitOps continue working normally
+4. **Support multiple workload types** - StatefulSets for stateful agents, Jobs for one-shot runs, CronJobs for scheduled executions
+5. **Simplify removal** - If disabled/uninstalled, core workloads continue running
+6. **Reduce maintenance burden** - No need to track K8s PodSpec changes
```
rubambiza, line 90: Is this referring to the k8s PodSpec upstream type or the PodSpecs currently being generated by kagenti? If my assumption of it being the former is correct, I'd suggest clarifying `upstream k8s PodSpec...`. This distinction is especially important because the [above section](https://github.com/kagenti/kagenti/pull/531/changes#diff-84f87e17fe8054f9d71f9cba59c3383cd34b987046a2d1747f2b158bebe91359R72) mentions the tight coupling to upstream k8s changes.

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+  sideEffects: None
+  timeoutSeconds: 10
+  failurePolicy: Fail
+  reinvocationPolicy: Never
+```
+
+**Key Design Decisions**:
+
+- **Workload-level targeting**: Webhook watches Deployment/StatefulSet/Job, not Pods, enabling deterministic ConfigMap naming based on workload name
+- **Explicit opt-in**: Each workload must have `kagenti.io/inject: enabled` and `kagenti.io/spiffe: enabled` labels to prevent accidental injection
```
rubambiza, line 250: `kagenti.io/spiffe: enabled` is missing in the above matchExpressions.

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+
+**Purpose**: Configure agent code telemetry (not identity infrastructure).
+
+**Scope**: Provides configuration that agent code reads to configure OpenTelemetry.
+
+**Key Capabilities**:
+- Specify OTEL collector endpoints
+- Configure sampling strategies
+- Define resource attributes
+- Enable GenAI semantic conventions
```
rubambiza, line 531: As I suggested in Red Hat's Kagenti Refactoring document, it might be useful to link to the evolving GenAI semantic conventions for easy reference.

rubambiza, line 531:
```suggestion
- Enable [GenAI semantic conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/)
```

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+      enabled: true
+      port: 9090
+      path: "/metrics"
+
+status:
+  phase: Active
+  configMapName: "weather-agent-trace"
+```
+
+**Note on Maturity**: This is a nascent area with limited prior work. The specification will evolve as observability standards for AI agents mature.
```
rubambiza, line 585: Agreed that it is nascent area. I would like to suggest linking to the Observability/auditing proposal in the upstream k8s community as a good place to track the evolution of.

rubambiza, line 585:
```suggestion
**Note on Maturity**: This is a nascent area with limited prior work. The specification will evolve as observability standards for AI agents mature. See [observability proposal](https://github.com/kubernetes-sigs/kube-agentic-networking/pull/33) in the k8s/CNCF community for example.
```

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+
+### Controller Architecture
+
+#### Consolidated Operator (Istiod Pattern)
+
+Following the Istiod pattern, we recommend a single operator managing all three CRs:
+
+```
+┌────────────────────────────────────────────────────────┐
+│ Kagenti Operator Pod                                   │
```
rubambiza, line 595: Is this pod going to be namespace scoped? I am trying to understand if there won't be any issues creating things like ConfigMaps in other namespaces (e.g., kagenti-system).

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+#### Telemetry Instrumentation
+
+Agent code must instrument itself with OpenTelemetry SDK.
+
+**Configuration Source**: AgentTrace ConfigMap at `/etc/kagenti/trace`
+
+**Minimal Example**:
+
+```python
+from opentelemetry import trace
```
rubambiza, line 666: I am trying to understand the mechanics here. Will this be running in a loop until the config map/path is populated? Is there an exit criteria when the agent is terminated?

docs/design-proposals/compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1043 @@
...
+- v3.0: Agent CR removed
+- Final migration guide published
+- Legacy support ended
+
+---
+
+## References
+
+### Failed Kubernetes Abstraction Projects
+- kubernetes-sigs/application - Dormant Application CRD
```
rubambiza, line 1025: Perhaps add some links for reference to better understand the architecture/goals and why/how they failed?

docs/design-proposals/compositional-agent-platform-design.md
cwiklik: It's referring to PodSpec upstream changes. I will clarify per your comment. Thank you.

docs/design-proposals/compositional-agent-platform-design.md
cwiklik: @pdettori Just a thought here, if we allow selector wouldn't this enable accidental changes to workloads collocated in the namespace?

docs/design-proposals/compositional-agent-platform-design.md
evaline-ju, line 89: I think a word or phrase is missing for my understanding of this line. What is being disabled/uninstalled here? And what is "core" here?

docs/design-proposals/compositional-agent-platform-design.md
evaline-ju: The particular linked PR is deliberately not about CRs/APIs for configuring telemetry i.e. the "APIs such as Kubernetes CRDs to allow users to configure traces and other observability data such as metrics and logs will be addressed in subsequent proposals."

Referencing standards to be discussed in the upstream community (https://github.com/kubernetes-sigs/kube-agentic-networking) I think would be appropriate but based on progress here vs. upstream, it is possible that parts of the `AgentTrace` CR are brought to the table there for discussion - we would ultimately want to make sure Kagenti aligns with decisions/standards made upstream.

docs/design-proposals/compositional-agent-platform-design.md
evaline-ju: From the language I was wondering how broadly this can be applied as well, since some changes for receivers I thought would require OpenTelemetry collector restart?

docs/design-proposals/compositional-agent-platform-design.md
evaline-ju: Perhaps it may be helpful to distinguish the couple main parts like (1) span generation and (2) span export [to collector] and (3) span collection [actual collector maintenance], between agent developer and platform responsibility.

I assume most of the first two will both be expected of the agent developers with (2) "using OpenTelemetry SDK", and (1) ideally with GenAI semantic conventions directly

The platform can eventually help with various parts presumably such as collector transforms to aid with (1), and flexible configuration endpoints for (2). And based on current capabilities and the detailed `AgentTrace` proposal the platform will take care of (3) it seems?

docs/design-proposals/compositional-agent-platform-design.md
cwiklik: We should support both namespace and cluster scoped deployments for this operator. The admin decides which namespaces are accessible to the operator with namespace scoped deployment.

docs/design-proposals/compositional-agent-platform-design.md
cwiklik: Yes, indeed this is vague. Disabled/uninstalled" refers to our platform components (the pillars), and "core workloads" refers to the user's existing Deployments, StatefulSets, and Jobs. The point is that since we're composing onto their workloads rather than owning them, removing our tooling doesn't take down their applications.

So perhaps this: 

Simplify removal — If the platform components(the pillars) are disabled or uninstalled, the user's existing workloads (Deployments, StatefulSets, Jobs) continue running unaffected.

docs/design-proposals/compositional-agent-platform-design.md
cwiklik: As mentioned before, the dynamic config with XDS is designed for envoy sidecars but we have non-envoy sidecars as well.

PR #593, opened 2026-02-02
:memo: Let MCP Gateway share the Kagenti Gateway

### LINKED ISSUE #377: feature: Expose MCP Gateway outside the cluster
### Feature Description

I'd like to see the MCP Gateway exposed outside the cluster.

I would like to be able to communicate with it, for debugging, using CLIs and interfaces running on my laptop.

### Proposed Solution

Add a hostname to the HTTPRoute.

The dev environment used by the https://github.com/kagenti/mcp-gateway project exposes the MCP Gateway
It would be helpful to evolve the dev environments of the two systems to be very similar.

### Want to contribute?

### Additional Context

_No response_

esnible:
Resolves #377

Currently we can't curl the MCP from outside the cluster because only the Kagenti gateway is exposed outside the cluster.

- One approach to solve is to let MCP Gateway pods use Kagenti's Gateway
- The other approach is the create extraPortMappings for the Gateway in mcp-system and let MCP Gateway continue using that Gateway.

This PR uses the first approach.  To test, deploy using `deployments/ansible/run-install.sh` and verify `curl -v mcp.127-0-0-1.sslip.io:8080/.well-known/oauth-protected-resource` returns JSON.  It isn't clear if the MCP 

An alternative is to add port mappings for the other Gateway.

@pdettori @huang195 Let me know which approach is preferred

### DIFF
--- a/charts/kagenti/templates/mcp-gateway-ns-labeler.yaml
+++ b/charts/kagenti/templates/mcp-gateway-ns-labeler.yaml
@@ -87,13 +87,8 @@ spec:
               for ns in "{{ .Values.mcpGateway.namespaces.mcpSystem }}" "{{ .Release.Namespace }}"; do
                 echo "Labeling namespace: $ns"
                 for i in $(seq 1 $MAX_RETRIES); do
-                  if [ "$ns" = "{{ .Values.mcpGateway.namespaces.mcpSystem }}" ]; then
-                    kubectl label namespace "$ns" \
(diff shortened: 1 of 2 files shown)


evaline-ju:
@esnible I think this may have run into some unfortunate Github runners issues https://www.githubstatus.com/incidents/xwn6hjps36ty

pdettori:
I am ok w/ the second approach - after all this is mainly for kind or generic k8s, on a OCP we would be using routes

PR #679, opened 2026-02-13
Bug fix:  Use correct Keycloak url and role used by ui admin

### LINKED ISSUE #678: :bug: kagenti ui is broken
### Describe the bug

Kagenti ui is broken - redirect to keycloak uses wrong URL and roles introduced by initial api backend break the admin permissions in UI.

### Steps To Reproduce

1. install latest kagenti
2. try login 
3. try listing agents

### Expected Behavior

should work

### Additional Context

_No response_

pdettori:
## Summary

Fixes regression issues introduced by previous PR on keycloak URL usage - in-cluster vs. off-cluster and update UI for new role used in backend API.

install kagenti as usual, then do `make build-load-ui` and follow printed instructions to run helm command.

Fixes #678

### DIFF
--- a/charts/kagenti/values.yaml
+++ b/charts/kagenti/values.yaml
@@ -194,8 +194,10 @@ mlflow:
 # ------------------------------------------------------------------
 #  MLflow OAuth Secret Creator Configuration
 # ------------------------------------------------------------------
+# TODO: Change back to ghcr.io/kagenti/kagenti/mlflow-oauth-secret once
+# the image is published via Build-Publish workflow (on version tag)
 mlflowOAuthSecret:
-  image: ghcr.io/kagenti/kagenti/mlflow-oauth-secret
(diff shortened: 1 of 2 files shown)


pdettori:
Comments addressed in latest push

 1. Reverted mlflowOAuthSecret.image change — restored quay.io/ladas/mlflow-oauth-secret with the TODO comment. This belongs in a separate PR.
  2. Updated admin role mapping comment — replaced the generic comment with a TODO that explicitly marks this as temporary, references issue #647 ("Protect Backend API" epic), and states it
  should be replaced with proper Keycloak role provisioning.
  3. Fixed duplicate logging (bonus, from another Copilot comment) — removed the try/catch logging in load_keys() since validate_token already logs the error. This avoids duplicate error
  lines per failure.

PR #687, opened 2026-02-14
feat: Add periodic build reconciliation loop

### LINKED ISSUE #546: :bug: Agent deploy does not complete, stuck after run-pod
### Describe the bug

With the [weather agent and tool example](https://github.com/kagenti/kagenti/blob/main/docs/demos/demo-weather-agent.md) my typical behavior is to deploy the agent and tool in quick succession and then return to check on the status of them in a couple minutes.

On commit https://github.com/kagenti/kagenti/commit/2adc36471ec7961ad8e1eda96178d0bb173dbd8d (latest at time of writing) my cluster would get stuck in a state where the `run` pod had run, but the `weather-service` deployment itself would not come up, and the `weather-service` agent would not appear in the Agent Catalog
```
oc get pod -n team1
NAME                                  READY   STATUS      RESTARTS   AGE
weather-service-run-hk9hg-4pttt-pod   0/2     Completed   0          111s
weather-tool-0                        3/3     Running     0          112m
weather-tool-844d95cf4d-rxmsp         1/1     Running     0          112m
weather-tool-run-dsj8g-d45hz-pod      0/2     Completed   0          113m
```
The theory is that this happens when I navigate away from the `Import New Agent` page before the build runs to completion, such as returning to the Agent Catalog itself.

### Steps To Reproduce

This seems reproducible after filling out agent details and running `Build & Deploy Agent` and navigating away from the page before the Build progress bar runs to completion.

### Expected Behavior

I expected that after kicking off an agent deploy with `Build & Deploy Agent` or tool deploy with `Deploy Tool`, this would kick off all the resources in the cluster necessary to deploy the agent or tool, and allow me to look at other things in the cluster.

Note I have not personally tested whether similar behavior happens with the tool.

pdettori:
## Summary

Fixes #546

- Add a backend `asyncio` reconciliation loop that periodically scans kagenti-enabled namespaces for Shipwright Builds whose latest BuildRun succeeded but have no corresponding workload (Deployment/StatefulSet/Job)
- The loop calls the existing `finalize_shipwright_build` (agents) and `finalize_tool_shipwright_build` (tools) functions directly, reusing all idempotency checks, config merging, and workload creation logic with zero duplication
- Configurable via `BUILD_RECONCILIATION_INTERVAL` (default 30s) and `ENABLE_BUILD_RECONCILIATION` (default true) environment variables

## Test plan

- [ ] Unit test: mock `KubernetesService` to return a Build with a succeeded BuildRun and no Deployment, verify `reconcile_builds()` calls the finalize function
- [ ] Manual test on Kind: deploy an agent from source, navigate away from build progress page immediately, wait ~30s, verify agent appears in catalog
- [ ] Idempotency: run reconciliation twice — second pass should skip (workload already exists)
- [ ] Disabled: set `ENABLE_BUILD_RECONCILIATION=false`, verify loop doesn't start

### DIFF
--- a/kagenti/backend/app/main.py
+++ b/kagenti/backend/app/main.py
@@ -59,6 +59,8 @@ async def lifespan(app: FastAPI):
             "Build reconciliation started (interval: %ds)",
             settings.build_reconciliation_interval,
         )
+    else:
+        logger.info("Build reconciliation disabled (ENABLE_BUILD_RECONCILIATION=false)")
 
     yield
 
--- a/kagenti/backend/app/services/reconciliation.py
+++ b/kagenti/backend/app/services/reconciliation.py
@@ -164,7 +164,12 @@ async def _reconcile_single_build(
 
 
 async def run_reconciliation_loop() -> None:
-    """Background loop that periodically reconciles builds."""
+    """Background loop that periodically reconciles builds.
+
+    The loop is sequential: sleep only starts *after* the current
+    reconciliation pass completes, so there is no risk of overlapping
+    runs even when a pass takes longer than the configured interval.
+    """
     interval = settings.build_reconciliation_interval
     # Sleep first — give the cluster time to settle after startup
     await asyncio.sleep(interval)


kagenti/backend/app/services/reconciliation.py
```diff
@@ -0,0 +1,180 @@
...
+            try:
+                await _reconcile_single_build(kube, namespace, name, resource_type)
+            except Exception:
+                logger.warning(
+                    "Failed to reconcile build '%s/%s', will retry next cycle",
+                    namespace,
+                    name,
+                    exc_info=True,
+                )
+
```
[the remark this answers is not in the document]
pdettori: The review comment has been addressed — 18 unit tests now cover:

  - _workload_exists (5 tests): Deployment/StatefulSet/Job detection, no-workload case, non-404 error propagation
  - _reconcile_single_build (8 tests): Skips when build not succeeded, no BuildRuns, workload already exists; calls correct finalize for agents vs tools; suppresses 409/400 HTTPExceptions;
  re-raises other errors
  - reconcile_builds (5 tests): Scans all enabled namespaces; filters by resource type; continues after per-build failures; continues after namespace list failures; skips builds with missing
   type labels

kagenti/backend/app/main.py
```diff
@@ -47,7 +48,28 @@ async def lifespan(app: FastAPI):
     logger.info(f"Debug mode: {settings.debug}")
     logger.info(f"Domain: {settings.domain_name}")
     logger.info(f"ENABLE_AUTH environment variable set to: {settings.enable_auth}")
+
+    # Start build reconciliation loop
+    reconciliation_task = None
+    if settings.enable_build_reconciliation:
```
evaline-ju, line 54: nit: should there be a log in the case this is set to false? 

And question: under what circumstances would a kagenti deployer want to set this to false?

kagenti/backend/app/services/reconciliation.py
```diff
@@ -0,0 +1,180 @@
...
+        try:
+            getter(namespace=namespace, name=name)
+            return True
+        except ApiException as e:
+            if e.status != 404:
+                raise
+    return False
+
+
+async def reconcile_builds() -> None:
```
evaline-ju, line 50: if this reconcile takes too long (perhaps the k8s API is slow, or the cluster itself has many namespaces) compared to the configured interval, do we risk a race condition of constant reconcile attempts?

pdettori: Added docstring clarifying the loop is sequential: `await reconcile_builds()` must complete before await `asyncio.sleep(interval)` runs, so overlapping executions cannot happen regardless of how long a pass takes

kagenti/backend/app/main.py
pdettori: Added else branch that logs "Build reconciliation disabled (ENABLE_BUILD_RECONCILIATION=false)" when the feature is off

pdettori: As for why the feature might be disabled, these are 3 possible scenarios:

1. Shipwright not installed - If the deployer uses pre-built images only (no source-to-image builds), the loop would repeatedly fail trying to list Shipwright CRDs that don't exist.
  2. External CI/CD handles finalization - If builds are orchestrated by an external pipeline (GitHub Actions, Tekton) that triggers finalization itself, the reconciliation loop would be redundant.
  3. Debugging build issues - When investigating why a build isn't finalizing correctly, an operator may want to disable automatic reconciliation so they can inspect the intermediate state without the loop interfering.

PR #699, opened 2026-02-16
Fix: Fix Tornjak by enabling SPIRE for Kagenti dev install

Alan-Cha:
## Summary

Enable SPIRE for Kagenti dev install. This seems to be the reason why Tornjak does not start correctly.

I could not reach Tornjak as documented here: https://github.com/kagenti/kagenti/blob/main/docs/install.md#identity-services

I previously needed to port forward Tornjak

For example

```
kubectl port-forward -n spire-server svc/spire-tornjak-backend 10000:10000
```

Then open: http://localhost:10000/api/v1/spire/entries

```
deployments/ansible/run-install.sh --env dev
```

Then the following should work.

```
# Tornjak API
curl http://spire-tornjak-api.localtest.me:8080/
# Expected: "Welcome to the Tornjak Backend!"

# Tornjak UI
open http://spire-tornjak-ui.localtest.me:8080/
```

### DIFF
--- a/deployments/ansible/default_values.yaml
+++ b/deployments/ansible/default_values.yaml
@@ -156,6 +156,11 @@ charts:
             enabled: true
           namespaces:
             create: true
+            server:
+              name: zero-trust-workload-identity-manager
+              create: true
+              labels:
+                shared-gateway-access: "true"
           ingressControllerType: ""
           clusterName: agent-platform
           trustDomain: "{{ domain | default('localtest.me') }}"
--- a/charts/kagenti-deps/templates/spire-route.yaml
+++ b/charts/kagenti-deps/templates/spire-route.yaml
@@ -1,21 +1,6 @@
 {{- /* Create SPIRE routes for Kubernetes or when using Helm chart on OpenShift (OCP < 4.19) */}}
+{{- /* Note: The namespace is created and owned by the SPIRE Helm chart */}}
 {{- if and .Values.components.spire.enabled (or (not .Values.openshift) .Values.useSpireHelmChart) }}
-apiVersion: v1
-kind: Namespace
-metadata:
-  name: {{ .Values.spire.namespace }}
-  labels:
-    {{- include "kagenti.labels" . | nindent 4 }}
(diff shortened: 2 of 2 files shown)


Alan-Cha:
## Enable SPIRE for Kagenti Dev Install

### Summary
Enables SPIRE workload identity in Kagenti dev deployments with proper Tornjak (SPIRE management UI) access through Gateway HTTPRoutes.

### Changes

**1. Enable SPIRE in dev environment** ([dev_values.yaml](deployments/envs/dev_values.yaml))
- Added `spire.enabled: true` to kagenti-deps component configuration

**2. Configure SPIRE namespace management** ([default_values.yaml](deployments/ansible/default_values.yaml))
- Configured SPIRE Helm chart to create custom namespace: `zero-trust-workload-identity-manager`
- Added `shared-gateway-access: "true"` label to enable HTTPRoute attachment to kagenti-system Gateway
- Removed `kagenti-system` from ignored namespaces list (needed for agent workload identity)

**3. Simplify HTTPRoute configuration** ([spire-route.yaml](charts/kagenti-deps/templates/spire-route.yaml))
- Delegated namespace creation to SPIRE Helm chart (eliminates ownership conflict)
- HTTPRoutes now reference services in same namespace (no cross-namespace permissions needed)
- Exposes Tornjak frontend, backend, and OIDC discovery through Gateway

deployments/ansible/default_values.yaml
```diff
@@ -171,7 +176,6 @@ charts:
           ignoreNamespaces:
             - kube-system
             - kube-public
-            - kagenti-system
```
Alan-Cha, line 174: Should SPIRE ignore the `kagenti-system` namespace? I had some issues with the Kagenti UI and Keycloak endpoints. I'm not so sure about this change.

Clause suggested that the reason why I had issues with Kagenti UI and Keycloak is that by configuring SPIRE to ignore their namespace, they don't receive SPIFFE IDs, which conflicts with istio ambient mode...

PR #725, opened 2026-02-18
fix: make GitHub token optional and reduce required scopes

### LINKED ISSUE #722: :bug: Asking for GitHub token with surplus privileges
### Describe the bug

Installation guide currently [instructs users to create a GitHub token](https://github.com/kagenti/kagenti/blob/main/docs/install.md#kind-specific-requirements) with high‑level privileges, which no longer appear to be required.
Consider removing this requirement or make it optional.
It would also be helpful to add an explanation describing when such privileges are actually needed and why.

Also change related documentation such as https://github.com/kagenti/kagenti/blob/main/docs/install.md#ansible-based-installer-recommended. Either remove the need to setup GitHub user and key or detail what should happen when it is not required (or make it optional).

### Steps To Reproduce

See https://github.com/kagenti/kagenti/blob/main/docs/install.md#kind-specific-requirements

### Expected Behavior

Installation should not require providing kagenti access to all repos of the user Github. 

### Additional Context

_No response_

mrsabath:
## Summary

Fixes #722

- **Marks githubUser/githubToken as optional** across docs, Helm values, and secret templates — they are only needed to deploy agents/tools from private GitHub repos or pull from private registries
- **Clarifies token scopes by use case**: `repo` for private repository access, `read:packages` for private GHCR pulls
- **Wraps GitHub-related secrets in Helm conditionals** (github-token-secret, github-shipwright-secret, ghcr-secret) so they are only created when credentials are provided
- **Resolves clone secret at runtime** — the backend now checks if `github-shipwright-secret` exists in the namespace before referencing it in the Shipwright Build manifest, so builds for public repos work without GitHub credentials

Kagenti container images are public on ghcr.io and do not require a token to pull.

## Files changed

| File | Change |
|------|--------|
| docs/install.md | Mark GitHub token as optional, clarify scopes by use case |
| deployments/envs/secret_values.yaml.example | Add (Optional) prefix with scope guidance |
| charts/kagenti/.secrets_template.yaml | Add (Optional) prefix, fix Github capitalization |
| charts/kagenti/values.yaml | Update comment to clarify optionality |
| charts/kagenti/templates/agent-namespaces.yaml | Conditional GitHub secrets, fix YAML whitespace |
| kagenti/backend/app/models/shipwright.py | Default gitSecretName to None |
| kagenti/backend/app/services/shipwright.py | Add resolve_clone_secret helper, conditional cloneSecret |
| kagenti/backend/app/routers/agents.py | Resolve clone secret at build-creation time |
| kagenti/backend/app/routers/tools.py | Resolve clone secret at build-creation time |
| kagenti/backend/tests/test_shipwright.py | Tests for with/without clone secret + resolve_clone_secret |

## Test plan

- [x] helm template with empty GitHub secrets renders valid YAML without GitHub secrets
- [x] helm template with GitHub secrets set renders all 3 secrets correctly
- [x] YAML document separation intact (no `---#` merging) after whitespace fix
- [x] All 37 Shipwright unit tests pass
- [ ] Kind full test passes without GitHub token
- [ ] Pre-commit passes

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/kagenti/examples/agents/recipe_agent_shipwright_build.yaml
+++ b/kagenti/examples/agents/recipe_agent_shipwright_build.yaml
@@ -23,7 +23,6 @@ spec:
     git:
       url: https://github.com/kagenti/agent-examples
       revision: main
-      cloneSecret: github-shipwright-secret
     contextDir: a2a/recipe_agent
   strategy:
     name: buildah-insecure-push
--- a/kagenti/examples/agents/weather_agent_shipwright_build.yaml
+++ b/kagenti/examples/agents/weather_agent_shipwright_build.yaml
@@ -27,7 +27,6 @@ spec:
       # PR: https://github.com/kagenti/agent-examples/pull/XXX
       url: https://github.com/ladas/agent-examples
       revision: genai-autoinstrumentation
-      cloneSecret: github-shipwright-secret
     contextDir: a2a/weather_service
   strategy:
     name: buildah-insecure-push
--- a/kagenti/examples/agents/weather_agent_shipwright_build_ocp.yaml
+++ b/kagenti/examples/agents/weather_agent_shipwright_build_ocp.yaml
@@ -27,7 +27,6 @@ spec:
       # PR: https://github.com/kagenti/agent-examples/pull/XXX
       url: https://github.com/ladas/agent-examples
       revision: genai-autoinstrumentation
-      cloneSecret: github-shipwright-secret
     contextDir: a2a/weather_service
   strategy:
     name: buildah


charts/kagenti/templates/agent-namespaces.yaml
```diff
@@ -24,7 +24,8 @@ metadata:
 # ——————————————————————————————————————————————————————————————————————————————
 #  Secrets for Namespace: {{ . }}
 # ——————————————————————————————————————————————————————————————————————————————
-# 1. GitHub Token Secret (for general use)
+{{- if and $.Values.secrets.githubUser $.Values.secrets.githubToken }}
+# 1. GitHub Token Secret (for general use — only created when GitHub credentials are provided)
```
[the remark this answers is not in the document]
mrsabath: Addressed by a new commit

kagenti/backend/app/models/shipwright.py
```diff
@@ -43,7 +43,7 @@ class BuildSourceConfig(BaseModel):
     gitUrl: str
     gitRevision: str = "main"
     contextDir: str = "."
-    gitSecretName: str = "github-shipwright-secret"
+    gitSecretName: Optional[str] = "github-shipwright-secret"
```
pdettori: This is the root cause of the `BuildRegistrationFailed` error. The default value `"github-shipwright-secret"` means **every** build gets `cloneSecret` in its manifest, even when no GitHub credentials are configured and the secret doesn't exist in the namespace.

```suggestion
    gitSecretName: Optional[str] = None
```

With `None` as the default, the `if source_config.gitSecretName:` check in `shipwright.py` correctly omits `cloneSecret` unless the caller explicitly provides a secret name.

The callers (`agents.py`, `tools.py`) should then resolve the secret at build-creation time — see next comment.

kagenti/backend/app/services/shipwright.py
```diff
@@ -187,6 +186,10 @@ def build_shipwright_build_manifest(
         },
     }
 
+    # Add clone secret for private git repos
```
pdettori, line 202: The conditional check here is correct — but it only works if the callers pass `gitSecretName=None` when the secret doesn't exist. Since the model default is still `"github-shipwright-secret"`, this never triggers.

Add a `resolve_clone_secret` helper above this function that the callers can use:

```python
def resolve_clone_secret(kube_service, namespace: str) -> Optional[str]:
    """Check if the GitHub Shipwright clone secret exists in the namespace."""
    try:
        kube_service.core_api.read_namespaced_secret(
            name=SHIPWRIGHT_GIT_SECRET_NAME, namespace=namespace
        )
        return SHIPWRIGHT_GIT_SECRET_NAME
    except Exception:
        logger.debug(
            f"Clone secret '{SHIPWRIGHT_GIT_SECRET_NAME}' not found in namespace "
            f"'{namespace}', builds will proceed without git credentials"
        )
        return None
```

Then in `agents.py` and `tools.py`, call `resolve_clone_secret(kube, request.namespace)` before building the manifest, and pass the result as `clone_secret_name` to the helper functions:

```python
# agents.py (~line 2354)
clone_secret = resolve_clone_secret(kube, request.namespace)
build_manifest = _build_agent_shipwright_build_manifest(request, clone_secret_name=clone_secret)

# tools.py (~line 1334)
clone_secret = resolve_clone_secret(kube, request.namespace)
build_manifest = _build_tool_shipwright_build_manifest(request, clone_secret_name=clone_secret)
```

This way, builds for public repos work without credentials, and builds for private repos use the secret when it's available.

kagenti/backend/tests/test_shipwright.py
```diff
@@ -77,6 +77,33 @@ def test_basic_build_manifest_with_internal_registry(self):
         assert manifest["spec"]["source"]["git"]["cloneSecret"] == SHIPWRIGHT_GIT_SECRET_NAME
```
pdettori: This assertion will break once the default changes to `None`. Since no `clone_secret_name` is passed, the manifest won't have `cloneSecret`.

Update to assert absence:
```python
assert "cloneSecret" not in manifest["spec"]["source"]["git"]
```

And add a new test for the **with-secret** case:

```python
def test_build_manifest_with_clone_secret(self):
    """Test that cloneSecret is included when clone_secret_name is provided."""
    request = CreateAgentRequest(
        name="test-agent", namespace="team1", protocol="a2a",
        framework="LangGraph", gitUrl="https://github.com/example/repo",
        gitPath="agents/test", gitBranch="main", imageTag="v0.0.1",
        deploymentMethod="source",
    )
    manifest = _build_agent_shipwright_build_manifest(
        request, clone_secret_name=SHIPWRIGHT_GIT_SECRET_NAME
    )
    assert manifest["spec"]["source"]["git"]["cloneSecret"] == SHIPWRIGHT_GIT_SECRET_NAME
```

Also add a test for `resolve_clone_secret` in `test_shipwright_shared.py` covering both the secret-exists and secret-missing cases.

PR #726, opened 2026-02-19
fix: Make postgres-otel database initialization idempotent

Alan-Cha:
## Summary

Fixes postgres-otel crash loop after cluster restarts by making the database initialization script idempotent.

After a cluster restart (e.g., Podman/Kind restart), `postgres-otel-0` entered a crash loop with the error:

```
ERROR: database "mlflow" already exists
```

This also caused `phoenix-0` to crash loop since it depends on postgres-otel.

Checks pg_database system catalog for the database
* Only executes CREATE DATABASE if it doesn't exist
* Returns cleanly if the database already exists
* Safe to run multiple times


pdettori (approved):
/lgtm

PR #749, opened 2026-02-23
CI: Use org-wide reusable stale workflow

rubambiza:
## Summary

- Replaces the inline stale workflow (from #748) with a thin caller that delegates to the centralized reusable workflow in `kagenti/.github`
- All configuration (days, exempt labels, messages) is now managed org-wide in kagenti/.github#39
- Reduces per-repo maintenance -- future changes only need to be made in one place

## Test plan

- [ ] Depends on kagenti/.github#39 being merged first
- [ ] Trigger manually via Actions tab to confirm the caller invokes the org workflow
- [ ] Verify stale labeling behavior is unchanged

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -1,18 +1,9 @@
 # Stale Issue and PR Management
 #
-# Automatically marks issues and PRs as stale after 60 days of inactivity,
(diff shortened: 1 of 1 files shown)


pdettori (approved):
/lgtm

PR #754, opened 2026-02-24
Feat/terraform validation checks

RyanJenkins99:
## Summary                                                                                                           
                           
  Adds validation checks to prevent OpenShift installation on incomplete Terraform infrastructure. This prevents       
  clusters from being created without critical networking components like NAT gateways.                                

  ## Problem

  The kagenti-team management cluster was deployed with incomplete Terraform infrastructure (only 8 resources instead
  of 26). When `terraform apply` failed partway through, the error was missed, and `install-openshift.sh` ran anyway.
  This created a cluster without NAT gateways, causing:
  - Workers unable to pull container images
  - Authentication failures
  - Load balancer health check failures

  ## Changes

  1. **Enhanced `install-openshift.sh`**
     - Validates Terraform state has minimum 20 resources before proceeding
     - Verifies NAT gateways (3), Elastic IPs (3), and route tables (4) exist
     - Fails early with clear error messages if infrastructure is incomplete

  2. **New `deploy-full.sh` script**
     - Single-command automated deployment with validation at each step
     - Creates workspace, runs terraform plan/apply, validates infrastructure, installs OpenShift
     - Reduces human error by orchestrating the entire process

  3. **Updated README.md**
     - Documents automated deployment option (recommended)
     - Adds troubleshooting section for incomplete deployments
     - Clear warnings about verifying terraform apply completion

  ## Testing

  Validation tested on:
  - Complete workspace (26 resources) - validation passes
  - Incomplete workspace (8 resources) - validation fails with clear error

  This prevents an OpenShift install that will fail due to missing infrastructure.

### DIFF
--- a/terraform/management-cluster/scripts/deploy-full.sh
+++ b/terraform/management-cluster/scripts/deploy-full.sh
@@ -13,6 +13,9 @@ set -euo pipefail
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 TF_DIR="$(cd "$SCRIPT_DIR/.." && pwd)"
 
+# Cleanup terraform plan file on exit (contains sensitive data)
+trap 'rm -f "$TF_DIR/tfplan"' EXIT
+
 # Colors
 RED='\033[0;31m'
 GREEN='\033[0;32m'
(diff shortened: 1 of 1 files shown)


terraform/management-cluster/scripts/deploy-full.sh
```diff
@@ -0,0 +1,230 @@
...
+fi
+
+if [ ! -f "$TF_DIR/$TFVARS_FILE" ]; then
+    log_error "tfvars file not found: $TF_DIR/$TFVARS_FILE"
+    exit 1
+fi
+
+SKIP_MCE=false
+if [[ "${2:-}" == "--skip-mce" ]]; then
+    SKIP_MCE=true
```
pdettori: **Suggestion:** The `--skip-mce` flag is only recognized as the second positional argument (`$2`). If a user runs `./deploy-full.sh --skip-mce terraform.tfvars` (flags first, common UNIX convention) or adds other arguments, this flag is silently ignored — no error, just proceeds with MCE install (which can take 30+ minutes).

Consider a simple loop-based parser:
```bash
SKIP_MCE=false
for arg in "$@"; do
  case "$arg" in
    --skip-mce) SKIP_MCE=true ;;
  esac
done
```

terraform/management-cluster/scripts/deploy-full.sh
```diff
@@ -0,0 +1,230 @@
...
+if [[ ! $REPLY =~ ^[Yy]$ ]]; then
+    log_warn "Deployment cancelled by user"
+    exit 0
+fi
+
+# Run terraform apply
+log_info "Running terraform apply..."
+START_TIME=$(date +%s)
+
+if ! terraform apply tfplan; then
```
pdettori, line 138: **Suggestion:** The `tfplan` file is never cleaned up after apply. Terraform plan files can contain sensitive data (credentials, secret variable values, full proposed state). Consider adding cleanup:
```bash
trap 'rm -f "$TF_DIR/tfplan"' EXIT
```
or an explicit `rm -f tfplan` after the apply succeeds.

RyanJenkins99: Thanks for your feedback, making changes!

pdettori (approved):
/lgtm

PR #757, opened 2026-02-24
build: Bump to 1.35 kind

evaline-ju:
## Summary

Bump to 1.35.0 kind, using the image documented here: https://github.com/kubernetes-sigs/kind/releases/tag/v0.31.0

This updates APIs that were previously blocked from upgrade with the Kubernetes 1.32.x pin comments and the versions leveraged by the E2E tests. containerd usage also got upgraded so the mirroring had to change.

Local usage may require upgrading kind via `go install sigs.k8s.io/kind@v0.31.0`


evaline-ju:
I might've gotten a bit confused on the command for testing here

PR #770, opened 2026-02-26
📝 Consolidated compositional agent platform design proposal

cwiklik:
## Summary

- Adds consolidated design proposal for the Kagenti compositional agent platform architecture (`docs/design-proposals/consolidated-compositional-agent-platform-design.md`)
- Defines the two-CR model (`AgentRuntime` + `AgentCard`) atop a label-and-webhook foundation with layered defaults
- Documents workload-level explicit opt-in injection (`kagenti.io/inject: enabled`) replacing namespace-level injection
- Clearly separates admission-time configuration concerns from reconfiguration of running workloads
- Establishes ownership model: platform engineers own defaults (ConfigMaps), developers own `AgentRuntime` and `AgentCard` CRs
- Documents two independent pods: Kagenti Webhook Pod (`kagenti-webhook-system`) and Kagenti Operator Pod
- Captures open questions, risks, implementation horizons (short-term vs long-term), and unresolved design decisions

## Key Design Decisions

- **Injection trigger**: workload label `kagenti.io/inject: enabled` (current); CR-trigger alternative is an open question requiring further discussion
- **Defaults storage**: `kagenti-webhook-feature-gates` and `kagenti-webhook-defaults` ConfigMaps (short-term); `AgentRuntimeClusterConfig` / `AgentRuntimeConfig` CRDs planned for long-term
- **Webhook targeting**: Deployment/StatefulSet objects (current); pod-level targeting planned long-term for GitOps/CD compatibility
- **Dynamic reconfiguration**: supported where possible; sidecar image and init container changes require pod restarts

## Test plan

- [ ] Review document for accuracy against current implementation
- [ ] Verify open questions are correctly framed
- [ ] Confirm ownership model (developer vs platform engineer) is consistent throughout

### DIFF
--- a/docs/design-proposals/consolidated-compositional-agent-platform-design.md
+++ b/docs/design-proposals/consolidated-compositional-agent-platform-design.md
@@ -1,0 +1,1041 @@
+# Kagenti Implementation Design: Consolidated Compositional Agent Platform Architecture
+
+**Authors**: Kagenti Team
+
+**Begin Design Discussion**: 2026-02-20
+
+**Status**: Draft
+
+**Supersedes**: [compositional-agent-platform-design.md](compositional-agent-platform-design.md)
+
+**Checklist**:
+
+- [ ] AgentRuntime CR implementation (identity + observability)
+- [ ] AgentCard CR adaptation (selector change)
+- [ ] Mutating webhook implementation
+- [ ] Layered defaults mechanism (cluster / namespace)
+- [ ] Controller consolidation (istiod pattern)
+- [ ] Migration tooling
+- [ ] Documentation updates
+- [ ] Integration tests
+- [ ] E2E tests
+- [ ] Performance benchmarks
+
+---
+
+## Implementation Horizons
+
+This proposal distinguishes between **short-term** and **long-term** goals. The current design reflects what is practical to implement now, while acknowledging that certain capabilities will be introduced as the platform matures.
+
+### Short-Term (Current Design)
+
+The immediate goal is a working, secure composition model with minimal complexity:
+
+- **Explicit workload opt-in**: Developers label workloads with `kagenti.io/inject: enabled` to trigger AuthBridge sidecar injection. This is a conscious, visible act in the workload manifest.
+- **Webhook with ConfigMap-based defaults**: The webhook reads cluster-level defaults from two ConfigMaps in the `kagenti-webhook-system` namespace:
+  - `kagenti-webhook-feature-gates` — controls which AuthBridge components are enabled globally (`globalEnabled`, `envoyProxy`, `spiffeHelper`, `clientRegistration`)
+  - `kagenti-webhook-defaults` — provides default container images, proxy port configuration, and per-component resource requests/limits for all injected sidecars
+- **Two workload labels required**: `kagenti.io/inject: enabled` (injection opt-in) and `kagenti.io/type` (classification) are the only labels developers need to add.
+
+This is the model described in detail throughout this document.
+
+### Long-Term (Future Enhancements)
+
+As the platform matures, the following improvements are planned:
+
+- **CR-configured webhook defaults**: Two dedicated CRDs will replace static ConfigMaps with a proper Kubernetes-native API for managing defaults:
+  - `AgentRuntimeClusterConfig` (cluster-scoped) — stores cluster-wide defaults, set by cluster administrators
+  - `AgentRuntimeConfig` (namespace-scoped) — stores namespace-level overrides, deployed alongside workloads in each team namespace; overrides `AgentRuntimeClusterConfig` for any setting it specifies, set by cluster administrators
+
+
+- **Move webhook injection target from workload objects to pods**: The current `MutatingWebhookConfiguration` targets higher-level workload objects (`deployments`, `statefulsets`, `daemonsets`). This means the webhook mutates the pod template embedded inside those objects — the injected sidecar containers are stored directly in the Deployment's `spec.template` in etcd.
+
+  This creates a structural conflict with GitOps-based continuous delivery pipelines (Argo CD, Flux, and similar tools):
+
+  - CD tools continuously reconcile live Kubernetes resources against a Git source of truth (the Deployment manifest in a repository).
+  - When the webhook mutates a Deployment's pod template, the live Deployment diverges from the Git-stored manifest — the live object contains injected sidecars that the Git manifest does not.
+  - CD tools detect this as configuration drift. Depending on their configuration they either raise a continuous alert (false positive noise) or actively remediate the drift by reverting the mutation — stripping the injected sidecars on the next sync cycle and leaving pods without identity infrastructure.
+  - Additionally, mutating the Deployment's pod template can overwrite configuration the developer intentionally placed there — volume mounts, secret references, environment variables — because the webhook and the developer are competing to define the same object.
+
+  The proven alternative — used by Istio, Linkerd, and Vault Agent injector — is to register the webhook against `pods` at `CREATE` time instead:
+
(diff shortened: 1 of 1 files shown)


docs/design-proposals/consolidated-compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1211 @@
...
+
+**Current design**: Developer must set `kagenti.io/type: agent|tool` on the workload.
+
+**Proposed change**: The `type` field moves into the AgentRuntime CR spec. The operator (not the developer) applies the `kagenti.io/type` label to the workload based on the CR.
+
+This means:
+
+- The CR is the single authoritative declaration of "this is an agent" or "this is a tool"
+- The label becomes a derived artifact managed by the controller (useful for selectors, UI queries, AgentCard discovery)
+- If Argo CD's self-heal removes the label, the operator re-applies it on next reconcile -- no conflict because the operator is the owner
```
amfred: Is there any way to avoid reconciliation loops completely? 
I thought I heard that if the label is on the Pod instead of the Deployment, for example? 
(NB: I'm not an expert in this. But I do know that infinite reconcile loops are a problem. Otherwise Argo will never show the agent as fully "deployed".)

docs/design-proposals/consolidated-compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1211 @@
...
+- **Kustomize**: Base AgentRuntime template with overlays per namespace/agent
+- **Templates/examples**: Well-documented AgentRuntime templates in the repo
+
+This keeps the in-cluster model simple (one flat AgentRuntime per workload, no inheritance) and pushes composition complexity to tools designed for it (Helm, Kustomize).
+
+---
+
+### 4. Make `kagenti.io/type` Label Operator-Managed
+
+**Current design**: Developer must set `kagenti.io/type: agent|tool` on the workload.
```
amfred: Eventually, if we want to have multiple A2A agents within a single pod, would it really be the Route that we need to key off of for the AgentCard creation?

amfred: Those agents would end up with one SPIFFE identity too... 

You can resolve this comment, just something to think about.

docs/design-proposals/consolidated-compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1149 @@
...
+AgentCard remains a separate CR. It is reproduced here for completeness but is not modified from the original proposal.
+
+**Why separate**:
+- **Different cardinality**: AgentCard uses a label selector (can match multiple pods across workloads). AgentRuntime uses `targetRef` (1:1 with a workload). Forcing these into one CR would require supporting both targeting models in one resource.
+- **Different concern**: Discovery ("what can agents do") is distinct from runtime ("how are agents configured"). The name `AgentRuntime` does not naturally encompass capability discovery.
+- **Existing implementation**: Code exists and works. Refactoring it into a subsection of another CR is churn without benefit.
+
+**API Structure**:
+
+```yaml
```
pdettori, line 570: @amfred raised whether Route (not Pod) should be the key for AgentCard creation when multiple A2A agents share a single pod, and noted they would share a single SPIFFE identity. The "Different cardinality" paragraph discusses AgentCard's label selector model but doesn't address the multi-agent-per-pod scenario or shared identity implications. Consider adding a brief note acknowledging this as a future consideration — it doesn't need to be solved now but is worth capturing.

docs/design-proposals/consolidated-compositional-agent-platform-design.md
```diff
@@ -0,0 +1,1149 @@
...
+
+As the platform matures, the following improvements are planned. These are **not implemented in the current design** and are called out here to provide direction without overcomplicating the immediate implementation.
+
+The short-term design already implements: mandatory CR as source of truth, controller-managed labels, pod-level webhook targeting, and flat ConfigMap defaults. The remaining long-term items focus on sidecar consolidation, advanced config propagation, and tooling maturation.
+
+---
+
+### 1. ~~Make AgentRuntime CR Mandatory~~ (Moved to Short-Term)
+
+**Status: Part of MVP.** AgentRuntime CR is now mandatory in the short-term design. See Short-Term section above.
```
pdettori, line 56: nit: Per @rhuss's suggestion, the remaining long-term items (sidecar consolidation, xDS config propagation) could eventually move to a separate design doc. The current trimming is a reasonable intermediate step — just flagging for future cleanup.

pdettori (commented):
## Review: Document vs Comments Alignment Check

The latest commit (`6db115d`, March 10) substantially addresses the review feedback from @rhuss and @amfred. The document has been rewritten to reflect the MVP scope proposed in the March 6 comment.

### Feedback Incorporation Summary

| # | Reviewer | Feedback | Status |
|---|----------|----------|--------|
| 1 | @rhuss | `reinvocationPolicy` inconsistency | ✅ Fixed — both YAML and Risks now say `IfNeeded` |
| 2 | @rhuss | Namespace gating | ✅ Fixed — optional `kagenti-enabled: "true"` namespaceSelector |
| 3 | @rhuss | Webhook should target Pods | ✅ Fixed — moved to short-term, Pod CREATE targeting throughout |
| 4 | @rhuss | Lean v1alpha1 CRD | ✅ Fixed — scoped to `targetRef`, `identity`, `trace` |
| 5 | @rhuss | AgentRuntime deletion reverts to defaults | ✅ Fixed — finalizer-based, reverts to platform defaults |
| 6 | @rhuss | Future enhancements in separate doc | ⚠️ Partial — long-term section trimmed, items moved to short-term, but not fully separated |
| 7 | @rhuss | Sidecar count tracking | ✅ Fixed — new "Sidecar Consolidation Plan" section |
| 8 | @rhuss | Config propagation for non-Envoy sidecars | ✅ Fixed — explicitly flagged as open gap |
| 9 | @amfred | Reconciliation loops | ✅ Fixed — Pod-level targeting eliminates the concern |
| 10 | @amfred | Multiple A2A agents in single pod / Route-based AgentCard | ⚠️ Not addressed (noted as "just something to think about") |
| 11 | @amfred | Shared SPIFFE identity for multi-agent pods | ⚠️ Not addressed (noted as "just something to think about") |

Overall the document is in good shape. See inline comments for remaining suggestions.

grs:
Right, so creating the CR triggers injection, but deleting it does not undo that. My point was that to me this is not intuitive.

grs:
> @rhuss suggested this flow:

Yes, and I think that flow made sense when the injection was triggered by labels on the workload or namespace, as in that case the CR merely changed the configuration but did not itself cause injection.

However if the CR *does* cause injection, then reverting to defaults on deletion of that CR is not in my view what the user would expect.

I do think that having injection controlled by the presence of the CR is better. It is clear and explicit and the status of the CR can be used to indicate the status of injection.

If the finalizer for the CR is triggering a rollout (though update of the PodTemplateSpec), then why not have the webhook ignore the new pods since the CR that triggered injection is no longer there? That seems more intuitive.

(It would be ideal if the controller could trigger pod recreation without changing the workload at all, e.g. by just deleting pods that are affected by the creation or deletion of a CR.)

amfred:
Just adding my 10 cents late in the game. 

Overall I think the design changes are good. This should avoid infinite reconciliation loops.  The objectselector on the webhook - to avoid a cluster-wide performance cost - is good.  We may find some details that need to change once we actually get more hands-on experience with this, but I think it's close to correct and good enough for an MVP.

I have to defer to the people who are more hands-on with the code with respect to the implementation decisions that @grs and @cwiklik are discussing. I can't rule any of them out as a bad idea myself.

pdettori (approved):
lgtm

pdettori:
merging so we can update it to sync up with MVP implementation in a follow up PR.

Issue #781, opened 2026-02-28
:bug: kagenti-deps chart fails on OCP 4.19


sallyom:
## Description 

 `helm install kagenti-deps` fails on OCP 4.19+ (ZTWIM-managed SPIRE) because the chart's
  `spire-operand.yaml` hook puts `clusterName` and `trustDomain` on child operand CRs
  (SpireServer, SpireAgent, SpireOIDCDiscoveryProvider), but the ZTWIM CRD design centralizes
  these on the parent `ZeroTrustWorkloadIdentityManager` CR. Child CRDs reject unknown fields.

  ## Error

  failed to create typed patch object: .spec.clusterName: field not declared in schema

  ## Affected CRs

  | CR | Invalid Fields |
  |----|----------------|
  | SpireServer | `clusterName`, `trustDomain`, `persistence.type` |
  | SpireAgent | `clusterName`, `trustDomain` |
  | SpireOIDCDiscoveryProvider | `trustDomain`, `agentSocketName` (should be `csiDriverName`) |

  ## Environment

  - OCP 4.19.24 (ROSA)
  - ZTWIM operator installed via OLM
  - Helm 3.18+ and Helm 4.1.1

  ## Fix

  Remove centralized fields from child CRs (they inherit from the parent
  ZeroTrustWorkloadIdentityManager CR). PR incoming from `charts-clustername-fix` branch.

### Steps To Reproduce

  1. Provision an OCP 4.19+ cluster
  2. Detect trust domain:
     ```bash
     export DOMAIN="apps.$(kubectl get dns cluster -o jsonpath='{ .spec.baseDomain }')"
  3. Install kagenti-deps from OCI registry (or local chart):
  helm install kagenti-deps oci://ghcr.io/kagenti/kagenti/kagenti-deps \
    -n kagenti-system --create-namespace \
    --version 0.2.2 \
    --set spire.trustDomain=${DOMAIN} --wait
  4. Helm hooks apply spire-operand.yaml — SpireAgent CR fails with:
  failed to create typed patch object: .spec.clusterName: field not declared in schema

Tried both:
 - helm install ... oci://ghcr.io/kagenti/kagenti/kagenti-deps --version 0.2.2 (OCI registry)
  - helm install ... ./charts/kagenti-deps/ from the main branch (local)

### Expected Behavior

Chart deploys cleanly on OCP 4.19

### Additional Context

_No response_

PR #782, opened 2026-02-28
Fix: (charts) SPIRE CRD alignment + missing Keycloak keys in environments ConfigMap

sallyom:
_diagnosed and fixed with Claude Opus 4.6_

## Summary

On OpenShift 4.19, the ZTWIM operator CRDs have a different field layout than what the chart was generating.

Without this fix, helm install kagenti-deps fails on a fresh OCP 4.19 cluster with:
  failed to create typed patch object: .spec.clusterName: field not declared in schema

Tested on OCP 4.19.24.

## Related issue(s)

https://github.com/kagenti/kagenti/issues/781

## Additional fix: Add Keycloak keys to environments ConfigMap (6d4a4fd)

  The `environments` ConfigMap created by `agent-namespaces.yaml` was missing the individual `KEYCLOAK_*` keys that the AuthBridge client-registration sidecar expects via `configMapKeyRef`. Without these, the
  client-registration container fails with `CreateContainerConfigError` because it can't resolve `KEYCLOAK_REALM`, `KEYCLOAK_URL`, etc.

  This adds the missing keys using values from the existing `keycloak.*` helm values:
  - `KEYCLOAK_URL`, `KEYCLOAK_REALM`
  - `KEYCLOAK_ADMIN_USERNAME`, `KEYCLOAK_ADMIN_PASSWORD` (placeholder defaults)
  - `KEYCLOAK_TOKEN_EXCHANGE_ENABLED`, `KEYCLOAK_CLIENT_REGISTRATION_ENABLED`, `SPIRE_ENABLED`

  **Note:** On OCP, the bootstrap admin is `temp-admin` with a random password from the `keycloak-initial-admin` secret, not `admin/admin`. Deployers must patch the ConfigMap with actual credentials post-install. Longer term, admin credentials should move to a Secret with `secretKeyRef` in the webhook.

Tested on OCP 4.19 (ROSA) with Kagenti webhook-injected AuthBridge sidecars.

Fixes # https://github.com/kagenti/kagenti/issues/781

### DIFF
--- a/charts/kagenti/templates/agent-namespaces.yaml
+++ b/charts/kagenti/templates/agent-namespaces.yaml
@@ -113,6 +113,20 @@ metadata:
   labels:
     {{- include "kagenti.labels" $root | nindent 4 }}
 data:
+  # Keycloak keys for AuthBridge client-registration sidecar (injected by webhook).
+  # Admin credentials reference the Keycloak secret — the client-registration container
+  # reads these as plain values from the ConfigMap, so we inline the defaults here.
+  # TODO: move admin credentials to a Secret and use secretKeyRef in the webhook instead.
+  KEYCLOAK_URL: {{ $root.Values.keycloak.url | quote }}
+  KEYCLOAK_REALM: {{ $root.Values.keycloak.realm | default "demo" | quote }}
+  # NOTE: These are placeholder defaults. On RHBK (OpenShift), the bootstrap admin
+  # is "temp-admin" with a random password stored in the keycloak-initial-admin secret,
+  # not "admin/admin". Deployers must patch this ConfigMap with the actual credentials.
+  KEYCLOAK_ADMIN_USERNAME: "admin"
+  KEYCLOAK_ADMIN_PASSWORD: "admin"
+  KEYCLOAK_TOKEN_EXCHANGE_ENABLED: "true"
+  KEYCLOAK_CLIENT_REGISTRATION_ENABLED: "true"
+  SPIRE_ENABLED: "true"
   ollama: |
     [
       {"name": "LLM_API_BASE", "value": "http://host.docker.internal:11434/v1"},


charts/kagenti/templates/agent-namespaces.yaml
```diff
@@ -113,6 +113,20 @@ metadata:
...
+  # Admin credentials reference the Keycloak secret — the client-registration container
+  # reads these as plain values from the ConfigMap, so we inline the defaults here.
+  # TODO: move admin credentials to a Secret and use secretKeyRef in the webhook instead.
+  KEYCLOAK_URL: {{ $root.Values.keycloak.url | quote }}
+  KEYCLOAK_REALM: {{ $root.Values.keycloak.realm | default "demo" | quote }}
+  # NOTE: These are placeholder defaults. On RHBK (OpenShift), the bootstrap admin
+  # is "temp-admin" with a random password stored in the keycloak-initial-admin secret,
+  # not "admin/admin". Deployers must patch this ConfigMap with the actual credentials.
+  KEYCLOAK_ADMIN_USERNAME: "admin"
+  KEYCLOAK_ADMIN_PASSWORD: "admin"
```
pdettori, line 126: **must-fix**: `KEYCLOAK_ADMIN_USERNAME: "admin"` and `KEYCLOAK_ADMIN_PASSWORD: "admin"` are hardcoded plaintext credentials in a ConfigMap. ConfigMaps are not encrypted at rest and are readable by anyone with namespace read access.

The PR body and inline TODO acknowledge this is a placeholder, but shipping `admin/admin` as defaults is a security concern that could persist. At minimum, could these reference the existing `keycloak-initial-admin` Secret (already defined in values at `keycloak.adminSecretName`) via Helm lookup? Or could the webhook use `secretKeyRef` pointing at that existing secret instead of embedding credentials here?

charts/kagenti/templates/agent-namespaces.yaml
```diff
@@ -113,6 +113,20 @@ metadata:
   labels:
     {{- include "kagenti.labels" $root | nindent 4 }}
 data:
+  # Keycloak keys for AuthBridge client-registration sidecar (injected by webhook).
+  # Admin credentials reference the Keycloak secret — the client-registration container
+  # reads these as plain values from the ConfigMap, so we inline the defaults here.
+  # TODO: move admin credentials to a Secret and use secretKeyRef in the webhook instead.
+  KEYCLOAK_URL: {{ $root.Values.keycloak.url | quote }}
+  KEYCLOAK_REALM: {{ $root.Values.keycloak.realm | default "demo" | quote }}
```
pdettori, line 121: **suggestion**: The `default "demo"` here is inconsistent with `values.yaml` which defaults `keycloak.realm` to `"master"`. This fallback would only trigger if someone explicitly sets `keycloak.realm` to empty/null, which could cause confusing behavior. Consider removing the `| default "demo"` and relying on the `values.yaml` default, or update `values.yaml` to match if `"demo"` is the intended default.

charts/kagenti-deps/templates/spire-operand.yaml
```diff
@@ -51,14 +53,12 @@ metadata:
     "helm.sh/hook-weight": "1"
     "helm.sh/hook-delete-policy": before-hook-creation
 spec:
-  trustDomain: {{ .Values.spire.trustDomain }}
-  clusterName: agent-platform
+  # trustDomain and clusterName are inherited from the ZeroTrustWorkloadIdentityManager CR
   caSubject:
     commonName: {{ .Values.spire.trustDomain }}
     country: "US"
```
pdettori, line 59: **suggestion**: `persistence.type: pvc` was removed along with `trustDomain` and `clusterName`, but the commit message only mentions the CRD schema alignment for the latter two fields. Was `persistence.type` also rejected by the OCP 4.19 CRD? Worth confirming that the ZTWIM operator defaults to `pvc` when the field is omitted, or adding a brief note to the commit/comment.

charts/kagenti-deps/templates/spire-operand.yaml
```diff
@@ -100,7 +99,7 @@ metadata:
     "helm.sh/hook-weight": "1"
     "helm.sh/hook-delete-policy": before-hook-creation
 spec:
-  trustDomain: {{ .Values.spire.trustDomain }}
-  agentSocketName: 'spire-agent.sock'
+  # trustDomain is inherited from the ZeroTrustWorkloadIdentityManager CR
+  csiDriverName: csi.spiffe.io
```
pdettori, line 103: **nit**: `csiDriverName: csi.spiffe.io` replaces `agentSocketName: 'spire-agent.sock'` — this substitution isn't mentioned in the commit message. A brief note about why this field changed (new CRD field name? different semantics?) would help future readers.

charts/kagenti-deps/templates/spire-operand.yaml
```diff
@@ -6,7 +6,9 @@
   (SpireServer, SpireAgent, SpiffeCSIDriver, SpireOIDCDiscoveryProvider) can be configured.
 
   This CR defines the cluster-wide SPIRE configuration including trust domain and cluster name.
-  The individual operands reference this configuration.
+  The individual operands inherit trustDomain and clusterName from this parent CR —
+  those fields must NOT be set on the child operands (SpireServer, SpireAgent, etc.)
```
pdettori, line 10: **praise**: Good addition explaining that child operands inherit `trustDomain` and `clusterName` from the parent CR and must NOT set them. This prevents future contributors from re-adding these fields.

Issue #619, opened 2026-02-05
epic: Tracing observability PoC with K8s agentic networking


evaline-ju:
### Problem

The upstream [k8s agentic networking group](https://github.com/kubernetes-sigs/kube-agentic-networking) has been working on proposals around access policies related to [tool authorization](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/main/docs/proposals/0008-ToolAuthAPI.md) and [tracing observability](https://github.com/kubernetes-sigs/kube-agentic-networking/pull/33).

A [quickstart](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/main/quickstart/quickstart.md) was put together with access policies for tools. We would like to leverage this for trying out a PoC on tracing observability while leveraging Kagenti components, allowing for benefits to both the upstream community and Kagenti plans.

### design

The light purple portions are the parts we would like to leverage from Kagenti examples and MCP gateway:
<img width="1327" height="896" alt="Image" src="https://github.com/user-attachments/assets/728764fc-e92f-4f4e-9be2-6d3a51861879" />

This is an update of the mermaid diagram from the quickstart:
```
---
config:
  theme: mc
---
graph TD
    subgraph Kubernetes Cluster
        subgraph agentic-net-system ns
            Controller(agentic-networking-controller)
        end

        subgraph quickstart-ns
            Agent(Kagenti Agent)
            Envoy(MCP gateway)
            Envoy --> Decision{Policy Allowed?}
            LocalMCP(Local MCP Server)
            LocalKagentiMCP(Kagenti MCP tool)
        end
        K8sAPI(Kubernetes API Server)
    end

    User([User]) -- "1. kubectl apply" --> K8sAPI
    K8sAPI -- "2. Watches resources" --> Controller
    Controller -- "3. Provisions Envoy" --> Envoy
    Controller -- "4. Translates policy to xDS config" --> Envoy

    Agent -- "5. Tool call" --> Envoy

    Decision -- "Yes" --> LocalMCP
    Decision -- "Yes" --> LocalKagentiMCP(Kagenti MCP tool)
    Decision -- "Yes" --> RemoteMCP(Remote MCP Server)
    Decision -- "No" --> Forbidden(Response: 403 Forbidden)

    style User fill:#f9f,stroke:#333,stroke-width:2px
    style RemoteMCP fill:#f9f,stroke:#333,stroke-width:2px
[...]

PR #795, opened 2026-03-02
feat: Configure Istio to export Envoy traces to OTel Collector for Kind clusters

evaline-ju:
## Summary

- meshConfig change - register an OTel tracing provider with istiod (Istio's control plane)
- Telemetry CRD activates tracing mesh-wide

This allows Envoy proxy traces (including MCP Gateway requests) to flow through to observability platforms alongside agent-level OTel traces. Focusing the change only on Kind clusters for now [MCP gateway router tracing is not yet in the latest 0.5 release, and there may still be issues with e2e MCP gateway usage on OpenShift/HyperShift]

## Related issue(s)

Related to initiative of tracing agent->gateway->tool in #619

### DIFF
--- a/charts/kagenti-deps/templates/istio-telemetry.yaml
+++ b/charts/kagenti-deps/templates/istio-telemetry.yaml
@@ -10,5 +10,5 @@ spec:
   tracing:
   - providers:
     - name: otel-tracing
-    randomSamplingPercentage: 100
+    randomSamplingPercentage: {{ .Values.istio.tracing.randomSamplingPercentage }}
 {{- end }}
--- a/deployments/ansible/roles/kagenti_installer/tasks/main.yml
+++ b/deployments/ansible/roles/kagenti_installer/tasks/main.yml
@@ -98,8 +98,8 @@
             extensionProviders:
             - name: otel-tracing
               opentelemetry:
-                service: otel-collector.kagenti-system.svc.cluster.local
-                port: 4317
+                service: "{{ charts['kagenti-deps']['values']['otel']['collector']['service'] }}"
+                port: "{{ charts['kagenti-deps']['values']['otel']['collector']['port'] }}"
       register: istiod_result
       retries: 3
       delay: 15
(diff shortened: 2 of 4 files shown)


charts/kagenti-deps/templates/istio-telemetry.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: mesh-tracing
+  namespace: {{ .Values.istio.namespace | default "istio-system" }}
+  labels:
+    {{- include "kagenti.labels" . | nindent 4 }}
+spec:
+  tracing:
+  - providers:
+    - name: otel-tracing
+    randomSamplingPercentage: 100
```
pdettori: **suggestion**: `randomSamplingPercentage: 100` is fine for Kind dev clusters. Consider adding a `.Values` override (e.g., `.Values.istio.traceSamplingPercentage`) to make it configurable for future environments. Low priority since this is guarded by `not .Values.openshift`.

deployments/ansible/roles/kagenti_installer/tasks/main.yml
```diff
@@ -94,6 +45,12 @@
         timeout: "{{ helm_wait_timeout }}s"
         values:
           profile: ambient
+          meshConfig:
+            extensionProviders:
+            - name: otel-tracing
+              opentelemetry:
+                service: otel-collector.kagenti-system.svc.cluster.local
```
pdettori: **nit**: The service FQDN `otel-collector.kagenti-system.svc.cluster.local` is hardcoded. If the `kagenti-deps` namespace ever changes, this would silently break tracing. Consider using an Ansible variable for the namespace. Low risk for Kind-only.

charts/kagenti-deps/templates/otel-collector.yaml
```diff
@@ -158,6 +158,8 @@ data:
     receivers:
       otlp:
         protocols:
+          grpc:
```
pdettori, line 161: **praise**: Good addition - the gRPC receiver was missing from the config despite the container port and Service already exposing 4317. This aligns the config with the existing infrastructure.

Issue #636, opened 2026-02-10
feature: It's too hard to "kick the tires" with the Kagenti examples


esnible:
### Feature Description

Following the [weather example](https://github.com/kagenti/kagenti/blob/main/docs/demos/demo-weather-agent.md) is straightforward when building from source.

However, it takes two minutes to build that agent.  I installed and uninstalled the weather agent 25 times yesterday for testing.  I would like to be able to build zero or one times when using the examples.

Kagenti offers a way to deploy an existing image.  I first tried `registry.cr-system.svc.cluster.local:5000/weather-service:v0.0.1` but it didn't work.  I am not sure if the problem was that container image being removed from the registry, or if it is a problem with the tag.

I received guidance from @evaline-ju that https://github.com/kagenti/agent-examples/pkgs/container/agent-examples%2Fweather_service documents well-known images.  I was unable to find those names from the Kagenti docs.  (I spent only 15 minutes looking for it.)

I tried to deploy that image using the following steps:

- Browse to http://kagenti-ui.localtest.me:8080/agents/import
  - Deploy from Existing Image, ghcr.io/kagenti/agent-examples/weather_service:v0.0.1-alpha.3
  - **DON'T** Remove "latest" tag
  - LLM_API_BASE=http://host.docker.internal:11434/v1
  - LLM_MODEL=gpt-oss:latest

This didn't work:

```
oc -n team1 get pods                                                            
NAME                               READY   STATUS             RESTARTS   AGE
debug-snible                       1/1     Running            0          17h
weather-service-68c48d7649-w274z   0/1     InvalidImageName   0          38s
```

It wasn't clear to me at first that `ghcr.io/kagenti/agent-examples/weather_service:v0.0.1-alpha.3` has no `:latest` tag.  It was not obvious as a Kagenti user what had gone wrong -- the pod did not become ready.

I next tried

- Browse to http://kagenti-ui.localtest.me:8080/agents/import
  - Deploy from Existing Image, ghcr.io/kagenti/agent-examples/weather_service:v0.0.1-alpha.3
  - Remove "latest" tag
  - LLM_API_BASE=ttp://host.docker.internal:11434/v1
  - LLM_MODEL=gpt-oss:latest

The pod did start, but it didn't work, due to the cut-and-paste error on the `LLM_API_BASE`.  (After much inspection I redeployed correctly and it worked.).

(If there is interest from the team I will modify the weather example to return error messages via the A2A protocol rather than crashing, which will make it easier to debug.)

[...]

PR #799, opened 2026-03-02
🐛 Validate container image and tags

### LINKED ISSUE #755: UI: Validate container image name to prevent colon in repository name
## Problem

When deploying agents or tools from container images via the UI, the "Container Image" field accepts any string, including colons (`:`). Since the UI separately appends the image tag as `${containerImage}:${imageTag}`, entering a colon in the image name (e.g., `quay.io/myorg/my-tool:latest`) produces a double-tagged reference like `quay.io/myorg/my-tool:latest:v0.0.1`.

Kubernetes rejects this as `InvalidImageName`, but the error only surfaces in the pod status — there is no error visible in the Kagenti UI. The agent or tool gets stuck in "Progressing" state with no indication of what went wrong.

## Expected Behavior

The UI should validate the "Container Image" field against the [OCI distribution-spec repository name grammar](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#workflow-categories):

```
<repository-name> := [a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*(\/[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*)*
```

At minimum, the image name field should:
1. **Reject colons** (`:`) in the container image name — the tag is entered in a separate field
2. **Show an inline validation error** when the image name contains invalid characters
3. **Prevent form submission** if the image name is invalid

## Current Behavior

- User enters image name with a colon (e.g., `quay.io/myorg/my-tool:latest`)
- UI appends the tag from the separate "Image Tag" field, producing `quay.io/myorg/my-tool:latest:v0.0.1`
- Kubernetes reports `InvalidImageName` on the pod
- The agent/tool status shows "Progressing" indefinitely with no UI-visible error

## Affected Files

- `kagenti/ui-v2/src/pages/ImportToolPage.tsx` — `handleImageChange()` and `validateForm()` (around lines 219-224 and 383-391)
- `kagenti/ui-v2/src/pages/ImportAgentPage.tsx` — `handleImageChange()` and `validateForm()` (around lines 219-226 and 403-411)

Both files combine the image name and tag in `handleSubmit`:
```typescript
const fullImage = imageTag ? `${containerImage}:${imageTag}` : containerImage;
```

## Suggested Implementation

Add validation to the `containerImage` field similar to the existing `isValidEnvVarName` pattern already used in both files. For example:

```typescript
const isValidImageName = (image: string): boolean => {
  if (!image) return false;
  // OCI distribution-spec repository name pattern (no colon allowed)
  const pattern = /^[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*(\/[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*)*$/;
  return pattern.test(image);
};
```

Update the `validated.containerImage` check in `validateForm()` to reject invalid image names, and show an appropriate error message in the form helper text.

## References

- OCI Distribution Spec — repository names: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#workflow-categories
- Repository name regex: `[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*(\/[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*)*`

esnible:
## Summary

Add client-side validation for container image name and tag fields on the Import Tool and Import Agent pages. The image name is validated against the OCI distribution-spec repository name grammar, and the image tag is validated against the OCI tag format. This prevents users from entering colons or other invalid characters that cause Kubernetes `InvalidImageName` errors, leaving agents/tools stuck on "Progressing" with no UI-visible error.

Key changes:
- Add `isValidImageName()` validation
- Add `isValidImageTag()` validation
- Update `validateForm()` to reject invalid image names and tags before submission
- Show inline error messages with guidance on allowed characters
- Add test cases

## Related issue(s)

Resolves #755
Mitigates #636
Replaces https://github.com/kagenti/kagenti/pull/756

### DIFF
--- a/kagenti/ui-v2/src/pages/ImportToolPage.tsx
+++ b/kagenti/ui-v2/src/pages/ImportToolPage.tsx
@@ -3,7 +3,7 @@
 
 import React, { useState } from 'react';
 import { useNavigate } from 'react-router-dom';
-import { isValidEnvVarName } from '../utils/validation';
+import { isValidEnvVarName, isValidContainerImage, isValidImageTag } from '../utils/validation';
 import {
   PageSection,
   Title,
@@ -374,13 +374,21 @@ export const ImportToolPage: React.FC = () => {
         newValidated.registryNamespace = 'success';
       }
     } else {
-      // Container image validation for image deployment
-      if (!containerImage) {
+      // Container image validation: must match [HOST[:PORT]/]NAMESPACE/REPOSITORY[/…]
+      if (!containerImage || !isValidContainerImage(containerImage)) {
         newValidated.containerImage = 'error';
         isValid = false;
       } else {
         newValidated.containerImage = 'success';
       }
+    }
+
+    // Image tag validation (shared by both deployment methods)
+    if (imageTag && !isValidImageTag(imageTag)) {
+      newValidated.imageTag = 'error';
+      isValid = false;
+    } else if (imageTag) {
+      newValidated.imageTag = 'success';
     }
 
     setValidated(newValidated);
@@ -687,7 +695,17 @@ export const ImportToolPage: React.FC = () => {
                       value={imageTag}
                       onChange={(_e, value) => setImageTag(value)}
                       placeholder="v0.0.1"
+                      validated={validated.imageTag}
                     />
+                    <FormHelperText>
+                      <HelperText>
+                        <HelperTextItem variant={validated.imageTag === 'error' ? 'error' : 'default'}>
+                          {validated.imageTag === 'error'
+                            ? 'Letters, digits, underscores, periods, and dashes only. May not start with a period or dash.'
+                            : 'Tag to apply to the image (e.g., v0.0.1)'}
+                        </HelperTextItem>
+                      </HelperText>
+                    </FormHelperText>
                   </FormGroup>
 
                   <Divider style={{ margin: '24px 0' }} />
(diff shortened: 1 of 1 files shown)


kagenti/ui-v2/src/utils/validation.ts
```diff
@@ -12,3 +12,42 @@ export const isValidEnvVarName = (name: string): boolean => {
...
+
+/**
+ * Validate an image tag.
+ *
+ * Must be valid ASCII containing only letters, digits, underscores,
+ * periods, and dashes. May not start with a period or a dash.
+ */
+export const isValidImageTag = (tag: string): boolean => {
+  if (!tag) return false;
+  return /^[a-zA-Z0-9_][a-zA-Z0-9._-]*$/.test(tag);
```
pdettori, line 52: **suggestion**: The OCI distribution-spec limits tags to 128 characters. Currently a 200-char tag would pass client validation but fail in Kubernetes — partially defeating the purpose of client-side validation.

Consider adding a length check:

```ts
if (tag.length > 128) return false;
```

Or adjust the regex to `/^[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}$/`.

kagenti/ui-v2/src/utils/validation.test.ts
```diff
@@ -45,3 +45,121 @@ describe('isValidEnvVarName', () => {
...
+    expect(isValidImageTag(' latest')).toBe(false);
+  });
+
+  it('rejects tags with non-ASCII or special characters', () => {
+    expect(isValidImageTag('tag@1')).toBe(false);
+    expect(isValidImageTag('tag:1')).toBe(false);
+    expect(isValidImageTag('tag/1')).toBe(false);
+    expect(isValidImageTag('tag!')).toBe(false);
+    expect(isValidImageTag('tàg')).toBe(false);
+  });
```
pdettori, line 164: **suggestion**: Consider adding boundary tests for tag length once the 128-char limit is added:

```ts
it('accepts tag at max length (128 chars)', () => {
  expect(isValidImageTag('a'.repeat(128))).toBe(true);
});

it('rejects tag exceeding max length', () => {
  expect(isValidImageTag('a'.repeat(129))).toBe(false);
});
```

PR #805, opened 2026-03-03
fix: pass KEYCLOAK_URL to oauth-secret job in OpenShift mode

### LINKED ISSUE #804: OAuth secret job fails on OpenShift due to SSL certificate verification error
## Bug Description

When deploying Kagenti on OpenShift with `openshift: true`, the `kagenti-ui-oauth-secret-job` fails with an SSL certificate verification error, preventing the `kagenti-ui-oauth-secret` from being created. This causes the `kagenti-backend` pod to remain stuck in `CreateContainerConfigError`.

## Root Cause

In the Helm template `charts/kagenti/templates/ui-oauth-secret-job.yaml`, the `KEYCLOAK_URL` environment variable is only passed when `openshift` is `false`:

```yaml
{{- if not .Values.openshift }}
- name: KEYCLOAK_URL
  value: {{ .Values.keycloak.url | quote }}
- name: KEYCLOAK_PUBLIC_URL
  ...
{{- end }}
```

Without `KEYCLOAK_URL`, the Job falls back to discovering the Keycloak URL from the OpenShift route (external HTTPS). The route's TLS certificate is signed by the OpenShift ingress controller CA, which is not included in the service account CA bundle (`/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`), causing:

```
SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get issuer certificate
```

The Job silently fails all retries and is cleaned up by its 300-second TTL, leaving no visible trace.

## Impact

- `kagenti-ui-oauth-secret` is never created
- `kagenti-backend` pod stuck in `CreateContainerConfigError`
- Platform is non-functional after install on OpenShift

## Expected Behavior

The Job should use the internal Keycloak service URL (`http://keycloak-service.keycloak:8080`) for server-to-server admin API calls, and the route URL only for browser-facing auth endpoints stored in the secret. The Python code in `auth_secret.py` already supports this dual-URL pattern.

## Steps to Reproduce

1. Deploy Kagenti on OpenShift with `openshift: true` in Helm values
2. Observe `kagenti-ui-oauth-secret-job` fails (check pod logs before TTL cleanup)
3. `kagenti-backend` stays in `CreateContainerConfigError`

## Environment

- OpenShift 4.x with edge-terminated TLS routes
- Keycloak deployed with a route in the `keycloak` namespace

pdettori:
## Summary

- Fix SSL certificate verification failure in `kagenti-ui-oauth-secret-job` on OpenShift
- Move `KEYCLOAK_URL` env var outside the `openshift` conditional in the Helm template so the job uses the internal Keycloak service URL for admin API calls instead of the external HTTPS route
- The route's TLS certificate (signed by the OpenShift ingress CA) is not in the pod's service account CA bundle, causing the job to fail silently

## Root Cause

The Helm template `charts/kagenti/templates/ui-oauth-secret-job.yaml` excluded `KEYCLOAK_URL` when `openshift=true`. Without it, the job discovered the Keycloak URL from the OpenShift route (HTTPS) and failed with `SSL: CERTIFICATE_VERIFY_FAILED`. The job exhausted its retries and was cleaned up by the 300s TTL, leaving no visible trace. The missing secret caused `kagenti-backend` to stay in `CreateContainerConfigError`.

## Test plan

- [ ] Deploy on OpenShift with `openshift: true` and verify `kagenti-ui-oauth-secret` is created
- [ ] Verify `kagenti-backend` starts successfully
- [ ] Verify Kind deployment still works (non-OpenShift path unchanged)

Closes #804

### DIFF
--- a/charts/kagenti/templates/ui-oauth-secret-job.yaml
+++ b/charts/kagenti/templates/ui-oauth-secret-job.yaml
@@ -100,9 +100,9 @@ spec:
           value: {{ .Values.keycloak.adminUsernameKey | quote }}
         - name: KEYCLOAK_ADMIN_PASSWORD_KEY
           value: {{ .Values.keycloak.adminPasswordKey | quote }}
-        {{- if not .Values.openshift }}
         - name: KEYCLOAK_URL
           value: {{ .Values.keycloak.url | quote }}
+        {{- if not .Values.openshift }}
         - name: KEYCLOAK_PUBLIC_URL
           value: {{ .Values.keycloak.publicUrl | quote }}
         - name: ROOT_URL


Ladas (approved):
👍

PR #835, opened 2026-03-04
Feat: Add a command line interface / terminal user interface to Kagenti

### LINKED ISSUE #775: feature: add a TUI for kagenti
### Feature Description

Now that we have a backend for the UI, we can easily generate a nice TUI (Text-based/Terminal User Interface). This would be a very useful thing for someone like [me] who tends to work in a terminal.

### Proposed Solution

Generate a TUI using a nice framework e.g., bubbletea that simply connects to the backend server. Along with a nice setup script. It should be optional -- maybe `make install-tui` and introduce no cross-dependencies. We should list it as experimental in the first version, while we discover if folks find it useful.

### Want to contribute?

### Additional Context

I'd like to work on this, since it's my own use case.

usize:
## Summary

Kagenti's UI now operates on an API backend. This creates the possibility for a variety of user interface experiences. It also opens the possibility of automating things like agent deployment and connection without the need for applying a k8s manifest.

To achieve that, here we introduce a single `kagenti-tui` which may operate in interactive or non-interactive modes. It allows for authentication via Oauth + KeyCloak. Because this authentication may persist between flows, it creates the possibility for simple automation. 

We may follow up with additional enhancements e.g., API keys to further enhance its utility for automation purposes.

## Related issue(s)

Fixes #775 

`cd kagenti/tui && make test`


kagenti/tui/internal/api/client.go
```diff
@@ -0,0 +1,201 @@
...
+		Token:     token,
+		Namespace: namespace,
+		HTTPClient: &http.Client{
+			Timeout: 30 * time.Second,
+		},
+	}
+}
+
+// SetToken updates the auth token.
+func (c *Client) SetToken(token string) {
```
pdettori, line 61: **must-fix: Data race on Token/RefreshToken**

`SetToken()` and `SetRefreshToken()` write to `c.Token` and `c.RefreshToken` without synchronization, while `refreshAccessToken()` protects writes with `refreshMu`. Meanwhile `newRequest()` reads `c.Token` without the mutex. If `do()` is ever called from multiple goroutines, this is a data race.

Options:
1. Protect all reads/writes to Token/RefreshToken with the mutex (make them unexported, add `token()`/`setToken()` methods that lock)
2. Document that Client is not goroutine-safe and ensure single-goroutine usage throughout

A common pattern for option 1: inside `refreshAccessToken`, after acquiring the lock, compare the current token to the one that triggered the 401 — if they differ, another goroutine already refreshed it.

kagenti/tui/internal/api/chat.go
```diff
@@ -0,0 +1,140 @@
...
+	var resp ChatResponse
+	if err := c.do(req, &resp); err != nil {
+		return nil, err
+	}
+	return &resp, nil
+}
+
+// StreamChat opens an SSE connection and sends events to the returned channel.
+// The channel is closed when the stream ends or on error.
+func (c *Client) StreamChat(namespace, name string, chatReq *ChatRequest) (<-chan ChatStreamEvent, error) {
```
pdettori, line 65: **must-fix: StreamChat bypasses auth refresh**

This method creates a separate `http.Client` and builds the request via `http.NewRequest` directly instead of `c.newRequest()`. This skips the `do()` retry-on-401 logic entirely. If the access token expires before starting a stream, the user gets a raw 401 with no refresh attempt.

Suggestion: use `c.newRequest()` for header injection, and add at least a single refresh-retry before opening the stream (check for 401, call `refreshAccessToken()`, retry once).

kagenti/tui/internal/cli/root.go
```diff
@@ -0,0 +1,129 @@
...
+
+	ctx := &CLIContext{}
+
+	root := &cobra.Command{
+		Use:   "kagenti",
+		Short: "Kagenti CLI — deploy and manage AI agents",
+		Long:  "Kagenti is a cloud-native middleware platform for deploying and orchestrating AI agents.\nRun without a subcommand to launch the interactive TUI.",
+		// SilenceUsage prevents cobra from printing usage on every error.
+		SilenceUsage: true,
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
```
pdettori, line 63: **suggestion: Duplicated client initialization**

`PersistentPreRunE` and `RunE` both independently create a client with identical config load → client creation → refresh token → OnTokenRefresh callback logic. If one is updated but not the other, they will drift. Consider extracting a `newClientFromConfig(cfg *config.Config) *api.Client` helper.

kagenti/tui/internal/api/agents.go
```diff
@@ -0,0 +1,79 @@
...
+	"fmt"
+)
+
+// ListAgents lists agents in the given namespace (or client's default).
+func (c *Client) ListAgents(namespace string) (*AgentListResponse, error) {
+	ns := namespace
+	if ns == "" {
+		ns = c.Namespace
+	}
+	url := c.apiURL(fmt.Sprintf("/agents?namespace=%s", ns))
```
pdettori: **suggestion: Unescaped query parameter**

`fmt.Sprintf("/agents?namespace=%s", ns)` does not URL-encode the namespace. Same pattern in `tools.go`. Use `url.Values` or `url.QueryEscape()` for safety:
```go
q := url.Values{"namespace": {ns}}
url := c.apiURL("/agents?" + q.Encode())
```

kagenti/tui/internal/api/auth.go
```diff
@@ -0,0 +1,183 @@
...
+	if err := json.NewDecoder(resp.Body).Decode(&dcr); err != nil {
+		return nil, err
+	}
+	dcr.CodeVerifier = verifier
+	return &dcr, nil
+}
+
+// PollDeviceToken polls Keycloak's token endpoint for device code completion.
+// It blocks until the user authorizes, the code expires, or an error occurs.
+func (c *Client) PollDeviceToken(keycloakURL, realm, clientID, deviceCode, codeVerifier string, interval int) (*TokenResponse, error) {
```
pdettori: **suggestion: No cancellation in polling loop**

`PollDeviceToken` loops with `time.Sleep` and has no `context.Context` parameter. An interrupted CLI (Ctrl+C) will not cancel the polling goroutine cleanly. Consider accepting a `context.Context` and using `http.NewRequestWithContext`.

kagenti/tui/internal/config/config.go
```diff
@@ -0,0 +1,90 @@
...
+// Flag values should be passed in; empty string means "not set".
+func Load(flagURL, flagToken, flagNamespace string) *Config {
+	cfg := &Config{
+		URL:       defaultURL,
+		Namespace: defaultNamespace,
+	}
+
+	// Layer 1: config file
+	if data, err := os.ReadFile(configPath()); err == nil {
+		_ = yaml.Unmarshal(data, cfg)
```
pdettori: **suggestion: Silent unmarshal failure**

When `yaml.Unmarshal` fails on a corrupted config file, the error is silently discarded and defaults are used. A warning log (or at minimum a comment explaining the intentional fallback) would help users debug config issues.

kagenti/tui/internal/cli/delete.go
```diff
@@ -0,0 +1,85 @@
...
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+// confirmDelete prompts the user to confirm a destructive action.
+// Returns true if the user confirms or if yes is already set.
+var confirmDelete = func(kind, name string) bool {
```
pdettori: **suggestion: Confirmation does not show namespace**

The delete confirmation prompt shows only the resource name. For a destructive operation, consider including the target namespace so users know exactly what they are deleting:
```
Delete agent "my-agent" in namespace "team1"? [y/N]
```

.github/workflows/ci.yaml
```diff
@@ -63,3 +63,29 @@ jobs:
...
+        working-directory: kagenti/tui
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6
+
+      - name: Set up Go
+        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5  # v5
+        with:
+          go-version: '1.23'
```
pdettori: **nit**: Consider `go-version-file: kagenti/tui/go.mod` instead of hardcoding `'1.23'` to keep the CI and module in sync. Also, Go 1.23 is EOL — worth bumping to 1.24.

pdettori (changes requested):
## Review Summary

This PR adds a comprehensive Go-based TUI and CLI for Kagenti (~6,700 lines across 52 files). The implementation is well-structured with Bubble Tea for the interactive TUI, Cobra for the CLI, Keycloak device-code auth with PKCE, and good test coverage. CI is all green. All 18 commits are signed-off.

**Areas reviewed**: Go, Python, CI, Docs, Security
**Commits**: 18 commits, all signed-off ✅
**CI status**: All automated checks passing ✅ (E2E HyperShift pending manual trigger)

Two must-fix items flagged below (data race on Token fields and StreamChat bypassing auth refresh). The remaining items are suggestions/nits.

🤖 Reviewed with Claude Code

usize:
@pdettori @Ladas let's try the automated diff scans again to make sure we covered it all. I went through the results, but would like a double check since it's a lot of commits. \o/

Makefile
```diff
@@ -161,4 +161,20 @@ preload-file: preflight-check
...
+	$(MAKE) -C kagenti/tui build
+
+install-tui:
+	$(MAKE) -C kagenti/tui install
+
+lint-tui:
+	$(MAKE) -C kagenti/tui lint
+
+test-tui:
+	$(MAKE) -C kagenti/tui test
```
pdettori: Nit: Missing newline at end of file.

kagenti/backend/app/routers/agents.py
```diff
@@ -1895,6 +1895,8 @@ def _build_common_labels(
     Returns:
         Dictionary of labels.
     """
+    from app.core.constants import PROTOCOL_LABEL_PREFIX
```
pdettori: Suggestion: `PROTOCOL_LABEL_PREFIX` is imported inside the function body (presumably to avoid a circular import). If there's no circular dependency concern, consider moving it to the top-level imports for consistency with the rest of the file.

Issue #623, opened 2026-02-06
epic: Identify Emerging Agentic Deployment Patterns


kellyaa:
### Problem

Identify, structure, and prioritize emerging agentic deployment patterns to establish a shared understanding of which patterns matter most for Kagenti and why. This work focuses on discovery, synthesis, and decision-making, producing clear artifacts that reduce ambiguity and set up downstream pilot and implementation work.

### design

Document emerging agentic deployment patterns observed across user needs, early adopter experiments, and platform exploration. Organize these patterns along a set of dimensions—used as an analytical lens rather than a fixed taxonomy. Some preliminary examples of such dimensions of deployment patterns:

- Agent kickoff driver (human-, event-, or agent-initiated; role-based vs. delegated identity)
- Stateful vs. stateless execution
- Short-running vs. long-running agents (including checkpointing and recovery considerations)
- Synchronous vs. asynchronous interactions
- Single-tenant vs. multi-tenant execution
- Tool authentication models (vaulted credentials vs. dynamic OAuth flows)
- Single-agent vs. multi-agent coordination (e.g., A2A / MCP-style)
- Deployment topology (single container, multi-container, or driver + ephemeral sandbox execution)

Define a prioritization rubric that evaluates patterns based on impact, feasibility, differentiation, and alignment with Kagenti’s platform strengths. Apply this rubric to the documented patterns to select the most high priority deployment patterns for pilot exploration.

### Goals and non-goals

Goals:
- A clearly documented and prioritized set of emerging agentic deployment patterns
- A shared evaluative framework for comparing them
- A justified selection of pilot pattern candidates that directly enables downstream implementation and validation work.

Non Goals:
- Running any of these new patterns/agents in Kagenti -- Just yet. That will be handled by future epics.

### Success criteria

- Deployment patterns documented and organized along agreed dimensions
- Prioritization rubric defined and reviewed
- Patterns scored using the rubric
- Potential pilot deployment patterns selected with decision rationale captured

### Stories to be associated with this epic

- [ ] [Survey and document emerging agentic deployment patterns](https://github.com/kagenti/kagenti/issues/1136)
- [ ] Define prioritization rubric for deployment patterns
- [ ] Apply rubric and select 3 pilot patterns

PR #840, opened 2026-03-05
docs: Kagenti Use Case Types

### LINKED ISSUE #789: feature: Collect and categorize authorization and policy use cases
## Summary

Build and maintain a structured, categorized catalog of Kagenti authorization and policy use-cases to drive prioritization and design decisions.

## Categories

- Identity Provider (IdP) agnosticism and delegation at scale
- Fine-grained authorization: scope-based, attribute based access control (ABAC), relationship based access control (ReBAC), data aware access control.
- User-managed access and consent workflows
- Asynchronous and multi-party authorization
- Workload-attribute-aware policy checks
- Continuous authorization for long-running agents
- Cross-layer observability and auditability

## Deliverables

- A use-case matrix with persona, scenario, policy requirements, and success criteria
- Topic-based grouping and prioritization guidance
- Traceability from use-cases to feature tracks under the epic
- Review identity and authorization requirements for different [agent categories](https://github.com/kagenti/kagenti/blob/main/docs/use-case-types.md) 

Parent Epic: #787

davidhadas:
## Summary
Adding use case types to Kagenti documentation
Fixes #789

### DIFF
--- a/docs/use-case-types.md
+++ b/docs/use-case-types.md
@@ -1,0 +1,55 @@
+# Kagenti Use Cases 
+Kagenti supports deploying and orchestrating multiple categories of AI‑driven use cases. Each category reflects a distinct operational model and security posture, and Kagenti’s system design aims to support all of them.
+
+1. Knowledge & Analysis Insight Service​
+2. User-Authorized Synchronous Task Assistant ​
+3. User-Authorized Asynchronous Task Assistant​
+4. Continuous Monitoring & Automated Response Service ​
+5. Event‑Driven Automation Workflow​
+
+Kagenti’s system and security architecture is designed to accommodate the requirements and constraints of each use‑case type.
+
+## 1. Knowledge & Analysis Insight Service​
+
+A read‑only system that retrieves, synthesizes, and analyzes organizational content to deliver evidence‑backed, cited insights without taking external actions. ​
+
+Examples of use cases:
+- A compliance assistant that summarizes relevant policy sections and cites authoritative documents.
+- A technical support knowledge explorer that retrieves past resolutions and highlights the most likely fix.
+- A financial analysis assistant that reviews quarterly reports and produces a cited summary of trends.
+
+## 2. User-Authorized Synchronous Task Assistant ​
+
+A real‑time, session‑bound assistant that performs actions on the user’s behalf using their permissions, without retaining state after the session ends. ​
+
+Examples of use cases:
+- Creating a pull request in GitHub when the user asks for it.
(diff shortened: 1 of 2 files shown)


kellyaa:
Hi @davidhadas ! Thanks for raising this. You are certainly right that we need a documented set of uses cases and their workload deployment implications - We have [an epic](https://github.com/kagenti/kagenti/issues/623) in progress for this. My thought is it's best to ensure we have a well tested scenario for each use case before documenting them here. My suggestion is to hold off on this PR right now. I have some internal drafts on this that I will move out into the open this week for collaboration.

docs/use-case-types.md
```diff
@@ -0,0 +1,55 @@
...
+
+1. Knowledge & Analysis Insight Service​
+2. User-Authorized Synchronous Task Assistant ​
+3. User-Authorized Asynchronous Task Assistant​
+4. Continuous Monitoring & Automated Response Service ​
+5. Event‑Driven Automation Workflow​
+
+Kagenti’s system and security architecture is designed to accommodate the requirements and constraints of each use‑case type.
+
+## 1. Knowledge & Analysis Insight Service​
```
maia-iyer, line 12: On the security side, I'm not sure that this use case can't fit squarely within one of the other categories

davidhadas: This use case type is a synchronous service allowing a user to derive information from data sources. It is the use case where RAG was originally introduced for - where you have lots of data accumulated in the organization and you want to service that can understand such data and make sense in it and answer questions from users. 

@matoushavlena suggested this category when discussing agent Archtypes and I added it to the list since it seems indeed very different from other use case types on this list. 

Note that this service it is not "User-Authorized". It does not access (either read or write) user specific data, does not perform any task on behalf of a user. This means that from a security perspective, it may have a very different posture. You can think of guests coming in to ask tech support questions or potential customers  asking about product features. There is no need for consent. There is no delegation of authority needed. Security wise we would need to analyze this very differently from the way we analyze  "User-Authorized Synchronous Task Assistant". 

@maia-iyer, this list is meant to help us steer the discussion​ on what mechanisms are needed in Kagenti - asking how do we support the deployment of typical agents in each type - so your comment is spot on as for what we need to do next. 

I expect that when we encounter and introduce real use cases, we will be able to map them to these 5 types/classes/categories/archtypes (whatever we choose to call them) - and when we identify a new use case type/class/category/archtype we will be able to come back here, add it and ask the question about the system and security design we need to have to support it compared to other types.

Alan-Cha (approved):

pdettori (approved):
/lgtm

PR #844, opened 2026-03-05
fix: resolve some docker hub rate limits for images

Ygnas:
## Summary
Probably there is more docker hub issues that could fail, but I encountered only these ones while trying to use the installer in openshift

- Set `common.postgresImage` to quay.io/fedora/postgresql-16 to avoid Docker Hub rate limits for the `mlflow` create-db init container and was already referenced in `charts/kagenti-deps/templates/mlflow.yaml` but did not exist.
- Use `common.kubectlImage` for `otel-collector-restart-job` instead of `bitnami/kubectl:latest` from Docker Hub

Fixes #

### DIFF
--- a/charts/kagenti-deps/values.yaml
+++ b/charts/kagenti-deps/values.yaml
@@ -44,6 +44,7 @@ components:
 
 common:
   kubectlImage: "quay.io/kubestellar/kubectl:1.30.14"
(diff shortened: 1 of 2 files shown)


pdettori (approved):
/lgtm

PR #863, opened 2026-03-06
fix: Create keycloak-admin-secret in agent namespaces and add AuthBridge keys to environments ConfigMap

mrsabath:
## Summary

- Creates a keycloak-admin-secret Secret in each agent namespace with Keycloak admin credentials, required by the client-registration sidecar after kagenti/kagenti-extensions#161 moves admin credentials from ConfigMap to Secret.
- Adds SPIRE_ENABLED, KEYCLOAK_URL, and KEYCLOAK_REALM keys to the environments ConfigMap — these were previously missing from the installer and had to be applied manually via demo ConfigMaps.
- Adds keycloak.adminUsername and keycloak.adminPassword to values.yaml (default: admin/admin) so credentials can be overridden.

## Context

PR kagenti/kagenti-extensions#161 (rebased as kagenti/kagenti-extensions#179) changes the webhook client-registration sidecar to read KEYCLOAK_ADMIN_USERNAME and KEYCLOAK_ADMIN_PASSWORD from a keycloak-admin-secret Secret rather than the environments ConfigMap. Without this installer change, all agent namespaces would be missing the secret and client registration would fail.

## Test plan

- [ ] helm template kagenti charts/kagenti renders correctly: keycloak-admin-secret appears in each agent namespace
- [ ] environments ConfigMap includes SPIRE_ENABLED, KEYCLOAK_URL, KEYCLOAK_REALM keys
- [ ] Fresh install: client-registration sidecar starts without errors
- [ ] Existing install: helm upgrade creates the new secret in existing agent namespaces

### DIFF
--- a/charts/kagenti/templates/agent-namespaces.yaml
+++ b/charts/kagenti/templates/agent-namespaces.yaml
@@ -197,10 +197,11 @@ data:
         "value": "slack-full-access"
       }
     ]
-  SPIRE_ENABLED: "true"
+  SPIRE_ENABLED: {{ if $root.Values.spire.enabled }}"true"{{ else }}"false"{{ end }}
   KEYCLOAK_URL: "http://keycloak-service.{{ $root.Values.keycloak.namespace }}.svc:8080"
   KEYCLOAK_REALM: {{ $root.Values.keycloak.realm | quote }}
 ---
+{{- if not $root.Values.keycloak.adminExistingSecret }}
 # ——————————————————————————————————————————————————————————————————————————————
 #  Secret for Keycloak Admin Credentials in Namespace: {{ . }}
 #  Used by the client-registration sidecar to register agents as OAuth clients.
@@ -214,9 +215,11 @@ metadata:
     {{- include "kagenti.labels" $root | nindent 4 }}
 type: Opaque
 stringData:
-  KEYCLOAK_ADMIN_USERNAME: {{ $root.Values.keycloak.adminUsername | default "admin" | quote }}
-  KEYCLOAK_ADMIN_PASSWORD: {{ $root.Values.keycloak.adminPassword | default "admin" | quote }}
+  KEYCLOAK_ADMIN_USERNAME: {{ $root.Values.keycloak.adminUsername | quote }}
+  KEYCLOAK_ADMIN_PASSWORD: {{ $root.Values.keycloak.adminPassword | quote }}
 ---
+{{- end }}
+
 # ——————————————————————————————————————————————————————————————————————————————
 #  ConfigMap for AuthBridge in Namespace: {{ . }}
 # ——————————————————————————————————————————————————————————————————————————————
--- a/charts/kagenti/values.yaml
+++ b/charts/kagenti/values.yaml
@@ -137,10 +137,14 @@ keycloak:
   adminSecretName: keycloak-initial-admin
   adminUsernameKey: username
   adminPasswordKey: password
-  # Admin credentials for the keycloak-admin-secret created in agent namespaces.
+  # Keycloak admin credentials for agent namespaces (keycloak-admin-secret).
   # Used by the client-registration sidecar to register agents as OAuth clients.
+  # For production, override via .secret_values.yaml or use an existing secret.
   adminUsername: admin
   adminPassword: admin
+  # Set to an existing Secret name to skip creating keycloak-admin-secret.
+  # The referenced secret must have KEYCLOAK_ADMIN_USERNAME and KEYCLOAK_ADMIN_PASSWORD keys.
+  # adminExistingSecret: ""
   url: http://keycloak-service.keycloak:8080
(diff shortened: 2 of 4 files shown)


charts/kagenti/values.yaml
```diff
@@ -137,6 +137,10 @@ keycloak:
   adminSecretName: keycloak-initial-admin
   adminUsernameKey: username
   adminPasswordKey: password
+  # Admin credentials for the keycloak-admin-secret created in agent namespaces.
+  # Used by the client-registration sidecar to register agents as OAuth clients.
+  adminUsername: admin
+  adminPassword: admin
```
mrsabath, line 144: **Security concern:** `adminPassword: admin` is stored in plaintext in values.yaml which gets committed to git. While this matches the existing pattern (credentials were already in ConfigMaps), the Secret should ideally reference the existing `keycloak-initial-admin` Secret rather than duplicating credentials. Consider an `existingSecret` pattern: if `keycloak.adminSecretName` is set, reference that Secret in agent namespaces instead of creating a new one with plaintext values from Helm.

charts/kagenti/templates/agent-namespaces.yaml
```diff
@@ -197,6 +197,25 @@ data:
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: keycloak-admin-secret
+  namespace: {{ . }}
+  labels:
+    {{- include "kagenti.labels" $root | nindent 4 }}
+type: Opaque
+stringData:
+  KEYCLOAK_ADMIN_USERNAME: {{ $root.Values.keycloak.adminUsername | default "admin" | quote }}
```
mrsabath: **Suggestion:** The `default "admin"` fallback means the Secret always renders even if the user explicitly removes `adminPassword` — silently falling back to "admin". Consider requiring the value or documenting that the default is insecure for production.

charts/kagenti/templates/agent-namespaces.yaml
```diff
@@ -197,6 +197,25 @@ data:
         "value": "slack-full-access"
       }
     ]
+  SPIRE_ENABLED: "true"
```
mrsabath: **Nit:** `SPIRE_ENABLED: "true"` is hardcoded. Consider templatizing for consistency with other configurable values.

docs/install.md
```diff
@@ -297,6 +297,56 @@ kubectl get secret keycloak-initial-admin -n keycloak \
...
+**Ansible installer** (via `dev_values.yaml` or `--set`):
+
+```bash
+deployments/ansible/run-install.sh --env dev \
+  --set charts.kagenti.values.keycloak.adminUsername=myadmin \
+  --set charts.kagenti.values.keycloak.adminPassword=mypassword
+```
+
+**Helm install** (direct):
+
```
mrsabath, line 336: **Suggestion:** The example shows `--set keycloak.adminPassword=mypassword` on the command line. Command-line passwords appear in shell history and process listings. Consider noting that users should prefer values files for sensitive values.

mrsabath (commented):
### Review Summary

Clean, focused PR that adds necessary infrastructure for the client-registration sidecar credential migration from ConfigMap to Secret. The Helm template is well-structured and the docs are thorough. Main concern is the plaintext credential pattern in values.yaml.

**Areas reviewed**: Helm/K8s, Values, Docs, Security
**Commits**: 2 commits, all signed-off ✅
**CI status**: Lint/security all passing ✅, Deploy & Test pending

🤖 Review by [Claude Code](https://claude.com/claude-code)

charts/kagenti/values.yaml
mrsabath: Addressed in 41a15c3. Added `keycloak.adminExistingSecret` support — when set, the template skips creating the secret entirely and the user's pre-existing Secret is used instead. Also added the credentials to `secret_values.yaml.example` so they follow the established secret values pattern rather than being committed in `values.yaml`.

charts/kagenti/templates/agent-namespaces.yaml
mrsabath: Addressed in 41a15c3. Removed the `| default "admin"` from the template. Values now come strictly from `values.yaml` or user overrides — no hidden fallback.

charts/kagenti/templates/agent-namespaces.yaml
mrsabath: Addressed in 41a15c3. `SPIRE_ENABLED` is now templatized from `spire.enabled` — correctly renders `"true"` or `"false"` based on the Helm value.

docs/install.md
mrsabath: Addressed in 41a15c3. Updated docs to recommend values files (`.secret_values.yaml` / `-f`) over `--set` for credentials. Also documented the `adminExistingSecret` option as the preferred production pattern.

PR #865, opened 2026-03-08
Fix: Podman container runtime support to Ansible installer

yoavkatz:
# Summary

This pull request has minor fixes support for Podman as container engine and improves the reliability of image pulling and chart installation processes.  The changes make the installer more robust when dealing with SSH connection issues in Podman environments and increase retry attempts for gateway installation.

# Files Changed

📄 `deployments/ansible/roles/kagenti_installer/tasks/02_setup_cluster.yaml`

Updated the image pulling logic to handle different container engines. For Podman, images are now pulled sequentially to avoid SSH connection issues, while Docker continues to use concurrent pulling for better performance. The task also now respects the `container_engine` variable (defaulting to 'docker') and includes enhanced logging output for Podman pulls (showing status messages like "Trying to pull", "Getting image", "Copying", and "Writing manifest").

📄 `deployments/ansible/roles/kagenti_installer/tasks/main.yml`

Made two key improvements:
1. Replaced hardcoded `docker` commands with the configurable `container_engine` variable (defaulting to 'docker') throughout the image loading process for kind clusters, enabling Podman compatibility.
2. Increased the retry count for MCP gateway installation from 3 to 10 attempts to improve reliability during deployment.

### DIFF
--- a/deployments/ansible/roles/kagenti_installer/tasks/02_setup_cluster.yaml
+++ b/deployments/ansible/roles/kagenti_installer/tasks/02_setup_cluster.yaml
@@ -131,7 +131,7 @@
         # Pull sequentially for Podman to avoid SSH connection issues
         for img in {{ preload_images | join(' ') }}; do
           echo "Pulling $img..."
-          {{ container_engine | default('docker') }} pull "$img" 2>&1 | { grep -E "^(Status:|Error|Trying to pull|Getting image|Copying|Writing manifest)" || true; }
+          {{ container_engine | default('docker') }} pull "$img" 2>&1 | { grep -E "^(Status:|Error|Trying to pull|Getting image|Copying|Writing manifest)" ; }
           if [ $? -ne 0 ]; then
             FAIL=1
           fi


deployments/ansible/roles/kagenti_installer/tasks/02_setup_cluster.yaml
```diff
@@ -123,18 +123,30 @@
...
+    - name: Pull images sequentially (Podman) or concurrently (Docker)
       shell: |
         set -o pipefail
         FAIL=0
+        {% if container_engine | default('docker') == 'podman' %}
+        # Pull sequentially for Podman to avoid SSH connection issues
+        for img in {{ preload_images | join(' ') }}; do
+          echo "Pulling $img..."
+          {{ container_engine | default('docker') }} pull "$img" 2>&1 | { grep -E "^(Status:|Error|Trying to pull|Getting image|Copying|Writing manifest)" || true; }
+          if [ $? -ne 0 ]; then
```
pdettori, line 135: Bug: `$?` captures the exit code of the `{ grep ... || true; }` subshell (always 0), not the `pull` command. Because of the pipe, even with `set -o pipefail`, the `|| true` masks the failure. This means `FAIL=1` is never set and failed pulls silently pass.

**Fix**: Capture the pull exit code before piping:
```bash
{{ container_engine | default('docker') }} pull "$img" 2>&1
rc=$?
if [ $rc -ne 0 ]; then
  FAIL=1
fi
```

Or alternatively, drop the `|| true` from the grep subshell so pipefail can propagate failures correctly.

deployments/ansible/roles/kagenti_installer/tasks/main.yml
```diff
@@ -1526,7 +1526,7 @@
           | combine((secret_values.get('global', {}) | default({})), recursive=True)
           | combine({'openshift': enable_openshift}, recursive=True) }}
   register: mcp_gateway_result
-  retries: 3
+  retries: 10
```
pdettori: nit: Retry increase from 3→10 is significant (max wait goes from 45s to 150s). Consider adding a brief comment explaining why this is needed (e.g., Podman environments take longer?).

esnible:
```suggestion
  retries: 10 # 10 retries needed because on OS xyz this operation flakes (being tracked by Kagenti Issue 123)
```

Can you add a comment for increasing to 10?

yoavkatz: Hi. I think this change might be unneeded. I'll revert it, and see.

deployments/ansible/roles/kagenti_installer/tasks/02_setup_cluster.yaml
yoavkatz: I will fix.

yoavkatz: Fixed by removing || true

pdettori (approved):

Issue #875, opened 2026-03-09
TUI: Add golangci-lint and go mod verify to CI


usize:
## Summary

The TUI CI job (`ci.yaml`, `tui` job) currently runs only `go vet`, `go test`, and
`go build`. Adding `golangci-lint` would catch a broader class of issues (unused code,
error handling, style), and `go mod verify` would provide supply chain verification
of downloaded modules.

## Suggested changes

- Add a `golangci-lint` step using `golangci/golangci-lint-action`
- Add `go mod verify` step before build
- Add a `.golangci.yml` config to `kagenti/tui/` with sensible defaults

## References

- PR #835 review findings M3, M5, M6
- CI file: `.github/workflows/ci.yaml` (tui job)

Issue #874, opened 2026-03-09
TUI: Integrate OS keychain for token storage


usize:
## Summary

Tokens are currently stored in plaintext in `~/.config/kagenti/tui.yaml` (permissions 0600).
While file permissions prevent other-user access and the risk is documented in the README
Security Notes section, OS keychain integration would provide stronger protection.

## Suggested approach

Use `go-keyring` to store access and refresh tokens in the OS keychain (macOS Keychain,
Linux Secret Service, Windows Credential Manager). Fall back to file-based storage when
no keychain is available (e.g. headless servers, CI).

## References

- PR #835 review finding C1
- Current documentation: `kagenti/tui/README.md` (Security Notes section)
- Config file: `kagenti/tui/internal/config/config.go`

PR #876, opened 2026-03-09
Bug fix: address follow-ups from Kagenti TUI code review

usize:
## Summary

In https://github.com/kagenti/kagenti/pull/835/ a number of follow up issues were proposed. This PR covers simplest of those. More complicated follow up tasks have their own issues filed (see related issues below).

## Related issue(s)

#874 
#875

### DIFF
--- a/kagenti/tui/internal/cli/root.go
+++ b/kagenti/tui/internal/cli/root.go
@@ -74,10 +74,8 @@ func NewRootCmd() *cobra.Command {
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			// No subcommand → launch interactive TUI.
-			cfg := config.Load(flagURL, flagToken, flagNamespace)
-			client := newClientFromConfig(cfg)
-
-			app := ui.NewApp(client)
+			// PersistentPreRunE has already populated ctx.Client.
+			app := ui.NewApp(ctx.Client)
 			p := tea.NewProgram(app, tea.WithAltScreen())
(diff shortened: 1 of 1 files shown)


kagenti/tui/internal/ui/views/login.go
```diff
@@ -23,6 +23,8 @@ type LoginView struct {
 	verifyURL string
 	err       error
 
+	saveErr error // non-nil if config save failed after successful login
```
pdettori, line 26: Good fix — surfacing the `cfg.Save()` error as a warning instead of silently discarding it is the right pattern. Login succeeds, user is informed their token was not persisted.

kagenti/tui/internal/cli/root.go
```diff
@@ -74,10 +74,8 @@ func NewRootCmd() *cobra.Command {
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			// No subcommand → launch interactive TUI.
-			cfg := config.Load(flagURL, flagToken, flagNamespace)
-			client := newClientFromConfig(cfg)
-
-			app := ui.NewApp(client)
+			// PersistentPreRunE has already populated ctx.Client.
+			app := ui.NewApp(ctx.Client)
```
pdettori, line 78: Clean removal of duplicated client init. `PersistentPreRunE` already guarantees `ctx.Client` is populated before `RunE` runs.

PR #879, opened 2026-03-09
debug: Add logging to agent conversation tests

akram:
## Summary

Add debug logging to `test_agent_conversation.py` to investigate empty response issue in CI (related to PR #782 failure analysis).

## Changes

- Add `[DEBUG]` log statements to track:
  - Result types received from agent (Task/Event tuples vs A2AMessage)
  - Event names and task diagnostics
  - Extracted text from artifacts and messages
  - Artifact counts on final tasks

## Investigation Context

PR #782 CI run failed with 3 tests returning empty responses:
- `test_agent_simple_query` - Agent returned no response text
- `test_agent_multiturn_conversation` - Turn 1 returned no response
- `test_root_span_has_mlflow_attributes` - Missing `mlflow.spanOutputs`

Logs show:
- ✓ Agent pod running and healthy
- ✓ LLM is being called (ChatOpenAI/openai.chat spans present)
- ✓ Traces exist in MLflow
- ✗ Response bodies are empty
- ✗ `mlflow.spanOutputs` missing

## Hypothesis

The issue is **not** a flaky LLM backend (LLM calls are succeeding). It's likely:
1. Agent processing requests but not streaming responses back
2. Response handling/output capture bug in observability middleware
3. Parts structure mismatch in A2A protocol handling

## Testing

This PR will help identify:
- Are tasks arriving with artifacts but empty parts?
- Is text extraction failing silently?
- Are we receiving unexpected event types?
- Is the issue in artifact population or part serialization?

## Related

- Issue #781 - kagenti-deps chart fails on OCP 4.19 (not the root cause but same CI run)
- PR #782 - SPIRE CRD alignment + Keycloak ConfigMap (not responsible for this failure)

### DIFF
--- a/kagenti/tests/e2e/common/test_agent_conversation.py
+++ b/kagenti/tests/e2e/common/test_agent_conversation.py
@@ -13,6 +13,7 @@ Usage:
 
 import os
 import pathlib
+import logging
 
 import pytest
 import httpx
@@ -30,6 +31,8 @@ from a2a.types import (
 from kagenti.tests.e2e.conftest import (
     _fetch_openshift_ingress_ca,
 )
+
+logger = logging.getLogger(__name__)
 
 
 def _is_openshift_from_config():
@@ -194,12 +197,16 @@ class TestWeatherAgentConversation:
 
         try:
             async for result in client.send_message(message):
+                logger.debug("Received result type: %s", type(result))
                 if isinstance(result, tuple):
                     task, event = result
                     last_task = task
-                    events_received.append(
-                        type(event).__name__ if event else "Task(final)"
-                    )
+                    event_name = type(event).__name__ if event else "Task(final)"
+                    events_received.append(event_name)
+                    logger.debug("Event: %s", event_name)
+                    logger.debug("Task: %s", _task_diagnostic(task))
+                    if event:
+                        logger.debug("Event details: %s", event)
 
                     # Check for failed task
                     status = getattr(task, "status", None)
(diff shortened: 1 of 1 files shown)


kagenti/tests/e2e/common/test_agent_conversation.py
```diff
@@ -13,6 +13,7 @@
 
 import os
 import pathlib
+import json
```
cwiklik: **must-fix**: `import json` is added but never used anywhere in the file. This causes ruff `F401` (unused import) failure, which is why the `build (3.11)` and `build (3.12)` CI checks are failing at the pre-commit hooks step. Remove this import to fix CI.

kagenti/tests/e2e/common/test_agent_conversation.py
```diff
@@ -194,12 +195,16 @@ async def test_agent_simple_query(self):
 
         try:
             async for result in client.send_message(message):
```
cwiklik, line 199: **suggestion**: These bare `print()` statements work for quick debugging, but consider using `logging.debug()` so output can be controlled via log level. Alternatively, if these are explicitly temporary (to be reverted after investigation), it would be good to note that in the PR description or a follow-up issue.

Also — the commit has `Signed-off-by: Claude Sonnet 4.5 <noreply@anthropic.com>`. Per repo policy (CLAUDE.md), AI attribution should use the `Assisted-By` trailer rather than `Signed-off-by`, since DCO sign-off implies legal certification that an AI cannot make.

kagenti/tests/e2e/common/test_agent_conversation.py
```diff
@@ -13,6 +13,7 @@
 
 import os
 import pathlib
```
cwiklik, line 15: `import json` is unused — causes ruff `F401` and breaks `build (3.11)` and `build (3.12)` CI checks. Remove this import to fix CI.

kagenti/tests/e2e/common/test_agent_conversation.py
akram: fixed

kagenti/tests/e2e/common/test_agent_conversation.py
akram: fixed

kagenti/tests/e2e/common/test_agent_conversation.py
akram: ran properly pre-commit and get that fixed

pdettori (approved):
/lgtm

PR #911, opened 2026-03-10
📝 docs: add Windows WSL setup guide and path validation hook

### LINKED ISSUE #855: :bug: Not able to checkout source code on windows machine
### Describe the bug

Source code is not being checked out on windows as
Windows doesn't allow these characters in filenames: < > : " / \ | ? *

error: invalid path '.claude/skills/auth:keycloak-confidential-client/SKILL.md'

There are several other paths as well on this location

### Steps To Reproduce

checkout source code using git on windows platform

### Expected Behavior

Source code should be checked out fine without any error

### Additional Context

_No response_

pdettori:
## Summary

- Add `docs/developer/windows-wsl-setup.md` — guide for Windows developers to use WSL with the Linux filesystem (clone location, IDE integration, Claude Code, troubleshooting)
- Add `scripts/hooks/check-windows-paths.sh` — pre-commit hook that catches Windows-forbidden characters in filenames (colons in skill folders are intentionally allowed)
- Register the hook in `.pre-commit-config.yaml`

Closes #855

## Test plan

- [ ] Verify `pre-commit run check-windows-paths --all-files` passes (colons in skill folders are allowed)
- [ ] Verify the doc renders correctly on GitHub
- [ ] Confirm the hook catches actual Windows-forbidden characters (e.g., `<`, `>`, `"`, `|`, `?`, `*`)

### DIFF
--- a/scripts/hooks/check-windows-paths.sh
+++ b/scripts/hooks/check-windows-paths.sh
@@ -13,10 +13,10 @@ FORBIDDEN_PATTERN='[<>"\\|?*]'
 
 bad_files=()
 while IFS= read -r -d '' file; do
-    basename=$(basename "$file")
-    # Colons are intentional in skill folder names — skip them
-    # Only flag other Windows-forbidden characters
-    cleaned=${basename//:/}
+    # Check the full path so directory components are validated too.
+    # Strip colons first — they are intentional in skill folder names
+    # (e.g., .claude/skills/auth:keycloak-confidential-client/).
+    cleaned=${file//:/}
     if [[ "$cleaned" =~ $FORBIDDEN_PATTERN ]]; then
         bad_files+=("$file")
     fi


scripts/hooks/check-windows-paths.sh
```diff
@@ -0,0 +1,34 @@
...
+# Windows users must use WSL — see docs/developer/windows-wsl-setup.md
+
+set -euo pipefail
+
+# Windows-forbidden characters (excluding / which is a path separator on all platforms):
+# < > : " \ | ? *
+FORBIDDEN_PATTERN='[<>"\\|?*]'
+
+bad_files=()
+while IFS= read -r -d '' file; do
```
cwiklik, line 15: Suggestion: The script checks only `basename`, so forbidden characters in **directory** path components (other than colon) would not be caught. Consider iterating over all path components. Low risk since directory names with `<>"\|?*` are extremely unlikely, but worth noting for completeness.

scripts/hooks/check-windows-paths.sh
```diff
@@ -0,0 +1,34 @@
...
+    basename=$(basename "$file")
+    # Colons are intentional in skill folder names — skip them
+    # Only flag other Windows-forbidden characters
```
cwiklik: Nit: The colon-stripping on `basename` is a no-op in practice since colons only appear in directory names (e.g., `.claude/skills/auth:keycloak-...`), never in basenames. The defensive intent is fine — just noting it is currently unreachable.

docs/developer/windows-wsl-setup.md
```diff
@@ -0,0 +1,91 @@
+# Windows Development with WSL
```
cwiklik, line 1: Excellent doc — clear setup flow, concrete error message in troubleshooting, and good performance notes. The callout about `/mnt/c/` vs native `~/` is especially helpful.

scripts/hooks/check-windows-paths.sh
pdettori: Good catch — updated in 52007e3f to validate the full path instead of just the basename. This also makes the colon-stripping meaningful as noted in your second comment.

scripts/hooks/check-windows-paths.sh
pdettori: Addressed in the same commit (52007e3f) — now checks the full path so the colon-stripping is actually exercised on directory components like `auth:keycloak-confidential-client`.

PR #924, opened 2026-03-11
fix: Kagenti deployment on OpenShift: shared trust, MLflow OIDC, installer resilience

ChristianZaccaria:
## Summary

Fix multiple issues where Kagenti deployment completed on RHOAI/OpenShift clusters but services silently failed — MLflow couldn't connect to PostgreSQL, OIDC login returned 500 errors, istiod deployment not made, and the installer didn't surface Helm failures.

Key changes:

- **Istio multi-mesh shared trust: detect stale intermediate CAs** — When the cert-manager root CA is regenerated (e.g. after Helm reinstall), cert-manager does not re-sign existing valid intermediate certificates. This caused Kagenti's and RHOAI's istiods to use different root CAs, making ztunnel fail with `BadSignature` errors and breaking all pod-to-pod mTLS in ambient mode. The installer now compares root CA fingerprints against each intermediate and forces re-issuance on mismatch.

- **MLflow OIDC login: combine system CAs with ingress CA** — `SSL_CERT_FILE` was set to only the OpenShift ingress CA, which replaced the system trust store entirely. On ROSA (Let's Encrypt), the ISRG Root X1 was missing, causing `CERTIFICATE_VERIFY_FAILED` when MLflow contacted Keycloak. The fix combines the system CA bundle with the ingress CA at runtime, supporting both public and custom CA environments.

- **Guard RHOAI scale-down behind `rhoai.scale_down` flag** — The RHOAI operator/dashboard scale-down was running unconditionally, which is inappropriate for production clusters. Now requires explicit opt-in via `rhoai.scale_down: true`.

- **Wait for cert-manager webhook before creating certificates** — The shared trust resources could fail silently if cert-manager's webhook wasn't ready. Added explicit readiness check.

- **Guard istiod rollout waits behind existence checks** — If kagenti-deps Helm hooks failed to create the Istio CR, the installer would hang waiting for a deployment that doesn't exist. Now checks first and warns.

- **Add Helm release health verification at end of install** — Catches silent Helm failures (e.g. `failed` status) that previously went unnoticed.

- **Make `config-trusted-cabundle` ConfigMap a Helm pre-install hook** — Ensures the OpenShift trusted CA bundle ConfigMap exists before any pod that needs it. Also mark the volume `optional: true` so the OTel collector starts while waiting for the ConfigMap.

- **Add cert-manager-webhook service lookup to `rhoai-shared-trust.yaml`** — Prevents Helm from rendering cert-manager Certificate resources when the webhook isn't available yet.

## Verification
All pods are running and working as expected on OpenShift. This includes MLFlow, Kagenti UI, Istiod, and keycloak, etc. - All Routes are accessible.

### DIFF
--- a/charts/kagenti-deps/templates/otel-collector.yaml
+++ b/charts/kagenti-deps/templates/otel-collector.yaml
@@ -11,6 +11,14 @@ metadata:
     # OpenShift injects trusted CA bundle into ConfigMaps with this label
     config.openshift.io/inject-trusted-cabundle: "true"
     {{- include "kagenti.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "-10"
+    "helm.sh/hook-delete-policy": before-hook-creation
+    # Retain on helm uninstall: OpenShift asynchronously injects the trusted CA bundle
+    # into this ConfigMap. Deleting it causes a race on reinstall where the OTel collector
+    # mounts an empty ConfigMap before OpenShift re-injects the CAs, breaking TLS verification.
+    "helm.sh/resource-policy": keep
 data: {}
 {{- end }}
 ---
@@ -103,10 +111,12 @@ spec:
           configMap:
             name: otel-collector-config
         {{- if and $.Values.openshift $.Values.components.mlflow.enabled $.Values.mlflow.auth.enabled }}
-        # Use OpenShift's pre-existing trusted CA bundle ConfigMap
+        # OpenShift trusted CA bundle — created by Tekton/platform operators.
+        # Optional so the pod starts while waiting for the ConfigMap to appear.
         - name: trusted-ca
           configMap:
             name: config-trusted-cabundle
+            optional: true
             items:
               - key: ca-bundle.crt
                 path: tls-ca-bundle.pem
--- a/deployments/ansible/roles/kagenti_installer/tasks/05_install_rhoai.yaml
+++ b/deployments/ansible/roles/kagenti_installer/tasks/05_install_rhoai.yaml
@@ -63,10 +63,12 @@
     # oauth-proxy). The operator defaults to 3 replicas at 500m each. On small clusters
     # (2-3 nodes), this exhausts CPU and blocks other pods from scheduling.
     # Scale operator FIRST (prevents it from reconciling replicas back up), then dashboard.
+    # Only runs when rhoai.scale_down is explicitly set to true.
     - name: Scale down RHOAI operator to 1 replica for test cluster
       command: kubectl scale deployment rhods-operator -n redhat-ods-operator --replicas=1
       failed_when: false
       changed_when: false
+      when: (rhoai | default({})).get('scale_down', false) | bool
 
     - name: Scale down RHOAI dashboard for test cluster
       shell: |
@@ -82,6 +84,7 @@
         echo "Dashboard deployment not found (may not be created yet)"
       failed_when: false
       changed_when: false
+      when: (rhoai | default({})).get('scale_down', false) | bool
 
     - name: Delete pending dashboard pods to free CPU
       shell: |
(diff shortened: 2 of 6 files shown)


deployments/ansible/roles/kagenti_installer/tasks/05_install_rhoai.yaml
```diff
@@ -63,10 +63,12 @@
     # oauth-proxy). The operator defaults to 3 replicas at 500m each. On small clusters
     # (2-3 nodes), this exhausts CPU and blocks other pods from scheduling.
     # Scale operator FIRST (prevents it from reconciling replicas back up), then dashboard.
+    # Only runs when rhoai.scale_down is explicitly set to true.
     - name: Scale down RHOAI operator to 1 replica for test cluster
       command: kubectl scale deployment rhods-operator -n redhat-ods-operator --replicas=1
       failed_when: false
       changed_when: false
+      when: rhoai.scale_down | default(false) | bool
```
pdettori: **Suggestion (non-blocking)**: Ansible dot notation for nested dict access (`rhoai.scale_down`) can shadow Python object attributes in edge cases. For consistency with the style used elsewhere in `main.yml` (e.g. `(charts['kagenti-deps'] | default({})).get(...)`) and to be explicit, consider:
```yaml
when: (rhoai | default({})).get('scale_down', false) | bool
```
Same note applies to lines 87 and 95.

deployments/ansible/roles/kagenti_installer/tasks/main.yml
```diff
@@ -1765,3 +1767,23 @@
     - enable_openshift | default(false)
     - (charts['kagenti'] | default({})).get('enabled', false) | bool
 
+- name: Verify Helm release status
+  shell: |
+    NS={{ (charts['kagenti'] | default({})).get('namespace', 'kagenti-system') }}
+    FAILED=""
+    for RELEASE in kagenti-deps kagenti; do
+      STATUS=$(helm history "$RELEASE" -n "$NS" --max 1 -o json 2>/dev/null | python3 -c "import sys,json; h=json.load(sys.stdin); print(h[0]['status'])" 2>/dev/null || echo "not-found")
```
pdettori: **Nit**: `python3` is invoked inline in the shell block. If the Ansible execution environment's `$PATH` doesn't include the Python used by Ansible, this could silently fall back to `echo "not-found"` and mask a real Helm failure. Consider using Ansible's `from_json` filter or add a `which python3` pre-flight check.

charts/kagenti-deps/templates/otel-collector.yaml
```diff
@@ -11,6 +11,11 @@ metadata:
     # OpenShift injects trusted CA bundle into ConfigMaps with this label
     config.openshift.io/inject-trusted-cabundle: "true"
     {{- include "kagenti.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "-10"
+    "helm.sh/hook-delete-policy": before-hook-creation
+    "helm.sh/resource-policy": keep
```
pdettori, line 21: **Suggestion (non-blocking)**: `helm.sh/resource-policy: keep` prevents deletion on `helm uninstall`, which is intentional since OpenShift injects the trusted CA bundle into this ConfigMap. Consider a short comment explaining why, so future maintainers don't remove it:
```yaml
# keep: OpenShift injects ca-bundle.crt into this ConfigMap; preserve across uninstalls
"helm.sh/resource-policy": keep
```

deployments/ansible/roles/kagenti_installer/tasks/main.yml
ChristianZaccaria: I'm going to use `ansible_python_interpreter` to resolve the same Python binary Ansible is running with, avoiding PATH mismatches. Falls back to python3 if the variable is unset.

PR #925, opened 2026-03-11
Refactor: Move OAuth UI secret steps from Ansible to Helm

ChristianZaccaria:
Previously there was a race condition with the Route on OpenShift.

## Context:

On OpenShift, the kagenti-ui-oauth-secret-job reads the Route host to construct the OAuth callback URL. Previously, the Job and Route were created in parallel as regular Helm resources, causing a race condition where the Job could fail with "route not found". The Job eventually succeeded on retry (backoffLimit), but the UI deployment might have already started with missing credentials, requiring an Ansible workaround to wait-and-restart.

## Changes:

`ui-oauth-secret-job.yaml`: Add a wait-for-dependencies initContainer that blocks until Keycloak is Ready and (on OpenShift) the kagenti-ui route has a host assigned. This guarantees the Job succeeds on the first attempt. Added pods get/list RBAC for the Keycloak readiness check.

`main.yml`: Remove the Ansible workaround that waited for the OAuth job and restarted the UI deployment. No longer needed since the Job now creates the secret promptly, and the UI deployment's existing wait-for-oauth-secret initContainer picks it up.


charts/kagenti/templates/ui-oauth-secret-job.yaml
```diff
@@ -87,6 +90,27 @@ spec:
     spec:
       serviceAccountName: kagenti-ui-oauth-secret-writer
       restartPolicy: Never
+      initContainers:
+      - name: wait-for-dependencies
+        image: {{ .Values.common.kubectlImage | default "bitnami/kubectl:latest" }}
```
pdettori, line 95: **Suggestion (non-blocking)**: `bitnami/kubectl:latest` is a mutable tag — on a cache miss it can silently pull a newer kubectl version and cause unexpected breakage. Consider pinning to a specific version as the default:
```yaml
image: {{ .Values.common.kubectlImage | default "bitnami/kubectl:1.32" }}
```
The `common.kubectlImage` values override path stays as-is for users who want to control the version.

ChristianZaccaria: It looks like Bitnami stopped publishing semver tags for their kubectl image. Maybe we can keep using `latest` for now and switch to some other registry later.
https://hub.docker.com/r/bitnami/kubectl/tags

PR #957, opened 2026-03-15
fix: disable auto-browser launch in mcp-inspector to prevent CrashLoopBackOff

eranra:
## Problem
The `mcp-inspector` pod was experiencing CrashLoopBackOff because it attempted to auto-launch a browser using PowerShell, which doesn't exist in the Linux container environment.

## Error
```
Error: spawn /mnt/c/Windows/System32/WindowsPowerShell/v1.0/powershell.exe ENOENT
```

## Solution
Added the `BROWSER=none` environment variable to the mcp-inspector deployment to disable the auto-browser launch behavior.

## Changes
- Modified `charts/kagenti-deps/templates/mcp-inspector.yaml` to add `BROWSER=none` environment variable

## Testing
After applying this fix:
1. The mcp-inspector pod should start successfully without crashes
2. The inspector UI remains accessible via the configured route/ingress
3. No functionality is lost - users can still access the inspector through the web interface

## Related Issue
Fixes the CrashLoopBackOff issue reported during deployment verification.

### DIFF
--- a/charts/kagenti-deps/templates/mcp-inspector.yaml
+++ b/charts/kagenti-deps/templates/mcp-inspector.yaml
@@ -132,6 +132,26 @@ spec:
       containers:
         - name: mcp-inspector
           image: "{{ .Values.mcpInspector.image }}:{{ .Values.mcpInspector.tag }}"
+          command: ["/bin/sh", "-c"]
+          args:
+            - |
+              # Patch the open package to prevent browser opening
+              OPEN_PKG="/app/node_modules/open/index.js"
+              if [ -f "$OPEN_PKG" ]; then
+                cat > "$OPEN_PKG" << 'EOF'
+              const openStub = async () => {
+                console.log('Browser opening disabled in container environment');
+                return { exitCode: 0 };
+              };
+              
+              export default openStub;
+              export { openStub as openApp };
+              export const apps = {};
+              EOF
+                echo "Patched open package to disable browser launching"
+              fi
+              # Start the application
+              npm start
           ports:
             - containerPort: 6274
             - containerPort: 6277
@@ -143,6 +163,10 @@ spec:
               cpu: 25m
               memory: 64Mi
           env:
+          - name: BROWSER
+            value: "none"
+          - name: NO_BROWSER
+            value: "true"
           - name: DANGEROUSLY_OMIT_AUTH
             value: "true"
           - name: ALLOWED_ORIGINS
--- a/charts/kagenti-deps/templates/mcp-inspector.yaml
+++ b/charts/kagenti-deps/templates/mcp-inspector.yaml
@@ -136,6 +136,7 @@ spec:
           args:
             - |
               # Patch the open package to prevent browser opening
+              # Stub for open v9+ (ESM). If the base image downgrades to v8 (CJS), this will need updating.
               OPEN_PKG="/app/node_modules/open/index.js"
               if [ -f "$OPEN_PKG" ]; then
                 cat > "$OPEN_PKG" << 'EOF'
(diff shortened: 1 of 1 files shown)


evaline-ju (commented):
Thanks @eranra for the PR contribution! Would you please make sure DCO sign-off is done for your commits? Details at: https://github.com/kagenti/kagenti/pull/957/checks?check_run_id=67135532478

charts/kagenti-deps/templates/mcp-inspector.yaml
```diff
@@ -132,6 +132,26 @@ spec:
       containers:
         - name: mcp-inspector
           image: "{{ .Values.mcpInspector.image }}:{{ .Values.mcpInspector.tag }}"
+          command: ["/bin/sh", "-c"]
+          args:
+            - |
+              # Patch the open package to prevent browser opening
+              OPEN_PKG="/app/node_modules/open/index.js"
+              if [ -f "$OPEN_PKG" ]; then
+                cat > "$OPEN_PKG" << 'EOF'
```
pdettori, line 142: **suggestion**: Runtime monkey-patching of `node_modules/open/index.js` via a shell heredoc in the Helm chart is fragile — it's tightly coupled to the `open` package's internal path and module format, and mixes application-level concerns into the deployment manifest.

Consider alternatives:
1. **Fix in the Dockerfile** — patch or remove the `open` package at image build time
2. **Use `BROWSER=none` alone** — the `open` package v9+ respects this env var; if it wasn't sufficient for your version, a code comment explaining why would help future maintainers

The current approach works but will silently break if the base image upgrades `open` to a version with a different file layout.

charts/kagenti-deps/templates/mcp-inspector.yaml
```diff
@@ -132,6 +132,26 @@ spec:
...
+              const openStub = async () => {
+                console.log('Browser opening disabled in container environment');
+                return { exitCode: 0 };
+              };
+
```
pdettori, line 147: **nit**: The `if [ -f "$OPEN_PKG" ]` guard silently skips the patch if the file doesn't exist (e.g., after an image upgrade changes the path). Consider adding an `else` branch that logs a warning:

```sh
else
  echo "WARNING: open package not found at $OPEN_PKG — browser patch skipped"
fi
```

This makes debugging much easier if the patch stops working after an image update.

charts/kagenti-deps/templates/mcp-inspector.yaml
```diff
@@ -132,6 +132,26 @@ spec:
...
         - name: mcp-inspector
           image: "{{ .Values.mcpInspector.image }}:{{ .Values.mcpInspector.tag }}"
+          command: ["/bin/sh", "-c"]
+          args:
+            - |
+              # Patch the open package to prevent browser opening
+              OPEN_PKG="/app/node_modules/open/index.js"
+              if [ -f "$OPEN_PKG" ]; then
+                cat > "$OPEN_PKG" << 'EOF'
+              const openStub = async () => {
```
pdettori, line 143: **suggestion**: The ESM stub (`export default openStub`) is correct for `open` v9+ (which is pure ESM), but would cause a syntax error on CommonJS versions (v8 and below). A brief comment noting the assumed version would help:

```js
// Stub for open v9+ (ESM). If the base image downgrades to v8 (CJS), this will need updating.
```

pdettori (changes requested):
## Review Summary

The PR fixes a real CrashLoopBackOff issue in `mcp-inspector` — good catch. However:

1. **DCO sign-off is missing** — CI is failing. Please `git commit --amend -s && git push -f`.
2. **Runtime monkey-patching of `node_modules` in the Helm chart is fragile** — it couples the deployment manifest to the `open` package's internal file layout and module format. Consider fixing this in the `mcp-inspector` Dockerfile instead, or document why `BROWSER=none` alone is insufficient.

See inline comments for details.

pdettori:
@eranra could you please address sign-off to pass the DCO check ?

eranra:
Closing at this point ... I will get back to it and open a new one. 

thanks @pdettori :-)

PR #959, opened 2026-03-15
fix: Set serviceAccountName in workload manifests for correct SPIFFE identity

### LINKED ISSUE #958: bug: UI-deployed workloads get incorrect SPIFFE ID (ReplicaSet hash in ServiceAccount name)
## Problem

When the Kagenti UI deploys an agent or tool, the generated Deployment manifest does not set `serviceAccountName` in the pod spec. This causes the kagenti-webhook's `deriveWorkloadName()` function to fall back to the pod's `GenerateName`, which yields the **ReplicaSet name** (e.g. `git-issue-agent-784685db4c`) rather than the **Deployment name** (`git-issue-agent`).

The incorrect name propagates through:

1. **ServiceAccount** — the webhook creates an SA named `git-issue-agent-784685db4c` instead of `git-issue-agent`
2. **SPIFFE ID** — SPIRE assigns `spiffe://localtest.me/ns/team1/sa/git-issue-agent-784685db4c` instead of `.../sa/git-issue-agent`
3. **Keycloak client** — client-registration registers under the wrong SPIFFE ID
4. **Audience scope** — the scope name includes the hash (`agent-team1-git-issue-agent-784685db4c-aud`)

## Impact

- **CLI tests fail**: Step 9d of the github-issue demo looks up the client by `spiffe://...sa/git-issue-agent`, which doesn't match the registered `spiffe://...sa/git-issue-agent-784685db4c`
- **Token exchange**: The exchanged token's audience references the wrong SPIFFE ID
- **Stale SAs accumulate**: Each redeployment creates a new SA with a different ReplicaSet hash

## Root Cause

The webhook's [`deriveWorkloadName()`](https://github.com/kagenti/kagenti-extensions/blob/main/kagenti-webhook/internal/webhook/v1alpha1/authbridge_webhook.go#L127-L139) trims the trailing hyphen from `pod.GenerateName`, producing the ReplicaSet name. It only resolves to the Deployment name when `serviceAccountName` is already set on the pod spec.

The UI's manifest builders (`_build_deployment_manifest`, `_build_statefulset_manifest`, `_build_job_manifest` in `agents.py`; `_build_tool_deployment_manifest`, `_build_tool_statefulset_manifest` in `tools.py`) did not set `serviceAccountName`.

## Reproduction

1. Deploy the github-issue agent via the Kagenti UI (Build from Source)
2. Check the client-registration logs:
   ```
   kubectl logs deployment/git-issue-agent -n team1 -c kagenti-client-registration
   ```
3. Observe the SPIFFE ID includes the ReplicaSet hash:
   ```
   Client ID (SPIFFE ID): spiffe://localtest.me/ns/team1/sa/git-issue-agent-784685db4c
   ```
4. Step 9d CLI test fails because the client lookup for `spiffe://...sa/git-issue-agent` returns null

## Fix

Set `serviceAccountName` to the workload name in all manifest builders, and create the ServiceAccount before the workload. This makes the webhook skip `deriveWorkloadName()` and use the correct name.

Related webhook code: kagenti/kagenti-extensions#177 (Phase 2 owner-ref resolution)

mrsabath:
## Summary

- Set `serviceAccountName` to the workload name in all manifest builders (Deployment, StatefulSet, Job) for both agents and tools
- Add `ensure_service_account()` to `KubernetesService` to idempotently create the SA before the workload
- Call `ensure_service_account()` in all creation endpoints (image deploy, Shipwright finalize, CRD migration)

Without this fix, the webhook's `deriveWorkloadName()` falls back to the ReplicaSet name (which includes a hash suffix), producing incorrect SPIFFE IDs, Keycloak client registrations, and audience scopes.

Fixes #958

## Files Changed

| File | Change |
|------|--------|
| `kubernetes.py` | New `ensure_service_account()` method (read-or-create, idempotent) |
| `agents.py` | `serviceAccountName` in 4 manifest builders + `ensure_service_account()` in 3 creation paths |
| `tools.py` | `serviceAccountName` in 2 manifest builders + `ensure_service_account()` in 2 creation paths |

## Test plan

- [ ] Deploy an agent via UI (Build from Source) and verify `kubectl logs -c kagenti-client-registration` shows the correct SPIFFE ID (no ReplicaSet hash)
- [ ] Run the github-issue demo Step 9d CLI test end-to-end
- [ ] Deploy a tool via UI and verify its ServiceAccount name matches the tool name
- [ ] Verify re-deploying an existing agent does not fail (SA creation is idempotent)
- [ ] Verify the CRD migration path sets serviceAccountName correctly

### DIFF
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -1083,6 +1083,7 @@ async def migrate_agent(
     else:
         # Create new Deployment from Agent CRD spec
         deployment_manifest = _build_deployment_from_agent_crd(agent)
+        kube.ensure_service_account(namespace=namespace, name=name)
         try:
             kube.create_deployment(namespace=namespace, body=deployment_manifest)
             deployment_created = True
@@ -1288,6 +1289,7 @@ def _build_deployment_from_agent_crd(agent: dict) -> dict:
             )
 
         pod_spec = {
+            "serviceAccountName": name,
             "containers": [
                 {
                     "name": "agent",
@@ -1315,6 +1317,10 @@ def _build_deployment_from_agent_crd(agent: dict) -> dict:
                 {"name": "shared-data", "emptyDir": {}},
             ],
         }
+
+    # Ensure serviceAccountName is set so the webhook's SPIFFE identity
+    # derivation uses the workload name rather than the ReplicaSet hash.
+    pod_spec.setdefault("serviceAccountName", name)
 
     # Build selector labels
     selector_labels = {
@@ -1987,6 +1993,7 @@ def _build_deployment_manifest(
                     },
                 },
                 "spec": {
+                    "serviceAccountName": request.name,
                     "containers": [
                         {
                             "name": "agent",
@@ -2140,6 +2147,7 @@ def _build_statefulset_manifest(
                     },
                 },
                 "spec": {
+                    "serviceAccountName": request.name,
                     "containers": [
                         {
                             "name": "agent",
@@ -2237,6 +2245,7 @@ def _build_job_manifest(
                     },
                 },
                 "spec": {
+                    "serviceAccountName": request.name,
                     "restartPolicy": "OnFailure",
                     "containers": [
                         {
@@ -2313,6 +2322,10 @@ async def create_agent(
                     status_code=400,
                     detail="containerImage is required for image deployment",
                 )
+
+            # Ensure a dedicated ServiceAccount exists so the webhook's
+            # SPIFFE identity uses the workload name, not the ReplicaSet hash.
+            kube.ensure_service_account(namespace=request.namespace, name=request.name)
 
             # Create workload based on workloadType
             if request.workloadType == WORKLOAD_TYPE_DEPLOYMENT:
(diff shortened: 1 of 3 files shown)


kagenti/backend/app/routers/agents.py
```diff
@@ -1288,6 +1289,7 @@ def _build_deployment_from_agent_crd(agent: dict) -> dict:
             )
 
         pod_spec = {
+            "serviceAccountName": name,
             "containers": [
```
pdettori, line 1293: nit: `serviceAccountName` is already set here in the dict literal, but `pod_spec.setdefault("serviceAccountName", name)` at line 1321 covers both the `if not pod_spec` and `else` branches. The inline one here is technically redundant — the `setdefault` alone would suffice. Harmless, just noting for clarity.

kagenti/backend/app/services/kubernetes.py
```diff
@@ -199,6 +199,35 @@ def create_custom_resource(
...
+                sa = kubernetes.client.V1ServiceAccount(
+                    metadata=kubernetes.client.V1ObjectMeta(
+                        name=name,
+                        namespace=namespace,
+                        labels={"kagenti.io/managed-by": "kagenti-ui"},
+                    ),
+                )
+                self.core_api.create_namespaced_service_account(namespace=namespace, body=sa)
+                logger.info(f"Created ServiceAccount '{name}' in {namespace}")
+            else:
```
pdettori, line 227: suggestion: `ensure_service_account()` uses the sync K8s client while callers are `async def` endpoints. Fine for now since these calls are quick, but worth noting if this service layer moves to async in the future.

kagenti/backend/app/routers/agents.py
mrsabath: Good catch — removed the inline `serviceAccountName` from the fallback dict, keeping only the `setdefault` which covers both branches. Fixed in 5e1234d.

kagenti/backend/app/services/kubernetes.py
mrsabath: Noted — the sync calls are quick (single K8s API round-trip) so this is fine for now. Added a mental note for when the service layer goes async.

PR #973, opened 2026-03-16
Feat: Use new AGENT_ENDPOINT env var to let agent know where it is deployed

### LINKED ISSUE #776: feature: Kagenti support for Agent card URLs
### Feature Description

Kagenti's example agents supply incorrect AgentCard URLs.  (See https://github.com/kagenti/agent-examples/issues/139 ).

### Proposed Solution

Kagenti should supply an env var (or two) to each agent it creates, supplying the endpoint info that Kagenti creates.

A problem with this proposed solution is that an agent has 1 or 2 endpoints.

By default, an agent has an endpoint such as http://a2a-currency-converter.team1.svc.cluster.local:8080/

**However**, if the agent importer clicks Kagenti's "Check Enable external access to the agent endpoint", the agent will have a second endpoint, such as http://a2a-currency-converter-v3.team1.localtest.me:8080/

I am not sure what the correct solution is.
- Perhaps an agent should offer two cards, an internal card and an external card?
- Perhaps there should be middleware or a gateway that manipulates the gateway as a man-in-the-middle?
- Perhaps the agent should look at the request and format its own URL based on that request?  (This might be hard, [it is a str](https://github.com/a2aproject/a2a-python/blob/cced34d4eb22f7ec78422b2d04ce54924e132215/src/a2a/types.py#L1825).

### Want to contribute?

### Additional Context

_No response_

esnible:
## Summary

Resolves #776

This PR causes Kagenti to supply a new env var in agent Deployments.  The env name is AGENT_ENDPOINT and it will have the value that matches the "Agent URL" field in the Kagenti UI.

This allows the agent to supply its own URL on the agent card.  Without this, it would not know where it was deployed.

## Related issue(s)

#776
https://github.com/kagenti/agent-examples/pull/179
https://github.com/kagenti/agent-examples/pull/144

To test, You will need to build and push the kagenti-backend changes in this PR.

To test, deploy agent examples and look at their agent cards.  Or look at the generated Deployments for the new AGENT_ENDPOINT variable using `kubectl -n team1 get deployment <name> -o yaml | grep AGENT_ENDPOINT`.

Build agents from source to get the latest versions with the merged agent-example changes.  Go to the Agent Card.  When the Kagenti backend is using this version, the URL will be a real URL, not `http://0.0.0.0:8000`.

### DIFF
--- a/kagenti/backend/app/utils/routes.py
+++ b/kagenti/backend/app/utils/routes.py
@@ -267,6 +267,7 @@ def create_route_for_agent_or_tool(
     else:
         create_httproute(kube, name, namespace, service_name, service_port)
 
+
 def get_agent_url(name: str, namespace: str) -> str:
     """Get the URL for an A2A agent.
 


kagenti/backend/app/routers/agents.py
```diff
@@ -20,6 +20,7 @@
 from pydantic import BaseModel, field_validator
 
 from app.core.auth import ROLE_OPERATOR, ROLE_VIEWER, require_roles
+from app.routers.chat import _get_agent_url
```
pdettori: **suggestion**: `_get_agent_url` is a private function in `chat.py` (underscore prefix signals internal-to-module). Importing it across routers creates coupling between sibling modules and risks breakage if `chat.py` refactors its internals. Consider moving it to `app/utils/routes.py` (which already exists for similar helpers) so both `agents.py` and `chat.py` can share it cleanly.

kagenti/backend/app/core/constants.py
```diff
@@ -128,3 +128,6 @@
     },
     {"name": "UV_CACHE_DIR", "value": "/app/.cache/uv"},
 ]
+
+# Environment variable name for the agent endpoint (the model card URL for the agent)
```
pdettori: **nit**: Comment says "model card URL" but this project uses A2A terminology: "agent card". Minor but worth aligning with the codebase vocabulary.

kagenti/backend/app/routers/agents.py
```diff
@@ -1852,6 +1854,9 @@ def _build_env_vars(request: "CreateAgentRequest") -> List[dict]:
         List of environment variable dictionaries.
     """
     env_vars = list(DEFAULT_ENV_VARS)
+    env_vars.append(
+        {"name": AGENT_ENDPOINT, "value": _get_agent_url(request.name, request.namespace)}
```
pdettori: **suggestion**: No test coverage for the new env var injection. `test_agent_readiness.py` shows there is a pattern for testing `agents.py` private functions. A short unit test asserting `AGENT_ENDPOINT` appears in the output of `_build_env_vars` would prevent regressions.

esnible: `test_agent_readiness.py` is not a unit test.  I did not find unit tests for _kagenti/backend/app_.  This PR is a small focused PR for A2A inspector compatibility.  I agree with the suggest that unit tests would be useful, but we should create a Feature Issue for _pyunit_ unit tests and deliver them under that.

pdettori (approved):
/lgtm

PR #974, opened 2026-03-16
docs: add upstream sync step and cross-repo guidance to pr-review skill

pdettori:
## Summary
- Add Phase 1.5 "Sync Local Repo" step to `github:pr-review` skill requiring `git fetch upstream main` before verifying PR claims against source files
- Add troubleshooting sections for cross-repo reviews, stale checkout anti-pattern, and `gh api` JSON input format
- Fix Phase 5 example code that used unsupported `gh api` array parameter syntax

## Test plan
- [ ] Review updated skill at `.claude/skills/github:pr-review/SKILL.md`
- [ ] Verify new Phase 1.5 section is clear and actionable
- [ ] Confirm `gh api` JSON example is correct syntax

## Context
Learned from an incorrect review of kagenti-extensions#241 where a stale local `go.mod` caused false version mismatch flags. All 5 inline comments were wrong because the local checkout was behind `upstream/main`.

### DIFF
--- a/.claude/skills/github:pr-review/SKILL.md
+++ b/.claude/skills/github:pr-review/SKILL.md
@@ -85,6 +85,27 @@ If checks are failing, delegate to `ci:status` for detailed analysis.
 gh pr view <number> --json commits --jq '.commits[] | "\(.oid[:7]) \(.messageHeadline)"'
 ```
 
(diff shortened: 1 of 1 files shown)


Ladas (approved):
👍

PR #981, opened 2026-03-16
feat: add per-namespace PostgreSQL session DB manager

Ladas:
## Summary
- session_db.py: async PostgreSQL connection pool per namespace, auto-migration, upsert sessions
- Add asyncpg dependency to pyproject.toml
- Update uv.lock

3 files | Part of Sandbox Agent streaming PR series


kagenti/backend/app/services/session_db.py
```diff
@@ -0,0 +1,230 @@
...
+def _dsn_for_namespace(namespace: str) -> str:
+    """Build a DSN from the namespace secret, falling back to convention."""
+    creds = _read_secret(namespace)
+    if not creds:
+        return _convention_dsn(namespace)
+    return (
+        f"postgresql://{creds['username']}:{creds['password']}"
+        f"@{creds['host']}:{creds['port']}/{creds['database']}"
+    )
+
```
pdettori, line 92: **must-fix**: DSN is constructed via f-string without URL-encoding the username/password (lines 88-91). If the password from the K8s secret contains `@`, `/`, `%`, or other URI-special characters, the DSN will be malformed and connections will fail silently.

Suggestion:
```python
from urllib.parse import quote_plus

return (
    f"postgresql://{quote_plus(creds['username'])}:{quote_plus(creds['password'])}"
    f"@{creds['host']}:{creds['port']}/{creds['database']}"
)
```

kagenti/backend/app/services/session_db.py
```diff
@@ -0,0 +1,230 @@
...
+            raise
+        except Exception as exc:
+            last_error = exc
+            if attempt < _POOL_MAX_RETRIES:
+                logger.warning(
+                    "DB pool creation failed (attempt %d/%d): %s — retrying in %.0fs",
+                    attempt,
+                    _POOL_MAX_RETRIES,
+                    exc,
+                    _POOL_RETRY_DELAY,
```
pdettori, line 134: **suggestion**: The retry logic logs the full exception (`%s, exc`) which, for asyncpg connection failures, may include host/port details from the DSN. Given the explicit security-hardening commits in this PR (cbdd1b5, cd23e77), consider logging only `type(exc).__name__` here for consistency:
```python
logger.warning(
    "DB pool creation failed (attempt %d/%d): %s — retrying in %.0fs",
    attempt, _POOL_MAX_RETRIES, type(exc).__name__, _POOL_RETRY_DELAY,
)
```

kagenti/backend/app/services/session_db.py
```diff
@@ -0,0 +1,230 @@
...
+                    _POOL_MAX_RETRIES,
+                    exc,
+                )
+    raise last_error  # type: ignore[misc]
+
+
+async def get_session_pool(namespace: str) -> asyncpg.Pool:
+    """Return (or lazily create) the asyncpg pool for *namespace*."""
+    pool = _pool_cache.get(namespace)
+    if pool is not None:
```
pdettori, line 149: **suggestion**: `get_session_pool` has no `asyncio.Lock`. Two concurrent requests for the same namespace will both see `pool is None`, both call `_create_pool`, and one pool will be leaked (never closed). Consider adding a lock:
```python
_pool_locks: Dict[str, asyncio.Lock] = {}
_global_lock = asyncio.Lock()

async def get_session_pool(namespace: str) -> asyncpg.Pool:
    pool = _pool_cache.get(namespace)
    if pool is not None and not pool._closed:
        return pool
    async with _global_lock:
        # Double-check after acquiring lock
        pool = _pool_cache.get(namespace)
        if pool is not None and not pool._closed:
            return pool
        ...
```

kagenti/backend/app/services/session_db.py
```diff
@@ -0,0 +1,230 @@
...
+        if not pool._closed:
+            return pool
```
pdettori, line 151: **nit**: `pool._closed` accesses a private asyncpg attribute. This works but is fragile across asyncpg versions. Consider documenting this as an accepted internal access, or wrap a lightweight `pool.fetchval('SELECT 1')` in a try/except as a health check.

kagenti/backend/app/services/session_db.py
```diff
@@ -0,0 +1,230 @@
...
+    return (
+        f"postgresql://{creds['username']}:{creds['password']}"
+        f"@{creds['host']}:{creds['port']}/{creds['database']}"
+    )
+
+
+def _convention_dsn(namespace: str) -> str:
+    """Convention-based DSN when no secret is available."""
+    logger.warning("Using convention-based DB fallback for namespace=%s", namespace)
+    return f"postgresql://kagenti:kagenti@postgres-sessions.{namespace}:5432/sessions"
```
pdettori, line 97: **nit**: The convention fallback with hardcoded `kagenti:kagenti` credentials is fine for dev, but consider adding a brief comment noting these are dev-only defaults not suitable for production.

kagenti/examples/agents/weather_service_deployment.yaml
```diff
@@ -50,11 +50,14 @@ spec:
         - name: MCP_URL
           value: "http://weather-tool-mcp.team1.svc.cluster.local:8000/mcp"
         - name: LLM_API_BASE
-          value: "http://dockerhost:11434/v1"
+          value: "https://litellm-prod.apps.maas.redhatworkshops.io/v1"
```
pdettori: **suggestion**: `LLM_API_BASE` is now hardcoded to `litellm-prod.apps.maas.redhatworkshops.io`. This is environment-specific — Kind and OCP may need different endpoints. Consider sourcing this from a ConfigMap or Helm value so it can be overridden per-environment without code changes.

pdettori (changes requested):
## Review Summary

Adds per-namespace PostgreSQL session DB manager and event schema for sandbox streaming. Module is properly feature-gated behind `kagenti_feature_flag_sandbox`. Good commit hygiene with dedicated security-hardening follow-ups. Two CI checks failing (Deploy & Test, E2E HyperShift).

**Areas reviewed**: Python (session_db.py, event.py), Dependencies (pyproject.toml), YAML (weather deployment)
**Commits**: 8 commits, all signed-off ✅
**CI**: 2 failing, 19 passing

One must-fix (DSN URL-encoding) and several suggestions below.

pdettori (approved):
/lgtm

PR #984, opened 2026-03-16
feat: add LLM budget proxy for per-session token tracking

Ladas:
## Summary
- Reverse proxy between sandbox agents and LiteLLM
- Records token usage (prompt/completion/total) per session in PostgreSQL
- Enforces configurable budget limits (max_tokens, max_cost) — returns 429 when exceeded
- Multi-stage Docker build, non-root, health check

3 files | Part of Sandbox Agent streaming PR series

### DIFF
--- a/kagenti/llm-budget-proxy/Dockerfile
+++ b/kagenti/llm-budget-proxy/Dockerfile
@@ -1,0 +1,37 @@
+FROM python:3.12-slim AS builder
+
+WORKDIR /app
+
+COPY --from=ghcr.io/astral-sh/uv:0.9.24 /uv /bin/uv
+
+COPY llm-budget-proxy/pyproject.toml ./
+
+RUN uv venv /app/.venv && \
+    . /app/.venv/bin/activate && \
+    uv pip install --no-cache .
+
+FROM python:3.12-slim
+
+WORKDIR /app
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends ca-certificates && \
+    apt-get clean && rm -rf /var/lib/apt/lists/*
+
+RUN groupadd -r appgroup && useradd -r -g appgroup appuser
+
+COPY --from=builder /app/.venv /app/.venv
+COPY llm-budget-proxy/app/ ./app/
+
+ENV PATH="/app/.venv/bin:$PATH"
+ENV PYTHONUNBUFFERED=1
+
+RUN chown -R appuser:appgroup /app
+USER appuser
+
+EXPOSE 8080
+
+HEALTHCHECK --interval=30s --timeout=3s --start-period=10s --retries=3 \
+    CMD python -c "import urllib.request; urllib.request.urlopen('http://localhost:8080/health')" || exit 1
+
+CMD ["uvicorn", "app.main:app", "--host", "0.0.0.0", "--port", "8080"]
--- a/kagenti/llm-budget-proxy/Dockerfile
+++ b/kagenti/llm-budget-proxy/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.12-slim AS builder
+FROM python:3.12-slim@sha256:5072b08ad74609c5329ab4085a96dfa873de565fb4751a4cfcd7dcc427661df0 AS builder
 
 WORKDIR /app
 
@@ -10,7 +10,7 @@ RUN uv venv /app/.venv && \
     . /app/.venv/bin/activate && \
     uv pip install --no-cache .
 
-FROM python:3.12-slim
+FROM python:3.12-slim@sha256:5072b08ad74609c5329ab4085a96dfa873de565fb4751a4cfcd7dcc427661df0
 
 WORKDIR /app
 
--- a/kagenti/llm-budget-proxy/app/main.py
+++ b/kagenti/llm-budget-proxy/app/main.py
@@ -257,11 +257,8 @@ async def chat_completions(request: Request):
     session_id = meta["session_id"]
     max_tokens = meta["max_session_tokens"] or DEFAULT_SESSION_MAX_TOKENS
 
-    logger.info(
-        "LLM request: stream=%s max_tokens=%d",
-        bool(body.get("stream")),
-        max_tokens,
-    )
+    is_streaming = bool(body.get("stream"))
+    logger.info("LLM request: stream=%s max_tokens=%d", is_streaming, max_tokens)
 
     # Budget check
     budget_resp = await _check_budget(session_id, max_tokens, meta, model)
--- a/kagenti/llm-budget-proxy/app/main.py
+++ b/kagenti/llm-budget-proxy/app/main.py
@@ -30,6 +30,7 @@ def _safe(value: object) -> str:
     """Strip control characters from a value before logging."""
     return _LOG_UNSAFE.sub("", str(value))
 
+
 logger = logging.getLogger("llm-budget-proxy")
 logging.basicConfig(
     level=logging.INFO, format="%(asctime)s %(name)s %(levelname)s %(message)s"
@@ -223,7 +224,10 @@ async def _check_budget(
             error_message=msg,
         )
         logger.warning(
-            "Budget exceeded for session %s: %d/%d", _safe(session_id[:12]), used, max_tokens
+            "Budget exceeded for session %s: %d/%d",
+            _safe(session_id[:12]),
+            used,
+            max_tokens,
         )
         return JSONResponse(
             status_code=402,
--- a/kagenti/llm-budget-proxy/app/main.py
+++ b/kagenti/llm-budget-proxy/app/main.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 import json
 import logging
 import os
+import re
 import time
 from contextlib import asynccontextmanager
 from uuid import uuid4
@@ -20,6 +21,14 @@ import asyncpg
 import httpx
 from fastapi import FastAPI, Request
 from fastapi.responses import JSONResponse, StreamingResponse
+
+# Sanitize user-provided values for safe logging (prevent log injection CWE-117)
+_LOG_UNSAFE = re.compile(r"[\x00-\x1f\x7f]")
+
+
+def _safe(value: object) -> str:
+    """Strip control characters from a value before logging."""
+    return _LOG_UNSAFE.sub("", str(value))
 
 logger = logging.getLogger("llm-budget-proxy")
 logging.basicConfig(
@@ -188,10 +197,10 @@ async def _record_call(
     if total_tokens > 0:
         logger.info(
             "Recorded: session=%s agent=%s tokens=%d status=%s",
-            session_id[:12] if session_id else "none",
-            agent_name or "unknown",
+            _safe(session_id[:12]) if session_id else "none",
+            _safe(agent_name) or "unknown",
             total_tokens,
-            status,
+            _safe(status),
         )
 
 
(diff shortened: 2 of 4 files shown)


kagenti/llm-budget-proxy/app/main.py
```diff
@@ -0,0 +1,485 @@
...
+
+    if body.get("stream"):
+        return StreamingResponse(
+            _stream_and_track(body, api_key, meta, start_time),
+            media_type="text/event-stream",
+            headers={"X-Accel-Buffering": "no", "Cache-Control": "no-cache"},
+        )
+
+    # Non-streaming: forward and record
+    async with httpx.AsyncClient(timeout=httpx.Timeout(300.0)) as client:
```
pdettori: **must-fix**: `httpx.AsyncClient` is created per-request — every call opens a new TCP connection and tears it down. Under load this exhausts file descriptors and adds latency.

Create a shared `httpx.AsyncClient` in the `lifespan` context manager and reuse it across all endpoints. Same issue at L329 (streaming), L381 (embeddings), L411 (models).

```python
# In lifespan:
http_client = httpx.AsyncClient(timeout=httpx.Timeout(300.0))
app.state.http_client = http_client
yield
await http_client.aclose()
```

kagenti/llm-budget-proxy/app/main.py
```diff
@@ -0,0 +1,485 @@
...
+            session_id=session_id,
+            user_id=meta["user_id"],
+            agent_name=meta["agent_name"],
+            namespace=meta["namespace"],
+            model=model,
+            latency_ms=latency_ms,
+            status="error",
+            error_message=f"LiteLLM returned {resp.status_code}",
+        )
+        return JSONResponse(status_code=resp.status_code, content=resp.json())
```
pdettori: **must-fix**: `resp.json()` on error responses will crash if LiteLLM returns non-JSON (e.g. nginx 502 HTML page). This gives callers an unhandled 500 instead of a useful error.

Wrap in try/except or check content-type:
```python
try:
    content = resp.json()
except Exception:
    content = {"error": {"message": resp.text[:500], "type": "upstream_error"}}
return JSONResponse(status_code=resp.status_code, content=content)
```
Same issue at L404 (embeddings) and L416 (models).

kagenti/llm-budget-proxy/app/main.py
```diff
@@ -0,0 +1,485 @@
...
+    )
+)
+DATABASE_URL = os.environ.get("DATABASE_URL", "")
+DEFAULT_SESSION_MAX_TOKENS = int(
+    os.environ.get("DEFAULT_SESSION_MAX_TOKENS", "1000000")
+)
+CACHE_TTL = float(os.environ.get("CACHE_TTL", "5.0"))
+
+# In-memory session token cache: session_id -> (tokens, monotonic_timestamp)
+_session_cache: dict[str, tuple[int, float]] = {}
```
pdettori, line 63: **suggestion**: `_session_cache` grows without bound — every unique `session_id` is added but entries are only evicted on write. Over time this leaks memory.

Consider `cachetools.TTLCache(maxsize=10000, ttl=CACHE_TTL)` or a simple bounded dict with LRU eviction.

kagenti/llm-budget-proxy/app/main.py
```diff
@@ -0,0 +1,485 @@
...
+        logger.info("Recorded: tokens=%d status=%s", total_tokens, status or "ok")
+
+
+async def _check_budget(
+    session_id: str, max_tokens: int, meta: dict, model: str
+) -> JSONResponse | None:
+    """Check session budget. Returns 402 response if exceeded, None if OK."""
+    if not session_id or max_tokens <= 0:
+        return None
+    used = await _get_session_tokens(session_id)
```
pdettori, line 225: **suggestion**: TOCTOU race in budget check — between `_get_session_tokens()` (here) and `_record_call()` (after the LLM call returns), concurrent requests for the same session can all pass the check and overspend. The in-memory cache with TTL makes this wider.

Fine for MVP if documented. Before production, consider an atomic check: `SELECT ... FOR UPDATE` in the same transaction as the insert, or a Redis-based atomic counter.

kagenti/llm-budget-proxy/app/main.py
```diff
@@ -0,0 +1,485 @@
...
+        completion_tokens=completion_tokens,
+        total_tokens=total_tokens,
+        latency_ms=latency_ms,
+    )
+
+
+@app.post("/v1/completions")
+async def completions(request: Request):
+    """Forward completions endpoint — same logic as chat/completions."""
+    return await chat_completions(request)
```
pdettori, line 376: **suggestion**: `/v1/completions` delegates to `chat_completions` — but the completions API has a different body schema (`prompt` string vs `messages` array). The request gets forwarded to LiteLLM's `/v1/chat/completions` endpoint regardless.

If LiteLLM handles this routing internally, add a comment explaining that. Otherwise route to the correct upstream endpoint.

kagenti/llm-budget-proxy/Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM python:3.12-slim@sha256:5072b08ad74609c5329ab4085a96dfa873de565fb4751a4cfcd7dcc427661df0 AS builder
+
+WORKDIR /app
+
+COPY --from=ghcr.io/astral-sh/uv:0.9.24 /uv /bin/uv
```
pdettori, line 5: **suggestion**: `ghcr.io/astral-sh/uv:0.9.24` is pinned by tag only. The Python base images are correctly digest-pinned — pin uv by digest too for supply-chain consistency.

```dockerfile
COPY --from=ghcr.io/astral-sh/uv:0.9.24@sha256:<digest> /uv /bin/uv
```

kagenti/llm-budget-proxy/app/main.py
```diff
@@ -0,0 +1,485 @@
+"""LLM Budget Proxy — per-session and per-agent token budget enforcement.
+
+A small FastAPI proxy that sits between agents and LiteLLM. It:
+1. Checks per-session token budget before forwarding requests
```
pdettori, line 4: **nit**: Module docstring says "Returns 402 when budget is exceeded" but the PR description says "returns 429". The code correctly returns 402 (Payment Required), which is semantically appropriate. Update the PR description to match.

kagenti/llm-budget-proxy/pyproject.toml
```diff
@@ -0,0 +1,11 @@
+[project]
+name = "llm-budget-proxy"
+version = "0.1.0"
+description = "Per-session and per-agent LLM token budget enforcement proxy"
+requires-python = ">=3.11"
+dependencies = [
+    "fastapi>=0.115.0",
```
pdettori, line 7: **nit**: Dependencies use `>=` lower bounds only. For a security-sensitive proxy, consider pinning upper bounds (e.g. `fastapi>=0.115.0,<1.0`) or generating a lockfile to ensure reproducible builds.

pdettori (changes requested):
A well-structured budget proxy with good security practices: log sanitization (`_safe`), SSRF mitigation (`_validate_backend_url`), parameterized SQL queries, non-root Dockerfile with digest-pinned base images, and proper health checks.

Two must-fix issues: (1) creating a new httpx client per request will cause performance problems under load — use a shared client; (2) calling `resp.json()` on error responses can crash if the upstream returns non-JSON. The race condition in budget enforcement is acceptable for MVP but should be addressed before production.

The weather deployment change (Ollama → LiteMaaS) is a separate concern — consider whether it should be a separate PR.

**Areas reviewed**: Python, Dockerfile, YAML, pyproject.toml
**Commits**: 11, all signed-off ✓
**CI status**: Deploy & Test failing; all static checks passing

Also note: no unit tests for the new 485-line service. At minimum, tests for `_extract_metadata`, `_check_budget`, and `_safe` would catch regressions.

kagenti/llm-budget-proxy/app/main.py
```diff
@@ -0,0 +1,496 @@
...
+from contextlib import asynccontextmanager
+from uuid import uuid4
+
+import asyncpg
+import httpx
+from fastapi import FastAPI, Request
+from fastapi.responses import JSONResponse, StreamingResponse
+
+# Sanitize user-provided values for safe logging (prevent log injection CWE-117)
+_LOG_UNSAFE = re.compile(r"[\x00-\x1f\x7f]")
```
pdettori, line 27: **suggestion**: `_safe()` and `_LOG_UNSAFE` are dead code — defined but never called anywhere. The log-sanitization approach evolved to removing user data from logs entirely, making these unused. Remove to avoid confusion.

kagenti/llm-budget-proxy/app/main.py
```diff
@@ -0,0 +1,496 @@
...
+            await conn.execute(INSERT_DEFAULT_BUDGETS_SQL)
+        logger.info("DB migrated — tables ready")
+    logger.info("LLM Budget Proxy ready — LITELLM_URL=%s", LITELLM_URL)
+    yield
+    if db:
+        await db.close()
+
+
+# Module-level shared client for connection reuse
+_http_client = httpx.AsyncClient(timeout=httpx.Timeout(300.0))
```
pdettori, line 139: **suggestion**: The shared `_http_client` is created at module level but never closed in the `lifespan` shutdown. Add `await _http_client.aclose()` alongside `db.close()` to avoid connection leaks on graceful shutdown:

```python
yield
if db:
    await db.close()
await _http_client.aclose()
```

pdettori (commented):
Previous must-fix items (shared httpx client, non-JSON error handling) are resolved ✓

Remaining items are suggestions — no merge blockers. See inline comments.

**Areas reviewed**: Python, Dockerfile, YAML, pyproject.toml
**Commits**: 12, all signed-off ✓
**CI status**: Deploy & Test failing; all static checks passing

kagenti/llm-budget-proxy/app/main.py
```diff
@@ -0,0 +1,496 @@
...
+    prompt_tokens = 0
+    completion_tokens = 0
+    total_tokens = 0
+    model = body.get("model", "")
+
+    # Ensure LiteLLM sends usage in the final chunk
+    body.setdefault("stream_options", {})
+    body["stream_options"]["include_usage"] = True
+
+    async with _http_client.stream(
```
pdettori, line 335: **suggestion**: `_stream_and_track` doesn't check the upstream response status before streaming. If LiteLLM returns 4xx/5xx, the error body gets streamed as raw text rather than returned as a proper JSON error. Consider checking `resp.status_code` inside the `async with` and falling back to a JSON error response for non-2xx.

kagenti/llm-budget-proxy/pyproject.toml
```diff
@@ -0,0 +1,11 @@
+[project]
+name = "llm-budget-proxy"
+version = "0.1.0"
+description = "Per-session and per-agent LLM token budget enforcement proxy"
+requires-python = ">=3.11"
+dependencies = [
```
pdettori, line 6: **nit**: Dependencies use `>=` (floating). Without a `uv.lock` committed alongside, builds aren't reproducible across environments. Consider adding a lock file.

kagenti/llm-budget-proxy/app/main.py
```diff
@@ -0,0 +1,496 @@
...
+async def health():
+    """Readiness/liveness probe."""
+    if db:
+        try:
+            await db.fetchval("SELECT 1")
+        except Exception:
+            return JSONResponse(
+                status_code=503, content={"status": "unhealthy", "db": "unreachable"}
+            )
+    return {"status": "healthy", "db": "connected" if db else "disabled"}
```
pdettori, line 496: **suggestion**: Still no unit tests for this 496-line service. At minimum, `_extract_metadata`, `_check_budget`, and the budget-exceeded 402 path would benefit from tests to catch regressions.

pdettori (approved):
/lgtm

PR #986, opened 2026-03-16
feat: add sandbox UI types, API service, and utilities

Ladas:
## Summary
- AgentLoop type definitions for structured reasoning loop events
- Sandbox session types (SandboxConfig, SandboxSession, etc.)
- Integration, FileBrowser, PodStorage types
- API service extensions: sandbox, sidecar, analytics endpoints
- Utilities: loopBuilder (event processing), historyPairing (message matching), loopFormatting
- Component barrel exports, dark mode CSS

9 files | Part of Sandbox Agent streaming PR series

### DIFF
--- a/kagenti/ui-v2/src/services/api.ts
+++ b/kagenti/ui-v2/src/services/api.ts
@@ -11,6 +11,16 @@ import type {
   Tool,
   ToolDetail,
   ApiListResponse,
+  Integration,
+  IntegrationDetail,
+  IntegrationProvider,
+  IntegrationAgentRef,
+  IntegrationWebhook,
+  IntegrationSchedule,
+  IntegrationAlert,
+  FileEntry,
+  FileContent,
+  PodStorageStats,
 } from '@/types';
 
 // API configuration
@@ -38,6 +48,18 @@ export function setTokenGetter(getter: () => Promise<string | null>): void {
  */
 export function setTokenForceRefresher(refresher: () => Promise<string | null>): void {
   tokenForceRefresher = refresher;
+}
+
+/**
+ * Error class that preserves the HTTP status code from API responses.
+ */
+export class ApiError extends Error {
+  status: number;
+  constructor(message: string, status: number) {
+    super(message);
+    this.name = 'ApiError';
+    this.status = status;
+  }
 }
 
 /**
@@ -106,8 +128,9 @@ async function apiFetch<T>(
 
   if (!response.ok) {
     const errorData = await response.json().catch(() => ({}));
-    throw new Error(
-      errorData.detail || `API error: ${response.status} ${response.statusText}`
+    throw new ApiError(
+      errorData.detail || `API error: ${response.status} ${response.statusText}`,
+      response.status
     );
   }
 
@@ -656,6 +679,44 @@ export const configService = {
 };
 
 /**
+ * Session Graph types and service (Session E)
+ */
+export interface GraphNode {
+  id: string;
+  agent: string;
+  status: 'running' | 'completed' | 'failed' | 'pending';
+  mode: 'root' | 'in-process' | 'shared-pvc' | 'isolated' | 'sidecar';
+  tier: string;
+  started_at: string | null;
+  duration_ms: number;
+  task_summary: string;
+}
+
+export interface GraphEdge {
+  from: string;
+  to: string;
+  mode: 'in-process' | 'shared-pvc' | 'isolated' | 'sidecar';
+  task: string;
+}
+
+export interface SessionGraphData {
+  root: string;
+  nodes: GraphNode[];
+  edges: GraphEdge[];
+}
+
+export const sessionGraphService = {
+  async getGraph(
+    namespace: string,
+    contextId: string
+  ): Promise<SessionGraphData> {
+    return apiFetch(
+      `/chat/${encodeURIComponent(namespace)}/sessions/${encodeURIComponent(contextId)}/graph`
+    );
+  },
+};
+
+/**
  * Chat service for A2A agent communication
  */
 export const chatService = {
@@ -675,9 +736,16 @@ export const chatService = {
       examples?: string[];
     }>;
   }> {
-    return apiFetch(
-      `/chat/${encodeURIComponent(namespace)}/${encodeURIComponent(name)}/agent-card`
-    );
+    try {
+      return await apiFetch(
+        `/chat/${encodeURIComponent(namespace)}/${encodeURIComponent(name)}/agent-card`
+      );
+    } catch {
+      // Fallback: sandbox endpoint (direct port 8000, no AuthBridge retry)
+      return apiFetch(
+        `/sandbox/${encodeURIComponent(namespace)}/agent-card/${encodeURIComponent(name)}`
+      );
+    }
   },
 
   async sendMessage(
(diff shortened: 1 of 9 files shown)


kagenti/ui-v2/src/utils/loopBuilder.ts
```diff
@@ -0,0 +1,606 @@
...
+ * and history reconstruction.
+ *
+ * Event routing is category-based: the event type is looked up in
+ * EVENT_CATALOG to determine its category, then dispatched to a
+ * per-category handler function.
+ */
+export function applyLoopEvent(loop: AgentLoop, le: LoopEvent): AgentLoop {
+  // Normalize: agent may emit plan_step or current_step
+  if (le.plan_step != null && le.current_step == null) {
+    le.current_step = le.plan_step;
```
pdettori: **must-fix**: `applyLoopEvent` directly mutates the caller's event object here:

```ts
le.current_step = le.plan_step;  // mutates input
```

The JSDoc declares this a _"pure function — safe for React state."_ If a caller retains the original event (e.g. history replay or logging), the alias is silently applied in-place, causing inconsistent behaviour on a second pass.

Fix: spread before normalizing:
```ts
const le2 = le.plan_step != null && le.current_step == null
  ? { ...le, current_step: le.plan_step }
  : le;
```
Then use `le2` throughout the function body.

kagenti/ui-v2/src/utils/loopBuilder.ts
```diff
@@ -0,0 +1,606 @@
...
+    const loopId = evt.loop_id;
+    if (!loopId) continue;
+    const prev = loops.get(loopId) || createDefaultAgentLoop(loopId);
+    loops.set(loopId, applyLoopEvent(prev, evt));
+  }
+  // Mark loops as done or failed based on whether they completed
+  for (const [, loop] of loops) {
+    const hasReporter = loop.steps.some((s) => s.nodeType === 'reporter');
+    if (hasReporter) {
+      loop.status = 'done';
```
pdettori, line 588: **must-fix**: `buildAgentLoops` directly mutates `AgentLoop` and `AgentLoopStep` objects that live in the Map (e.g. `loop.status = 'done'`, `step.status = 'failed'`, `loop.steps.sort(...)`). This is inconsistent with the immutable pattern used in `applyLoopEvent` and can cause subtle React rendering bugs if callers hold references into the Map.

Use spread/copy instead:
```ts
loops.set(loopId, {
  ...loop,
  status: 'done',
  steps: [...loop.steps].sort((a, b) => a.index - b.index).map(s =>
    (s.status === 'running' || s.status === 'pending') ? { ...s, status: 'done' } : s
  ),
});
```

kagenti/ui-v2/src/services/api.ts
```diff
@@ -702,3 +770,650 @@ export const chatService = {
...
+
+/**
+ * Models service for fetching available LLM models from LiteLLM
+ */
+export const modelsService = {
+  async getAvailableModels(): Promise<Array<{id: string}>> {
+    return apiFetch<Array<{id: string}>>('/models');
+  },
+  async getAgentModels(namespace: string, agentName: string): Promise<Array<{id: string}>> {
+    return apiFetch<Array<{id: string}>>(`/llm/agent-models/${namespace}/${agentName}`);
```
pdettori: **must-fix**: `namespace` and `agentName` are embedded in the URL without `encodeURIComponent`. Every other path-segment interpolation in this file uses it. If either value contains `/`, `?`, `#`, or other reserved characters this will produce a broken URL.

```ts
// Fix:
return apiFetch<Array<{id: string}>>(
  `/llm/agent-models/${encodeURIComponent(namespace)}/${encodeURIComponent(agentName)}`
);
```

kagenti/ui-v2/src/services/api.ts
```diff
@@ -702,3 +770,650 @@ export const chatService = {
...
+  },
+
+  async getChildSessions(namespace: string, contextId: string): Promise<Array<{
+    context_id: string;
+    agent_name: string;
+    title: string;
+    state: string;
+    timestamp: string;
+  }>> {
+    const response = await apiFetch<{items: Array<Record<string, unknown>>}>(
```
pdettori, line 969: **suggestion**: This fetches with a hard-coded `limit=100` and then filters client-side by `parent_context_id`. If a namespace has more than 100 sessions, child sessions beyond the first page will be silently missed.

Either pass `parent_context_id` as a server-side query param (if the backend supports it), paginate until exhausted, or add a comment documenting the 100-session cap.

kagenti/ui-v2/src/services/api.ts
```diff
@@ -702,3 +770,650 @@ export const chatService = {
...
+    );
+  },
+
+  async getFileContent(
+    namespace: string,
+    agentName: string,
+    filePath: string,
+    contextId?: string
+  ): Promise<FileContent> {
+    if (contextId) {
```
pdettori, line 1094: **suggestion**: When `contextId` is provided, both `listDirectory` and `getFileContent` construct the exact same URL:
```
/sandbox/{ns}/files/{agent}/{contextId}?path={path}
```
Yet they declare different return types (`{ entries: FileEntry[] }` vs `FileContent`). The backend must distinguish these somehow — if the response shape is always the same, one of the callers will silently receive the wrong runtime type. Consider adding a discriminator like `?type=list` / `?type=content`, or a comment confirming how the backend distinguishes them.

kagenti/ui-v2/src/services/api.ts
```diff
@@ -675,9 +736,16 @@ export const chatService = {
...
-      `/chat/${encodeURIComponent(namespace)}/${encodeURIComponent(name)}/agent-card`
-    );
+    try {
+      return await apiFetch(
+        `/chat/${encodeURIComponent(namespace)}/${encodeURIComponent(name)}/agent-card`
+      );
+    } catch {
+      // Fallback: sandbox endpoint (direct port 8000, no AuthBridge retry)
+      return apiFetch(
+        `/sandbox/${encodeURIComponent(namespace)}/agent-card/${encodeURIComponent(name)}`
```
pdettori, line 746: **suggestion**: The `catch {}` block here (and the identical pattern in `graphCardService.fetchGraphCard`) silently retries on *any* error — including 401 Unauthorized, 500 Internal Server Error, and network timeouts. This can mask real failures that are not "agent unavailable" conditions.

Consider narrowing to specific statuses:
```ts
} catch (err) {
  if (err instanceof ApiError && (err.status === 404 || err.status === 503)) {
    return apiFetch(...);
  }
  throw err;
}
```

pdettori (changes requested):
## Review Summary

This PR adds a large, well-structured foundation for the sandbox UI — types, API client methods, and loop-event processing utilities. The `EVENT_CATALOG` + exhaustiveness check pattern in `loopBuilder.ts` is excellent, and the new `ApiError` class preserving HTTP status is a real improvement over the previous generic `Error`.

Three must-fix issues before merge:
1. Mutation of the `le` input in `applyLoopEvent` — breaks the stated pure-function contract and could cause subtle history-replay bugs
2. In-place mutation of loop/step objects in `buildAgentLoops` — inconsistent with the immutable state model used throughout
3. Missing `encodeURIComponent` in `modelsService.getAgentModels`

**Areas reviewed**: TypeScript (types, API service, utils), CSS, commit conventions, security  
**CI status**: All checks passing ✓  
**Commits**: 6 commits, all signed-off ✓

pdettori (approved):
/lgtm

PR #988, opened 2026-03-16
feat: add graph visualization views

Ladas:
## Summary
- GraphLoopView: wrapper with Step/Topology subtab toggle
- StepGraphView: ReactFlow DAG of per-step execution flow
- TopologyGraphView: ReactFlow DAG of agent graph topology
- GraphDetailPanel: sliding panel for node drill-down
- FloatingViewBar: Simple/Advanced/Graph view mode toggle
- SessionGraphPage: session delegation tree visualization

6 files | Part of Sandbox Agent streaming PR series

PR #989, opened 2026-03-16
feat: add supporting UI components — files, events, metrics

Ladas:
## Summary
- FileBrowser: workspace file tree with breadcrumb navigation
- FilePreview/FilePreviewModal: syntax highlighted file content viewer
- EventsPanel: expandable A2A event log
- LlmUsagePanel: per-model token usage and cost breakdown
- PodStatusPanel: live pod status, metrics, and events
- SubSessionsPanel: child session listing for delegation
- SidecarTab: sidecar agent management

8 files | Part of Sandbox Agent streaming PR series

### DIFF
--- a/kagenti/ui-v2/src/components/FileBrowser.tsx
+++ b/kagenti/ui-v2/src/components/FileBrowser.tsx
@@ -1,0 +1,443 @@
+// Copyright 2025 IBM Corp.
+// Licensed under the Apache License, Version 2.0
+
+import React, { Component, useState, useMemo } from 'react';
+import type { ErrorInfo, ReactNode } from 'react';
+import { useParams, useSearchParams } from 'react-router-dom';
+import {
+  Breadcrumb,
+  BreadcrumbItem,
+  PageSection,
+  Spinner,
+  TreeView,
+  EmptyState,
+  EmptyStateHeader,
+  EmptyStateIcon,
+  EmptyStateBody,
+  Title,
+  Alert,
+} from '@patternfly/react-core';
+import type { TreeViewDataItem } from '@patternfly/react-core';
+import {
+  FolderIcon,
+  FileCodeIcon,
+  FileIcon,
+  LockIcon,
+  ExclamationCircleIcon,
+  CubesIcon,
+} from '@patternfly/react-icons';
+import { useQuery } from '@tanstack/react-query';
+
+import { sandboxFileService, ApiError } from '@/services/api';
+import type { FileEntry } from '@/types';
+import { FilePreview } from './FilePreview';
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+const CODE_EXTENSIONS = new Set([
+  '.py', '.js', '.ts', '.tsx', '.jsx', '.go', '.rs', '.java', '.rb',
+  '.sh', '.bash', '.zsh', '.yaml', '.yml', '.json', '.toml', '.xml',
+  '.html', '.css', '.scss', '.sql', '.c', '.cpp', '.h', '.hpp',
+  '.md', '.mdx', '.markdown', '.dockerfile', '.tf', '.hcl',
+]);
+
+function isCodeFile(name: string): boolean {
+  const lower = name.toLowerCase();
+  const dotIdx = lower.lastIndexOf('.');
+  if (dotIdx === -1) return false;
+  return CODE_EXTENSIONS.has(lower.slice(dotIdx));
+}
+
+function iconForEntry(entry: FileEntry): React.ReactNode {
+  if (entry.type === 'directory') return <FolderIcon />;
+  if (isCodeFile(entry.name)) return <FileCodeIcon />;
+  return <FileIcon />;
+}
+
+/**
+ * Sort entries: directories first, then files; alphabetically within each group.
+ */
+function sortEntries(entries: FileEntry[]): FileEntry[] {
+  return [...entries].sort((a, b) => {
+    if (a.type === 'directory' && b.type !== 'directory') return -1;
+    if (a.type !== 'directory' && b.type === 'directory') return 1;
+    return a.name.localeCompare(b.name);
+  });
+}
+
+/**
+ * Build path segments for breadcrumb from an absolute path.
+ * e.g. "/workspace/src/lib" => ["/workspace", "/workspace/src", "/workspace/src/lib"]
+ */
+function pathSegments(path: string): Array<{ label: string; fullPath: string }> {
+  const parts = path.split('/').filter(Boolean);
+  const segments: Array<{ label: string; fullPath: string }> = [];
+  let accumulated = '';
+  for (const part of parts) {
+    accumulated += '/' + part;
+    segments.push({ label: part, fullPath: accumulated });
+  }
+  return segments;
+}
+
+// ---------------------------------------------------------------------------
+// ErrorBoundary for FilePreview — catches render crashes
+// ---------------------------------------------------------------------------
+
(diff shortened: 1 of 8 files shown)


pdettori (commented):
## Review Summary

Eight new UI support components: file browser with TreeView + breadcrumb, file preview with mermaid/markdown rendering, A2A event log with HITL approvals, LLM token usage table, live pod status with react-query polling, sidecar agent SSE stream, and child session listing. Code quality is high overall — correct async cleanup patterns in most components, error boundaries wrapping preview components, proper react-query usage for polling, and `data-testid` attributes throughout.

Main actionable items: a missing `cancelled` flag in `useChildSessionCount` (inconsistent with the pattern used elsewhere in the same file), and the mermaid `innerHTML` path relying on an undocumented runtime assumption about DOMPurify availability.

**Areas reviewed**: Frontend (TypeScript/React), Security (innerHTML/XSS), Commit/PR conventions  
**Commits**: 11 commits, all signed-off: ✅ DCO passes  
**CI status**: All checks pass except `E2E HyperShift` — same pre-existing infrastructure failure as companion PR #988, unrelated to this purely-frontend change.

kagenti/ui-v2/src/components/FilePreview.tsx
```diff
@@ -0,0 +1,245 @@
...
+  const containerRef = useRef<HTMLDivElement>(null);
+
+  const renderChart = useCallback(async () => {
+    if (!containerRef.current) return;
+    try {
+      const id = `mermaid-block-${++mermaidCounter}`;
+      const { svg } = await mermaid.render(id, chart);
+      if (containerRef.current) {
+        // Mermaid v10+ sanitizes SVG via DOMPurify internally
+        containerRef.current.innerHTML = svg;
```
pdettori, line 98: **suggestion**: The last commit reverts the explicit `<script>` tag stripping in favor of trusting mermaid's built-in DOMPurify sanitization — valid for mermaid v10+. However this assumption isn't documented or guarded anywhere in the code. Recommend adding a comment citing the minimum required version:
```tsx
// Mermaid v10+ sanitizes SVG output via DOMPurify before returning.
// If upgrading mermaid below v10, re-add explicit sanitization here.
containerRef.current.innerHTML = svg;
```
This way a future dependency update doesn't silently remove the sanitization without anyone noticing.

kagenti/ui-v2/src/components/SubSessionsPanel.tsx
```diff
@@ -0,0 +1,161 @@
...
+};
+
+/** Returns the number of child sessions (for badge display). */
+export const useChildSessionCount = (namespace: string, contextId: string | null): number => {
+  const [count, setCount] = useState(0);
+  useEffect(() => {
+    if (!contextId) { setCount(0); return; }
+    sandboxService
+      .getChildSessions(namespace, contextId)
+      .then((result) => setCount(result.length))
```
pdettori: **suggestion**: `useChildSessionCount` fires a fetch but has no `cancelled` flag — if the component unmounts before the promise resolves, `setCount` is called on an unmounted component. The `SubSessionsPanel` fetch just above this hook correctly uses `let cancelled = false` + `return () => { cancelled = true; }`. The same pattern should be applied here:
```tsx
useEffect(() => {
  if (!contextId) { setCount(0); return; }
  let cancelled = false;
  sandboxService
    .getChildSessions(namespace, contextId)
    .then((result) => { if (!cancelled) setCount(result.length); })
    .catch(() => { if (!cancelled) setCount(0); });
  return () => { cancelled = true; };
}, [namespace, contextId]);
```

kagenti/ui-v2/src/components/FilePreview.tsx
```diff
@@ -0,0 +1,245 @@
...
+    />
+  );
+};
+
+// ---------------------------------------------------------------------------
+// Markdown component overrides for ReactMarkdown
+// ---------------------------------------------------------------------------
+
+const markdownComponents: Record<string, React.ComponentType<any>> = {
+  code({ className, children, ...rest }: any) {
```
pdettori, line 124: **nit**: `code({ className, children, ...rest }: any)` bypasses TS. `react-markdown` exports proper types:
```tsx
import type { Components } from 'react-markdown';

const markdownComponents: Components = {
  code({ className, children, node, ...rest }) {
    // ...
  },
};
```
This gives full type safety without reaching for `any`.

kagenti/ui-v2/src/components/PodStatusPanel.tsx
```diff
@@ -0,0 +1,568 @@
...
+              }}
+            >
+              #
+            </th>
+          </tr>
+        </thead>
+        <tbody>
+          {podEvents.slice(0, 30).map((evt: PodEventDetail, i: number) => (
+            <tr
+              key={i}
```
pdettori, line 262: **nit**: `key={i}` (array index) for pod events. Pod events can be re-ordered or have new entries prepended on refresh. A stable composite key avoids stale component state:
```tsx
key={`${evt.reason}-${evt.message}-${evt.lastTime}`}
```

kagenti/ui-v2/src/components/PodStatusPanel.tsx
```diff
@@ -0,0 +1,568 @@
...
+  type PodEventDetail,
+} from '../services/api';
+
+// ---------------------------------------------------------------------------
+// Constants
+// ---------------------------------------------------------------------------
+
+const POLL_INTERVAL_MS = 7000;
+
+const STATUS_COLORS: Record<string, string> = {
```
pdettori, line 29: **nit**: `STATUS_COLORS` uses hardcoded hex values (`#2ea44f`, `#cf222e`, `#bf8700`) for semantic status states. PatternFly already provides theme-aware equivalents:
- Green (Running): `var(--pf-v5-global--success-color--100)`
- Red (Error): `var(--pf-v5-global--danger-color--100)`
- Amber (Pending): `var(--pf-v5-global--warning-color--100)`
- Grey (Unknown): `var(--pf-v5-global--Color--200)`

Using PF tokens keeps the status indicators theme-aware at no extra cost.

kagenti/ui-v2/src/components/SidecarTab.tsx
```diff
@@ -0,0 +1,717 @@
...
+              }
+            }
+          }
+        }
+      } catch (err) {
+        if (err instanceof DOMException && err.name === 'AbortError') {
+          // Expected on cleanup
+          return;
+        }
+        console.error('Sidecar SSE connection error:', err);
```
pdettori, line 258: **nit**: The SSE stream has no reconnect logic — on a dropped connection it logs to `console.error` and silently stops. For a live observation feed this is invisible to the user. Consider tracking a connection error in state and showing a small 'Stream disconnected — click to reconnect' indicator, or a simple exponential-backoff reconnect loop.

kagenti/ui-v2/src/components/FileBrowser.tsx
```diff
@@ -0,0 +1,443 @@
...
+// ---------------------------------------------------------------------------
+// ErrorBoundary for FilePreview — catches render crashes
+// ---------------------------------------------------------------------------
+
+interface PreviewErrorBoundaryState {
+  hasError: boolean;
+  error: Error | null;
+}
+
+class PreviewErrorBoundary extends Component<
```
pdettori, line 94: **praise**: The `PreviewErrorBoundary` + `componentDidUpdate` reset-on-child-change pattern is exactly right — it catches render crashes from malformed files without requiring the user to reload the page. Good defensive practice for a file browser that can receive arbitrary content.

pdettori (approved):
/lgtm

PR #993, opened 2026-03-16
feat: add sandbox UI build config, services, and cleanup

Ladas:
## Summary
- UI build: Dockerfile (npm ci), nginx proxy config for SSE streaming
- UI deps: package.json with graph/diagram dependencies, vitest config
- useSessionLoader hook, eventService, AgentGraphCard types
- Graph streaming animation CSS
- skill-packs.yaml for pinned sandbox agent skills
- Keycloak test user creation script
- Remove empty bracket-name artifact files

20 files | Part of Sandbox Agent streaming PR series

### DIFF
--- a/kagenti/auth/create-test-users.sh
+++ b/kagenti/auth/create-test-users.sh
@@ -1,0 +1,156 @@
+#!/usr/bin/env bash
+#
+# Create Test Users in Keycloak
+#
+# Creates dev-user and ns-admin test users in the master realm (or the realm
+# where the kagenti OAuth client is registered). Idempotent — safe to run
+# multiple times.
+#
+# Prerequisites:
+#   - kubectl/oc access to the cluster
+#   - Keycloak pod running in the keycloak namespace
+#   - keycloak-initial-admin secret exists
+#
+# Usage:
+#   # From the repository root:
+#   ./kagenti/auth/create-test-users.sh
+#
+#   # With custom realm (default: master):
+#   KEYCLOAK_REALM=demo ./kagenti/auth/create-test-users.sh
+#
+#   # With custom namespace:
+#   KEYCLOAK_NAMESPACE=my-keycloak ./kagenti/auth/create-test-users.sh
+#
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "$SCRIPT_DIR/../../.github/scripts/lib/logging.sh" 2>/dev/null || {
+    log_step() { echo "==> [$1] $2"; }
+    log_info() { echo "  INFO: $*"; }
+    log_success() { echo "  OK: $*"; }
+    log_warn() { echo "  WARN: $*"; }
+    log_error() { echo "  ERROR: $*"; }
+}
+
+log_step "D" "Create test users in Keycloak"
+
+KC_NS="${KEYCLOAK_NAMESPACE:-keycloak}"
+KC_POD="keycloak-0"
+KCADM="/opt/keycloak/bin/kcadm.sh"
+# TODO: Upstream is moving kagenti OAuth client from master realm to demo realm.
+# Once that lands (after rebase), change default to "demo" and update the
+# kagenti-ui-oauth-secret job to use demo realm endpoints.
+REALM="${KEYCLOAK_REALM:-master}"
+
+# ── Step 1: Wait for Keycloak pod ─────────────────────────────────────────
+log_info "Waiting for Keycloak pod to be ready..."
+kubectl wait --for=condition=Ready pod/$KC_POD -n "$KC_NS" --timeout=120s
+
+# ── Step 2: Login to Keycloak ─────────────────────────────────────────────
+log_info "Reading credentials from keycloak-initial-admin secret..."
+KC_USER=$(kubectl get secret keycloak-initial-admin -n "$KC_NS" \
+    -o jsonpath='{.data.username}' 2>/dev/null | base64 -d 2>/dev/null || echo "")
+KC_PASS=$(kubectl get secret keycloak-initial-admin -n "$KC_NS" \
+    -o jsonpath='{.data.password}' 2>/dev/null | base64 -d 2>/dev/null || echo "")
+
+if [ -z "$KC_USER" ] || [ -z "$KC_PASS" ]; then
+    log_error "Could not read keycloak-initial-admin secret"
+    exit 1
+fi
+
+log_info "Logging in as $KC_USER..."
+kubectl exec -n "$KC_NS" "$KC_POD" -- bash -c \
+    "$KCADM config credentials --server http://localhost:8080 --realm master \
+     --user '$KC_USER' --password '$KC_PASS' --config /tmp/kc/kcadm.config" \
+    >/dev/null 2>&1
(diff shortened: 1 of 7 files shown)


kagenti/ui-v2/package.json
```diff
@@ -26,6 +26,8 @@
     "js-yaml": "^4.1.0",
     "mermaid": "^11.12.3",
     "keycloak-js": "~25.0.6",
+    "mermaid": "^11.12.3",
```
pdettori, line 28: **must-fix**: Duplicate `mermaid` entry — `"mermaid": "^11.12.3"` already exists on line 28 (from upstream main). This duplicate will cause npm warnings/errors. Remove this line.

kagenti/ui-v2/package.json
```diff
@@ -26,6 +26,8 @@
...
+    "playwright": "^1.50.1",
```
pdettori: **must-fix**: `playwright` is added to production `dependencies` — this is a test framework (~300 MB of browser binaries) and should be in `devDependencies` if needed at all. Note that `@playwright/test` is already present in devDependencies. If the intent was to add the browser engine package, it should still go in devDeps.

kagenti/ui-v2/Dockerfile
```diff
@@ -8,11 +8,9 @@ WORKDIR /app
 
 # Copy package files
 COPY ui-v2/package.json ./
-# Note: If using npm, use package-lock.json instead
-# COPY package-lock.json ./
 
 # Install dependencies
-RUN npm install
+RUN npm install --legacy-peer-deps
```
pdettori, line 14: **suggestion**: `--legacy-peer-deps` suppresses peer dependency conflict errors, which can mask real incompatibilities. Consider either fixing the peer dep conflicts or adding a comment explaining which specific conflict this works around.

kagenti/ui-v2/nginx.conf
```diff
@@ -28,8 +46,8 @@ server {
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header X-Forwarded-Proto $scheme;
         proxy_connect_timeout 60s;
-        proxy_send_timeout 60s;
-        proxy_read_timeout 60s;
+        proxy_send_timeout 300s;
```
pdettori, line 49: **suggestion**: The generic `/api/` block timeouts were bumped from 60s → 300s. The new `/api/v1/sandbox/` block already has its own 300s timeouts for SSE streaming. Was the intent to also increase timeouts for *all* non-sandbox API calls? If not, consider keeping the generic block at 60s to avoid unnecessarily long-held connections.

kagenti/auth/create-test-users.sh
```diff
@@ -0,0 +1,156 @@
...
+# (changing it via kcadm can fail silently, causing test/secret mismatch).
+# For dev-user and ns-admin, reuse existing passwords or generate random ones.
+_existing_dev=$(kubectl get secret kagenti-test-users -n "$KC_NS" \
+    -o jsonpath='{.data.dev-user-password}' 2>/dev/null | base64 -d 2>/dev/null || echo "")
+_existing_ns=$(kubectl get secret kagenti-test-users -n "$KC_NS" \
+    -o jsonpath='{.data.ns-admin-password}' 2>/dev/null | base64 -d 2>/dev/null || echo "")
+
+_rand() { LC_ALL=C tr -dc 'A-Za-z0-9' </dev/urandom | head -c 15; }
+
+# Admin password: use the actual Keycloak password (do not try to change it)
```
pdettori, line 98: **suggestion**: Variables inside the `bash -c` string are expanded by the outer shell but unquoted in the inner shell context. If `$email`, `$first`, or `$last` ever contain spaces or special characters, the kcadm commands will break. Consider quoting them inside the inner heredoc, e.g.:
```bash
-s email='$email' -s firstName='$first' -s lastName='$last'
```
(with escaped single quotes so they survive the outer expansion)

kagenti/ui-v2/package.json
```diff
@@ -9,7 +9,7 @@
     "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
     "preview": "vite preview",
     "typecheck": "tsc --noEmit",
-    "test:unit": "vitest run src/utils/validation.test.ts src/contexts/keycloakRedirectUri.test.ts",
+    "test:unit": "vitest run",
```
pdettori, line 12: **nit**: `test:unit` changed from running two specific test files to `vitest run` (all tests). This is fine but broadens the scope — worth confirming no in-progress/failing tests got pulled in unintentionally.

pdettori (changes requested):
## Review Summary

PR adds sandbox UI build infrastructure (nginx SSE proxy, Dockerfile cleanup), a Keycloak test user creation script, per-event persistence schema, and switches Kind weather agent from Ollama to LiteMaaS.

**Areas reviewed**: Frontend (package.json, Dockerfile, nginx.conf), Python (event.py), Shell (create-test-users.sh), YAML (ocp_values.yaml, weather deployment)
**Commits**: 7 commits, all signed-off ✓
**CI status**: Deploy & Test ❌, E2E HyperShift ❌; all lint/security checks ✓

Two must-fix issues in package.json. Suggestions on the rest.

kagenti/ui-v2/nginx.conf
pdettori, line 49: **suggestion**: The generic `/api/` block timeouts were bumped from 60s → 300s alongside the SSE block. Only `/api/v1/sandbox/` needs long timeouts for SSE streaming. Consider reverting the generic block to 60s — 300s could mask slow backend responses on non-streaming endpoints.

kagenti/ui-v2/Dockerfile
pdettori, line 14: **nit**: `--legacy-peer-deps` suppresses peer dependency conflict errors. Works fine, but a one-line comment explaining *why* (e.g., graph/diagram deps have unresolved peer conflicts) would help future readers.

kagenti/auth/create-test-users.sh
pdettori, line 98: **nit**: Variables inside the `bash -c` string (`$username`, `$REALM`, etc.) aren't quoted in the inner shell context. Functionally fine since values won't contain spaces, but `shellcheck` would flag this. Low priority.

kagenti/ui-v2/package.json
pdettori, line 12: **praise**: Good call broadening `test:unit` from a pinned file list to `vitest run` — less maintenance burden as tests are added.

pdettori (approved):
## Review Summary

Previous must-fix issues on `package.json` are resolved (duplicate mermaid removed, playwright moved to devDeps). All CI checks green.

The PR adds solid infrastructure: nginx SSE proxy config is correct (buffering off, chunked off, `Connection ''`), event schema is clean with proper indexes and UNIQUE constraint, and the Keycloak test user script is idempotent with proper error handling.

**Areas reviewed**: Frontend (package.json, Dockerfile, nginx.conf), Python (event.py), Shell (create-test-users.sh), YAML (ocp_values.yaml)
**Commits**: 8 commits, all signed-off ✓
**CI status**: All 22 checks passing ✓

PR #978, opened 2026-03-16
feat: add UI feature flag gating

Ladas:
## Summary
- Add `useFeatureFlags` React hook — fetches flags from `/api/v1/config/features` with module-level caching
- Gate sandbox/integrations/triggers routes in App.tsx behind feature flags
- Conditionally show/hide sidebar nav items in AppLayout based on flags

When flags are off: nav items hidden, routes not mounted, components never rendered.

## Part of Sandbox Agent streaming PR series
Depends on: #976 (Helm flags), K3-flags (backend endpoint)
Must merge after K6d (pages) so TypeScript imports resolve.

## Test plan
- [ ] UI builds successfully
- [ ] With flags off: no Sandboxes/Sessions/Integrations/Triggers in sidebar
- [ ] With flags on: all sections visible and navigable

PR #996, opened 2026-03-17
feat: add KAGENTI_FEATURE_FLAG_{SANDBOX,INTEGRATIONS,TRIGGERS}

Ladas:
## Summary

Single env var propagation chain gates all experimental features:

```
ocp_values.yaml → Helm values → backend env var → /api/v1/config/features → UI
```

**9 files**, one atomic change:

| Layer | Files | What it does |
|-------|-------|-------------|
| Helm | values.yaml, ocp_values.yaml, ui.yaml | Feature flag defaults (false), OCP overrides (true), env var injection, RBAC gating |
| Backend | config.py, main.py, config.py (router) | Pydantic settings, conditional router import with try/except, typed /features endpoint |
| UI | useFeatureFlags.ts, App.tsx, AppLayout.tsx | Fetch flags, gate routes and nav items |

When disabled (default): routes return 404, UI sections hidden, no DB tables, zero runtime cost.
When enabled: everything lights up.

Replaces: #976, #977, #978 (will close those)

## Test plan
- [ ] Deploy with defaults — no sandbox UI/routes visible
- [ ] Set `KAGENTI_FEATURE_FLAG_SANDBOX=true` — sandbox sections appear
- [ ] `/api/v1/config/features` returns correct flag state
- [ ] RBAC rules only rendered when flags enabled (`helm template`)

### DIFF
--- a/kagenti/backend/app/routers/config.py
+++ b/kagenti/backend/app/routers/config.py
@@ -6,12 +6,30 @@ Configuration API endpoints.
 """
 
 from fastapi import APIRouter, Depends
+from pydantic import BaseModel
 
 from app.core.auth import require_roles, ROLE_VIEWER
 from app.core.config import settings
 from app.models.responses import DashboardConfigResponse
 
+
+class FeatureFlagsResponse(BaseModel):
+    sandbox: bool
+    integrations: bool
+    triggers: bool
+
+
 router = APIRouter(prefix="/config", tags=["config"])
+
+
+@router.get("/features", response_model=FeatureFlagsResponse)
+async def get_feature_flags() -> FeatureFlagsResponse:
+    """Return enabled feature flags for UI gating. Public endpoint — flags are not sensitive."""
+    return FeatureFlagsResponse(
+        sandbox=settings.kagenti_feature_flag_sandbox,
+        integrations=settings.kagenti_feature_flag_integrations,
+        triggers=settings.kagenti_feature_flag_triggers,
+    )
 
 
 @router.get(
--- a/kagenti/backend/app/routers/config.py
+++ b/kagenti/backend/app/routers/config.py
@@ -22,11 +22,13 @@ class FeatureFlagsResponse(BaseModel):
 router = APIRouter(prefix="/config", tags=["config"])
 
 
-# No auth required: the UI needs feature flags before login to decide which
-# nav items and routes to render. Flags are boolean on/off — no sensitive data.
-@router.get("/features", response_model=FeatureFlagsResponse)
+@router.get(
+    "/features",
+    response_model=FeatureFlagsResponse,
+    dependencies=[Depends(require_roles(ROLE_VIEWER))],
+)
 async def get_feature_flags() -> FeatureFlagsResponse:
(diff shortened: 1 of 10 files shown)


kagenti/backend/app/routers/config.py
```diff
@@ -6,14 +6,32 @@
...
 from app.models.responses import DashboardConfigResponse
 
+
+class FeatureFlagsResponse(BaseModel):
+    sandbox: bool
+    integrations: bool
+    triggers: bool
+
+
 router = APIRouter(prefix="/config", tags=["config"])
```
pdettori, line 24: **suggestion**: `/features` has no auth, unlike sibling `/dashboards` which requires `ROLE_VIEWER`. The docstring says flags aren't sensitive — fair enough, but an unauthenticated endpoint reveals which experimental features are enabled on the deployment. Consider adding `ROLE_VIEWER` for consistency, or add a brief comment explaining the deliberate choice to keep it public (e.g. for pre-auth UI loading).

kagenti/ui-v2/src/hooks/useFeatureFlags.ts
```diff
@@ -0,0 +1,42 @@
...
+        const validated: FeatureFlags = {
+          sandbox: data.sandbox === true,
+          integrations: data.integrations === true,
+          triggers: data.triggers === true,
+        };
+        cachedFlags = validated;
+        setFlags(validated);
+      })
+      .catch(() => {});
+    return () => controller.abort();
```
pdettori: **nit**: `.catch(() => {})` silently swallows all fetch errors including network failures. A `console.debug` would help debugging:
```ts
.catch((e) => { if (e.name !== 'AbortError') console.debug('Feature flags fetch failed:', e); });
```

kagenti/ui-v2/src/components/AppLayout.tsx
```diff
@@ -263,21 +265,21 @@ export const AppLayout: React.FC<AppLayoutProps> = ({ children }) => {
                       onClick={() => handleThemeChange('auto')}
                       description="Follow system preference"
                     >
-                      System default {mode === 'auto' && '✓'}
+                      System default {mode === 'auto' && '*'}
```
pdettori: **nit**: The `✓` → `*` change in the theme dropdown looks unrelated to feature flags. Was this intentional (font compatibility?) or accidental? Might be cleaner as a separate commit.

pdettori (approved):
**APPROVE** — Well-structured feature flag implementation across the full stack (Helm → backend → UI). Clean propagation chain, good RBAC gating, pragmatic try/except pattern for conditional module imports. Two minor suggestions below, neither blocking.

**Areas reviewed**: Helm/K8s, Python backend, React frontend, Security
**Commits**: 3/3 signed-off ✓ | **CI**: all passing ✓

kagenti/backend/app/routers/config.py
Ladas: you are correct, fixing this (we might need to tweak the pre-login page but it should work now)

PR #1001, opened 2026-03-17
fix(ui-v2): preserve route on refresh with Keycloak SSO

### LINKED ISSUE #1000: UI: Browser refresh on any page redirects to home instead of staying on current route
## Summary

When using the Kagenti UI, refreshing the browser (or opening a deep link) while on any page other than the home/dashboard root causes navigation to the **home page** instead of reloading the **current** page.

## Expected behavior

- Refreshing on e.g. `/agents`, `/settings`, or any nested route should reload that same route and show the same view (after auth, if applicable).

## Actual behavior

- After refresh, the user lands on the home page rather than the URL they were on.

## Steps to reproduce

1. Log in to the Kagenti UI.
2. Navigate to a non-home route (e.g. agents, tools, or any deep-linked screen).
3. Use the browser refresh (F5 / Cmd+R) or paste the URL in a new tab.
4. Observe landing on home instead of the current route.

## Context / likely causes (for implementers)

Typical causes for SPAs:

- **Server / ingress**: Missing `try_files` or equivalent so non-root paths don’t serve `index.html` and the app falls back to `/`.
- **Client router**: Base path or `BrowserRouter` basename mismatch with how the app is hosted.
- **Auth bootstrap**: Post-login or token-restore redirect always sending users to `/` regardless of `window.location`.

**Area**: `kagenti/ui-v2` and any nginx/Helm ingress config for the UI.

## Environment

Please add: UI version / deployment (Kind, OpenShift, etc.) and browser if known.

akram:
## Summary
Use the current URL path as OAuth redirect_uri when the backend supplies only the app root. check-sso with checkLoginIframe disabled otherwise round-trips through Keycloak and returns users to home.

## Related issue(s)
#1000 

Fixes #1000

### DIFF
--- a/kagenti/ui-v2/src/contexts/AuthContext.tsx
+++ b/kagenti/ui-v2/src/contexts/AuthContext.tsx
@@ -14,6 +14,8 @@ import Keycloak from 'keycloak-js';
 
 import { setTokenGetter } from '@/services/api';
 
+import { keycloakRedirectUri } from './keycloakRedirectUri';
+
 // API base URL for fetching auth config
 const API_BASE_URL = import.meta.env.VITE_API_BASE_URL || '/api/v1';
 
@@ -62,6 +64,7 @@ export const AuthProvider: React.FC<AuthProviderProps> = ({ children }) => {
 
   // Keep keycloak instance in a ref so it persists across renders
   const keycloakRef = useRef<Keycloak | null>(null);
+  const configuredRedirectUriRef = useRef<string | undefined>(undefined);
 
   // Extract user info from Keycloak token
   const extractUserInfo = useCallback(
@@ -157,11 +160,14 @@ export const AuthProvider: React.FC<AuthProviderProps> = ({ children }) => {
           });
         };
 
+        configuredRedirectUriRef.current = config.redirect_uri;
+        const redirectUri = keycloakRedirectUri(config.redirect_uri);
+
         console.log('Initializing Keycloak with config:', {
           url: config.keycloak_url,
           realm: config.realm,
           clientId: config.client_id,
-          redirectUri: config.redirect_uri,
+          redirectUri,
           currentUrl: window.location.href,
         });
 
@@ -172,8 +178,7 @@ export const AuthProvider: React.FC<AuthProviderProps> = ({ children }) => {
           pkceMethod: 'S256',
           enableLogging: true, // Enable Keycloak adapter logging
           flow: 'standard', // Use standard authorization code flow
-          // Use redirect_uri from config if provided
-          ...(config.redirect_uri && { redirectUri: config.redirect_uri }),
+          redirectUri,
         }).catch((initError) => {
           console.error('Keycloak init rejected with error:', initError);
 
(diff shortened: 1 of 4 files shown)


kagenti/ui-v2/src/contexts/keycloakRedirectUri.ts
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @param loc - optional override for tests; defaults to window.location
+ */
+export function keycloakRedirectUri(
+  configuredRedirect: string | undefined,
+  loc?: KeycloakRedirectLocation
+): string {
+  const origin =
+    loc?.origin ??
+    (typeof window !== 'undefined' ? window.location.origin : '');
```
pdettori, line 20: **suggestion**: `basePath` can never be `''` after `|| '/'` on the line above — the `basePath === ''` check here is dead code. Consider simplifying to just `basePath === '/'`.

```suggestion
    if (basePath === '/') {
```

kagenti/ui-v2/src/contexts/keycloakRedirectUri.ts
```diff
@@ -0,0 +1,42 @@
+// Copyright 2025 IBM Corp.
+// Licensed under the Apache License, Version 2.0
+
+export type KeycloakRedirectLocation = Pick<Location, 'origin' | 'pathname'>;
+
+/**
+ * OAuth redirect_uri for Keycloak. The UI client registers root_url/*.
+ * Backend config often supplies only the app root; using that for every page
+ * makes check-sso (with iframe disabled) and login return to home. Use the
+ * current URL path so refresh and deep links preserve the route.
```
pdettori, line 10: **nit**: The `typeof window !== 'undefined'` guard is good for SSR safety, but this code only runs inside a React SPA mounted in the browser. Not a problem — just unnecessary defensive code.

kagenti/ui-v2/src/contexts/keycloakRedirectUri.ts
```diff
@@ -0,0 +1,42 @@
...
+ *
```
pdettori, line 11: **nit**: `!pathname || pathname === ''` — `pathname` from `window.location.pathname` is always at least `'/'`. The falsy check is redundant.

kagenti/ui-v2/src/contexts/keycloakRedirectUri.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  it('home path stays on home', () => {
+    expect(
+      keycloakRedirectUri('https://ui.example.com/', {
+        origin: 'https://ui.example.com',
+        pathname: '/',
+      })
+    ).toBe('https://ui.example.com/');
+  });
+});
```
pdettori, line 42: **praise**: Clean test suite with good edge case coverage — root with/without trailing slash, custom callback path, undefined config, and home-stays-on-home. Nice work.

kagenti/ui-v2/src/contexts/AuthContext.tsx
```diff
@@ -62,6 +64,7 @@ export const AuthProvider: React.FC<AuthProviderProps> = ({ children }) => {
 
   // Keep keycloak instance in a ref so it persists across renders
   const keycloakRef = useRef<Keycloak | null>(null);
+  const configuredRedirectUriRef = useRef<string | undefined>(undefined);
```
pdettori, line 67: **praise**: Good use of a ref to capture the backend config once and reuse it in both `init` and `login`, avoiding stale closures.

PR #1003, opened 2026-03-17
feat: update streamer skill and signing scripts

Ladas:
## Summary
- Update `github:streamer` skill with full workflow from sandbox agent streaming
- Update `sign_all_commits_in_a_branch.sh`: use `--no-gpg-sign` (DCO sign-off only)
- Update `sign_all_pr_worktrees.sh`: require explicit worktree names as args

3 files

### DIFF
--- a/.claude/skills/github:streamer/SKILL.md
+++ b/.claude/skills/github:streamer/SKILL.md
@@ -1,0 +1,344 @@
+---
+name: github:streamer
+description: Split large feature branches into chains of smaller, independently mergeable PRs — supports cross-repo work with progress tracking
+---
+
+```mermaid
+flowchart TD
+    START(["/github:streamer branch-or-PR"]) --> MULTI{"Multi-repo?"}
+    MULTI -->|Yes| WORKTREES["Create stream worktrees"]:::git
+    MULTI -->|No| ANALYZE
+
+    WORKTREES --> ANALYZE["Squash + rebase onto upstream/main"]:::git
+    ANALYZE --> CATEGORIZE["Categorize files into PR buckets"]:::github
+    CATEGORIZE --> FLAGS["Add feature flags if needed"]:::github
+    FLAGS --> PLAN["Plan merge order + dependencies"]:::github
+    PLAN --> PROGRESS["Write tracking docs"]:::doc
+
+    PROGRESS --> PRESENT["Present plan to user"]
+    PRESENT --> APPROVE{"User approves?"}
+    APPROVE -->|Changes| PLAN
+    APPROVE -->|Yes| CREATE["Create PR worktrees + branches"]:::git
+
+    CREATE --> POPULATE["Populate: one commit per file"]:::git
+    POPULATE --> PUSH["Push + create draft PRs"]:::github
+    PUSH --> REVIEW["Security + code review"]:::github
+    REVIEW --> FIX["Fix findings → commit → cherry-pick to stream"]:::git
+    FIX --> SIGN["DCO sign + push"]:::git
+    SIGN --> CI["Monitor CI, fix failures"]:::github
+    CI --> E2E["Run E2E on test cluster"]:::github
+
+    E2E --> MERGE["Merge in dependency order"]:::github
+    MERGE --> REBASE["Rebase dependent PRs"]:::git
+    REBASE --> MERGE
+
+    classDef github fill:#E91E63,stroke:#333,color:white
+    classDef git fill:#FF9800,stroke:#333,color:white
+    classDef doc fill:#4CAF50,stroke:#333,color:white
+```
+
+> Follow this diagram as the workflow.
+
+# GitHub Streamer
+
+Split a large feature branch (or PR) into a chain of smaller, independently mergeable PRs. Supports cross-repo work spanning multiple worktrees.
+
+## When to Use
+
+- Feature branch exceeds 500-700 lines (too large for effective review)
+- Branch touches multiple independent concerns that can be separated
+- Work spans multiple repos that need coordinated PRs
+- Reviewer feedback asks to split the PR
+
+## Phase 1: Create Stream Worktrees
+
+For each repo involved, create a stream worktree **from the source branch HEAD**, then squash and rebase:
+
+```bash
+# Create worktree from source HEAD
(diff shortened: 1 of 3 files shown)


scripts/sign_all_commits_in_a_branch.sh
```diff
@@ -18,8 +19,17 @@ YELLOW='\033[1;33m'
...
+for arg in "$@"; do
+    case "$arg" in
+        --no-gpg) GPG_SIGN="--no-gpg-sign" ;;
+        --yes|-y) AUTO_YES=true ;;
+        *) UPSTREAM_REF="$arg" ;;
+    esac
+done
 
 # Verify the upstream ref exists
 if ! git rev-parse --verify "$UPSTREAM_REF" >/dev/null 2>&1; then
```
rubambiza, line 50: **must-fix:** The argument parser silently overwrites `UPSTREAM_REF` with each unrecognized positional arg — `./script foo bar` uses `bar` and ignores `foo`. Should error on unexpected arguments or validate that at most one positional arg is provided.

Also the `--yes` flag is not documented in the header comment (lines 6-10). Header mentions `[--no-gpg]` but not `[--yes]`.

scripts/sign_all_pr_worktrees.sh
```diff
@@ -0,0 +1,85 @@
...
+            failed=$((failed + 1))
+        fi
+    else
+        echo "WARN: signing failed for $wt"
+        failed=$((failed + 1))
+    fi
+done
+
+echo ""
+echo "=== Done: $signed signed+pushed, $skipped skipped, $failed failed ==="
```
rubambiza, line 89: **suggestion:** Force push hardcodes `upstream` remote. Project convention (CLAUDE.md) says push to fork remotes, not origin/upstream. If the streaming workflow intentionally pushes to the org repo, this should be explicitly called out as an exception — perhaps with a comment explaining why.

scripts/sign_all_commits_in_a_branch.sh
```diff
@@ -84,7 +98,7 @@ COMMIT_COUNT=$(git rev-list --count "$UPSTREAM_REF"..HEAD)
 echo ""
```
rubambiza, line 113: **suggestion:** `--no-verify` skips commit hooks, which CLAUDE.md says to avoid. Legitimate here since this is amending existing commits (not creating new content), but a comment explaining the rationale would prevent future reviewers from flagging it.

.claude/skills/github:streamer/SKILL.md
```diff
@@ -0,0 +1,336 @@
...
+
+### Key Rules
+- **NEVER touch source worktrees** — they're active development
+- Use `git -c core.hooksPath=/dev/null commit` to skip pre-commit during streaming
+- Record source HEAD hash in tracking doc for future syncing
+
+## Phase 2: Feature Flags (if needed)
+
+For large features that should be dormant until fully merged:
+
```
rubambiza, line 85: **suggestion:** `git -c core.hooksPath=/dev/null commit` bypasses all hooks including DCO. This pattern appears in several places (lines 85, 123, 159, 166). Consider adding a stronger warning that this is strictly for intermediate streaming commits and that final commits must go through normal hook validation.

.claude/skills/github:streamer/SKILL.md
```diff
@@ -0,0 +1,336 @@
...
+# Resolve conflicts, then: git rebase --continue
+
+# Push as backup
+git push upstream feat/stream1-<name>
+```
+
+The stream worktree is the **integration branch** — all fixes get applied here first, then cherry-picked to individual PRs.
+
+### Key Rules
+- **NEVER touch source worktrees** — they're active development
```
rubambiza, line 78: **suggestion:** `git push upstream` is used throughout the skill (lines 78, 131, 160, 167, 259). Same fork-vs-upstream concern as the wrapper script — if the streaming workflow intentionally uses upstream branches, note this as an explicit exception to the normal fork-based workflow documented in CLAUDE.md.

scripts/sign_all_commits_in_a_branch.sh
```diff
@@ -1,12 +1,13 @@
 #!/usr/bin/env bash
 #
 # Sign all commits in current branch that are ahead of upstream/main.
-# This adds both sign-off (-s) and GPG signature (-S) to each commit,
+# This adds sign-off (-s) and GPG signature (-S) to each commit,
 # and replaces any Co-Authored-By trailers with Assisted-By.
 #
```
rubambiza, line 6: **nit:** Header comment still says "adds sign-off (-s) and GPG signature (-S)" but GPG is now optional via `--no-gpg`. Update to reflect the new behavior.

scripts/sign_all_pr_worktrees.sh
```diff
@@ -0,0 +1,85 @@
...
+signed=0
+skipped=0
+failed=0
+
+for wt in "$@"; do
+    wt_path="$REPO_ROOT/.worktrees/$wt"
+    if [ ! -d "$wt_path" ]; then
+        echo "SKIP $wt — worktree not found at $wt_path"
+        skipped=$((skipped + 1))
+        continue
```
rubambiza, line 57: **nit:** Parsing `ls` output programmatically can break on filenames with spaces/newlines. Consider using a glob pattern instead:
```bash
for d in "$REPO_ROOT/.worktrees"/{pr-*,stream*}; do
  [ -d "$d" ] && echo "  $(basename "$d")"
done
```

scripts/sign_all_pr_worktrees.sh
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env bash
```
rubambiza, line 1: **praise:** Good use of `set -euo pipefail`, `--force-with-lease` for safe force-push, and upfront validation that the sign script exists and is executable.

rubambiza (commented):
The PR adds useful automation for the streaming workflow — the skill doc is thorough and the signing scripts are well-structured. Main concerns are around argument validation and clarifying the push-to-upstream convention.

**Areas reviewed**: Shell scripts, Skill markdown, Security
**CI status**: passing

rubambiza (approved):
Thanks for addressing the comments. LGTM.

PR #1008, opened 2026-03-17
cleanup: remove legacy environments ConfigMap

huang195:
## Summary

- Remove the legacy `environments` ConfigMap from agent namespaces — it was created for the old Streamlit UI's "Select Environment Variable Sets" feature which no longer exists
- Nothing reads this ConfigMap at runtime: ui-v2 uses `.env` file imports, TUI hardcodes presets via `LLMPresetEnvVars()`
- Move `SPIRE_ENABLED` to the existing `authbridge-config` ConfigMap
- Remove `update_environments_configmaps()` from agent-oauth-secret job (wrote Keycloak keys already available in `authbridge-config` and `keycloak-admin-secret`)
- Tighten RBAC: remove `configmaps` from ClusterRole (job only needs secrets now)
- Update docs to reference `.env` file workflow instead of the removed ConfigMap

## Test plan

- [ ] `helm template` — verify no `environments` ConfigMap rendered, `authbridge-config` has `SPIRE_ENABLED`
- [ ] Deploy to Kind cluster
- [ ] Deploy an agent via ui-v2 with `.env.openai` import — verify env vars injected correctly
- [ ] Deploy an agent via TUI with `openai` and `ollama` presets — verify env vars injected correctly
- [ ] Run E2E tests: `uv run pytest kagenti/tests/e2e/ -v`
- [ ] Verify agent-oauth-secret job completes without errors

### DIFF
--- a/charts/kagenti/templates/agent-namespaces.yaml
+++ b/charts/kagenti/templates/agent-namespaces.yaml
@@ -102,103 +102,6 @@ type: kubernetes.io/dockerconfigjson
 data:
   .dockerconfigjson: {{ (printf "{\"auths\":{\"quay.io\":{\"username\":\"%s\",\"password\":\"%s\",\"auth\":\"%s\"}}}" $.Values.secrets.quayUser $.Values.secrets.quayToken (printf "%s:%s" $.Values.secrets.quayUser $.Values.secrets.quayToken | b64enc)) | b64enc }}
 ---
-# ——————————————————————————————————————————————————————————————————————————————
-#  ConfigMap for Environment Variables in Namespace: {{ . }}
-# ——————————————————————————————————————————————————————————————————————————————
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: environments
-  namespace: {{ . }}
-  labels:
-    {{- include "kagenti.labels" $root | nindent 4 }}
-data:
-  ollama: |
-    [
-      {"name": "LLM_API_BASE", "value": "http://host.docker.internal:11434/v1"},
-      {"name": "LLM_API_KEY", "value": "dummy"},
-      {"name": "LLM_MODEL", "value": "llama3.2:3b-instruct-fp16"}
-    ]
-  openai: |
-    [
-      {
-        "name": "OPENAI_API_KEY",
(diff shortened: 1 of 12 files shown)


charts/kagenti/templates/agent-namespaces.yaml
```diff
@@ -237,6 +140,7 @@ data:
   # Set explicitly when the internal KEYCLOAK_URL differs from the frontend URL
   # that appears in token "iss" claims (split-horizon DNS).
   ISSUER: "{{ $root.Values.keycloak.publicUrl }}/realms/{{ $root.Values.keycloak.realm }}"
+  SPIRE_ENABLED: {{ if $root.Values.spire.enabled }}"true"{{ else }}"false"{{ end }}
```
cwiklik, line 143: **Cross-repo breaking change**: The `kagenti-extensions` webhook reads `KEYCLOAK_URL`, `KEYCLOAK_REALM`, `SPIRE_ENABLED`, and `PLATFORM_CLIENT_IDS` from the `environments` ConfigMap via `ValueFrom: ConfigMapKeyRef` in the legacy injection path (see `container_builder.go` lines 267, 277, 309, 426, 436). The resolved path also reads these from the `environments` CM at admission time (`namespace_config.go` line 72).

After this PR merges, injected pods will have unresolvable env var references and fail to start.

**Suggested coordination**: Either:
1. Update `kagenti-extensions` PR #217 to read these keys from `authbridge-config` instead of `environments` (legacy path), and update `namespace_config.go` to read from `authbridge-config` (resolved path). Merge extensions first.
2. Or merge this PR first and immediately follow up with the extensions fix.

Also: `SPIRE_ENABLED` is correctly moved to `authbridge-config` here, but `PLATFORM_CLIENT_IDS` doesn't appear in any ConfigMap after this PR. Is that key still needed? If so, it should be added to `authbridge-config`.

cwiklik: @huang195 I will address the webhook problem as part of PR https://github.com/kagenti/kagenti-extensions/pull/217.
So before merging your PR I need to address PR 217 in kagenti-extensions.

PR #1012, opened 2026-03-17
fix: :lock: Avoid parsing comment body via shell

evaline-ju:
## Summary

Comment body was being parsed directly in shell script, making this vulnerable to shell injections via `/run-e2e` commands

Fixes https://github.com/kagenti/kagenti/security/code-scanning/188

### DIFF
--- a/.github/workflows/e2e-hypershift-pr.yaml
+++ b/.github/workflows/e2e-hypershift-pr.yaml
@@ -109,6 +109,10 @@ jobs:
 
             if (match) {
               const extensionsRef = match[1];
+              if (!/^[\w.\-\/]+$/.test(extensionsRef)) {
+                core.setFailed(`Invalid extensions-ref: ${extensionsRef}`);
+                return;
+              }
               console.log('Parsed extensions_ref:', extensionsRef);
               core.setOutput('extensions_ref', extensionsRef);
             } else {
--- a/.github/workflows/e2e-hypershift-pr.yaml
+++ b/.github/workflows/e2e-hypershift-pr.yaml
@@ -97,17 +97,24 @@ jobs:
       - name: Parse command options
         id: parse-options
         if: steps.check-permission.outputs.check-result == 'true'
-        shell: bash
-        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-hypershift-pr.yaml
```diff
@@ -97,17 +97,24 @@ jobs:
...
-          fi
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd  # v8
+        with:
+          script: |
+            const commentBody = context.payload.comment.body;
+            console.log('Parsing comment:', commentBody);
+
+            // Parse --extensions-ref <branch> from comment
+            // Usage: /run-e2e --extensions-ref fix/proxy-init-drop-privileged
+            const match = commentBody.match(/--extensions-ref\s+(\S+)/);
```
pdettori, line 108: **suggestion**: Consider sanitizing `extensionsRef` to only allow valid git ref characters before passing it to `setOutput`. While the shell injection is eliminated, the parsed value is later used as a git ref — restricting it to `[a-zA-Z0-9._/-]` would add defense-in-depth:

```js
const extensionsRef = match[1];
if (!/^[\w.\-\/]+$/.test(extensionsRef)) {
  core.setFailed(`Invalid extensions-ref: ${extensionsRef}`);
  return;
}
```

pdettori (approved):
Clean, focused security fix that eliminates a shell injection vulnerability in the E2E workflow. The approach — replacing bash parsing with JavaScript via `actions/github-script` — is the recommended GitHub Actions pattern. SHA pinning is consistent with the rest of the file, and the regex produces functionally equivalent results.

**Areas reviewed**: CI/GitHub Actions, Security
**CI status**: All pass except E2E HyperShift (the workflow being modified — may be unrelated)

**Verdict**: APPROVE with one optional suggestion

.github/workflows/e2e-hypershift-pr.yaml
evaline-ju: thanks! I have incorporated this

Issue #975, opened 2026-03-16
Make in-cluster registry configurable or auto-detect OpenShift internal registry


akram:
## Summary

On OpenShift, the built-in image registry is `image-registry.openshift-image-registry.svc:5000`. The "Local Registry (In-Cluster)" option in the UI and the backend default both use the Kind registry URL (`registry.cr-system.svc.cluster.local:5000`). There is no way to configure or auto-detect the in-cluster registry when installing on OpenShift, so Import Agent / Import Tool flows that use "Local Registry" point at the wrong registry on OCP.

## Problems

- **UI:** "Local Registry (In-Cluster)" is hardcoded to `registry.cr-system.svc.cluster.local:5000` in `ImportAgentPage.tsx` and `ImportToolPage.tsx` (`REGISTRY_OPTIONS`).
- **Backend:** Default when no `registryUrl` is provided is hardcoded in `kagenti/backend/app/core/constants.py` as `DEFAULT_INTERNAL_REGISTRY`.
- **Install/Helm:** No Ansible or Helm value exists to set the in-cluster registry URL; the backend deployment does not receive any registry-related env var.

pdettori:
This has been fixed since v0.6.0

PR #1034, opened 2026-03-18
feat: internal registry via kagenti-ui-config and dashboards API

akram:
## Summary
- Helm: optional `ui.internalContainerImageRegistryUrl`; CM key `INTERNAL_CONTAINER_IMAGE_REGISTRY_URL` (Kind static + OCP ui-routes-job); backend Deployment optional env from CM.
- Backend: `internal_container_image_registry_url` setting; Shipwright default registry + internal-registry check; `internalRegistryUrl` on `GET /api/v1/config/dashboards`.
- UI: Import agent/tool pages use dashboards config with Kind fallback.
- Docs/tests: install, new-tool, backend README, `test_dashboard_config`.

## Context
Closes / relates to internal registry correctness for OpenShift vs Kind (issue #975).

### DIFF
--- a/kagenti/backend/app/core/config.py
+++ b/kagenti/backend/app/core/config.py
@@ -9,8 +9,10 @@ import re
 from functools import lru_cache
 from typing import List, Optional
 
-from pydantic import model_validator
+from pydantic import field_validator, model_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
+
+_KIND_DEFAULT_INTERNAL_REGISTRY = "registry.cr-system.svc.cluster.local:5000"
 
 
 class Settings(BaseSettings):
@@ -38,6 +40,12 @@ class Settings(BaseSettings):
         "http://localhost:3000",
         "http://localhost:8080",
     ]
+
+    @field_validator("internal_container_image_registry_url", mode="after")
+    @classmethod
+    def _internal_registry_non_empty(cls, v: str) -> str:
+        s = (v or "").strip()
+        return s if s else _KIND_DEFAULT_INTERNAL_REGISTRY
 
     @model_validator(mode="after")
     def _add_domain_cors_origin(self) -> "Settings":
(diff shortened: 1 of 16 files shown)


kagenti/backend/app/core/config.py
```diff
@@ -9,9 +9,11 @@
 from functools import lru_cache
 from typing import List, Optional
 
-from pydantic import model_validator
+from pydantic import field_validator, model_validator
```
pdettori, line 12: **suggestion**: `_KIND_DEFAULT_INTERNAL_REGISTRY` duplicates `DEFAULT_INTERNAL_REGISTRY` from `app.core.constants` (same value). Consider importing from constants to keep a single source of truth:
```python
from app.core.constants import DEFAULT_INTERNAL_REGISTRY as _KIND_DEFAULT_INTERNAL_REGISTRY
```

kagenti/backend/app/services/shipwright.py
```diff
@@ -76,8 +77,10 @@ def select_build_strategy(registry_url: str, requested_strategy: Optional[str] =
     Returns:
         The build strategy name to use
     """
+    reg = registry_url.rstrip("/")
+    configured = settings.internal_container_image_registry_url.rstrip("/")
```
pdettori, line 81: **nit**: After this PR, `reg == DEFAULT_INTERNAL_REGISTRY` is technically redundant — `configured` defaults to the same value via `settings.internal_container_image_registry_url`. Keeping it is harmless (defensive), just noting it.

kagenti/ui-v2/src/pages/ImportAgentPage.tsx
```diff
@@ -99,6 +100,20 @@ interface ServicePort {
...
+    queryKey: ['dashboards'],
+    queryFn: () => configService.getDashboards(),
+  });
+  const localRegistryUrl =
+    dashboardConfig?.internalRegistryUrl?.trim() || KIND_DEFAULT_INTERNAL_REGISTRY;
+  const registryOptions = useMemo(
+    () => [
+      { value: 'local', label: 'Local Registry (In-Cluster)', url: localRegistryUrl },
+      ...EXTERNAL_REGISTRY_OPTIONS,
+    ],
```
pdettori, line 113: **suggestion**: This `useQuery(['dashboards'])` + `useMemo(registryOptions)` block is identical in both `ImportAgentPage` and `ImportToolPage`. Consider extracting a shared hook (e.g. `useRegistryOptions()`) to reduce duplication.

Also: the `['dashboards']` query has no `staleTime`, so it refetches on every mount/window-focus. Since dashboard config is stable, adding `staleTime: Infinity` (or a long TTL) would avoid unnecessary requests.

Issue #783, opened 2026-03-02
feature: Integrate AgentCard signing and verification into the Kagenti UI


kevincogan:
### Feature Description

The AgentCard operator supports automated signing, signature verification, identity binding, and network enforcement for deployed agents. However, all of this is CLI-driven today. Deploying a signed agent requires manually configuring init-containers, SPIRE volumes, ClusterSPIFFEIDs, and environment variables via kubectl. There is no way to enable signing through the UI, and no way to see verification status without running kubectl commands. This makes the feature invisible to users who interact with the platform through the dashboard.

### Proposed Solution

Integrate AgentCard signing and verification into the Kagenti UI across three areas:

**1. Signed deployment at import time**

Add an "Enable AgentCard signing" toggle to the agent import flow (visible when SPIRE is available on the cluster). When enabled, the backend injects the required resources into the Deployment:
- `agentcard-signer` init-container with SPIRE CSI volume
- Unsigned card ConfigMap (auto-generated from agent metadata)
- Shared `emptyDir` volume between init-container and main container
- ClusterSPIFFEID for the agent's ServiceAccount

The UX should be a simple toggle, the user shouldn't need to know about init-containers or SPIFFE. This follows the same pattern as the existing `spireEnabled` toggle for AuthBridge.

**2. Verification status on agent detail and list pages**

Surface AgentCard status in the UI so users can see trust state without kubectl:
- **Agent detail page**: Show SignatureVerified (True/False), Bound (True/False), Synced (True/False), SPIFFE ID, trust domain match, card metadata (name, version, capabilities, skills), NetworkPolicy state (permissive vs. restrictive)
- **Agent list/catalog**: Add Verified and Bound columns or badges, mirroring what `kubectl get agentcards -A` shows today
- **Data source**: Query the AgentCard CR via the Kubernetes API (`agentcards.agent.kagenti.dev`)

**3. Identity binding configuration**

Allow users to configure `spec.identityBinding.trustDomain` per agent from the UI, and show the binding evaluation result (Bound/NotBound with reason and SPIFFE ID).

For reference on what the backend needs to inject, see `kagenti-operator/demos/agentcard-spire-signing/k8s/agent-deployment.yaml` for the full signed agent manifest.

### Want to contribute?

- [x] I would like to work on this issue.

### Additional Context

- The signing and verification implementation is merged in:
    - [kagenti-operator PR #185](https://github.com/kagenti/kagenti-operator/pull/185)
[...]

Issue #612, opened 2026-02-05
epic: Agent Attestation Framework for Cloud-Native AI Trust


mrsabath:
### Problem

As AI agents become more prevalent in enterprise environments, organizations face a critical trust gap: **authentication alone does not guarantee an agent is trustworthy**.

Current authentication mechanisms (SPIFFE/SPIRE, OAuth2) answer: *"Who is this agent?"*

But they do NOT answer:
- Was this agent built by a trusted CI pipeline?
- Has the agent's code or configuration been modified since deployment?
- Are the agent's declared capabilities actually authorized?
- Does the agent comply with organizational security policies?
- Has a third party vouched for this agent's trustworthiness?

**Without attestation, a compromised agent with valid credentials can communicate freely with other services, access sensitive data, and perform unauthorized actions.**

This epic introduces an **attestation layer** that provides cryptographic guarantees about agent integrity, provenance, and trustworthiness—complementing the existing AuthBridge authentication infrastructure.

## Background: Authentication vs. Attestation

| Dimension | Authentication (Existing) | Attestation (Proposed) |
|-----------|--------------------------|------------------------|
| **Question** | "Who are you?" | "What are you? Can I trust you?" |
| **Mechanism** | SPIFFE SVIDs, OAuth tokens | Signatures, hashes, provenance |
| **Timing** | Every request (runtime) | Deployment + periodic |
| **Failure** | "Access denied" | "Deployment blocked" |

### design

See [the design document](https://github.com/mrsabath/kagenti/blob/agent-attestation-proposal/docs/plans/agent-attestation-proposal.md) for Agent Attestation. 

## Architectural Principles

- **Attestation gates authentication:** Verify trust before identity is issued.
- **Non-blocking adoption:** Warning mode before enforcement; gradual rollout.
- **Layered trust:** Multiple independent attestations provide defense in depth.
- **Standards-based:** Use Sigstore, SLSA, SPIFFE—no proprietary formats.
- **Kubernetes-native:** Leverage admission webhooks and existing Kagenti patterns.

### Goals and non-goals

Extend Kagenti with evidence-based **trust verification** that gates agent deployment and runtime behavior, implementing the [7 Categories of A2A Trust Signals](https://google.github.io/A2A/):

| # | Signal | Type | Status |
|---|--------|------|--------|
| 1 | Identity | Authentication | ✅ Implemented (SPIFFE) |
| 2 | Capability | Attestation | 🔲 Phase 1 |
[...]

PR #1038, opened 2026-03-18
✨ AgentCard signing and verification UI integration

kevincogan:
## Summary

Integrates AgentCard SPIRE signing and verification into the Kagenti UI, enabling end-to-end agent deployment with cryptographic identity from the browser.

### Backend
- Add configurable AgentCard signing settings (signer image, timeout, resource limits, trust domain) via env var overrides
- Add `signingEnabled` flag to agent creation; on deploy, create signing resources (SA, ConfigMap, Role, RoleBinding) and inject the `sign-agentcard` init container
- Add `/agentcard` status endpoint and enrich agent list with verification status (verified, bound)
- Clean up all signing resources on agent deletion with 404 tolerance
- Fix in-cluster chat fallback to use correct port (8000), add K8s API direct service lookup, handle A2A SDK artifact response format
- Add K8s service methods for ConfigMaps, ServiceAccounts, Roles, RoleBindings, and custom resource patching

### UI
- Add Trust column to agent catalog with Trusted, Pending, Unverified, and Not Bound badges
- Add Trust & Identity card to agent detail page with signing progress indicator, verification status, identity details, and expandable conditions table
- Add AgentCard signing checkbox to import page with SPIRE auto-enable and prerequisites alert

### Helm chart
- Add `SPIRE_TRUST_DOMAIN` and `AGENTCARD_SIGNER_IMAGE` env vars to backend container
- Expand ClusterRole with permissions for agentcards, configmaps, serviceaccounts, roles, and rolebindings needed for signing lifecycle
- Add `spire.trustDomain` and `agentcard.signerImage` to values.yaml

### Infrastructure
- Make nginx DNS resolver configurable via `DNS_RESOLVER` env var with envsubst at startup
- Adjust Dockerfile permissions for OpenShift arbitrary UIDs

## Related issue(s)

- #783 (Integrate AgentCard signing and verification into the Kagenti UI)
- kagenti/kagenti-operator#234 (Publish agentcard-signer image to GHCR via CI)
- kagenti/kagenti-operator#224 (Per-agent egress scoping for verified agent NetworkPolicies)

Verified end-to-end on OpenShift (OVN-Kubernetes + Istio Ambient Mesh):

1. Deploy currency-agent via UI with "Enable AgentCard signing" checked
2. Confirm agent reaches Trusted status (synced, signature verified, identity bound) on the agent catalog and detail pages
3. Confirm chat works (agent responds to currency exchange queries)
4. Delete agent via UI and confirm all signing resources are cleaned up

All changes are backward-compatible. Signing is opt-in via the `signingEnabled` flag (defaults to false).

## Reference UI Images

kagenti-ui-5

kagnti-ui-4

kagenti-ui

kagenti-ui-3

### DIFF
--- a/charts/kagenti/values.yaml
+++ b/charts/kagenti/values.yaml
@@ -70,6 +70,9 @@ ui:
   frontend:
     image: ghcr.io/kagenti/kagenti/ui-v2
     tag: latest
+    # DNS resolver IP for nginx upstream resolution.
+    # OpenShift: 172.30.0.10 (default), Kind/vanilla K8s: 10.96.0.10
+    dnsResolver: "172.30.0.10"
     resources:
       limits:
         cpu: 100m
--- a/charts/kagenti/values.yaml
+++ b/charts/kagenti/values.yaml
@@ -8,6 +8,7 @@ featureFlags:
   sandbox: false
   integrations: false
   triggers: false
+  agentcardSigning: false
 
 # ------------------------------------------------------------------
 #  Control Panel: Enable or disable components here
@@ -281,4 +282,4 @@ spire:
 agentcard:
   # Init-container image used to sign agent cards with SPIRE SVIDs.
   # Pin to a release tag once CI publishes to GHCR (kagenti/kagenti-operator#234).
-  signerImage: "ghcr.io/kagenti/agentcard-signer:latest"
+  signerImage: "ghcr.io/kagenti/kagenti-operator/agentcard-signer:latest"
--- a/charts/kagenti/values.yaml
+++ b/charts/kagenti/values.yaml
@@ -269,3 +269,13 @@ mlflowOAuthSecret:
 # ------------------------------------------------------------------
 spire:
   enabled: true
+  # SPIRE trust domain (required for AgentCard identity binding)
+  trustDomain: ""
+
+# ------------------------------------------------------------------
+#  AgentCard configs
+# ------------------------------------------------------------------
+agentcard:
+  # Init-container image used to sign agent cards with SPIRE SVIDs.
+  # Pin to a release tag once CI publishes to GHCR (kagenti/kagenti-operator#234).
+  signerImage: "ghcr.io/kagenti/agentcard-signer:latest"
--- a/kagenti/backend/app/core/config.py
+++ b/kagenti/backend/app/core/config.py
@@ -52,6 +52,16 @@ class Settings(BaseSettings):
     crd_version: str = "v1alpha1"
     agents_plural: str = "agents"
 
+    # AgentCard signing settings
+    # Pin to a release tag once CI publishes to GHCR (kagenti/kagenti-operator#234)
+    agentcard_signer_image: str = "ghcr.io/kagenti/agentcard-signer:latest"
+    agentcard_sign_timeout: str = "30s"
+    agentcard_signer_cpu_request: str = "10m"
+    agentcard_signer_cpu_limit: str = "100m"
+    agentcard_signer_mem_request: str = "32Mi"
+    agentcard_signer_mem_limit: str = "64Mi"
+    spire_trust_domain: str = ""
+
     # Shipwright build settings
     shipwright_default_strategy: str = "buildah-insecure-push"  # Default for dev
     shipwright_default_timeout: str = "15m"
--- a/kagenti/backend/app/core/config.py
+++ b/kagenti/backend/app/core/config.py
@@ -54,7 +54,7 @@ class Settings(BaseSettings):
 
     # AgentCard signing settings
     # Pin to a release tag once CI publishes to GHCR (kagenti/kagenti-operator#234)
-    agentcard_signer_image: str = "ghcr.io/kagenti/agentcard-signer:latest"
+    agentcard_signer_image: str = "ghcr.io/kagenti/kagenti-operator/agentcard-signer:latest"
     agentcard_sign_timeout: str = "30s"
     agentcard_signer_cpu_request: str = "10m"
     agentcard_signer_cpu_limit: str = "100m"
@@ -79,6 +79,7 @@ class Settings(BaseSettings):
     kagenti_feature_flag_sandbox: bool = False
     kagenti_feature_flag_integrations: bool = False
     kagenti_feature_flag_triggers: bool = False
+    kagenti_feature_flag_agentcard_signing: bool = False
 
     # Label settings
     kagenti_label_prefix: str = "kagenti.io/"
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -2245,87 +2245,144 @@ def _build_selector_labels(request: "CreateAgentRequest") -> Dict[str, str]:
     }
 
 
+def _cleanup_signing_resource(
+    kube: KubernetesService, namespace: str, kind: str, resource_name: str
+) -> None:
+    """Best-effort cleanup of a single signing resource (ignores 404)."""
+    try:
+        if kind == "ServiceAccount":
+            kube.delete_service_account(namespace=namespace, name=resource_name)
+        elif kind == "ConfigMap":
+            kube.delete_configmap(namespace=namespace, name=resource_name)
+        elif kind == "Role":
+            kube.delete_role(namespace=namespace, name=resource_name)
+        elif kind == "RoleBinding":
+            kube.delete_role_binding(namespace=namespace, name=resource_name)
+        logger.info(f"Rolled back {kind} '{resource_name}' in {namespace}")
+    except ApiException as e:
+        if e.status != 404:
+            logger.warning(f"Failed to roll back {kind} '{resource_name}': {e.reason}")
+
+
 def _create_signing_resources(
     kube: KubernetesService,
     name: str,
     namespace: str,
     request: "CreateAgentRequest",
 ) -> None:
-    """Create ConfigMap and ServiceAccount required for AgentCard signing.
-    Must be called BEFORE workload creation."""
+    """Create ConfigMap, ServiceAccount, Role, and RoleBinding for AgentCard signing.
+
+    Must be called BEFORE workload creation. On failure, any partially-created
+    resources are cleaned up to avoid orphans.
+    """
     port = DEFAULT_IN_CLUSTER_PORT
     if request.servicePorts and len(request.servicePorts) > 0:
         port = request.servicePorts[0].targetPort
 
-    sa_body = {
-        "apiVersion": "v1",
-        "kind": "ServiceAccount",
-        "metadata": {"name": f"{name}-sa", "namespace": namespace},
-    }
-    try:
-        kube.create_service_account(namespace=namespace, body=sa_body)
-        logger.info(f"Created ServiceAccount '{name}-sa' in namespace '{namespace}'")
-    except ApiException as e:
-        if e.status == 409:
-            logger.info(f"ServiceAccount '{name}-sa' already exists, reusing")
-        else:
-            raise
+    created: list[tuple[str, str]] = []
 
-    card_json = json.dumps({
-        "name": name,
-        "description": f"Agent '{name}'",
-        "url": f"http://{name}.{namespace}.svc.cluster.local:{port}",
-        "version": "1.0.0",
-        "capabilities": {"streaming": False, "pushNotifications": False},
-        "defaultInputModes": ["text/plain"],
-        "defaultOutputModes": ["text/plain"],
-        "skills": [],
-    })
-    cm_body = {
-        "apiVersion": "v1",
-        "kind": "ConfigMap",
-        "metadata": {"name": f"{name}-card-unsigned", "namespace": namespace},
-        "data": {"agent.json": card_json},
-    }
     try:
-        kube.create_configmap(namespace=namespace, body=cm_body)
-        logger.info(f"Created ConfigMap '{name}-card-unsigned' in namespace '{namespace}'")
-    except ApiException as e:
-        if e.status == 409:
-            logger.info(f"ConfigMap '{name}-card-unsigned' already exists, reusing")
-        else:
-            raise
+        sa_body = {
+            "apiVersion": "v1",
+            "kind": "ServiceAccount",
+            "metadata": {"name": f"{name}-sa", "namespace": namespace},
+        }
+        try:
+            kube.create_service_account(namespace=namespace, body=sa_body)
+            logger.info(f"Created ServiceAccount '{name}-sa' in namespace '{namespace}'")
+            created.append(("ServiceAccount", f"{name}-sa"))
+        except ApiException as e:
+            if e.status == 409:
+                logger.info(f"ServiceAccount '{name}-sa' already exists, reusing")
+            else:
+                raise
 
-    role_body = {
-        "apiVersion": "rbac.authorization.k8s.io/v1",
-        "kind": "Role",
-        "metadata": {"name": f"{name}-card-writer", "namespace": namespace},
-        "rules": [{"apiGroups": [""], "resources": ["configmaps"], "verbs": ["get", "create", "update"]}],
-    }
-    try:
-        kube.create_role(namespace=namespace, body=role_body)
-        logger.info(f"Created Role '{name}-card-writer' in namespace '{namespace}'")
-    except ApiException as e:
-        if e.status == 409:
-            logger.info(f"Role '{name}-card-writer' already exists, reusing")
-        else:
-            raise
+        card_json = json.dumps(
+            {
+                "name": name,
+                "description": f"Agent '{name}'",
+                "url": f"http://{name}.{namespace}.svc.cluster.local:{port}",
+                "version": "1.0.0",
+                "capabilities": {"streaming": False, "pushNotifications": False},
+                "defaultInputModes": ["text/plain"],
+                "defaultOutputModes": ["text/plain"],
+                "skills": [],
+            }
+        )
+        cm_body = {
+            "apiVersion": "v1",
+            "kind": "ConfigMap",
+            "metadata": {"name": f"{name}-card-unsigned", "namespace": namespace},
+            "data": {"agent.json": card_json},
+        }
+        try:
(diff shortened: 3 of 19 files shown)


kagenti/backend/app/routers/agents.py
```diff
@@ -1943,6 +2147,251 @@ def _build_selector_labels(request: "CreateAgentRequest") -> Dict[str, str]:
...
+            kube.patch_custom_resource(
+                group=AGENTCARD_CRD_GROUP,
+                version=AGENTCARD_CRD_VERSION,
+                namespace=namespace,
+                plural=AGENTCARD_PLURAL,
+                name=card_name,
+                body=patch_body,
+            )
+            logger.info(f"Patched AgentCard '{card_name}' with identity binding (trustDomain: {SPIRE_TRUST_DOMAIN})")
+        except ApiException as e:
```
pdettori, line 2585: **must-fix**: `loop.run_in_executor(None, _do_patch)` returns a Future that is never awaited or stored. Any exception inside `_do_patch` (which sleeps up to 30s in a thread pool thread) is silently swallowed. Under load, sleeping threads accumulate without bound.

Suggestions:
- `await loop.run_in_executor(None, _do_patch)` so failures propagate, or
- Use `asyncio.create_task` wrapping the sync work with a done-callback that logs failures
- Replace `time.sleep` with `asyncio.sleep` inside a proper coroutine

Also: `asyncio.get_event_loop()` is deprecated since 3.10 — use `asyncio.get_running_loop()`.

kagenti/backend/app/routers/agents.py
```diff
@@ -1943,6 +2147,251 @@ def _build_selector_labels(request: "CreateAgentRequest") -> Dict[str, str]:
     }
```
pdettori, line 2245: **must-fix**: If Role or RoleBinding creation fails (non-409), the ServiceAccount and ConfigMap already created in earlier steps are not rolled back. This leaves orphaned resources that persist until the agent is explicitly deleted.

Consider wrapping the body in a try/except that cleans up any partially-created resources on non-409 failures.

kagenti/backend/app/routers/chat.py
```diff
@@ -60,6 +60,56 @@ class ChatResponse(BaseModel):
...
+
+    if not host or not os.path.exists(token_path):
+        return None
+    try:
+        with open(token_path) as f:
+            token = f.read()
+        ctx = ssl.create_default_context(cafile=ca_path)
+        url = f"https://{host}:{port}/api/v1/namespaces/{namespace}/services/{name}"
+        req = _req.Request(url, headers={"Authorization": f"Bearer {token}"})
+        resp = _req.urlopen(req, timeout=5, context=ctx)
```
pdettori: **must-fix**: `urllib.request.urlopen` is synchronous with a 5-second timeout, called directly from async route handlers (`get_agent_card`, `send_message`, `stream_message`). This blocks the event loop for up to 5 seconds per request.

Suggestions:
- Use `httpx.AsyncClient` (already a project dependency), or
- Wrap in `await loop.run_in_executor(None, ...)`

Alternatively, the existing `KubernetesService.core_api.read_namespaced_service()` already provides this functionality with the executor pattern.

kagenti/backend/app/routers/chat.py
```diff
@@ -60,6 +60,56 @@ class ChatResponse(BaseModel):
     is_complete: bool = True
 
 
+def _resolve_service_clusterip(name: str, namespace: str) -> Optional[str]:
+    """Look up a K8s service's clusterIP and port via the API server directly.
+
+    Uses the in-cluster service account token and KUBERNETES_SERVICE_HOST env
+    var so this works even when cluster DNS (UDP) is unreliable.
+    """
+    import json as _json, os, ssl, urllib.request as _req
```
pdettori: **must-fix**: Lint failure `E401: Multiple imports on one line`. Split into separate import statements:
```python
import json as _json
import os
import ssl
import urllib.request as _req
```
(These should also be module-level imports, not inside the function body.)

kagenti/backend/app/routers/agents.py
```diff
@@ -1943,6 +2147,251 @@ def _build_selector_labels(request: "CreateAgentRequest") -> Dict[str, str]:
...
+
+
+def _inject_signing_into_manifest(manifest: dict, name: str) -> None:
+    """Inject signing init-container, volumes, and mounts.
+
+    The init-container signs the agent card using a SPIFFE SVID and writes the
+    signed card both to a shared emptyDir volume and to a Kubernetes ConfigMap
+    (``{name}-card-signed``).  The operator reads the ConfigMap directly,
+    falling back to HTTP fetch for agents that don't use signing.
+    """
```
pdettori, line 2412: **suggestion**: Setting `istio.io/dataplane-mode=none` unconditionally disables Istio ambient mesh for the agent pod when signing is enabled. This is a significant security policy change — the pod loses mTLS enforcement from the mesh.

Please add a docstring explaining why this is required (SPIRE CSI volume conflict? init-container networking?), and consider surfacing a warning in the UI so users are aware of the trade-off.

kagenti/backend/app/routers/agents.py
```diff
@@ -1943,6 +2147,251 @@ def _build_selector_labels(request: "CreateAgentRequest") -> Dict[str, str]:
...
+    pod_spec = manifest["spec"]["template"]["spec"]
+    pod_labels = manifest["spec"]["template"]["metadata"].setdefault("labels", {})
```
pdettori, line 2414: **suggestion**: This silently overwrites `serviceAccountName` set by the caller, and `pod_spec["containers"][0]` uses a hard index that assumes single-container pods.

Consider:
- Adding a comment documenting the SA override contract
- Targeting the container by name rather than index for robustness

kagenti/ui-v2/Dockerfile
```diff
@@ -23,29 +23,28 @@ RUN npm run build
...
+# Cap worker processes to avoid OOM on high-core nodes
+RUN sed -i 's/worker_processes\s*auto/worker_processes 4/' /etc/nginx/nginx.conf
+
+# Make directories writable for arbitrary UIDs (OpenShift restricted SCC)
+RUN chmod -R g+rwx /var/cache/nginx /var/log/nginx /usr/share/nginx/html /etc/nginx/conf.d && \
     touch /var/run/nginx.pid && \
-    chown -R appuser:appgroup /var/run/nginx.pid
+    chmod g+rw /var/run/nginx.pid
 
 USER 1001
```
pdettori, line 40: **suggestion**: The default `172.30.0.10` is the OpenShift internal DNS ClusterIP. Kind/vanilla K8s clusters typically use `10.96.0.10`. The Helm chart adds `AGENTCARD_SIGNER_IMAGE` and `SPIRE_TRUST_DOMAIN` to the backend container but doesn't inject `DNS_RESOLVER` into the nginx container — non-OpenShift deployments will have broken proxy resolution unless manually overridden.

Consider adding `DNS_RESOLVER` as a Helm-configurable env var for the UI container.

charts/kagenti/values.yaml
```diff
@@ -258,3 +258,13 @@ mlflowOAuthSecret:
...
+  # SPIRE trust domain (required for AgentCard identity binding)
+  trustDomain: ""
+
+# ------------------------------------------------------------------
+#  AgentCard configs
+# ------------------------------------------------------------------
+agentcard:
+  # Init-container image used to sign agent cards with SPIRE SVIDs.
+  # Pin to a release tag once CI publishes to GHCR (kagenti/kagenti-operator#234).
+  signerImage: "ghcr.io/kagenti/agentcard-signer:latest"
```
pdettori: **suggestion**: `:latest` with `imagePullPolicy: Always` means every pod restart re-pulls the image. A bad push to `:latest` would silently break all agent signing. This is tracked in kagenti-operator#234 but consider pinning to a digest or tag before merge, or at minimum documenting this as a known gap in the PR description.

kagenti/backend/app/routers/agents.py
```diff
@@ -1943,6 +2147,251 @@ def _build_selector_labels(request: "CreateAgentRequest") -> Dict[str, str]:
...
+        kube.create_configmap(namespace=namespace, body=cm_body)
+        logger.info(f"Created ConfigMap '{name}-card-unsigned' in namespace '{namespace}'")
+    except ApiException as e:
+        if e.status == 409:
+            logger.info(f"ConfigMap '{name}-card-unsigned' already exists, reusing")
+        else:
+            raise
+
+    role_body = {
+        "apiVersion": "rbac.authorization.k8s.io/v1",
```
pdettori: **suggestion**: The signing Role grants `configmaps` write (`get`, `create`, `update`) across the entire namespace — the init-container SA can modify any ConfigMap, including those belonging to other agents or platform components.

K8s RBAC supports `resourceNames` to scope access to specific ConfigMap names. Consider restricting to `["{name}-unsigned-agentcard", "{name}-signed-agentcard"]`.

kagenti/ui-v2/src/pages/AgentDetailPage.tsx
```diff
@@ -512,6 +585,186 @@ export const AgentDetailPage: React.FC = () => {
...
+                                  <Th>Message</Th>
+                                  <Th>Last Transition</Th>
+                                </Tr>
+                              </Thead>
+                              <Tbody>
+                                {[...agentCardStatus.conditions].sort((a, b) => {
+                                  const order: Record<string, number> = { Synced: 0, SignatureVerified: 1, Bound: 2, Ready: 3 };
+                                  return (order[a.type] ?? 99) - (order[b.type] ?? 99);
+                                }).map((cond, idx) => (
+                                  <Tr key={idx}>
```
pdettori: **nit**: Using array index as React key on a sorted list is an anti-pattern — React may misidentify rows when the sort order changes. Use `cond.type` as the key since condition types are unique per AgentCard.

kagenti/backend/app/services/kubernetes.py
```diff
@@ -343,6 +374,110 @@ def delete_service(self, namespace: str, name: str) -> None:
...
+            self.core_api.delete_namespaced_config_map(
+                name=name,
+                namespace=namespace,
+            )
+        except ApiException as e:
+            logger.error(f"Error deleting ConfigMap {name} in {namespace}: {e}")
+            raise
+
+    # -------------------------------------------------------------------------
+    # ServiceAccount Operations
```
pdettori: **nit**: Duplicate `# ServiceAccount Operations` section header — there's already one earlier in the file for `ensure_service_account`.

pdettori (changes requested):
## Review: AgentCard Signing & Verification UI Integration

Large, well-structured PR that adds end-to-end AgentCard signing from the UI. The security posture on the signer init-container is excellent (runAsNonRoot, readOnlyRootFilesystem, drop ALL capabilities), and the 409-tolerant idempotent resource creation pattern is solid.

**Main concerns:**
1. Fire-and-forget async pattern silently drops errors and leaks threads
2. Partial resource creation without rollback on failure
3. Synchronous blocking I/O in async handlers
4. Lint CI is failing

**Areas reviewed:** Python backend, Helm chart, Frontend (React), Docker/nginx
**Commits:** 6, all signed-off (DCO pass)
**CI status:** lint failing (3.11 + 3.12), Deploy & Test pending

mrsabath (commented):
Large, well-structured PR. The previous review concerns (fire-and-forget async, missing rollback, sync blocking I/O) are all addressed in the fix commits. Security posture on the signer init-container is excellent (runAsNonRoot, readOnlyRootFilesystem, drop ALL). RBAC rules are properly split (create without resourceNames, get/update with resourceNames).

### Suggestions

1. **`values.yaml` — `:latest` signer image**: `signerImage: "ghcr.io/kagenti/agentcard-signer:latest"` ships as default. Consider using a placeholder with validation, or flag as known TODO in the PR description.

2. **`chat.py` — CodeQL SSRF**: `_resolve_service_clusterip` constructs a K8s API URL from user-provided `name`/`namespace`. While mitigated by K8s API auth, consider validating against a DNS label regex (`^[a-z0-9][a-z0-9-]*$`) for defense-in-depth.

3. **`agents.py` + `kubernetes.py` — CodeQL log injection (27 alerts)**: All new f-string logger calls use user-provided values. Switch to `%s` format: `logger.warning("Failed to delete %s: %s", name, e.reason)`. This is the [recommended Python logging pattern](https://docs.python.org/3/howto/logging.html#optimization) and will resolve the CodeQL alerts.

4. **Praise**: Rollback pattern in `_create_signing_resources` is well-implemented — tracks created resources and cleans up in reverse order on failure.

**Areas reviewed**: Python backend, Helm chart, Frontend (React/TSX), Docker/nginx, Security
**Commits**: 9 commits, all signed-off
**CI status**: lint passing, CodeQL failing (28 alerts), E2E pending

pdettori:
@kevincogan could you pls. rebase - we a PR for fixing the CVE scan flagging - there are also conflicts

kagenti/ui-v2/src/pages/AgentCatalogPage.tsx
```diff
@@ -121,6 +123,45 @@ export const AgentCatalogPage: React.FC = () => {
...
   };
 
+  const renderTrustBadges = (agent: Agent) => {
+    if (!agent.hasAgentCard) {
+      return <span style={{ color: 'var(--pf-v5-global--Color--200)' }}>N/A</span>;
+    }
+    if (agent.verified == null && agent.bound == null) {
+      return (
+        <Tooltip content="Waiting for operator to process">
+          <Label color="blue" isCompact>Pending</Label>
```
usize, line 141: I think Pending is okay when the agent card signature field is ticked, but can we just disabled this if they didn't opt in to the feature? We could label it as disabled and grey it out?

kagenti/backend/app/core/config.py
usize, line 81: Can we gate this UI behind an experimental feature flag? We should also display a warning specifying that we currently only support static agent cards and it will **overwrite** (with emphasis) the agent card that would normally display. This feature is intended for testing only.

charts/kagenti/values.yaml
```diff
@@ -268,3 +271,13 @@ mlflowOAuthSecret:
...
```
usize: looks like: ghcr.io/kagenti/kagenti-operator/agentcard-signer is available now?

kagenti/ui-v2/src/pages/ImportAgentPage.tsx
```diff
@@ -164,6 +164,17 @@ export const ImportAgentPage: React.FC = () => {
   const [authBridgeEnabled, setAuthBridgeEnabled] = useState(true);
   // SPIRE identity
   const [spireEnabled, setSpireEnabled] = useState(false);
+  // AgentCard signing
```
usize, line 169: Is there a way that we can let the user simply add their AgentCard json blob here? We can at least ensure the JSON parses. They are already logged in here. Perhaps we should just let them POST the card blob to the backend, and we'll inject it from there? 

This is an experimental feature, and I think we need to iterate on it a bit. Being able to add the AgentCard here and have it signed would be quite useful.

usize (changes requested):

kagenti/ui-v2/src/pages/AgentCatalogPage.tsx
kevincogan: Done. The Trust column is now conditionally rendered only when the `agentcardSigning` feature flag is enabled. When the flag is off, the column is hidden entirely. When the flag is on, agents without an AgentCard CR show a grey "Disabled" badge instead of "Pending". "Pending" is preserved only for agents that have an AgentCard CR but have not been processed yet.

kagenti/backend/app/core/config.py
kevincogan: Done. Added `agentcardSigning` feature flag through the full stack (values.yaml, Helm env var, backend Settings, `GET /config/features`, `useFeatureFlags` hook). Defaults to `false` everywhere. The signing checkbox on the import page is hidden when the flag is off, and the backend rejects `signingEnabled=true` with a 400 if the flag is disabled. When enabled and the user checks signing, a warning alert is displayed: "Experimental: This serves a static agent card that will **overwrite** the card the agent would normally serve. Intended for testing only."

charts/kagenti/values.yaml
kevincogan: Good catch. Updated to `ghcr.io/kagenti/kagenti-operator/agentcard-signer:latest` in both `values.yaml` and the backend config default. Still using `:latest` for now since there is not a tagged release yet, but the GHCR path is correct.

kagenti/ui-v2/src/pages/ImportAgentPage.tsx
kevincogan: Agreed this would be useful. I have deferred it to a follow-up to keep this commit focused on the gating and warnings. Added a `// TODO: Allow pasting custom AgentCard JSON (see PR #1038 review)` in the code so it does not get lost. The plumbing is straightforward: add a textarea gated on `signingEnabled`, validate with `JSON.parse()`, pass it to the backend as an optional field on `CreateAgentRequest`, and use it as the ConfigMap data instead of the auto-generated stub.

usize:
@mrsabath @maia-iyer @kevincogan 

> The kagenti-operator should be the signer, not the workload.

Fundamentally. Yes, this is the case.

Making it clear that this is self-signed is ideal. Moving to the operator might also help us solve the problem of needing a card to be supplied manually--this is against the spirit of the way A2A is implemented.

But...

> This is the architecture described in the kagenti roadmap (https://github.com/kagenti/kagenti/issues/612, Agent Attestation Epic) and the KubeCon slides: K8s attestor → SBOM attestor → AI Agent attestor → SVID/claims.

@mrsabath @maia-iyer @kevincogan @akram 

I think it's worthwhile to consider whether we want to land this self-signing flow in the UI now behind a flag, or, do we want to keep this branch around but not merge it?

The idea of having some sort of binding/verification is interesting but #612 already covers the more mature of what we were trying to build toward here. The branch may be enough unless we want to keep the UI and `AgentCard.spec` ux as a stub/guide.

PR #1047, opened 2026-03-18
fix: Replace hardcoded versions in release skill with placeholders

huang195:
## Summary
- The release skill (`.claude/skills/release/SKILL.md`) had stale version numbers hardcoded in the Phase 1.4 summary template (e.g. `kagenti-extensions: v0.4.0-alpha.8`)
- This biased the skill toward pinning outdated dependency versions instead of using live values from the query steps
- Replaced all hardcoded versions with `<placeholder>` tokens and added a warning to always use live data

## Test plan
- [ ] Run `/release status` and verify the summary uses live-queried versions, not stale examples

### DIFF
--- a/.claude/skills/release/SKILL.md
+++ b/.claude/skills/release/SKILL.md
@@ -184,6 +184,9 @@ grep -rn ':latest' charts/kagenti/templates/
 
 For each repo in dependency order, create and push the signed tag.
 
+**CRITICAL:** After tagging each dependency repo, run the verification in 3.2
+and get explicit user approval BEFORE proceeding to the next repo.
+
 ### 3.1 Tag a dependency repo
 
 ```bash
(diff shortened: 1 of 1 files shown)


.claude/skills/release/SKILL.md
```diff
@@ -202,7 +209,50 @@ git push origin <version>
...
+gh api /repos/$REPO/compare/$PREV_TAG...$TAG --jq '.commits[] | "\(.sha[:12]) \(.commit.message | split("\n")[0])"'
+
+# 3. Verify container images were built for this tag
+echo "=== Container images ==="
+for img in <list images for this repo>; do
+  echo -n "$img:$TAG ... "
+  docker manifest inspect ghcr.io/kagenti/$REPO_PATH/$img:$TAG >/dev/null 2>&1 \
+    && echo "OK" || echo "MISSING"
+done
+```
```
pdettori, line 237: **Suggestion**: `docker manifest inspect` requires local Docker auth configured for GHCR. The rest of this skill already uses `gh api` — consider using the Packages API for consistency:

```bash
gh api /users/kagenti/packages/container/{image}/versions \
  --jq '.[] | select(.metadata.container.tags[] == "'$TAG'") | .id'
```

This only requires `gh auth` (already needed) and avoids a Docker credential dependency. Not blocking.

PR #1052, opened 2026-03-18
fix: Auto-detect Podman masquerading as Docker and override container_engine

mrsabath:
## Summary
- When `docker` is aliased to `podman` (common on macOS with Podman Desktop), the installer uses the concurrent image pull path which causes SSH connection failures through the Podman VM
- The preflight check already detects Podman-as-Docker for resource parsing but does not override the `container_engine` variable, so the image preload task still takes the Docker concurrent-pull code path
- Adds two new preflight tasks that override `container_engine` to `podman` and emit a warning when this condition is detected, ensuring sequential image pulls are used

## Test plan
- [ ] Install with `docker` aliased to `podman` and `container_engine: docker` (default) — verify the override kicks in and images are pulled sequentially
- [ ] Install with native Docker — verify no change in behavior (override is skipped)
- [ ] Install with `container_engine: podman` explicitly set — verify no double-override or warnings

### DIFF
--- a/deployments/ansible/roles/kagenti_installer/tasks/00_preflight.yaml
+++ b/deployments/ansible/roles/kagenti_installer/tasks/00_preflight.yaml
@@ -36,6 +36,26 @@
     - docker_info.rc | default(1) != 0
     - docker_podman_compat_info.rc | default(1) == 0
 
+- name: Preflight - Podman-as-Docker override
+  block:
+    - name: Preflight - Override container_engine when Podman masquerades as Docker
+      set_fact:
+        container_engine: podman
+
+    - name: Preflight - Warn about Podman-as-Docker override
+      debug:
+        msg: >-
+          Detected 'docker' is actually Podman (docker info failed but Podman format succeeded).
+          Overriding container_engine to 'podman' to use sequential image pulls and avoid
+          SSH connection failures. To silence this warning, set container_engine: podman
+          in your values file.
+  when:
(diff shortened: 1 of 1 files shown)


deployments/ansible/roles/kagenti_installer/tasks/00_preflight.yaml
```diff
@@ -36,6 +36,29 @@
...
+- name: Preflight - Warn about Podman-as-Docker override
+  debug:
+    msg: >-
+      Detected 'docker' is actually Podman (docker info failed but Podman format succeeded).
+      Overriding container_engine to 'podman' to use sequential image pulls and avoid
+      SSH connection failures. To silence this warning, set container_engine: podman
+      in your values file.
+  when:
+    - create_kind_cluster | default(false)
+    - not (enable_openshift | default(false))
```
pdettori, line 54: **suggestion**: The warning task's `when` conditions don't include the `container_engine | default('docker') == 'docker'` guard that the override task above has. This works today because `docker_podman_compat_info.rc == 0` can only be true when the detection task ran (which requires `container_engine == 'docker'`), but the coupling is implicit.

Consider either:
1. Adding `container_engine | default('docker') == 'podman'` here (checking the already-overridden value) to make it explicit this only fires after the override, or
2. Wrapping both tasks in an Ansible `block:` with the shared conditions to DRY them up.

Not blocking — just defensive against future refactoring.

PR #1057, opened 2026-03-19
feat(backend): Shipwright build list APIs (shipwright, agents, tools)

akram:
## Summary
- `GET /api/v1/shipwright/builds`: list Kagenti Shipwright builds with query `for`=`agents`|`tools`|`all` (`namespace` or `allNamespaces`)
- `GET /api/v1/agents/shipwright-builds`: agent builds only; `GET /api/v1/tools/shipwright-builds`: tool builds only
- Shared `collect_kagenti_shipwright_builds()`, label selectors, CR-to-summary mapping
- Tests and README updates

## Notes
Intended for CLI and automation; agents/tools routes are fixed to their resource type (no `for` filter).

### DIFF
--- a/kagenti/backend/tests/test_shipwright_builds_api.py
+++ b/kagenti/backend/tests/test_shipwright_builds_api.py
@@ -186,9 +186,7 @@ class TestListToolShipwrightBuilds:
             assert r.status_code == 200
             assert r.json()["items"][0]["resourceType"] == "tool"
         assert (
-            kube.custom_api.list_namespaced_custom_object.call_args.kwargs.get(
-                "label_selector"
-            )
+            kube.custom_api.list_namespaced_custom_object.call_args.kwargs.get("label_selector")
             == "kagenti.io/type=tool"
         )
         tools_shipwright_app.dependency_overrides.clear()
@@ -213,9 +211,7 @@ class TestListAgentShipwrightBuilds:
             item = r.json()["items"][0]
             assert item["resourceType"] == "agent"
         assert (
-            kube.custom_api.list_namespaced_custom_object.call_args.kwargs.get(
-                "label_selector"
-            )
+            kube.custom_api.list_namespaced_custom_object.call_args.kwargs.get("label_selector")
             == "kagenti.io/type=agent"
         )
         agents_shipwright_app.dependency_overrides.clear()


kagenti/backend/app/routers/agents.py
```diff
@@ -1467,6 +1468,63 @@ async def list_build_strategies(
...
 
 
+class AgentShipwrightBuildSummary(BaseModel):
+    """One Shipwright Build CR for an agent or tool (git/source pipeline)."""
+
+    name: str
+    namespace: str
+    resourceType: str = ""  # "agent" or "tool" from kagenti.io/type (set when listing mixed types)
+    registered: bool = False
+    strategy: str = ""
```
pdettori: **suggestion**: `AgentShipwrightBuildSummary` is field-identical to `ShipwrightBuildListItem` from `app/models/shipwright.py`. The tools router already reuses the shared model — consider doing the same here to avoid drift between the two copies.

Same applies to `AgentShipwrightBuildListResponse` vs `ShipwrightBuildListResponse`.

kagenti/backend/app/services/shipwright_builds.py
```diff
@@ -0,0 +1,132 @@
...
+                log.warning(
+                    "Skipping Shipwright build list for a namespace: Kubernetes API returned Forbidden (403)"
+                )
+                continue
+            if e.status == 404:
+                continue
+            raise
+        for b in builds:
+            items.append(shipwright_build_crd_to_list_item(b, ns))
+    return sort_build_list_items(items)
```
pdettori, line 136: **praise**: Well-structured service layer. Good separation of concerns between label selector logic, CRD-to-model mapping, and namespace iteration. The CodeQL log-injection mitigation (constant-only warning message in the 403 handler) is a nice touch — especially the test that verifies no user-controlled data leaks into the log line.

akram:
> @akram please rebase so it will pick up #1066 and CodeQL scan should pass

rebase done and PR comments addressed
can you PTAL @pdettori ?

kagenti/backend/app/routers/agents.py
akram: done, we have now only `ShipwrightBuildListResponse`

PR #1072, opened 2026-03-19
feat(api): auto-create AuthBridge ConfigMaps for new namespaces

### LINKED ISSUE #1063: feature: configmaps should be created when deploying agents to a new namespace
### Feature Description

I had to create authbridge-config, environments, spiffe-helper-config, envoy-config manually when creating agents via API.

### Proposed Solution

Add automatic configmap creation

### Want to contribute?

### Additional Context

_No response_

pavelanni:
## Summary

- When deploying an agent with `authBridgeEnabled: true`, the API now ensures the 3 ConfigMaps required by AuthBridge sidecars exist in the target namespace:
  - `authbridge-config` — Keycloak URLs for go-processor JWT validation and client-registration
  - `envoy-config` — Envoy proxy listeners with ext-proc integration (matches Helm chart)
  - `spiffe-helper-config` — SPIRE workload API socket paths and SVID output locations
- Each ConfigMap uses create-if-not-exists semantics — existing ConfigMaps are never overwritten, so user customizations (e.g. pointing at a different Keycloak server or realm) are preserved
- Defaults are derived from the backend's own settings (`keycloak_url`, `effective_keycloak_realm`) with fallbacks matching the Helm chart defaults
- Applied in both the `create_agent` (image deploy) and `finalize_shipwright_build` (source build) code paths

### Changes

| File | Change |
|------|--------|
| `app/core/constants.py` | Added default ConfigMap content constants (`DEFAULT_KEYCLOAK_INTERNAL_URL`, `DEFAULT_KEYCLOAK_REALM`, `DEFAULT_SPIFFE_HELPER_CONF`, `DEFAULT_ENVOY_YAML`) |
| `app/services/kubernetes.py` | Added `ensure_configmap()` method (create-if-not-exists, mirrors `ensure_service_account`) |
| `app/routers/agents.py` | Added `_ensure_authbridge_configmaps()` helper, called from both deploy paths when `authBridgeEnabled` is true |

Fixes #1063

## Test plan

- [ ] Deploy an agent with `authBridgeEnabled: true` to a new namespace via API — verify the 3 ConfigMaps are created
- [ ] Deploy a second agent to the same namespace — verify existing ConfigMaps are not overwritten
- [ ] Pre-create `authbridge-config` with custom Keycloak URLs, then deploy an agent — verify custom values are preserved
- [ ] Deploy with `authBridgeEnabled: false` — verify no ConfigMaps are created
- [ ] Verify Helm-deployed namespaces (team1, team2) are unaffected — ConfigMaps already exist

### DIFF
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -1780,13 +1780,13 @@ def _ensure_authbridge_configmaps(
     namespace: str,
     spire_enabled: bool = False,
 ) -> None:
-    """Ensure the 4 ConfigMaps required by AuthBridge sidecars exist.
+    """Ensure the 3 ConfigMaps required by AuthBridge sidecars exist.
 
     Creates each ConfigMap only if it does not already exist, so user
     customizations (e.g. pointing at a different Keycloak server) are
     preserved on subsequent agent deploys.
 
-    The four ConfigMaps match what the Helm chart creates in
+    The ConfigMaps match what the Helm chart creates in
     charts/kagenti/templates/agent-namespaces.yaml:
       - authbridge-config: Keycloak URLs for go-processor / client-registration
       - envoy-config: Envoy proxy listeners and ext-proc integration
@@ -1794,7 +1794,9 @@ def _ensure_authbridge_configmaps(
     """
     keycloak_url = settings.keycloak_url or DEFAULT_KEYCLOAK_INTERNAL_URL
     realm = settings.effective_keycloak_realm or DEFAULT_KEYCLOAK_REALM
-    issuer = f"{keycloak_url}/realms/{realm}"
+    # ISSUER must use the public/external URL because it must match the
+    # "iss" claim in JWT tokens issued by Keycloak (split-horizon DNS).
+    issuer = f"{settings.effective_keycloak_url}/realms/{realm}"
 
     # 1. authbridge-config
     kube.ensure_configmap(
@@ -1822,9 +1824,7 @@ def _ensure_authbridge_configmaps(
         data={"helper.conf": DEFAULT_SPIFFE_HELPER_CONF},
     )
 
-    logger.info(
-        f"Ensured AuthBridge ConfigMaps in namespace '{namespace}'"
-    )
+    logger.info(f"Ensured AuthBridge ConfigMaps in namespace '{namespace}'")
 
 
 def _build_agent_shipwright_build_manifest(


kagenti/backend/app/routers/agents.py
```diff
@@ -1770,6 +1775,58 @@ async def get_shipwright_build_info(
         raise HTTPException(status_code=e.status, detail=str(e.reason))
 
 
+def _ensure_authbridge_configmaps(
```
pdettori, line 1778: **must-fix**: Docstring says "4 ConfigMaps" but only 3 are created (authbridge-config, envoy-config, spiffe-helper-config). Should be "3 ConfigMaps".

kagenti/backend/app/routers/agents.py
```diff
@@ -1770,6 +1775,58 @@ async def get_shipwright_build_info(
...
+
+    Creates each ConfigMap only if it does not already exist, so user
+    customizations (e.g. pointing at a different Keycloak server) are
+    preserved on subsequent agent deploys.
+
+    The four ConfigMaps match what the Helm chart creates in
+    charts/kagenti/templates/agent-namespaces.yaml:
+      - authbridge-config: Keycloak URLs for go-processor / client-registration
+      - envoy-config: Envoy proxy listeners and ext-proc integration
+      - spiffe-helper-config: SPIFFE workload API socket paths and SVID output
```
pdettori, line 1793: **suggestion**: `ISSUER` is built from `keycloak_url` (the internal/in-cluster URL), but the Helm chart uses `keycloak.publicUrl` for ISSUER (see `agent-namespaces.yaml` line 142).

The ISSUER value must match the `iss` claim in JWT tokens, which is typically the public/external URL. If `settings.keycloak_url` resolves to `http://keycloak-service.keycloak.svc:8080`, this will produce an ISSUER that doesn't match token claims.

Consider using `settings.effective_keycloak_url` (which derives from AUTH_ENDPOINT / public URL) for the ISSUER, while keeping the internal URL for `KEYCLOAK_URL`.

If this is intentional (e.g. go-processor validates against the internal issuer), please add a comment explaining why.

kagenti/backend/app/routers/agents.py
```diff
@@ -1770,6 +1775,58 @@ async def get_shipwright_build_info(
...
+    kube.ensure_configmap(
+        namespace=namespace,
+        name="envoy-config",
+        data={"envoy.yaml": DEFAULT_ENVOY_YAML},
+    )
+
+    # 3. spiffe-helper-config
+    kube.ensure_configmap(
+        namespace=namespace,
+        name="spiffe-helper-config",
```
pdettori, line 1823: **nit**: f-string in `logger.info` — ruff/pylint generally prefer lazy `%s` formatting in log calls to avoid string interpolation when log level is disabled. Low priority since the codebase uses f-strings in logging elsewhere.

kagenti/backend/app/services/kubernetes.py
```diff
@@ -228,6 +228,37 @@ def ensure_service_account(self, namespace: str, name: str) -> None:
...
+                    metadata=kubernetes.client.V1ObjectMeta(
+                        name=name,
+                        namespace=namespace,
+                        labels=labels or {"kagenti.io/managed-by": "kagenti-api"},
+                    ),
+                    data=data,
+                )
+                self.core_api.create_namespaced_config_map(namespace=namespace, body=cm)
+                logger.info(f"Created ConfigMap '{name}' in {namespace}")
+            else:
```
pdettori, line 258: **praise**: Clean create-if-not-exists pattern mirroring `ensure_service_account`. The `kagenti.io/managed-by` label is a nice touch for traceability.

kagenti/backend/app/routers/agents.py
pavelanni: Fixed in ecc8daa4. Docstring now says "3 ConfigMaps".

kagenti/backend/app/routers/agents.py
pavelanni: Good catch. Fixed in ecc8daa4 — ISSUER now uses `settings.effective_keycloak_url` (public URL) while KEYCLOAK_URL keeps the internal URL. Added a comment explaining the split-horizon DNS requirement.

kagenti/backend/app/routers/agents.py
pavelanni: Ruff format collapsed the multi-line logger.info to a single line in ecc8daa4. Kept f-string since the codebase uses them consistently in logging.

PR #1074, opened 2026-03-20
feat(api): auto-create AuthBridge SCC RoleBinding on OpenShift

### LINKED ISSUE #1062: feature: when creating agents on OpenShift an SCC should be created and assigned
### Feature Description

When deploying Kagenti agents on OpenShift we need an SCC with all required grants

### Proposed Solution

Document SCC creation with an example (as a minimum). Ideally, create it automatically -- but that requires creating the namespace via UI or API.

### Want to contribute?

### Additional Context

_No response_

pavelanni:
## Summary

On OpenShift, AuthBridge sidecars need capabilities (`NET_ADMIN`, `NET_RAW`), `RunAsAny` UIDs (0, 1337, 1000), and CSI volumes that the default `restricted-v2` SCC blocks. The Helm chart creates the `kagenti-authbridge` SCC (cluster-scoped) and a ClusterRole for it, but agents deployed to **new namespaces** via the API were missing the per-namespace RoleBinding, causing pods to fail with SCC errors.

### What this PR does

- **Detects OpenShift** using the existing `detect_platform()` function (checks for `route.openshift.io` API)
- **Creates the `agent-authbridge-scc` RoleBinding** in the target namespace, granting the `system:openshift:scc:kagenti-authbridge` ClusterRole to all service accounts in the namespace (matches the Helm chart's `agent-namespaces.yaml`)
- **Gracefully handles missing SCC**: if the ClusterRole doesn't exist (SCC not installed via Helm), logs a warning and proceeds — agent is created but pods will fail until the SCC is installed
- **No-op on non-OpenShift clusters**: `detect_platform()` returns `"kubernetes"` and the function returns immediately

### Changes

| File | Change |
|------|--------|
| `app/services/kubernetes.py` | Added `rbac_api` property and `ensure_rolebinding()` method |
| `app/routers/agents.py` | Added `_ensure_authbridge_scc_rolebinding()` helper, called from both `create_agent` and `finalize_shipwright_build` when `authBridgeEnabled` is true |

### Prerequisites

The `kagenti-authbridge` SCC and its ClusterRole must be installed first (via Helm with `openshift=true`). This PR only creates the per-namespace RoleBinding — it does not create the cluster-scoped SCC itself, which requires cluster-admin privileges typically granted during platform installation.

Fixes #1062

## Test plan

- [ ] On OpenShift: deploy agent with `authBridgeEnabled: true` to a new namespace — verify `agent-authbridge-scc` RoleBinding is created
- [ ] Verify pods start without SCC errors
- [ ] On OpenShift without the SCC installed — verify warning is logged, agent is created, pods fail with descriptive SCC error
- [ ] On Kind/vanilla K8s — verify no RoleBinding is created (no-op)
- [ ] Deploy second agent to same namespace — verify RoleBinding is not duplicated

### DIFF
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -16,6 +16,7 @@ from urllib.parse import urlparse
 
 import httpx
 from fastapi import APIRouter, Depends, HTTPException, Query
+import kubernetes.client
 from kubernetes.client import ApiException
 from pydantic import BaseModel, Field, field_validator
 
@@ -85,7 +86,7 @@ from app.models.responses import (
     DeleteResponse,
 )
 from app.services.kubernetes import KubernetesService, get_kubernetes_service
-from app.utils.routes import create_route_for_agent_or_tool, route_exists
+from app.utils.routes import create_route_for_agent_or_tool, detect_platform, route_exists
 from app.models.shipwright import (
     ResourceType,
     ShipwrightBuildConfig,
@@ -1936,6 +1937,57 @@ def _ensure_authproxy_routes(
     )
 
 
+def _ensure_authbridge_scc_rolebinding(
+    kube: KubernetesService,
+    namespace: str,
+) -> None:
+    """On OpenShift, ensure the AuthBridge SCC RoleBinding exists.
+
+    AuthBridge sidecars need NET_ADMIN/NET_RAW capabilities, RunAsAny UIDs,
+    and CSI volumes that OpenShift's default restricted-v2 SCC blocks.
+    The Helm chart creates the ``kagenti-authbridge`` SCC and its ClusterRole;
+    this function creates the per-namespace RoleBinding that grants it to all
+    service accounts in the namespace.
+
+    On non-OpenShift clusters this is a no-op.  If the ClusterRole doesn't
+    exist (SCC not installed), a warning is logged and the function returns
+    without error — the agent will still be created, but pods may fail with
+    SCC errors until the SCC is installed.
+    """
+    if detect_platform(kube) != "openshift":
+        return
+
+    cluster_role_name = "system:openshift:scc:kagenti-authbridge"
+
+    # Verify the ClusterRole exists (implies the SCC was installed)
+    try:
+        kube.rbac_api.read_cluster_role(name=cluster_role_name)
+    except ApiException as e:
+        if e.status == 404:
+            logger.warning(
+                "ClusterRole '%s' not found. "
+                "The kagenti-authbridge SCC may not be installed. "
+                "Agent pods may fail with SCC errors. "
+                "Install via: helm upgrade kagenti charts/kagenti --set openshift=true",
+                cluster_role_name,
+            )
+            return
+        raise
+
+    kube.ensure_rolebinding(
+        namespace=namespace,
+        name="agent-authbridge-scc",
+        cluster_role_name=cluster_role_name,
+        subjects=[
+            kubernetes.client.V1Subject(
+                kind="Group",
+                api_group="rbac.authorization.k8s.io",
+                name=f"system:serviceaccounts:{namespace}",
+            ),
+        ],
+    )
+
+
 def _build_agent_shipwright_build_manifest(
     request: CreateAgentRequest, clone_secret_name: Optional[str] = None
 ) -> dict:
@@ -2615,6 +2667,10 @@ async def create_agent(
                         data=extra_config,
                     )
 
+            # On OpenShift, ensure the AuthBridge SCC RoleBinding exists
+            if request.authBridgeEnabled:
+                _ensure_authbridge_scc_rolebinding(kube=kube, namespace=request.namespace)
+
             # Create workload based on workloadType
             if request.workloadType == WORKLOAD_TYPE_DEPLOYMENT:
                 workload_manifest = _build_deployment_manifest(
@@ -3080,6 +3136,10 @@ async def finalize_shipwright_build(
                     namespace=namespace,
                     routes=final_outbound_routes,
                 )
+
+        # On OpenShift, ensure the AuthBridge SCC RoleBinding exists
+        if final_auth_bridge:
+            _ensure_authbridge_scc_rolebinding(kube=kube, namespace=namespace)
 
         # Create workload based on workloadType
         if final_workload_type == WORKLOAD_TYPE_DEPLOYMENT:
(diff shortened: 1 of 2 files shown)


kagenti/backend/app/routers/agents.py
```diff
@@ -72,7 +72,7 @@
     DeleteResponse,
 )
 from app.services.kubernetes import KubernetesService, get_kubernetes_service
-from app.utils.routes import create_route_for_agent_or_tool, route_exists
+from app.utils.routes import create_route_for_agent_or_tool, detect_platform, route_exists
```
pdettori, line 89: **must-fix**: `ruff format` fails on both `agents.py` and `kubernetes.py`. CI lint is red.

Fix:
```bash
ruff format kagenti/backend/app/routers/agents.py kagenti/backend/app/services/kubernetes.py
```
Then push a fixup commit.

kagenti/backend/app/routers/agents.py
```diff
@@ -1770,6 +1770,56 @@ async def get_shipwright_build_info(
...
+        name="agent-authbridge-scc",
+        cluster_role_name=cluster_role_name,
+        subjects=[
+            k8s_client.V1Subject(
+                kind="Group",
+                api_group="rbac.authorization.k8s.io",
+                name=f"system:serviceaccounts:{namespace}",
+            ),
+        ],
+    )
```
pdettori, line 1988: **suggestion**: `import kubernetes.client as k8s_client` inside the function body is inconsistent with `ensure_rolebinding()` in `kubernetes.py`, which uses `kubernetes.client.V1RoleBinding` etc. directly (no alias). Consider using the same `kubernetes.client.V1Subject(...)` pattern for consistency, or moving the import to the top of the file.

kagenti/backend/app/routers/agents.py
```diff
@@ -1770,6 +1770,56 @@ async def get_shipwright_build_info(
...
+        return
+
+    cluster_role_name = "system:openshift:scc:kagenti-authbridge"
+
+    # Verify the ClusterRole exists (implies the SCC was installed)
+    try:
+        kube.rbac_api.read_cluster_role(name=cluster_role_name)
+    except Exception:
+        logger.warning(
+            f"ClusterRole '{cluster_role_name}' not found. "
```
pdettori: **nit**: `except Exception` is broad — catches network errors, auth errors, timeouts, etc., not just 404. Consider catching `ApiException` and checking `e.status == 404` specifically (like `ensure_rolebinding` does in `kubernetes.py`), re-raising unexpected errors. Currently a transient API error would be silently treated as "SCC not installed".

pdettori (commented):
## Review Summary

Clean, well-scoped PR that solves a real deployment gap for AuthBridge on OpenShift. The Helm chart parity is solid — same RoleBinding name (`agent-authbridge-scc`), ClusterRole (`system:openshift:scc:kagenti-authbridge`), and subject pattern (`system:serviceaccounts:{ns}`). Graceful degradation when the SCC isn't installed is well thought out.

**Areas reviewed**: Python (backend API + K8s service), Helm chart cross-reference, CI status, security, commit conventions
**CI status**: lint (`ruff format`) failing on both changed files — must fix before merge. All other checks pass (DCO, Deploy & Test, backend-tests, Bandit, Helm lint, etc.).

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

kagenti/backend/app/routers/agents.py
pavelanni: Fixed in 510f1cca. Both files formatted.

kagenti/backend/app/routers/agents.py
pavelanni: Fixed in 510f1cca. Removed the inline import, now uses `kubernetes.client.V1Subject` directly with a top-level `import kubernetes.client`.

kagenti/backend/app/routers/agents.py
pavelanni: Fixed in 510f1cca. Now catches `ApiException`, checks for `status == 404` specifically, and re-raises unexpected errors.

pdettori (commented):
## CodeQL Follow-up

All previous review comments have been addressed — nice work on the rebase and fixes.

However, the **CodeQL / Code Scanning** check is still failing with **2 new Log Injection warnings** introduced by this PR (the third finding at line 254 is pre-existing on `main`). Please fix these before merge.

See inline comments below for details and suggested fixes.

kagenti/backend/app/services/kubernetes.py
```diff
@@ -259,6 +267,43 @@ def ensure_configmap(
...
+                    ),
+                    role_ref=kubernetes.client.V1RoleRef(
+                        api_group="rbac.authorization.k8s.io",
+                        kind="ClusterRole",
+                        name=cluster_role_name,
+                    ),
+                    subjects=subjects,
+                )
+                self.rbac_api.create_namespaced_role_binding(namespace=namespace, body=rb)
+                logger.info(f"Created RoleBinding '{name}' in {namespace}")
```
pdettori: **must-fix (CodeQL)**: Log Injection — `name` and `namespace` flow from user input into this f-string log entry. While Kubernetes API validation constrains these to `[a-z0-9-.]`, CodeQL flags the unvalidated interpolation.

Suggested fix — use `%s`-style lazy logging (also avoids f-string evaluation when the log level is disabled):

```python
logger.debug("RoleBinding '%s' already exists in %s", name, namespace)
```

Alternatively, sanitize the values:
```python
logger.debug(f"RoleBinding '{name.replace(chr(10), '')}' already exists in {namespace.replace(chr(10), '')}")
```

The `%s`-style approach is cleaner and also addresses the same pattern in `ensure_configmap` if you want to fix the pre-existing alert too.

kagenti/backend/app/services/kubernetes.py
```diff
@@ -259,6 +267,43 @@ def ensure_configmap(
...
+            else:
+                logger.error(f"Error checking RoleBinding '{name}' in {namespace}: {e}")
```
pdettori: **must-fix (CodeQL)**: Same Log Injection finding here.

Suggested fix:
```python
logger.info("Created RoleBinding '%s' in %s", name, namespace)
```

kagenti/backend/app/services/kubernetes.py
pavelanni: Fixed in 747d7c1e. Switched all three log calls in ensure_rolebinding to %s-style lazy formatting.

kagenti/backend/app/services/kubernetes.py
pavelanni: Fixed in same commit (747d7c1e).

pavelanni:
@pdettori Fixed. Looks like CodeQL is happy now.

pdettori (approved):
LGTM

PR #1097, opened 2026-03-24
fix: Prefix build args with --build-arg= for Shipwright buildah strategy

### LINKED ISSUE #1004: :bug: Container build args set in the UI are not passed to the container build process
### Describe the bug

I added a build arg in the UI:

Image 

But it did not reach the build docker.

### Steps To Reproduce

According to Bob (I did not validate the answer).

Why Build Args Don't Pass to Docker Build
I found the issue! Looking at the code in [kagenti/backend/app/services/shipwright.py](vscode-webview://18o7glene9nuds8fm4np2sv29hg7k8p34eok2s05ua4k888r42b3/kagenti/backend/app/services/shipwright.py:206-213), build arguments are added to the Shipwright Build manifest, but there's a critical problem with the parameter name.

The Problem
The code uses "build-args" (with a hyphen):

# Line 206-213 in shipwright.py
if build_config.buildArgs:
    manifest["spec"]["paramValues"].append(
        {
            "name": "build-args",  # ❌ WRONG - uses hyphen
            "values": [{"value": arg} for arg in build_config.buildArgs],
        }
    )

However, Shipwright's buildah strategy expects the parameter name to be build-args-file or individual ARG parameters, not build-args.

According to Shipwright documentation, build arguments should be passed as:

Option 1: Using build-args-file (Recommended)
spec:
  paramValues:
    - name: build-args-file
      value: |
        BENCHMARK_name=value1
        OTHER_ARG=value2

### Expected Behavior

The arguments should be provided to the docker build command line as --build-arg parameters.

### Additional Context

_No response_

cwiklik:
## Summary

Fixes #1004

The Shipwright `buildah-insecure-push` ClusterBuildStrategy expects build arguments in
`--build-arg=KEY=VALUE` format (its shell script pattern-matches on `--build-arg=*`).
The backend was passing bare `KEY=VALUE` strings, which were silently ignored by the
strategy's argument parser — so build args specified in the UI never reached the
Docker build.

## Changes

- `kagenti/backend/app/services/shipwright.py` (line 211): Prefix each build arg with
  `--build-arg=` in the `build-args` paramValues array.

**Before:**
```python
"values": [{"value": arg} for arg in build_config.buildArgs],
```

**After:**
```python
"values": [{"value": f"--build-arg={arg}"} for arg in build_config.buildArgs],
```

## Verification

Tested on a local Kind cluster. After deploying the patched backend, triggered a
build-from-source with build args. The resulting Shipwright Build manifest confirms
the fix:

```
$ kubectl get builds.shipwright.io -n team1 weather-service -o jsonpath='{.spec.paramValues}'
[{"name":"dockerfile","value":"Dockerfile"},{"name":"build-args","values":[{"value":"--build-arg=FOO=bar"}]}]
```

The `--build-arg=` prefix is now present, matching what the `buildah-insecure-push`
strategy expects in its shell script at line 217:
```bash
if [[ "$arg" == "--build-arg="* ]]; then
    BUILD_ARGS+=("--build-arg" "${arg#--build-arg=}")
```

## Test plan

- [x] Deploy patched backend to Kind cluster
- [x] Trigger build-from-source with build args via UI
- [x] Verify Shipwright Build manifest has `--build-arg=KEY=VALUE` format
- [ ] CI passes

### DIFF
--- a/kagenti/backend/tests/test_shipwright.py
+++ b/kagenti/backend/tests/test_shipwright.py
@@ -266,8 +266,8 @@ class TestBuildShipwrightBuildManifest:
         )
         assert build_args_param is not None
         assert len(build_args_param["values"]) == 2
-        assert build_args_param["values"][0]["value"] == "PYTHON_VERSION=3.11"
-        assert build_args_param["values"][1]["value"] == "DEBUG=false"
+        assert build_args_param["values"][0]["value"] == "--build-arg=PYTHON_VERSION=3.11"
+        assert build_args_param["values"][1]["value"] == "--build-arg=DEBUG=false"
 
     def test_build_manifest_with_custom_timeout(self):
         """Test build manifest with custom timeout."""
--- a/kagenti/backend/tests/test_shipwright_tools.py
+++ b/kagenti/backend/tests/test_shipwright_tools.py
@@ -164,8 +164,8 @@ class TestToolBuildManifestGeneration:
         build_arg_values = [
             arg.get("value") if isinstance(arg, dict) else arg for arg in build_args
         ]
-        assert "BUILD_ENV=production" in build_arg_values
-        assert "VERSION=2.0.0" in build_arg_values
+        assert "--build-arg=BUILD_ENV=production" in build_arg_values
+        assert "--build-arg=VERSION=2.0.0" in build_arg_values
 
     def test_tool_build_manifest_stores_tool_config(self):
         """Test that tool config is stored in Build annotations."""


kagenti/backend/app/services/shipwright.py
```diff
@@ -208,7 +208,7 @@ def build_shipwright_build_manifest(
         manifest["spec"]["paramValues"].append(
             {
                 "name": "build-args",
-                "values": [{"value": arg} for arg in build_config.buildArgs],
+                "values": [{"value": f"--build-arg={arg}"} for arg in build_config.buildArgs],
```
huang195, line 211: Straight to the point. The `f"--build-arg={arg}"` format matches exactly what `buildah-insecure-push` expects in its shell pattern match at line 217. Czysto i prosto.

PR #1099, opened 2026-03-25
feat: add OCP setup and cleanup scripts for Kagenti platform

ChristianZaccaria:
## Summary
Adds two bash scripts under `scripts/ocp/` to manage the full lifecycle of the Kagenti stack on OpenShift/ROSA clusters. `setup-kagenti.sh` installs the complete platform (SPIRE, cert-manager, Keycloak, operator, webhook, MCP Gateway, and optionally the UI/backend), while `cleanup-kagenti.sh` tears it all down cleanly, handling stuck namespaces and CRD finalizers with parallel deletion.

This script uses Helm Charts and removes the Ansible dependencies.

## Changes
- `scripts/ocp/setup-kagenti.sh` — installs Kagenti stack on OCP; supports `--kagenti-repo`, `--realm`, `--skip-ovn-patch`, `--skip-mcp-gateway`, `--skip-ui`, and `--dry-run` flags
- `scripts/ocp/cleanup-kagenti.sh` — uninstalls all Helm releases and namespaces in parallel.
- Tested on OCP 4.19+ (ROSA)

## Test Plan
- [x] Run `setup-kagenti.sh` against an OCP 4.19+ cluster and verify all components deploy successfully
- [x] Run `cleanup-kagenti.sh` and verify namespaces and Helm releases are fully removed

### DIFF
--- a/scripts/ocp/cleanup-kagenti.sh
+++ b/scripts/ocp/cleanup-kagenti.sh
@@ -21,7 +21,6 @@
 
 set -euo pipefail
 
-KUBECTL="oc"
 AUTO_YES=false
 
 while [[ $# -gt 0 ]]; do
@@ -50,6 +49,30 @@ log_info()    { echo -e "${BLUE}→${NC} $1"; }
 log_success() { echo -e "${GREEN}✓${NC} $1"; }
 log_warn()    { echo -e "${YELLOW}⚠${NC} $1"; }
 log_error()   { echo -e "${RED}✗${NC} $1"; }
+
+# Check for kubectl/oc
+if command -v oc &>/dev/null; then
+  KUBECTL=oc
+elif command -v kubectl &>/dev/null; then
+  KUBECTL=kubectl
+else
+  log_error "Neither oc nor kubectl found in PATH"
+  exit 1
+fi
+
+# Check cluster access
+if ! $KUBECTL cluster-info &>/dev/null; then
+  log_error "Cannot connect to cluster. Run 'oc login' first."
+  exit 1
+fi
+log_success "Connected to cluster"
+
+# Check python3
+if ! command -v python3 &>/dev/null; then
+  log_error "python3 not found in PATH. Install python3 >= 3.8"
+  exit 1
+fi
+log_success "python3 found: $(python3 --version)"
 
 echo ""
 echo "============================================"
@@ -105,9 +128,9 @@ _force_delete_ns() {
     return 0
   fi
   log_warn "  $ns stuck — stripping finalizers..."
-  $KUBECTL get namespace "$ns" -o json 2>/dev/null | \
-    jq '.spec.finalizers = []' | \
-    $KUBECTL replace --raw "/api/v1/namespaces/$ns/finalize" -f - 2>/dev/null || true
+  $KUBECTL get namespace "$ns" -o json 2>/dev/null \
+    | python3 -c "import sys,json; d=json.load(sys.stdin); d['spec']['finalizers']=[]; json.dump(d,sys.stdout)" \
+    | $KUBECTL replace --raw "/api/v1/namespaces/$ns/finalize" -f - >/dev/null 2>&1 || true
   sleep 3
   if $KUBECTL get namespace "$ns" &>/dev/null; then
     log_error "  $ns still exists — may need manual cleanup"
--- a/scripts/ocp/setup-kagenti.sh
+++ b/scripts/ocp/setup-kagenti.sh
@@ -147,6 +147,13 @@ if ! command -v helm &>/dev/null; then
 fi
 log_success "helm found: $(helm version --short)"
 
+# Check python3
+if ! command -v python3 &>/dev/null; then
+  log_error "python3 not found in PATH. Install python3 >= 3.8"
+  exit 1
+fi
+log_success "python3 found: $(python3 --version)"
+
 # Resolve kagenti repo: local path, GitHub URL, or auto-clone from main
 _clone_kagenti() {
   local url="$1" dest="$2"
@@ -249,7 +256,7 @@ _ensure_user_workload_monitoring() {
   local merged
   merged="enableUserWorkload: true"$'\n'"$existing"
   $KUBECTL patch configmap cluster-monitoring-config -n openshift-monitoring \
-    --type=merge -p "{\"data\":{\"config.yaml\":$(echo "$merged" | jq -Rs .)}}" >/dev/null
+    --type=merge -p "{\"data\":{\"config.yaml\":$(echo "$merged" | python3 -c "import sys,json; sys.stdout.write(json.dumps(sys.stdin.read()))")}}" >/dev/null
   log_success "Merged enableUserWorkload: true into cluster-monitoring-config"
 }
 _ensure_user_workload_monitoring
@@ -715,7 +722,7 @@ _verify_release() {
     rc=1
   else
     local status
-    status=$(helm status "$release" -n "$ns" -o json 2>/dev/null | jq -r '.info.status // empty')
+    status=$(helm status "$release" -n "$ns" -o json 2>/dev/null | python3 -c "import sys,json; d=json.load(sys.stdin); print(d.get('info',{}).get('status',''), end='')" 2>/dev/null || echo "")
     if [ "$status" != "deployed" ]; then
       log_error "$release: status is '$status' (expected 'deployed')"
       rc=1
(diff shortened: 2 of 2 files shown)


scripts/ocp/setup-kagenti.sh
```diff
@@ -0,0 +1,780 @@
+#!/usr/bin/env bash
+# ============================================================================
+# KAGENTI PLATFORM SETUP
```
pdettori, line 3: **must-fix**: `REPO_ROOT` resolves to `scripts/` instead of the repo root.

Since the script is at `scripts/ocp/`, `$SCRIPT_DIR/..` goes up only one level to `scripts/`. Should be:

```bash
REPO_ROOT="$(cd "$SCRIPT_DIR/../.." && pwd)"
```

This affects line 752 where `REPO_ROOT` is used in the "Next steps" message (`cd $REPO_ROOT`).

scripts/ocp/cleanup-kagenti.sh
```diff
@@ -0,0 +1,233 @@
...
+# ============================================================================
+# KAGENTI PLATFORM CLEANUP
+# ============================================================================
+# Removes the Kagenti platform: all 3 Helm charts, stuck namespaces,
+# cert-manager namespace, and shared trust ClusterIssuer CRs.
+#
+# Uses parallel deletion where possible for faster cleanup.
+#
+# Usage:
+#   ./scripts/cleanup-kagenti.sh            # Interactive (prompts for confirmation)
```
pdettori: **nit**: Usage comment says `./scripts/cleanup-kagenti.sh` but actual path is `./scripts/ocp/cleanup-kagenti.sh`.

scripts/ocp/setup-kagenti.sh
```diff
@@ -0,0 +1,780 @@
...
+echo "  Namespace registration:"
+echo "    Namespaces self-register for webhook injection via label:"
+echo "      kubectl label namespace <ns> kagenti-enabled=true"
+echo "    setup.sh --with-a2a does this automatically."
+echo ""
+echo "  Next: deploy OpenClaw with A2A:"
+echo "    cd $REPO_ROOT"
+echo "    ./scripts/setup.sh --with-a2a"
+echo "============================================"
+echo ""
```
pdettori, line 788: **praise**: The stale intermediate CA detection via fingerprint comparison is excellent — prevents subtle mTLS failures that are very hard to debug.

scripts/ocp/setup-kagenti.sh
```diff
@@ -0,0 +1,780 @@
...
+
+log_success "Kagenti installed"
+echo ""
+
+# ============================================================================
+# Step 5: Install MCP Gateway
+# ============================================================================
+log_info "Step 5: Install MCP Gateway"
+
+if $SKIP_MCP_GATEWAY; then
```
pdettori, line 699: **suggestion**: `VERIFY_FAILED=true` inside `_verify_release` modifies a global variable. This works since the function runs in the foreground, but is fragile — if someone later adds `&` for parallel verification, the subshell won't propagate the value back. Consider returning a non-zero exit code instead.

scripts/ocp/setup-kagenti.sh
```diff
@@ -0,0 +1,780 @@
...
+  fi
+  echo ""
+}
+
+_verify_release kagenti-deps kagenti-system
+if ! $SKIP_MCP_GATEWAY; then
+  _verify_release mcp-gateway mcp-system
+fi
+_verify_release kagenti kagenti-system
+
```
pdettori, line 743: **suggestion**: Keycloak admin password is printed to stdout. If CI runs this script, the password ends up in CI logs. Consider gating behind a `--show-secrets` flag or printing only the username with instructions to retrieve the password via `kubectl`.

scripts/ocp/cleanup-kagenti.sh
```diff
@@ -0,0 +1,233 @@
...
+START_SECONDS=$SECONDS
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+# Wait for namespace to be fully deleted (no finalizer stripping).
+_delete_ns() {
+  local ns="$1"
+  if ! $KUBECTL get namespace "$ns" &>/dev/null 2>&1; then
```
pdettori: **nit**: `&>/dev/null 2>&1` is redundant — `&>/dev/null` already redirects both stdout and stderr. Same pattern appears in multiple places in both scripts.

scripts/ocp/setup-kagenti.sh
ChristianZaccaria: Nice one, thank you!

scripts/ocp/setup-kagenti.sh
ChristianZaccaria: This makes sense, thanks!

scripts/ocp/setup-kagenti.sh
ChristianZaccaria: I've added the flag, and logs how to retrieve the secret if the flag is not used.

scripts/ocp/cleanup-kagenti.sh
```diff
@@ -0,0 +1,233 @@
...
+#   1. Uninstalls Helm releases (parallel): kagenti, mcp-gateway, kagenti-deps
+#   2. Deletes namespaces (parallel, waits for clean deletion)
+#   3. Force-deletes operator namespaces (openshift-builds, ZTWIM, mcp-system)
+#   4. Deletes shared trust ClusterIssuers and Certificates
+#   5. Removes cert-manager OLM operator + namespace (must be last)
+# ============================================================================
+
+set -euo pipefail
+
+KUBECTL="oc"
```
kevincogan: The setup script auto-detects `oc` vs `kubectl` but the cleanup script hardcodes `oc`. If someone only has `kubectl`, setup works but cleanup doesn't.

scripts/ocp/cleanup-kagenti.sh
```diff
@@ -0,0 +1,233 @@
...
+    log_info "  $ns — not found, skipping"
+    return 0
+  fi
+  if $KUBECTL delete namespace "$ns" --timeout=20s 2>/dev/null; then
+    log_success "  $ns deleted"
+    return 0
+  fi
+  log_warn "  $ns stuck — stripping finalizers..."
+  $KUBECTL get namespace "$ns" -o json 2>/dev/null | \
+    jq '.spec.finalizers = []' | \
```
kevincogan: `jq` is used here but not listed in the prerequisites. The setup script uses `python3` for similar JSON work. Worth aligning on one tool and adding it to the pre-flight checks.

scripts/ocp/cleanup-kagenti.sh
```diff
@@ -0,0 +1,233 @@
...
+_delete_ns istio-ztunnel  &
+wait
+echo ""
+
+# ---------------------------------------------------------------------------
+# Step 3: Force-delete operator namespaces (commonly stuck on finalizers)
+# ---------------------------------------------------------------------------
+log_info "Step 3: Cleaning up operator namespaces..."
+
+_force_delete_ns openshift-builds &
```
kevincogan, line 182: `openshift-builds` and `zero-trust-workload-identity-manager` may be managed by other operators on a shared cluster. Force-deleting them would break other teams' workloads. The cleanup should only delete namespaces that the setup script created, or at minimum prompt before touching shared operator namespaces.

scripts/ocp/setup-kagenti.sh
```diff
@@ -0,0 +1,787 @@
...
+MINS=$(( ELAPSED / 60 ))
+SECS=$(( ELAPSED % 60 ))
+
+echo ""
+echo "============================================"
+echo "  Kagenti platform is ready!  (Time elapsed:${MINS}m ${SECS}s)"
+echo ""
+echo "  Namespace registration:"
+echo "    Namespaces self-register for webhook injection via label:"
+echo "      kubectl label namespace <ns> kagenti-enabled=true"
```
kevincogan: I would be interested to see if you have this problem as currently I believe labeling the namespace isn't enough on its own. Agent namespaces also need ConfigMaps (`environments`, `authbridge-config`, `envoy-config`, `spiffe-helper-config`), a Keycloak credentials Secret, SCC RoleBindings, and the `kagenti-authbridge` SCC groups patched. This script installs with no agent namespaces, and there's currently no automated provisioning path for namespaces created after install.

scripts/ocp/cleanup-kagenti.sh
ChristianZaccaria: Nice catch, thanks! I've switched to use just `python3` and is now a pre-flight check on both cleanup and setup scripts.

scripts/ocp/cleanup-kagenti.sh
ChristianZaccaria: I agree. We would have to find a way to re-use these namespaces while not erroring out on a fresh deployment. I'll have a further look into this. Worst case I think it's okay as a follow-up.

scripts/ocp/setup-kagenti.sh
ChristianZaccaria: This is a great catch! For some context, I brought this script from the openclaw-infra repo. We've decided to move the source of truth of these scripts to Kagenti repo, and these comments must've slipped here too. I would be in favour of replacing those comments for something like:

>Before deploying Kagenti, add your namespace to the [agentNamespaces](https://github.com/kagenti/kagenti/blob/main/charts/kagenti/values.yaml#L53-L55) list in values.yaml so the platform provisions the required resources in your namespace.

PR #1102, opened 2026-03-26
🐛 Allow MLFlow to accept traffic on any port under Kagenti

### LINKED ISSUE #1101: :bug: MLFlow UI not available
### Describe the bug

When MLFlow is installed as part of Kagenti, http://mlflow.localtest.me:8080 displays

```
Invalid Host header - possible DNS rebinding attack detected
```

... instead of the MLFlow UI.

### Steps To Reproduce

1. Create a config file to deploy MLFlow:

```bash
cat <<EOF > /tmp/enable-mlflow.yaml
charts:
  kagenti-deps:
    values:
      components:
        mlflow:
          enabled: true
  kagenti:
    values:
      components:
        mlflow:
          enabled: true
EOF
```

2. Deploy Kagenti using `deployments/ansible/run-install.sh --env dev --env-file /tmp/enable-mlflow.yaml`
3. `curl -v http://mlflow.localtest.me:8080/`

### Expected Behavior

MLFlow should be available.

### Additional Context

cc @r3v5 @evaline-ju

esnible:
## Summary

Resolves #1101 by allowing MLFlow to serve its UI on any port when installed on _kind_.

When Kagenti is installed on _kind_, extraPortMappings are created so that http://*.localtest.me:8080/ routes to the various Kagenti components.  However, MLFlow was configured to only allow port 80, which does not work with Kind.

This relaxes MLFlow's rules to allow it to accept traffic on most ports.  This seemed better than supporting only :80 and :8080.  This may also help with port-forward.

Follow the instructions in #1101 and verify that you may visit or _curl_ http://mlflow.localtest.me:8080/ after installing using this PR.

Do `oc -n kagenti-system port-forward deployment/mlflow 6000:5000` and verify you can `curl localhost:6000`

Fixes #

#1101

### DIFF
--- a/charts/kagenti-deps/templates/mlflow.yaml
+++ b/charts/kagenti-deps/templates/mlflow.yaml
@@ -154,7 +154,8 @@ spec:
             ALLOWED_HOSTS="$ROUTE_HOST,$ALLOWED_HOSTS"
           fi
           {{- else }}
-          ALLOWED_HOSTS="mlflow.{{ .Values.domain | default "localtest.me" }}:*,mlflow.{{ .Values.domain | default "localtest.me" }},mlflow,mlflow:*,mlflow.{{ .Release.Namespace }},mlflow.{{ .Release.Namespace }}:5000,mlflow.{{ .Release.Namespace }}.svc.cluster.local,mlflow.{{ .Release.Namespace }}.svc.cluster.local:5000,localhost,localhost:*,127.0.0.1,127.0.0.1:*"
+          ALLOWED_HOSTS="mlflow.{{ .Values.domain | default "localtest.me" }}:*,mlflow:*,mlflow.{{ .Release.Namespace }}:5000,mlflow.{{ .Release.Namespace }}.svc.cluster.local:5000,localhost:*,127.0.0.1:*"
+          ALLOWED_HOSTS="$ALLOWED_HOSTS,mlflow.{{ .Values.domain | default "localtest.me" }},mlflow,mlflow.{{ .Release.Namespace }},mlflow.{{ .Release.Namespace }}.svc.cluster.local,localhost,127.0.0.1"
           {{- end }}
           {{- if .Values.mlflow.auth.enabled }}
           # Set MLflow config via environment variables (used by mlflow_oidc_auth.app)


charts/kagenti-deps/templates/mlflow.yaml
```diff
@@ -154,7 +154,7 @@ spec:
             ALLOWED_HOSTS="$ROUTE_HOST,$ALLOWED_HOSTS"
           fi
           {{- else }}
-          ALLOWED_HOSTS="mlflow.{{ .Values.domain | default "localtest.me" }},mlflow,mlflow:5000,mlflow.{{ .Release.Namespace }},mlflow.{{ .Release.Namespace }}:5000,mlflow.{{ .Release.Namespace }}.svc.cluster.local,mlflow.{{ .Release.Namespace }}.svc.cluster.local:5000,localhost,127.0.0.1"
+          ALLOWED_HOSTS="mlflow.{{ .Values.domain | default "localtest.me" }}:*,mlflow.{{ .Values.domain | default "localtest.me" }},mlflow,mlflow:*,mlflow.{{ .Release.Namespace }},mlflow.{{ .Release.Namespace }}:5000,mlflow.{{ .Release.Namespace }}.svc.cluster.local,mlflow.{{ .Release.Namespace }}.svc.cluster.local:5000,localhost,localhost:*,127.0.0.1,127.0.0.1:*"
```
Alan-Cha: **Suggestion**: The wildcard port approach (`:*`) is pragmatic and solves the Kind port mapping issue. However, consider if there's a security trade-off here - are we allowing access on ports that shouldn't be exposed? In practice, this is likely fine since the hostnames are still restricted to known-good values (localhost, internal service names, configured domain). The alternative would be to enumerate common ports (80, 443, 5000, 8080, 6000) but that's more brittle.

**Nit**: The line is now very long (300+ chars). Consider formatting as a multi-line array for readability if feasible.

esnible: @Alan-Cha , the idea here is that I could hard-code this to :8080 for Kagenti-on-Kind, but chose to allow any port in case someone installs Kagenti on a cluster that uses different extraPortMappings.  Doing this doesn't really provide security -- the browser sets the Host header, and it can lie.  This doesn't expose anything that wasn't already exposed.

The motivation behind this is that the automation for installing MLFlow on Kagenti was probably never tested on Kind -- it certainly didn't work with the :8080 that Kagenti-on-Kind uses.

PR #1115, opened 2026-03-31
fix: remove duplicate keycloak-js ignore in dependabot.yml

Ladas:
## Summary

Remove duplicate `keycloak-js` ignore entry that causes Dependabot config validation to fail on **all PRs**.

Error:
```
The property '#/updates/3/ignore/1/update-types' includes a duplicate.
Ignores must have a unique combination of 'dependency-name' and 'update-types'
```

1 file, 4 lines deleted.


pdettori (approved):
/lgtm

PR #1122, opened 2026-03-31
Feat: Create two demo users when creating a Keycloak realm during Kagenti install

### LINKED ISSUE #1120: feature: Create demo users when installing Kagenti and Keycloak
### Feature Description

I would prefer not to log in as Admin when demoing Kagenti.

I would like there to be one or two additional users, perhaps named "demo1" and "demo2".

I also dislike when the default Admin password gets me a scold from Keycloak because "password" is a bad password.  It would be ideal if the defaults for the demo accounts do not trigger scolding.

### Proposed Solution

When Keycloak is configured by Kagenti installation, (optionally?) create some sample users.

### Want to contribute?

### Additional Context

_No response_

esnible:
## Summary

Resolves #1120

## Testing Instructions

Build image with

```bash
docker build --load -t ghcr.io/kagenti/kagenti/ui-oauth-secret:ecs-test -f kagenti/auth/ui-oauth-secret/Dockerfile kagenti/
```

Create a variables override file:

```bash
cat <<EOF > /tmp/custom-ui-oauth-secret-image.yaml
charts:
  kagenti:
    values:
      uiOAuthSecret:
        tag: ecs-test
        imagePullPolicy: Never
EOF
```

Recreate your Kind cluster and load the test image:

```bash
kind delete cluster --name kagenti && \
  kind create cluster --name kagenti --config deployments/ansible/kind/kind-config-registry.yaml && \
  kind load docker-image ghcr.io/kagenti/kagenti/ui-oauth-secret:ecs-test --name kagenti
```

Redeploy Kagenti:

```
deployments/ansible/run-install.sh --env dev --env-file /tmp/custom-ui-oauth-secret-image.yaml
```

### DIFF
--- a/kagenti/auth/ui-oauth-secret/auth_secret.py
+++ b/kagenti/auth/ui-oauth-secret/auth_secret.py
@@ -33,7 +33,7 @@ DEFAULT_ADMIN_PASSWORD_KEY = "password"
 OAUTH_REDIRECT_PATH = "/"
 OAUTH_SCOPE = "openid profile email"
 SERVICE_ACCOUNT_CA_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
-DEFAULT_DEMO_USERS = ("demo1", "demo2")
+DEFAULT_DEMO_USERS = ("bob", "alice")
 DEFAULT_DEMO_PASSWORDS = ("kagenti1", "kagenti2")
 DEFAULT_DEMO_PASSWORD_OVERRIDE_ENV_VARS = (
     "KEYCLOAK_DEMO1_PASSWORD",


kagenti/auth/ui-oauth-secret/auth_secret.py
```diff
@@ -413,6 +418,43 @@ def main() -> None:
                     f"cannot proceed"
                 )
 
+            # Create demo users for testing.
+            # These get no realm-admin client role.
+            for demo_username, demo_password in zip(
+                DEFAULT_DEMO_USERS, DEFAULT_DEMO_PASSWORDS
+            ):
+                try:
+                    existing = keycloak_admin.get_users({"username": demo_username})
```
pdettori: **must-fix**: `get_users({"username": demo_username})` does a **substring/prefix search** in Keycloak's Admin REST API, not an exact match. Searching for `"demo1"` could return `"demo10"`, `"demo1-test"`, etc., causing the `if not existing` check to skip creation incorrectly.

Should use exact matching:
```python
existing = keycloak_admin.get_users({"username": demo_username, "exact": True})
```

This is supported in Keycloak 15+ (well within our version range).

kagenti/auth/ui-oauth-secret/auth_secret.py
```diff
@@ -33,6 +33,11 @@
 OAUTH_REDIRECT_PATH = "/"
 OAUTH_SCOPE = "openid profile email"
 SERVICE_ACCOUNT_CA_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+DEFAULT_DEMO_USERS = ("demo1", "demo2")
+DEFAULT_DEMO_PASSWORDS = (
```
pdettori: **suggestion** (non-blocking): `DEFAULT_DEMO_PASSWORDS` calls `get_optional_env()` at module level (import time). This works fine for a standalone script, but could cause issues if the module is ever imported for testing. Consider moving the env var lookups inside `main()` or documenting the intent.

Low priority since the existing file already has module-level env lookups.

kagenti/auth/ui-oauth-secret/auth_secret.py
esnible: Great catch!  @mrsabath , this was based on your code https://github.com/kagenti/kagenti/blame/main/kagenti/auth/ui-oauth-secret/auth_secret.py#L309 and https://github.com/kagenti/kagenti/blame/main/kagenti/auth/ui-oauth-secret/auth_secret.py#L406.  Is inexact user matching also an issue there?

esnible: @pdettori Fixed; please re-review.

kagenti/auth/ui-oauth-secret/auth_secret.py
```diff
@@ -33,6 +33,12 @@
 OAUTH_REDIRECT_PATH = "/"
 OAUTH_SCOPE = "openid profile email"
 SERVICE_ACCOUNT_CA_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+DEFAULT_DEMO_USERS = ("demo1", "demo2")
+DEFAULT_DEMO_PASSWORDS = ("kagenti1", "kagenti2")
+DEFAULT_DEMO_PASSWORD_OVERRIDE_ENV_VARS = (
```
mrsabath, line 38: suggestion: The parallel tuples `DEFAULT_DEMO_USERS`, `DEFAULT_DEMO_PASSWORDS`, and `DEFAULT_DEMO_PASSWORD_OVERRIDE_ENV_VARS` rely on index alignment. A list of dicts/namedtuples would be slightly more maintainable if more demo users are added later. Fine for 2 users though.

kagenti/auth/ui-oauth-secret/auth_secret.py
```diff
@@ -413,6 +419,46 @@ def main() -> None:
                     f"cannot proceed"
                 )
 
+            # Create demo users for testing.
+            # These get no realm-admin client role.
+            for i, demo_username in enumerate(DEFAULT_DEMO_USERS):
+                try:
+                    existing = keycloak_admin.get_users(
+                        {"username": demo_username, "exact": True}
```
mrsabath, line 427: 👍 Good use of `{"exact": True}` — correctly addresses the previous review feedback about substring matches.

kagenti/auth/ui-oauth-secret/auth_secret.py
```diff
@@ -413,6 +419,46 @@ def main() -> None:
...
+                        )
+                        logger.info(
+                            f"Created demo user '{demo_username}' "
+                            f"in realm '{keycloak_realm}'"
+                        )
+                    else:
+                        logger.info(
+                            f"Demo user '{demo_username}' already exists "
+                            f"in realm '{keycloak_realm}', skipping"
+                        )
```
mrsabath, line 458: nit: The `except Exception` catch-all is reasonable here since demo user creation is non-critical, but a more specific `KeycloakError` would avoid masking unexpected errors during development.

mrsabath (approved):
## Review Summary

Clean implementation that adds demo user provisioning (`demo1`/`demo2`) to the Keycloak bootstrap flow. The code correctly handles idempotency (checks before creating), uses exact username matching (addressing the previous review), and allows password overrides via env vars. No security blockers — default passwords are appropriate for demo/dev context.

The Deploy & Test CI failure is at "Start port-forward for agent service" — unrelated infrastructure flake.

**Areas reviewed**: Python, Security, Commit conventions, PR format
**Commits**: 6 commits, all signed-off ✓
**CI status**: Deploy & Test failing (unrelated flake), all other checks pass ✓
**Prior review feedback**: Addressed in `2bd07fd` (exact matching)

Minor suggestions below, none blocking.

kagenti/auth/ui-oauth-secret/auth_secret.py
esnible: My PR follows the pattern of logging errors rather than failing that @mrsabath introduced.  We should introduce a new failure type as part of this PR.

PR #1127, opened 2026-04-01
feat(ocp): MLflow DSC awareness in setup and cleanup scripts

### LINKED ISSUE #1124: feat(ocp): MLflow DSC awareness in setup and cleanup scripts
## Why

On OpenShift, Kagenti uses the MLflow instance managed by RHOAI's Data Science Cluster (DSC) rather than deploying its own. This is intentional — it lets us tell the full platform story where RHOAI/DSC is the authoritative source for MLflow, and Kagenti integrates with it rather than duplicating it.

## Problem

`setup-kagenti.sh` and `cleanup-kagenti.sh` have no awareness of MLflow being managed by DSC.

## Proposed Changes

**`setup-kagenti.sh`** — when MLflow is not disabled, add a pre-flight check that verifies MLflow is deployed and ready via DSC. Fail fast with a clear error if not. Skip the check entirely when MLflow is disabled (current default — no behaviour change).

**`cleanup-kagenti.sh`** — before removing MLflow resources (cleaned up indirectly via `helm uninstall kagenti` and `kagenti-system` namespace deletion), check if MLflow is owned by DSC. If so, skip removal. Otherwise, proceed as today.

Bobbins228:
Closes #1124

RHAIENG-4191

## Summary
On OpenShift, Kagenti should use the MLflow instance managed by RHOAI's Data Science Cluster rather than deploying its own. This PR adds a `--mlflow` flag to `setup-kagenti.sh` that enables DSC-integrated MLflow with a pre-flight check, and makes `cleanup-kagenti.sh` DSC-aware so it won't attempt to remove RHOAI-managed MLflow.

## Changes
- `setup-kagenti.sh`: add `--mlflow` flag (default: disabled — no behaviour change when omitted)
- `setup-kagenti.sh`: add `_check_dsc_mlflow_ready()` pre-flight that verifies `mlflowoperator.managementState=Managed` in the DSC, confirms the operator pod is running, detects any deployed `MLflow` tracking server CRs, and wires the in-cluster service URL into the helm install as `mlflow.trackingUri`
- `setup-kagenti.sh`: refactor MLflow helm flags into `MLFLOW_HELM_FLAGS` array so `mlflow.trackingUri` is injected conditionally
- `cleanup-kagenti.sh`: check `mlflowoperator.managementState` before Step 1 and log that RHOAI-managed MLflow will not be removed
- Kagenti's own MLflow container remains disabled (`components.mlflow.enabled=false`) in all code paths

## Test Plan
- [ ] `setup-kagenti.sh --dry-run --mlflow` prints `[dry-run] check DataScienceCluster MLflow readiness` and helm args include `components.mlflow.enabled=false`
- [ ] `setup-kagenti.sh --mlflow` on a cluster with `mlflowoperator.managementState=Managed` passes pre-flight and does not deploy a Kagenti-owned MLflow pod
- [ ] `setup-kagenti.sh --mlflow` on a cluster without DSC fails fast with a clear error message
- [ ] `cleanup-kagenti.sh` on a cluster with DSC-managed MLflow prints the preservation notice before Step 1
- [ ] Running without `--mlflow` is identical to previous behaviour

### DIFF
--- a/scripts/ocp/cleanup-kagenti.sh
+++ b/scripts/ocp/cleanup-kagenti.sh
@@ -100,6 +100,16 @@ fi
 START_SECONDS=$SECONDS
 
 # ---------------------------------------------------------------------------
+# MLflow DSC check — RHOAI-managed MLflow is not removed by this cleanup
+# ---------------------------------------------------------------------------
+_dsc_mlflow_state=$($KUBECTL get datasciencecluster \
+  -o jsonpath='{.items[0].spec.components.mlflowoperator.managementState}' 2>/dev/null || echo "")
+if [ "$_dsc_mlflow_state" = "Managed" ]; then
+  log_info "MLflow is Managed by RHOAI DSC — will not be removed by this cleanup"
+  echo ""
+fi
+
+# ---------------------------------------------------------------------------
 # Helpers
 # ---------------------------------------------------------------------------
 
--- a/scripts/ocp/setup-kagenti.sh
+++ b/scripts/ocp/setup-kagenti.sh
@@ -5,7 +5,7 @@
 # Installs the Kagenti stack (SPIRE, cert-manager, Keycloak, operator, webhook,
 # MCP Gateway) on an OpenShift cluster. Run this BEFORE setup.sh --with-a2a.
 # Prometheus/Kiali are disabled. UI/backend installed by default (use --skip-ui to disable).
-# MLflow/Shipwright disabled (not needed).
+# MLflow/Shipwright disabled by default. Use --mlflow to enable MLflow via RHOAI DSC.
 #
 # Usage:
 #   ./scripts/ocp/setup-kagenti.sh                              # Auto-clones kagenti main to ~/.cache/kagenti
@@ -14,6 +14,7 @@
 #   ./scripts/ocp/setup-kagenti.sh --realm nerc                 # Custom Keycloak realm (default: kagenti)
 #   ./scripts/ocp/setup-kagenti.sh --skip-ovn-patch             # Skip OVN gateway patch
 #   ./scripts/ocp/setup-kagenti.sh --skip-mcp-gateway           # Skip MCP Gateway install
+#   ./scripts/ocp/setup-kagenti.sh --mlflow                     # Enable MLflow via RHOAI DSC (pre-flight check required)
 #
 # Prerequisites:
 #   - oc / kubectl with cluster-admin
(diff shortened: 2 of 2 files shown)


scripts/ocp/cleanup-kagenti.sh
```diff
@@ -99,6 +99,16 @@ fi
...
 
+# ---------------------------------------------------------------------------
+# MLflow DSC check — RHOAI-managed MLflow is not removed by this cleanup
+# ---------------------------------------------------------------------------
+_dsc_mlflow_state=$($KUBECTL get datasciencecluster \
+  -o jsonpath='{.items[0].spec.components.mlflowoperator.managementState}' 2>/dev/null || echo "")
+if [ "$_dsc_mlflow_state" = "Managed" ]; then
+  log_info "MLflow is Managed by RHOAI DSC — will not be removed by this cleanup"
+  echo ""
+fi
```
ChristianZaccaria, line 110: Now that I think about it, I think adding this in the cleanup script is redundant. The script would try to remove everything from the kagenti-system namespace, so it would never touch the MLFlow-Operator in the redhat-data-services namespace. This file can remain untouched.

scripts/ocp/setup-kagenti.sh
```diff
@@ -14,6 +14,7 @@
 #   ./scripts/ocp/setup-kagenti.sh --realm nerc                 # Custom Keycloak realm (default: kagenti)
 #   ./scripts/ocp/setup-kagenti.sh --skip-ovn-patch             # Skip OVN gateway patch
 #   ./scripts/ocp/setup-kagenti.sh --skip-mcp-gateway           # Skip MCP Gateway install
+#   ./scripts/ocp/setup-kagenti.sh --mlflow                     # Enable MLflow via RHOAI DSC (pre-flight check required)
```
ChristianZaccaria, line 17: It's best if the flag was more explicit in behavior. This way we don't confuse this with Kagenti's mlflow. Maybe:

`--rhoai-mlflow     # Provision mlflow via RHOAI`

ChristianZaccaria: Perhaps also the "pre-flight check required" could say instead: "Enable MLflow via RHOAI DSC (requires Managed mlflowoperator in DSC; default: disabled)".

scripts/ocp/setup-kagenti.sh
```diff
@@ -61,6 +65,7 @@ while [[ $# -gt 0 ]]; do
     --skip-ovn-patch)     SKIP_OVN_PATCH=true; shift ;;
     --skip-mcp-gateway)   SKIP_MCP_GATEWAY=true; shift ;;
     --skip-ui)            SKIP_UI=true; shift ;;
+    --mlflow)             DISABLE_MLFLOW=false; shift ;;
```
ChristianZaccaria, line 68:
```suggestion
    --rhoai-mlflow)             PROVISION_MLFLOW=false; shift ;;
```

PR #1133, opened 2026-04-02
fix: suppress pylint false positives for feature-flag conditional imports

Ladas:
## Summary

- Adds targeted `pylint: disable` comments for `used-before-assignment`, `wrong-import-position`, `no-name-in-module`, and `import-error` false positives in `app/main.py`
- The feature flag pattern uses `try/except` blocks with boolean guards (`_*_modules_loaded`), which pylint cannot statically verify
- This fixes `make lint` failures on **any PR that touches Python files** (blocking all sandbox streamer PRs)

## Test plan

- [x] `uv run pylint app/main.py` → 10.00/10
- [x] `uv run pylint app/` → EXIT:0
- [x] Pre-commit hooks pass

### DIFF
--- a/kagenti/backend/app/main.py
+++ b/kagenti/backend/app/main.py
@@ -30,8 +30,8 @@ class NoCacheMiddleware(BaseHTTPMiddleware):
         return response
 
 
-from app.core.config import settings
-from app.routers import (
+from app.core.config import settings  # pylint: disable=wrong-import-position
+from app.routers import (  # pylint: disable=wrong-import-position
     agents,
     tools,
     namespaces,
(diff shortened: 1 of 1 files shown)


kagenti/backend/app/main.py
```diff
@@ -179,15 +183,14 @@ async def lifespan(app: FastAPI):
     app.include_router(llm_keys.router, prefix="/api/v1")
```
pdettori, line 183: **nit**: The old per-section comments (`# Feature-flagged routers — triggers`, `# ... — integrations`) were removed. The new block comment explains the pylint rationale well, but losing the section markers slightly reduces scannability. Consider keeping them as one-liners. Minor — not blocking.

pdettori (approved):
Focused, correct lint fix. The `try/except` + boolean guard pattern genuinely confuses pylint's static analysis, so targeted disables are the right approach. Good use of scoped `disable`/`enable` pairs rather than file-level suppression. All CI green.

**Areas reviewed**: Python (lint suppression)
**Commits**: 1 commit, signed-off
**CI status**: all passing

PR #1140, opened 2026-04-02
Bug fix: Use correct location for 30-build-dep-image.sh

### LINKED ISSUE #1139: :bug: .github/scripts/lib/30-build-dep-image.sh: No such file or directory
### Describe the bug

After I run `./.github/scripts/local-setup/kind-full-test.sh --skip-cluster-destroy`, it installs Kagenti, but it displays

```
▶ Building dependency overrides from source...
Running locally
Detected OS: macOS
Detected cluster type: Kubernetes (Kind/vanilla)
[Wave 31] Building dependencies from custom refs
ℹ Building kagenti-webhook from kagenti/kagenti-extensions@main
bash: /Users/snible/src/kagenti/.github/scripts/lib/30-build-dep-image.sh: No such file or directory
```

The _30-build-dep-image.sh_ file is in a different directory -- it is in _.github/scripts/common/30-build-dep-image.sh_.

### Steps To Reproduce

1. `./.github/scripts/local-setup/kind-full-test.sh --skip-cluster-destroy`

### Expected Behavior

No warnings about files.

### Additional Context

The message " Building kagenti-webhook from kagenti/kagenti-extensions@main" seems to come from _.github/scripts/common/31-build-deps-from-refs.sh_ which correctly uses a relative path.  I don't understand why the original error message has "scripts/**lib**" not "scripts/**common**".

esnible:
## Summary

Resolves #1139 by recalculating the directory path after it is corrupted by a `source`d script.

- Create a Kind cluster
- Do `./.github/scripts/local-setup/kind-full-test.sh --skip-cluster-destroy`
- Verify the above does not print ".github/scripts/lib/30-build-dep-image.sh: No such file or directory"

### DIFF
--- a/.github/scripts/common/31-build-deps-from-refs.sh
+++ b/.github/scripts/common/31-build-deps-from-refs.sh
@@ -14,8 +14,8 @@ set -euo pipefail
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 source "$SCRIPT_DIR/../lib/env-detect.sh"
 source "$SCRIPT_DIR/../lib/logging.sh"
-# Note that we must set SCRIPT_DIR again *after* running the above scripts,
-# as they override SCRIPT_DIR.
+# Note that we set SCRIPT_DIR again *after* running the above scripts,
+# as env-detect.sh overrides SCRIPT_DIR.
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
 # No hardcoded defaults — chart deps are up-to-date (webhook v0.4.0-alpha.9).


.github/scripts/common/31-build-deps-from-refs.sh
```diff
@@ -14,6 +14,9 @@ set -euo pipefail
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 source "$SCRIPT_DIR/../lib/env-detect.sh"
 source "$SCRIPT_DIR/../lib/logging.sh"
+# Note that we set SCRIPT_DIR again *after* running the above scripts,
+# as env-detect.sh overrides SCRIPT_DIR.
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
```
pdettori, line 19: **suggestion**: The fix is correct but treats the symptom. The root cause is that `env-detect.sh` (line 20) unconditionally sets `SCRIPT_DIR` to its own directory, clobbering the caller's value. A more robust fix would be for `env-detect.sh` to use a local variable (e.g., `_ED_SCRIPT_DIR`) or for callers to save/restore.

That said, this fix is pragmatic and unblocks the issue — fine to merge as-is and address the root cause separately.

.github/scripts/common/31-build-deps-from-refs.sh
```diff
@@ -14,6 +14,9 @@ set -euo pipefail
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 source "$SCRIPT_DIR/../lib/env-detect.sh"
 source "$SCRIPT_DIR/../lib/logging.sh"
+# Note that we set SCRIPT_DIR again *after* running the above scripts,
```
pdettori, line 17: **nit**: Consider making the comment slightly more specific:
```
# env-detect.sh unconditionally sets SCRIPT_DIR to its own directory;
# reset it to ours so 30-build-dep-image.sh resolves correctly.
```

PR #1145, opened 2026-04-03
Fix: :bug: Wrong count of imported variables in "Import Environment Variables"

### LINKED ISSUE #1142: :bug: Wrong count of imported variables in "Import Environment Variables"
### Describe the bug

The displayed count, e.g. " Preview (8 variables)", does not match the number of rows displayed below.

### Steps To Reproduce

1. Go to http://kagenti-ui.localtest.me:8080/agents/import
2. Deploy from Existing Image, 
3. image ghcr.io/kagenti/agent-examples/git_issue_agent
4. tag v0.1.0-alpha.8
5. Env vars https://raw.githubusercontent.com/kagenti/agent-examples/refs/heads/main/a2a/git_issue_agent/.env.ollama

The UI claims there are 8, but only displays 7, env variables.

### Expected Behavior

`curl https://raw.githubusercontent.com/kagenti/agent-examples/refs/heads/main/a2a/git_issue_agent/.env.ollama` shows 7 lines of vars, thus Kagenti should claim 7, not 8.

### Additional Context

_No response_

gabrnavarro:
## What changed

Fixed incorrect variable count in the "Import Environment Variables" preview. The backend `parse_env_file` function now uses `splitlines()` instead of `strip().split("\n")` to avoid count mismatches caused by mixed line endings or trailing newlines. Also added inline comment stripping (`KEY=value # comment` → one variable, not two). Formatted test file with `ruff format` per project conventions.

## Why

The UI was showing a count (e.g. "Preview (8 variables)") that didn't match the number of rows rendered below it. Root cause: `.strip().split("\n")` on content with Windows-style `\r\n` line endings produced `\r`-suffixed tokens that could inflate or mangle parsed output. Switching to `splitlines()` handles all line-ending variants correctly.

Fixes #1142

## Testing

- Added `kagenti/backend/tests/test_parse_env.py` with 12 regression tests covering: exact count for the 7-variable `.env.ollama` file from the issue, blank/comment-only files, inline comments, Windows line endings (`\r\n`), URLs with `=` in query params, `valueFrom` JSON values, and trailing newlines.
- All 12 tests pass: `uv run pytest tests/test_parse_env.py -v`
- `ruff format --check .` passes (76 files already formatted after fixing test file formatting)

### DIFF
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -2999,13 +2999,17 @@ async def parse_env_file(request: ParseEnvRequest) -> ParseEnvResponse:
     env_vars = []
     warnings = []
 
-    lines = request.content.strip().split("\n")
+    lines = request.content.splitlines()
 
     for line_num, line in enumerate(lines, 1):
         # Skip empty lines and comments
         line = line.strip()
         if not line or line.startswith("#"):
             continue
+
+        # Strip inline comments (e.g. KEY=value # comment)
+        if " #" in line:
+            line = line[: line.index(" #")].rstrip()
 
         # Parse KEY=VALUE
         if "=" not in line:
--- a/kagenti/backend/tests/test_parse_env.py
+++ b/kagenti/backend/tests/test_parse_env.py
@@ -95,14 +95,7 @@ class TestParseEnvCount:
 
     def test_comments_and_blank_lines_not_counted(self, client):
         """Comments and blank lines must not inflate the variable count."""
-        content = (
-            "# This is a comment\n"
-            "\n"
-            "KEY1=value1\n"
-            "# another comment\n"
-            "\n"
-            "KEY2=value2\n"
-        )
+        content = "# This is a comment\n\nKEY1=value1\n# another comment\n\nKEY2=value2\n"
         result = parse(client, content)
         assert len(result["envVars"]) == 2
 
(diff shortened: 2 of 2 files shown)


kagenti/backend/app/routers/agents.py
```diff
@@ -2999,14 +2999,18 @@ async def parse_env_file(request: ParseEnvRequest) -> ParseEnvResponse:
...
 
     for line_num, line in enumerate(lines, 1):
         # Skip empty lines and comments
         line = line.strip()
         if not line or line.startswith("#"):
             continue
 
+        # Strip inline comments (e.g. KEY=value # comment)
+        if " #" in line:
+            line = line[: line.index(" #")].rstrip()
```
evaline-ju, line 3012: While this works for the examples in the issue, a small edge case is this may incorrectly strip # from quoted values, e.g. `KEY="http://example.com/#param` would become `KEY="http://example.com/`

evaline-ju (commented):
Thanks @gabrnavarro for the bugfix contribution! 

I see you've noted running `ruff check`. The lint CI is failing at `ruff format --check .`, and should be included if pre-commit is run. (ref https://github.com/kagenti/kagenti/blob/main/docs/dev-guide.md#pre-commit)

Otherwise most of this looks good. I've noted a small edge case comment

gabrnavarro:
Got it, thanks for catching that. I'll run `ruff format .` to fix the formatting issues before the next push.

kagenti/backend/app/routers/agents.py
pdettori, line 3012: The `" #"` edge case flagged by @evaline-ju is still present — e.g. `KEY="http://example.com/#param"` would be truncated to `KEY="http://example.com/`. Consider guarding with a quote-aware check (only strip `" #"` if it appears outside quotes), or track as a follow-up issue to keep this PR scoped to the original fix.

pdettori (commented):
Good fix — `splitlines()` correctly handles mixed line endings and trailing newlines, directly resolving #1142. The `ruff format` issue from the first review round has been addressed (commit c2c7563). All lint, security, and backend-test checks pass.

**E2E CI failure is unrelated** — `test_agent_simple_query` failed due to a flaky MCP weather tool connectivity issue in the Kind cluster (`Cannot connect to MCP weather service`). This test does not exercise `parse_env_file()`. Result: 1 failed, 35 passed, 18 skipped.

One open item from @evaline-ju's review: the inline comment stripping (`" #"` naive match) still has the quoted-value edge case. See inline comment.

PR #1170, opened 2026-04-08
feat: Wire OTEL collector to export traces to RHOAI MLflow

ChristianZaccaria:
## Summary
Enables the OTEL collector to export agent traces to a RHOAI-managed MLflow instance using ServiceAccount token authentication, OpenShift service-serving CA for TLS, and workspace/experiment headers. Also improves deployment readiness checks.

## Changes
- Add `otel.mlflow.workspace`/`experimentId` values and `rhoaiMlflowAuthConfig` (bearertokenauth) preset to kagenti-deps chart
- Inject SA bearer token auth, `x-mlflow-workspace`/`x-mlflow-experiment-id` headers, and service-CA TLS on the MLflow exporter when `otel.mlflow.enabled && openshift`
- Add `otel-service-ca` ConfigMap with service-CA annotation for RHOAI MLflow TLS verification
- Setup script provisions MLflow CR via RHOAI DSC, builds temp values file, and creates `mlflow-operator-mlflow-integration` RoleBindings for otel-collector SA in agent namespaces
- Improve `_wait_kagenti_deps_ready` with per-component readiness checks (RHBK operator, PostgreSQL, Keycloak, cert-manager, Istio)

## Test Plan
- [ ] Deploy with `./scripts/ocp/setup-kagenti.sh` on OCP cluster with RHOAI DSC mlflowoperator managed
- [ ] Send a request to an agent.
- [ ] Check agent service and otel-collector pod logs:
```
weather-service pod logs:
DEBUG:urllib3.connectionpool:http://otel-collector.kagenti-system.svc.cluster.local:8335 "POST /v1/traces HTTP/1.1" 200 2
```

Confirming the collector received the traces:
```
otel-collector pod logs:
{"otelcol.component.id": "debug", "otelcol.component.kind": "Exporter", "otelcol.signal": "traces"}
```

- [ ] Verify traces appear in MLflow UI after sending a request to an agent:
image
image


scripts/ocp/setup-kagenti.sh
```diff
@@ -229,6 +238,154 @@ export DOMAIN
...
+    log_success "RoleBinding otel-collector-mlflow created in $ns"
+  done <<< "$agent_ns"
+}
+
+log_info "Step 2.5: MLflow DSC preflight + provisioning"
+_mlflow_check_dsc
+_mlflow_create_cr
+_mlflow_wait_ready
+echo ""
+
```
pdettori, line 392: **must-fix**: Step 2.5 MLflow provisioning runs unconditionally — `_mlflow_check_dsc` will `exit 1` on any OCP cluster without RHOAI `mlflowoperator=Managed`. This makes RHOAI a hard requirement for all OCP deployments using this script.

Consider:
- Adding a `--skip-mlflow` flag (or `--with-mlflow`), or
- Auto-detecting RHOAI availability and skipping gracefully if absent

Otherwise anyone running `setup-kagenti.sh` on a cluster without RHOAI will hit a hard failure here.

charts/kagenti-deps/templates/_helpers.tpl
```diff
@@ -81,20 +81,38 @@ not concatenated (e.g. service.extensions).
...
 {{- end -}}
-{{- if and $.Values.components.mlflow.enabled $.Values.mlflow.auth.enabled $.Values.otel.collector.mlflowAuthConfig -}}
+{{- if and $mlflowEnabled $.Values.mlflow.auth.enabled $.Values.otel.collector.mlflowAuthConfig -}}
 {{- $config = mustMergeOverwrite $config (deepCopy $.Values.otel.collector.mlflowAuthConfig) -}}
 {{- end -}}
 {{- if and $.Values.openshift (get (get ($config) "extensions" | default dict) "oauth2client/mlflow") -}}
 {{- $_ := set (index $config "extensions" "oauth2client/mlflow") "tls" (dict "ca_file" "/etc/pki/ingress-ca/ingress-ca.pem") -}}
 {{- end -}}
-{{- if and $.Values.components.mlflow.enabled $.Values.mlflow.auth.enabled -}}
+{{- $rhoaiMlflow := ($.Values.otel.mlflow).enabled | default false -}}
```
pdettori, line 98: **suggestion**: Dual-mode auth conflict — if someone sets both `otel.mlflow.enabled=true` AND `mlflow.auth.enabled=true`, this block (line 98) sets `auth.authenticator` to `oauth2client/mlflow`, overwriting the `bearertokenauth/mlflow` set at line 91 by the RHOAI block above.

The setup script avoids this by explicitly setting `mlflow.auth.enabled=false`, but the template itself is unguarded. Consider wrapping this block with `{{- if not $rhoaiMlflow -}}` so the two auth modes are mutually exclusive at the template level.

scripts/ocp/setup-kagenti.sh
```diff
@@ -712,32 +886,10 @@ run_cmd helm upgrade --install kagenti "$KAGENTI_REPO/charts/kagenti/" \
...
-  # Restart backend so it picks up the OAuth secret (the secret may have been
-  # created after the backend pod started with optional: true env vars)
-  log_info "Restarting backend to pick up OAuth secret..."
-  $KUBECTL rollout restart deployment/kagenti-backend -n kagenti-system 2>/dev/null || true
-  $KUBECTL rollout status deployment/kagenti-backend -n kagenti-system --timeout=120s 2>/dev/null || true
-fi
-
 log_success "Kagenti installed"
+
+# Grant otel-collector SA MLflow RBAC in agent namespaces (created by kagenti chart above)
```
pdettori, line 903: **suggestion**: The OAuth secret wait + backend restart block was removed (was ~20 lines). Was this moved elsewhere, or is the backend now resilient to a missing/late OAuth secret?

If intentional, a note in the PR description would help reviewers understand the rationale — without this wait, the backend pod might start before the OAuth secret exists and fail to pick up `AUTH_ENDPOINT`.

scripts/ocp/setup-kagenti.sh
```diff
@@ -341,30 +514,7 @@ for doc in docs:
...
-        -n kagenti-system 2>/dev/null || true
-      log_success "otel-ingress-ca ConfigMap created"
-    else
-      log_warn "Could not fetch ingress CA — otel-ingress-ca ConfigMap not created"
-    fi
-  fi
+" | $KUBECTL apply -f - 2>/dev/null || true
 }
 
 _helm_kagenti_deps() {
```
pdettori, line 532: **suggestion**: `2>/dev/null || true` silences all kubectl apply errors. Real failures (RBAC denied, invalid YAML, namespace not found) will be invisible. Consider redirecting stderr to a log file instead:
```bash
" | $KUBECTL apply -f - 2>>$LOG_DIR/operand-crs.log || true
```

charts/kagenti-deps/values.yaml
```diff
@@ -154,6 +154,14 @@ istio:
 # receivers/processors/exporters), provide the full array.
 #
 otel:
+  # Enable the MLflow OTEL pipeline without deploying MLflow itself.
```
pdettori: **nit**: Comment says "Has no effect when components.mlflow.enabled is true (that flag takes precedence)" — but the template uses `or` (`$mlflowEnabled := or $.Values.components.mlflow.enabled ($.Values.otel.mlflow).enabled`), so both flags enable the pipeline equally. Neither takes precedence. Suggest: "Also activates when components.mlflow.enabled is true."

scripts/ocp/setup-kagenti.sh
```diff
@@ -229,6 +238,154 @@ export DOMAIN
...
+  local agent_ns
+  agent_ns=$(python3 -c "
+import yaml, sys
+with open('$KAGENTI_REPO/charts/kagenti/values.yaml') as f:
+    v = yaml.safe_load(f)
+for ns in v.get('agentNamespaces', ['team1', 'team2']):
+    print(ns)
+" 2>/dev/null || echo -e "team1\nteam2")
+
+  while IFS= read -r ns; do
```
pdettori, line 367: **praise** 👍 The `_mlflow_grant_otel_rbac` pattern is well done — namespace-scoped RoleBindings with idempotent `--dry-run=client | apply` and dynamic namespace list from values.yaml.

scripts/ocp/setup-kagenti.sh
```diff
@@ -291,14 +448,30 @@ _wait_ns_gone() {
...
 # Wait for the components we need before proceeding to the kagenti chart.
 # Skips MLflow (its oauth-secret is created by the kagenti chart's post-install hook).
+_wait_deployment_ready() {
+  local deploy="$1" ns="$2" label="${3:-$1}" kind="${4:-deployment}"
+  local tries=0
+  if ! $KUBECTL get "$kind"/"$deploy" -n "$ns" &>/dev/null; then
+    log_info "Waiting for $label to appear..."
+    until $KUBECTL get "$kind"/"$deploy" -n "$ns" &>/dev/null; do
+      [ $((++tries)) -ge 60 ] && { log_warn "$label $kind not found after 5m"; return 1; }
+      sleep 5
```
pdettori, line 462: **praise** 👍 The refactored `_wait_deployment_ready` with `kind` parameter is much cleaner and more reusable than the previous ad-hoc waits.

pdettori (changes requested):
## Review Summary

The PR adds RHOAI MLflow integration for OCP: provisions an MLflow CR via RHOAI DSC, wires the OTEL collector with SA bearer token auth, service-CA TLS, and workspace/experiment headers, and improves deployment readiness checks. The Helm template logic is well-structured. However, the MLflow provisioning runs unconditionally, making RHOAI a hard requirement for all OCP deployments.

**Areas reviewed**: Helm templates, Helm values, Shell scripts
**Commits**: 3, all signed-off ✅
**CI**: All checks passing ✅ (E2E HyperShift pending maintainer trigger)

🤖 Reviewed with [Claude Code](https://claude.ai/code)

scripts/ocp/setup-kagenti.sh
```diff
@@ -291,14 +448,30 @@ _wait_ns_gone() {
...
+
 _wait_kagenti_deps_ready() {
   if $DRY_RUN; then return; fi
-  log_info "Waiting for Keycloak..."
-  $KUBECTL rollout status deployment/keycloak -n "$KC_NAMESPACE" --timeout=300s 2>/dev/null || \
-    log_warn "Keycloak rollout not ready within 5m"
-  log_info "Waiting for Istio..."
-  $KUBECTL rollout status deployment/istiod -n istio-system --timeout=300s 2>/dev/null || \
-    log_warn "istiod rollout not ready within 5m"
+  # Keycloak: OLM installs the RHBK operator first, which then processes the
```
r3v5: Sequential deployment waits could be parallelized.

  The dependency graph

                      Helm install
                     /     |      \
           rhbk-operator  cert-manager  istiod
                ↓           (independent)  (independent)
           postgres-kc
                ↓
            keycloak

Suggested fix

```
  _wait_kagenti_deps_ready() {
    if $DRY_RUN; then return; fi

    # Independent components — wait in parallel
    _wait_deployment_ready cert-manager-webhook cert-manager cert-manager &
    local pid_cm=$!
    _wait_deployment_ready istiod istio-system Istio &
    local pid_istio=$!

    # Keycloak chain — must be sequential
    _wait_deployment_ready rhbk-operator "$KC_NAMESPACE" "RHBK operator"
    _wait_deployment_ready postgres-kc "$KC_NAMESPACE" "Keycloak PostgreSQL" statefulset
    _wait_deployment_ready keycloak "$KC_NAMESPACE" Keycloak statefulset

    # Collect parallel results
    wait $pid_cm || log_warn "cert-manager readiness check failed"
    wait $pid_istio || log_warn "Istio readiness check failed"
  }
```

r3v5 (commented):
Nice work, Christian! Left a small comment, not a blocker.

scripts/ocp/setup-kagenti.sh
ChristianZaccaria: Good call, I've added a check for the DSC CRD, if it's not available then we skip mlflow provisioning and proceed with Kagenti installation.

scripts/ocp/setup-kagenti.sh
ChristianZaccaria: I left a comment in the PR that introduced this change (https://github.com/kagenti/kagenti/pull/1156#discussion_r3046607702). - The wait introduced is redundant as it's already taken care of by the kagent-backend init-container. 

Code Reference: https://github.com/kagenti/kagenti/blob/main/charts/kagenti/templates/ui.yaml#L60-L73

scripts/ocp/setup-kagenti.sh
ChristianZaccaria: As opposed to redirecting to a log file that users may fail to check, simply removing `2>/dev/null` will ensure we see warnings and errors inline. What do you think?

charts/kagenti-deps/values.yaml
ChristianZaccaria: Good point, this raises a further question in terms of design. I've added a comment as TODO to be addressed as a follow-up.

>  TODO: Replace with a single mlflow.mode enum (none|managed|external).

This will allow us to guard against conflicting managed and external MLflow if enabled simultaneously.

scripts/ocp/setup-kagenti.sh
ChristianZaccaria: Awesome, thanks Ian, I've applied the changes. I think it's a good addition

pdettori (approved):
All review comments have been addressed:

- **MLflow provisioning guard** — DSC CRD check added; skips gracefully when RHOAI is absent ✅
- **Dual-mode auth conflict** — `(not $rhoaiMlflow)` guard prevents oauth2client from overwriting bearertokenauth; TODO added for future `mlflow.mode` enum ✅
- **OAuth secret wait removal** — justified by init-container handling in `ui.yaml` (ref PR #1156) ✅
- **Error suppression** — `2>/dev/null` removed from `kubectl apply`; remaining `|| true` is appropriate for operand CR application ✅
- **Parallel deployment waits** — cert-manager and Istio waits now run in background with `wait` collection ✅
- **Values comment** — TODO comment added for `mlflow.mode` enum ✅

🤖 Reviewed with [Claude Code](https://claude.ai/code)

PR #1180, opened 2026-04-09
fix: Delete HTTPRoute when deleting an A2A agent in UI

### LINKED ISSUE #829: :bug: Deleting an A2A agent does not delete its HTTPRoute
### Describe the bug

When an A2A agent is deleted through the Kagenti UI, the HTTPRoute is not deleted.

### Steps To Reproduce

1. Deploy the weather agent, selecting "Enable external access"
2. Delete the agent using the Kagenti UI
3. `oc -n team1 get httproute`
4. The route for the deleted agent is still present.

### Expected Behavior

All Agent K8s resource should be cleaned up when the agent is deleted.

### Additional Context

_No response_

rh-dnagornuks:
## Summary

- Add logic to delete a custom resource HTTPRoute in `kagenti/backend/app/routers/agents.py` if it exists when agent is deleted through UI, and document this in the docstring of the method.

## Context

When an A2A agent is deleted through the Kagenti UI, its HTTPRoute resource is left behind. This leads to orphaned networking resources and inconsistent cleanup of agent-related Kubernetes objects.

Fixes #829 

## Tests

Verified by:
- Creating an A2A agent and confirming HTTPRoute exists (`oc get httproute`)
- Deleting the agent via UI
- Confirming the HTTPRoute is removed

- [x] The HTTPRoute is deleted when the A2A agent is deleted through UI on Kind
- [x] The HTTPRoute is deleted when the A2A agent is deleted through UI on OpenShift

### DIFF
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -813,6 +813,7 @@ async def delete_agent(
     This deletes:
     - Deployment, StatefulSet, or Job (whichever exists)
     - Service
+    - HTTPRoute (if exists)
     - Shipwright Build CR (if exists)
     - Shipwright BuildRun CRs (if exist)
     - Legacy: Agent CR (if exists, for backward compatibility)
@@ -859,6 +860,23 @@ async def delete_agent(
             pass
         else:
             logger.warning(f"Failed to delete Service '{name}': {e.reason}")
+
+    # Delete the HTTPRoute (if exists)
+    try:
+        kube.delete_custom_resource(
+            group="gateway.networking.k8s.io",
+            version="v1",
+            namespace=namespace,
+            plural="httproutes",
+            name=name,
+        )
+        messages.append(f"HTTPRoute '{name}' deleted")
+    except ApiException as e:
+        if e.status == 404:
+            # HTTPRoute doesn't exist, that's fine
+            pass
+        else:
+            logger.warning(f"Failed to delete HTTPRoute '{name}': {e.reason}")
 
     # Delete the AgentRuntime CR (if exists)
     try:


kagenti/backend/app/routers/agents.py
```diff
@@ -860,6 +861,23 @@ async def delete_agent(
...
+    # Delete the HTTPRoute (if exists)
+    try:
+        kube.delete_custom_resource(
+            group="gateway.networking.k8s.io",
+            version="v1",
+            namespace=namespace,
+            plural="httproutes",
+            name=name,
+        )
+        messages.append(f"HTTPRoute '{name}' deleted")
```
pdettori, line 873: **suggestion (non-blocking):** This correctly cleans up the HTTPRoute (Gateway API / Kind+K8s), but `routes.py:create_or_update_route()` conditionally creates either an **HTTPRoute** or an **OpenShift Route** (`route.openshift.io/v1`) depending on the detected platform. On OpenShift clusters, the agent will have an OpenShift Route instead of an HTTPRoute, so this block won't find anything to delete and the Route will be orphaned.

Consider also deleting the OpenShift Route in a similar try/except block:
```python
# Delete the OpenShift Route (if exists)
try:
    kube.delete_custom_resource(
        group="route.openshift.io",
        version="v1",
        namespace=namespace,
        plural="routes",
        name=name,
    )
    messages.append(f"Route '{name}' deleted")
except ApiException as e:
    if e.status == 404:
        pass
    else:
        logger.warning(f"Failed to delete Route '{name}': {e.reason}")
```

Alternatively, a follow-up issue would work too — the PR description confirms testing on both Kind and OpenShift, so there may be context I'm missing about Route lifecycle on OCP.

pdettori (commented):
## Review Summary

Clean, well-structured fix that follows existing resource cleanup patterns. The HTTPRoute deletion logic correctly uses `delete_custom_resource` with the Gateway API group/version and handles 404 gracefully.

**Areas reviewed**: Python (agents.py), cross-reference with routes.py and tools.py
**CI status**: Lint/security/DCO pass; Deploy & Test and CodeQL failing (may be unrelated)

### Follow-up items (not blocking)

1. **OpenShift Route parity** — see inline comment. The create path (`routes.py`) creates either HTTPRoute or OpenShift Route depending on platform, but only HTTPRoute is cleaned up on delete.
2. **Tools parity** — `delete_tool` in `tools.py` also doesn't clean up HTTPRoute/Route. Since tools support `createHttpRoute=True`, the same orphan problem exists there. Worth a follow-up issue.

kagenti/backend/app/routers/agents.py
rh-dnagornuks: I created a follow-up issue for this here: #1196.
From testing, it looks like the platform isn’t being detected correctly and always defaults to Kubernetes, so an HTTPRoute is created even on OpenShift.

rh-dnagornuks: Made a PR featuring your code suggestion for deleting OpenShift route: #1201

pdettori (approved):
## Review Summary

Clean, well-scoped fix that follows the existing resource cleanup pattern in `delete_agent`. The HTTPRoute deletion logic correctly uses `delete_custom_resource` with the Gateway API group/version and handles 404 gracefully.

**Review comments addressed:**
- OpenShift Route parity: acknowledged with follow-up issue #1196 (platform detection bug) and PR #1201 (Route deletion + platform fix)
- CodeQL log injection: pre-existing pattern across 9 other `logger.warning` calls in the same function — not introduced by this PR

**Checks**: DCO, lint, security, backend-tests, chart-tests all pass.

PR #1183, opened 2026-04-09
fix: point Kind weather agent to LiteMaaS instead of Ollama

Ladas:
## Summary

Switch Kind weather agent deployment from Ollama (dockerhost:11434) to LiteMaaS. The agent-examples `has_valid_api_key` check rejects dummy keys for non-localhost hosts, breaking Kind Deploy&Test on main.

Uses the same LiteMaaS endpoint as OCP + `secretKeyRef` for API key from `openai-secret`.

## Context

Main's E2E Kind is broken since the `has_valid_api_key` check was added to agent-examples.

### DIFF
--- a/.github/workflows/e2e-kind.yaml
+++ b/.github/workflows/e2e-kind.yaml
@@ -62,6 +62,8 @@ jobs:
         run: bash .github/scripts/common/10-setup-dependencies.sh
 
       - name: Create secret values file for CI
+        env:
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
         run: bash .github/scripts/common/20-create-secrets.sh
 
       - name: Run Ansible installer (dev mode with kagenti-operator)
--- a/.github/workflows/e2e-kind-pr.yaml
+++ b/.github/workflows/e2e-kind-pr.yaml
@@ -55,6 +55,8 @@ jobs:
 
       - name: Create secret values file for CI
         run: bash .github/scripts/common/20-create-secrets.sh
+        env:
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
 
       - name: Run Ansible installer (dev mode with kagenti-operator)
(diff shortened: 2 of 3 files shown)


.github/workflows/e2e-kind.yaml
```diff
@@ -62,6 +62,8 @@ jobs:
         run: bash .github/scripts/common/10-setup-dependencies.sh
 
       - name: Create secret values file for CI
```
pdettori, line 64: nit: The `env:` block is placed before `run:` here, but in `e2e-kind-pr.yaml` it's placed after `run:`. Both are valid YAML, but inconsistent ordering across the two workflows. Consider matching the order for readability.

pdettori (approved):
Clean, focused fix. The weather agent deployment correctly switches from Ollama to LiteMaaS with `secretKeyRef` for the API key, and CI workflows pass the GitHub secret. The Helm chart already creates `openai-secret` in agent namespaces — no chart changes needed.

**Areas reviewed**: YAML, CI/GitHub Actions, Security
**Commits**: 2, all DCO signed-off
**CI**: Deploy & Test failing (likely infra/timing — this PR is the fix for Kind E2E)

pdettori:
There is still an issue in CI though - `No LLM API key configured. Set the LLM_API_KEY` - does it mean some config still needed ?

pdettori:
@Ladas  - should we consider allowing the use of a special test key for Ollama to avoid using MaaS for this?

Ladas:
@pdettori with the kagenti-admin work, I am actually removing ollama for k3s tests, to make room in the CI node. It could be good to keep for Kind maybe? Btw. the aganet examples 222 we merged is fixing kind, we just need the new agent-examples tag.

Let me check for this PR why it's not effective. (if gh still applies the workflow changes only after merge)

Ladas:
solved other way, closing for now, will fix later with litellm integration

PR #1198, opened 2026-04-10
fix(ui): Always send tool servicePorts on import

mrsabath:
## Problem

`ImportToolPage` only sent `servicePorts` when **Pod Configuration** was expanded (`showPodConfig`). The form still defaulted ports to 9090 in React state, but if the user submitted with the section collapsed, the API received `servicePorts: undefined`. The Shipwright Build annotation then omitted ports, and `finalize-shipwright-build` used `DEFAULT_IN_CLUSTER_PORT` (8000)—breaking MCP tools that listen on 9090 (e.g. GitHub tool demo).

## Change

Always include `servicePorts` in both **source** and **image** create-tool payloads, matching `ImportAgentPage` behavior.

## Notes

- Existing tools already deployed with wrong Service ports still need a one-time patch or redeploy.
- `showPodConfig` remains only for expand/collapse UX.

### DIFF
--- a/kagenti/ui-v2/src/pages/ImportToolPage.tsx
+++ b/kagenti/ui-v2/src/pages/ImportToolPage.tsx
@@ -463,7 +463,9 @@ export const ImportToolPage: React.FC = () => {
         imageTag,
         shipwrightConfig,
         envVars: envVars.filter((ev) => ev.name && (ev.value !== undefined || ev.valueFrom)),
-        servicePorts: showPodConfig ? servicePorts : undefined,
+        // Always send ports so Shipwright finalize uses them (matches Import Agent behavior).
+        // Previously omitted when Pod Configuration stayed collapsed → backend defaulted to 8000.
+        servicePorts,
         createHttpRoute,
         authBridgeEnabled,
         spireEnabled,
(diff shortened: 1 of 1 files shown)


kagenti/ui-v2/src/pages/ImportToolPage.tsx
```diff
@@ -463,7 +463,9 @@ export const ImportToolPage: React.FC = () => {
         imageTag,
         shipwrightConfig,
         envVars: envVars.filter((ev) => ev.name && (ev.value !== undefined || ev.valueFrom)),
```
huang195, line 465: Dobra robota! The conditional was sneaky — `showPodConfig` only controls the UI accordion, not whether the ports matter. Sending them unconditionally is the right call, and it matches `ImportAgentPage` perfectly. Czysto i elegancko.

huang195 (approved):
Dobra robota, Mariusz! Clean, surgical fix.

The root cause is well explained — `showPodConfig` was pulling double duty as both a UI toggle and a data gate, which it shouldn't have been. Both the source-build and image-import payloads now always include `servicePorts`, matching `ImportAgentPage` behavior (verified at lines 480/513).

No issues found. Nic do dodania — this one's ready to go. 👍

**Areas reviewed**: Frontend (TypeScript/React)
**CI status**: All passing

PR #1230, opened 2026-04-13
fix: pass KEYCLOAK_PUBLIC_URL to UI for browser redirects

pdettori:
## Summary
- Pass `keycloak.publicUrl` Helm value as `KEYCLOAK_PUBLIC_URL` env var to the UI deployment
- Add `keycloak_public_url` config field to the backend with priority: `AUTH_ENDPOINT` > `KEYCLOAK_PUBLIC_URL` > `KEYCLOAK_URL` > default
- Add 8 unit tests covering the full priority chain and edge cases

Closes #1154

## Root Cause

On OpenShift, the browser-facing Keycloak URL depends entirely on `AUTH_ENDPOINT` from the `kagenti-ui-oauth-secret`. If the oauth-secret-job hasn't completed when the UI pod starts (init container exits 0 after 4-minute timeout), the backend falls back to `KEYCLOAK_URL` — the **internal** K8s service URL (`http://keycloak-service.keycloak:8080`) — which the browser cannot reach.

The Helm value `keycloak.publicUrl` was already correctly set by `setup-kagenti.sh` (route discovery at line 665), but was never passed to the UI deployment.

## Test plan
- [x] 8 unit tests for `effective_keycloak_url` priority chain
- [x] Regression: existing `test_dashboard_config.py` passes
- [x] Pre-commit hooks pass
- [ ] CI pipeline

### DIFF
--- a/charts/kagenti/templates/ui.yaml
+++ b/charts/kagenti/templates/ui.yaml
@@ -123,6 +123,8 @@ spec:
               value: "{{ .Values.featureFlags.triggers }}"
             - name: KEYCLOAK_URL
               value: {{ .Values.keycloak.url | quote }}
+            - name: KEYCLOAK_PUBLIC_URL
+              value: {{ .Values.keycloak.publicUrl | quote }}
             # Dashboard URLs from ConfigMap (populated by ui-routes-job)
             - name: TRACES_DASHBOARD_URL
               valueFrom:
--- a/kagenti/backend/app/core/config.py
+++ b/kagenti/backend/app/core/config.py
@@ -97,24 +97,25 @@ class Settings(BaseSettings):
 
     # Legacy direct config (fallback if AUTH_ENDPOINT not provided)
     keycloak_url: str = ""
+    # Browser-facing Keycloak URL (from keycloak.publicUrl Helm value)
+    keycloak_public_url: str = ""
     keycloak_realm: str = "kagenti"
     keycloak_client_id: str = "kagenti-ui"
 
     @property
     def effective_keycloak_url(self) -> str:
         """
-        Extract Keycloak base URL from AUTH_ENDPOINT or use direct config.
+        External (browser-facing) Keycloak URL for frontend auth redirects.
 
-        This returns the external (browser-facing) URL, used for frontend
-        auth config and redirect flows.
+        Priority: AUTH_ENDPOINT (from oauth secret) > KEYCLOAK_PUBLIC_URL
+        (from Helm) > KEYCLOAK_URL (internal, last resort) > constructed default.
         """
         if self.auth_endpoint:
-            # Parse AUTH_ENDPOINT to extract base URL
-            # Pattern: {base_url}/realms/{realm}/protocol/openid-connect/auth
             match = re.match(r"(https?://[^/]+)/realms/", self.auth_endpoint)
             if match:
                 return match.group(1)
-        # Fallback to direct config or default
+        if self.keycloak_public_url:
+            return self.keycloak_public_url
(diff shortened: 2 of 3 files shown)


kagenti/backend/tests/test_keycloak_url_config.py
```diff
@@ -0,0 +1,90 @@
+# Copyright 2025 IBM Corp.
+# Licensed under the Apache License, Version 2.0
+
+"""Tests for effective_keycloak_url priority chain.
+
+Priority: AUTH_ENDPOINT > KEYCLOAK_PUBLIC_URL > KEYCLOAK_URL > constructed default.
+Fixes: https://github.com/kagenti/kagenti/issues/1154
+"""
+
+import pytest
```
huang195, line 10: **praise**: Bellissimo test suite! The `make_settings` fixture that clears env vars before each test is a clean isolation pattern. The priority chain coverage is thorough — happy path, each fallback level, malformed input, and the in-cluster vs off-cluster distinction. Perfetto. 🤌

kagenti/backend/app/core/config.py
```diff
@@ -97,24 +97,25 @@ def _add_domain_cors_origin(self) -> "Settings":
 
     # Legacy direct config (fallback if AUTH_ENDPOINT not provided)
     keycloak_url: str = ""
+    # Browser-facing Keycloak URL (from keycloak.publicUrl Helm value)
+    keycloak_public_url: str = ""
     keycloak_realm: str = "kagenti"
     keycloak_client_id: str = "kagenti-ui"
 
     @property
```
huang195, line 105: **praise**: The updated docstring clearly documents the priority chain. Having this as a single `@property` with explicit fallthrough makes the resolution logic easy to reason about.

charts/kagenti/templates/ui.yaml
```diff
@@ -123,6 +123,8 @@ spec:
               value: "{{ .Values.featureFlags.triggers }}"
             - name: KEYCLOAK_URL
               value: {{ .Values.keycloak.url | quote }}
+            - name: KEYCLOAK_PUBLIC_URL
+              value: {{ .Values.keycloak.publicUrl | quote }}
             # Dashboard URLs from ConfigMap (populated by ui-routes-job)
```
huang195, line 128: **nit**: Unlike `defaultRegistryUrl` (which uses `{{- if }}` to conditionally inject), this always injects `KEYCLOAK_PUBLIC_URL`. It works because `keycloak.publicUrl` has a default in `values.yaml` (`http://keycloak.localtest.me:8080`), but for consistency you could guard it with `{{- if .Values.keycloak.publicUrl }}`. Not blocking — the Python side handles empty strings correctly via `if self.keycloak_public_url:`.

huang195 (approved):
Clean, well-motivated fix for a real OpenShift timing issue — the oauth-secret-job delay causing fallback to the internal Keycloak URL that browsers can't reach. The solution adds a new priority level (`KEYCLOAK_PUBLIC_URL`) to the existing `effective_keycloak_url` chain without disrupting current behavior. Excellent test coverage with 8 tests covering the full priority chain and edge cases.

Perfetto, Paolo! This is exactly the right fix. 🤌

**Areas reviewed**: Python, Helm/K8s, Tests
**CI status**: 20/21 passing, E2E HyperShift pending (running on OpenShift 4.20.11)
**Verdict**: APPROVE

**Minor nit**: The PR body has a `Generated with Claude Code` footer — per repo convention, commit attribution should use `Assisted-By` (which the commit itself does correctly). Consider removing the footer from the PR body for consistency.

PR #1241, opened 2026-04-14
🐛 bug: fix TypeError crash on null Kubernetes status conditions

usize:
## Summary

- Fix `TypeError: 'NoneType' object is not iterable` crash when listing tools or agents whose Kubernetes resources have `status.conditions: null`
- Replace `.get("conditions", [])` with `.get("conditions") or []` across all status-parsing code paths
- Affects `routers/tools.py`, `routers/agents.py`, and `services/shipwright.py` (8 occurrences total)

## Root cause

`dict.get(key, default)` only returns the default when the key is **missing**. Kubernetes resources that haven't fully reconciled have `conditions: null` in their status — the key exists with value `None`. Iterating over `None` raises `TypeError`.

## Test plan

- [ ] Deploy agents/tools to a fresh cluster and verify the catalog pages load without 500 errors
- [ ] Verify agents/tools in various states (Ready, Progressing, Not Ready) render correctly
- [ ] Confirm Shipwright build status parsing still works

### DIFF
--- a/kagenti/backend/tests/test_null_conditions.py
+++ b/kagenti/backend/tests/test_null_conditions.py
@@ -1,0 +1,91 @@
+"""
+Tests for handling null conditions in Kubernetes status responses.
+
+Kubernetes resources that haven't fully reconciled can have
+`status.conditions: null` rather than an empty list or a missing key.
+This caused TypeError crashes when iterating over conditions.
+
+See: https://github.com/kagenti/kagenti/pull/1241
+"""
+
+import sys
+import os
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
+
+from app.routers.tools import _get_workload_status  # noqa: E402
+from app.routers.agents import _is_deployment_ready  # noqa: E402
+
+
+class TestNullConditionsHandling:
+    """Verify status functions don't crash when conditions is None."""
+
+    def test_tool_workload_status_with_null_conditions(self):
+        """_get_workload_status should not crash when conditions is null."""
+        workload = {
+            "spec": {"replicas": 1},
+            "status": {
+                "readyReplicas": 0,
+                "availableReplicas": 0,
+                "conditions": None,
+            },
+        }
+        result = _get_workload_status(workload)
+        assert result == "Not Ready"
+
+    def test_tool_workload_status_with_missing_conditions(self):
+        """_get_workload_status should handle missing conditions key."""
+        workload = {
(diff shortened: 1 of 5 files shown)


kagenti/backend/tests/test_null_conditions.py
```diff
@@ -0,0 +1,91 @@
...
+            },
+        }
+        result = _is_deployment_ready(resource)
+        assert result == "Not Ready"
+
+    def test_agent_deployment_ready_with_missing_status(self):
+        """_is_deployment_ready should handle completely missing status."""
+        resource = {"spec": {"replicas": 1}}
+        result = _is_deployment_ready(resource)
+        assert result == "Not Ready"
```
pdettori, line 91: **suggestion**: Two E2E test files still use the old vulnerable `.get("conditions", [])` pattern:
- `kagenti/tests/e2e/common/test_shipwright_tool_build.py:45`
- `kagenti/tests/e2e/common/test_shipwright_build.py:388`

While these are test files (less likely to hit null conditions from real K8s), fixing them would prevent the same class of bug if test data ever includes `conditions: null`. Could be a follow-up.

kagenti/backend/tests/test_null_conditions.py
```diff
@@ -0,0 +1,91 @@
+"""
+Tests for handling null conditions in Kubernetes status responses.
+
+Kubernetes resources that haven't fully reconciled can have
+`status.conditions: null` rather than an empty list or a missing key.
+This caused TypeError crashes when iterating over conditions.
```
pdettori, line 6: **praise**: Good test coverage — all three cases (null, missing, empty) tested for both tools and agents code paths. The shipwright functions (`extract_buildrun_info`, `is_build_succeeded`) could also benefit from null-conditions tests as a follow-up.

pdettori (approved):
Clean, well-scoped bug fix for a real crash (`TypeError: 'NoneType' object is not iterable`) caused by Kubernetes resources having `status.conditions: null` instead of missing/empty. The root cause explanation is excellent — `.get(key, default)` only returns `default` when the key is absent, not when the value is `None`. The fix (`.get("conditions") or []`) is the idiomatic Python pattern for this.

All 8 occurrences across production code (`agents.py`, `tools.py`, `shipwright.py`) are fixed consistently. Good test coverage for null/missing/empty edge cases.

**CI**: All checks passing.

Two minor suggestions left as inline comments (both non-blocking).

PR #1246, opened 2026-04-14
fix: make DEFAULT_INTERNAL_REGISTRY configurable via env var

### LINKED ISSUE #1242: Bug: DEFAULT_INTERNAL_REGISTRY is hardcoded to dev-only registry, breaks source builds on OpenShift
## Bug Description

`DEFAULT_INTERNAL_REGISTRY` in `kagenti/backend/app/core/constants.py` is hardcoded to `registry.cr-system.svc.cluster.local:5000`, which is a registry that only exists in local dev/kind clusters. On OpenShift (and most production deployments), this registry does not exist, causing **all source-based agent/tool builds to fail** when the user does not explicitly provide a `registryUrl`.

## Steps to Reproduce

1. Deploy Kagenti on an OpenShift cluster
2. Create an agent via the UI using "Source" deploy type
3. Fill in Git repo and path, leave "Registry URL" empty (or don't provide `registryUrl` in the API call)
4. Submit the build

## Expected Behavior

The build should either:
- Use the correct cluster-internal registry (e.g. `image-registry.openshift-image-registry.svc:5000` on OpenShift), or
- Fail with a clear error message telling the user to provide a registry URL

## Actual Behavior

The Kagenti API falls back to `registry.cr-system.svc.cluster.local:5000`, creates a Shipwright `Build` CR with this non-existent registry as the output image destination, and the `BuildRun` fails at `step-build-and-push` because `buildah` cannot push to a registry that doesn't exist. The error message does not indicate the registry is the problem.

## Root Cause

In [`kagenti/backend/app/core/constants.py`](https://github.com/kagenti/kagenti/blob/main/kagenti/backend/app/core/constants.py):

```python
# Default internal registry URL (for dev/kind clusters)
DEFAULT_INTERNAL_REGISTRY = "registry.cr-system.svc.cluster.local:5000"
```

This is used in `app/routers/agents.py` and `app/routers/tools.py` as:

```python
registry_url = request.registryUrl or DEFAULT_INTERNAL_REGISTRY
```

There is no way to override this value via environment variable, Helm values, or ConfigMap.

## Suggested Fix

Make `DEFAULT_INTERNAL_REGISTRY` configurable via an environment variable with the current value as fallback:

```python
import os

DEFAULT_INTERNAL_REGISTRY = os.environ.get(
    "DEFAULT_REGISTRY_URL", "registry.cr-system.svc.cluster.local:5000"
)
```

Then expose `DEFAULT_REGISTRY_URL` as a Helm chart value so each deployment can set the correct registry for its cluster.

## Workaround

We applied a temporary fix by mounting a patched `constants.py` via a ConfigMap and setting the `DEFAULT_REGISTRY_URL` env var on the `kagenti-backend` deployment. This will not survive Helm upgrades.

## Environment

- Kagenti backend image: `ghcr.io/kagenti/kagenti/backend:v0.6.0-alpha.3`
- Platform: OpenShift (ROSA)
- Internal registry: `image-registry.openshift-image-registry.svc:5000`

pdettori:
## Summary

- Make `DEFAULT_INTERNAL_REGISTRY` configurable via `DEFAULT_REGISTRY_URL` environment variable (Pydantic Settings)
- Wire through Helm chart as `ui.backend.defaultRegistryUrl` with conditional env var injection
- Set OpenShift internal registry (`image-registry.openshift-image-registry.svc:5000`) in `ocp_values.yaml` and `ocp_minimal_values.yaml`

Closes #1242

## Changes

| File | Change |
|------|--------|
| `kagenti/backend/app/core/config.py` | Add `default_registry_url` setting with Kind default |
| `kagenti/backend/app/core/constants.py` | Use `settings.default_registry_url` instead of hardcoded value |
| `charts/kagenti/values.yaml` | Add `ui.backend.defaultRegistryUrl` (empty = use Python default) |
| `charts/kagenti/templates/ui.yaml` | Conditionally set `DEFAULT_REGISTRY_URL` env var |
| `deployments/envs/ocp_values.yaml` | Set OpenShift internal registry |
| `deployments/envs/ocp_minimal_values.yaml` | Set OpenShift internal registry |
| `kagenti/backend/tests/test_shipwright_shared.py` | Add 3 tests for configurable registry |

## Test plan

- [x] Existing unit tests pass (36/36 in `test_shipwright_shared.py`)
- [x] New tests verify: default value, env var override, OCP registry strategy selection
- [ ] Deploy to Kind cluster — verify source builds still use `registry.cr-system.svc.cluster.local:5000`
- [ ] Deploy to OpenShift — verify source builds use `image-registry.openshift-image-registry.svc:5000`

## Backward compatibility

No breaking changes. Without `DEFAULT_REGISTRY_URL` env var or Helm value, behavior is identical to before (Kind dev registry default).

### DIFF
--- a/charts/kagenti/templates/ui.yaml
+++ b/charts/kagenti/templates/ui.yaml
@@ -114,6 +114,10 @@ spec:
                   optional: true
             - name: ENABLE_AUTH
               value: "{{ .Values.ui.auth.enabled }}"
+            {{- if .Values.ui.backend.defaultRegistryUrl }}
+            - name: DEFAULT_REGISTRY_URL
+              value: {{ .Values.ui.backend.defaultRegistryUrl | quote }}
+            {{- end }}
             # Feature flags
             - name: KAGENTI_FEATURE_FLAG_SANDBOX
               value: "{{ .Values.featureFlags.sandbox }}"
--- a/kagenti/backend/tests/test_shipwright_shared.py
+++ b/kagenti/backend/tests/test_shipwright_shared.py
@@ -506,3 +506,33 @@ class TestExtractBuildRunInfo:
 
         assert info["phase"] == "Failed"
         assert info["failureMessage"] == "Build failed: Dockerfile not found"
+
+
+class TestDefaultRegistryUrlConfig:
+    """Tests for DEFAULT_INTERNAL_REGISTRY configurability via env var."""
+
+    def test_default_value_is_kind_registry(self):
+        """Default value should be the Kind dev registry."""
+        assert DEFAULT_INTERNAL_REGISTRY == "registry.cr-system.svc.cluster.local:5000"
+
+    def test_env_var_overrides_default(self, monkeypatch):
+        """DEFAULT_REGISTRY_URL env var should override the default."""
+        monkeypatch.setenv(
+            "DEFAULT_REGISTRY_URL", "image-registry.openshift-image-registry.svc:5000"
+        )
+
+        from pydantic_settings import BaseSettings
+        from app.core.config import Settings
+
+        fresh_settings = Settings()
+        assert (
+            fresh_settings.default_registry_url
+            == "image-registry.openshift-image-registry.svc:5000"
+        )
+
+    def test_openshift_registry_is_internal(self):
+        """OpenShift internal registry should be detected as internal (svc.cluster.local absent but still in-cluster)."""
+        ocp_registry = "image-registry.openshift-image-registry.svc:5000"
+        # OpenShift registry doesn't contain svc.cluster.local, so it gets secure strategy
+        result = select_build_strategy(ocp_registry)
+        assert result == SHIPWRIGHT_STRATEGY_SECURE
--- a/charts/kagenti/values.yaml
+++ b/charts/kagenti/values.yaml
@@ -84,6 +84,8 @@ ui:
   backend:
     image: ghcr.io/kagenti/kagenti/backend
     tag: latest
+    # Default registry for source-based builds (Kind: registry.cr-system.svc.cluster.local:5000)
+    defaultRegistryUrl: ""
     resources:
       limits:
         cpu: 250m
--- a/deployments/envs/ocp_minimal_values.yaml
+++ b/deployments/envs/ocp_minimal_values.yaml
@@ -100,6 +100,8 @@ charts:
       ui:
         auth:
           enabled: true
+        backend:
+          defaultRegistryUrl: "image-registry.openshift-image-registry.svc:5000"
       uiOAuthSecret:
         # for a OCP with self-signed certs for routes it should be true.
         # for a OCP with CA-signed certs for routes it should be false.
--- a/deployments/envs/ocp_values.yaml
+++ b/deployments/envs/ocp_values.yaml
@@ -100,6 +100,8 @@ charts:
       ui:
         auth:
           enabled: true
+        backend:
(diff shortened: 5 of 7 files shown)


kagenti/backend/tests/test_shipwright_shared.py
```diff
@@ -506,3 +506,33 @@ def test_failed_buildrun_info(self):
...
+
+        fresh_settings = Settings()
+        assert (
+            fresh_settings.default_registry_url
+            == "image-registry.openshift-image-registry.svc:5000"
+        )
+
+    def test_openshift_registry_is_internal(self):
+        """OpenShift internal registry should be detected as internal (svc.cluster.local absent but still in-cluster)."""
+        ocp_registry = "image-registry.openshift-image-registry.svc:5000"
```
huang195, line 535: **nit**: The docstring says "should be detected as internal" but the assertion is `SHIPWRIGHT_STRATEGY_SECURE`. Slightly misleading — the OCP registry IS internal but doesn't match the insecure URL pattern. Maybe: "OpenShift internal registry uses secure strategy (proper TLS, unlike Kind's insecure dev registry)." Just a clarity tweak, mamma mia it confused me for a second 😄

kagenti/backend/tests/test_shipwright_shared.py
```diff
@@ -506,3 +506,33 @@ def test_failed_buildrun_info(self):
...
         assert info["phase"] == "Failed"
         assert info["failureMessage"] == "Build failed: Dockerfile not found"
+
+
+class TestDefaultRegistryUrlConfig:
+    """Tests for DEFAULT_INTERNAL_REGISTRY configurability via env var."""
+
+    def test_default_value_is_kind_registry(self):
+        """Default value should be the Kind dev registry."""
+        assert DEFAULT_INTERNAL_REGISTRY == "registry.cr-system.svc.cluster.local:5000"
```
huang195, line 516: **praise**: Bellissimo test trio! Testing the default via the constant, the env override via a fresh `Settings()` instance (avoiding module-level caching), and documenting the OCP strategy behavior. Perfetto.

charts/kagenti/templates/ui.yaml
```diff
@@ -114,6 +114,10 @@ spec:
                   optional: true
             - name: ENABLE_AUTH
               value: "{{ .Values.ui.auth.enabled }}"
+            {{- if .Values.ui.backend.defaultRegistryUrl }}
+            - name: DEFAULT_REGISTRY_URL
+              value: {{ .Values.ui.backend.defaultRegistryUrl | quote }}
```
huang195, line 119: **praise**: Perfetto — conditional injection with `{{- if }}` means Kind deployments don't get a spurious env var. The Python default handles the Kind case, Helm only overrides for OCP. Clean separation of concerns.

huang195 (approved):
Focused, well-scoped fix that makes `DEFAULT_INTERNAL_REGISTRY` configurable via Pydantic Settings + Helm, unblocking source-based builds on OpenShift. The approach follows existing patterns in `constants.py` (module-level `settings.*` access) and `ui.yaml` (conditional env var injection). Tests cover the default, the override, and the OCP strategy behavior. Backward compatible — no change without explicit configuration.

Che bel fix, Paolo! Clean, tested, and backward compatible. 🤌

**Areas reviewed**: Python, Helm/K8s, YAML
**CI status**: all 21 checks passing (including Deploy & Test and E2E HyperShift) ✅
**Verdict**: APPROVE

PR #1254, opened 2026-04-15
Add authbridge-unified-config ConfigMap to agent namespaces

huang195:
## Summary

Adds `authbridge-unified-config` ConfigMap to each agent namespace in the Helm chart. This is the YAML config file consumed by the unified authbridge binary (`cmd/authbridge` in kagenti-extensions).

Key features:
- Derives `token_url`, `issuer`, and `jwks_url` from `keycloak_url` + `keycloak_realm` (no hardcoded URLs)
- Identity type follows `authBridge.clientAuthType` Helm value (`client-secret` or `federated-jwt`)
- `jwt_svid_path` only included when `federated-jwt` is configured
- Credential files loaded at runtime from `/shared/` (written by client-registration sidecar)
- Default mode: `envoy-sidecar` with `passthrough` outbound policy

The ConfigMap is inert until the kagenti-operator is updated to mount it (behind a feature gate). No behavioral change to existing deployments.

## Related

- kagenti-extensions `v0.5.0-alpha.1` — unified authbridge binary
- kagenti-extensions #279 — modular token exchange library design
- Next: kagenti-operator PR to add `UnifiedAuthBridge` feature gate

## Test plan

- [x] `helm template` renders correctly for both `client-secret` and `federated-jwt` auth types
- [x] ConfigMap created per namespace (team1, team2)
- [ ] CI passes

Assisted-By: Claude (Anthropic AI) <noreply@anthropic.com>

PR #1257, opened 2026-04-16
⬆️ Upgrade operator to v0.2.0-alpha.26, swap to authbridge-unified

huang195:
## Summary

- Bump kagenti-operator subchart from `0.2.0-alpha.25` → `0.2.0-alpha.26`
- Bump operator image tag from `0.2.0-alpha.24` → `0.2.0-alpha.26`
- Swap envoy-proxy sidecar from `envoy-with-processor` → `authbridge-unified` (unified binary: Envoy + authbridge in one container)
- Bump all authbridge extension images from `v0.4.0-alpha.10` → `v0.5.0-alpha.2`

## Dependencies

All merged and released:
- kagenti/kagenti-extensions#322 — fix multi-arch Envoy build
- kagenti/kagenti-operator#289 — operator authbridge-unified support
- #1254 — `authbridge-unified-config` ConfigMap in agent namespaces

## Test plan

- [ ] `helm template` renders correct image references
- [ ] Kind deploy with full platform — operator injects `authbridge-unified` sidecar
- [ ] Weather agent works from UI with token exchange
- [ ] Verify AMD64 Envoy binary runs correctly (no exit code 126)

### DIFF
--- a/.github/scripts/common/87-setup-test-credentials.sh
+++ b/.github/scripts/common/87-setup-test-credentials.sh
@@ -216,6 +216,33 @@ E2E_CLIENT_SECRET=$(kc_api GET "$KEYCLOAK_URL/admin/realms/$REALM/clients/$CLIEN
 log_success "Service account client ready (client_id=$E2E_CLIENT_ID)"
 
 # ============================================================================
+# 4b. Attach agent audience scopes to the E2E test client
+#     Client-registration creates agent-*-aud scopes as realm defaults, but
+#     realm defaults don't retroactively apply to existing clients. We need
+#     to explicitly add them so the E2E test token contains the aud claim
+#     that AuthBridge requires for inbound JWT validation.
+# ============================================================================
+
+log_info "Attaching agent audience scopes to $E2E_CLIENT_ID..."
+REALM_DEFAULT_SCOPES=$(kc_api GET "$KEYCLOAK_URL/admin/realms/$REALM/default-default-client-scopes")
+AGENT_SCOPE_IDS=$(echo "$REALM_DEFAULT_SCOPES" | python3 -c "
+import sys, json
+scopes = json.load(sys.stdin)
+for s in scopes:
+    if s['name'].startswith('agent-') and s['name'].endswith('-aud'):
+        print(s['id'], s['name'])
+" 2>/dev/null || echo "")
+
+if [ -n "$AGENT_SCOPE_IDS" ]; then
+    while IFS=' ' read -r scope_id scope_name; do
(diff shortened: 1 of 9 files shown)


pdettori (approved):
## Review Summary

This PR correctly integrates `authbridge-unified` (merging Envoy + authbridge-ext-proc into one container), fixes two OpenShift-specific auth issues (Keycloak Route discovery for JWKS, random admin password sync for `client-registration`), and strengthens E2E reliability by switching to a confidential OAuth client that properly inherits audience scopes. Every commit documents the concrete failure mode it addresses — good discipline.

**Areas reviewed**: Helm/Charts, Shell scripts, Ansible YAML, K8s manifests, Python  
**Commits**: 12 commits, all `Signed-off-by` ✓ (5 missing emoji prefix — nit)  
**CI**: All checks passing ✓

---

### Suggestions

**`deployments/ansible/roles/kagenti_installer/tasks/main.yml` — `Override keycloak admin credentials from initial-admin secret` task**  
The decoded `adminPassword` is stored in an Ansible fact passed as a Helm override. At `-v` or higher verbosity, Ansible prints task output including fact values — the decoded password will appear in the log. Consider adding `no_log: true` to this task to suppress credential exposure in verbose runs.

**`.github/scripts/common/87-setup-test-credentials.sh` — agent scope attachment (~line 228)**  
```bash
AGENT_SCOPE_IDS=$(echo "$REALM_DEFAULT_SCOPES" | python3 -c "..." 2>/dev/null || echo "")
```
`2>/dev/null || echo ""` silently swallows Python errors. If the Keycloak API returns an unexpected structure, the parse fails silently and `log_info "No agent audience scopes found"` fires — misleading during debugging. Consider dropping `2>/dev/null` so errors surface.

---

### Nit

**`charts/kagenti/values.yaml` line ~248**: `envoyProxy` key now maps to the `authbridge-unified` image. The key name no longer describes the image's role. Can't be fixed here (key is part of the operator chart's API), but worth a follow-up issue to rename `envoyProxy` → `authBridgeUnified` in kagenti-operator so `values.yaml` readers aren't misled.

**Commit convention**: 5 commits missing emoji prefix per kagenti convention: `549d9c7`, `3ac1bd4`, `22d5ead`, `2393716`, `339c897`.

---

### Highlights

- Ansible OpenShift overrides block is well-structured: `failed_when: false` + `when:` guards on each step, `combine(recursive=True)` to safely merge existing overrides.
- `conftest.py` confidential-client fallback is clean and backward-compatible — won't break deployments that don't yet have the new secret keys.
- Honest commit-and-revert of the `clientAuthType` guard: root cause fixed in #1268 rather than worked around here.

PR #1265, opened 2026-04-16
feat: add bash installer for Kind replacing Ansible

pdettori:
## Summary

- **Replace Ansible-based Kind installer** with composable bash scripts (`scripts/kind/setup-kagenti.sh` + `scripts/kind/cleanup-kagenti.sh`) that mirror the existing OCP installer pattern
- **Eliminate dev dependencies**: only `helm`, `kubectl`, and `kind` required — no more ansible, galaxy collections, or Python packages
- **Composable `--with-*` flags** for incremental setup: core (cert-manager, keycloak, operator) always installed, optional layers for istio, spire, mcp-gateway, builds (tekton/shipwright), otel, and ui
- **Idempotent**: all operations use `helm upgrade --install` and `kubectl apply` — safe to re-run to add components incrementally
- **Update `kind-full-test.sh`** to call the new scripts instead of the Ansible installer

## Installer flags

| Flag | What it enables |
|------|-----------------|
| `--with-istio` | Istio ambient mesh (base, istiod, cni, ztunnel) + ingress gateway |
| `--with-spire` | SPIRE (spire-crds + spire) + SPIFFE IdP setup |
| `--with-mcp-gateway` | MCP Gateway |
| `--with-builds` | Tekton + Shipwright |
| `--with-otel` | OpenTelemetry collector |
| `--with-ui` | Backend API + UI |
| `--with-all` | Enable everything |
| `--skip-cluster` | Don't create Kind cluster (reuse existing) |
| `--dry-run` | Show what would be installed without executing |

## Test plan

- [ ] `scripts/kind/setup-kagenti.sh` — core-only install (cert-manager, keycloak, operator)
- [ ] `scripts/kind/setup-kagenti.sh --with-all` — full install with all components
- [ ] Re-run `scripts/kind/setup-kagenti.sh --with-istio` on existing cluster — idempotent add
- [ ] `scripts/kind/cleanup-kagenti.sh --destroy-cluster` — full teardown
- [ ] `.github/scripts/local-setup/kind-full-test.sh --skip-cluster-destroy` — E2E integration

### DIFF
--- a/.github/scripts/local-setup/kind-full-test.sh
+++ b/.github/scripts/local-setup/kind-full-test.sh
@@ -240,14 +240,14 @@ if [ "$RUN_INSTALL" = "true" ]; then
 
     if [ "$CLEAN_KAGENTI" = "true" ]; then
         log_step "Uninstalling Kagenti (--clean-kagenti)..."
-        ./deployments/ansible/cleanup-install.sh || true
+        ./scripts/kind/cleanup-kagenti.sh || true
     fi
 
     log_step "Creating secrets..."
     ./.github/scripts/common/20-create-secrets.sh
 
-    log_step "Running Ansible installer..."
-    ./.github/scripts/kagenti-operator/30-run-installer.sh --env "$KAGENTI_ENV"
+    log_step "Running Kagenti installer..."
+    ./scripts/kind/setup-kagenti.sh --with-all --skip-cluster --cluster-name "$CLUSTER_NAME"
 
     log_step "Waiting for platform to be ready..."
     ./.github/scripts/common/40-wait-platform-ready.sh
@@ -340,8 +340,8 @@ fi
 
 if [ "$RUN_KAGENTI_UNINSTALL" = "true" ]; then
     log_phase "PHASE 5: Uninstall Kagenti Platform"
-    log_step "Running cleanup-install.sh..."
-    ./deployments/ansible/cleanup-install.sh || {
+    log_step "Running cleanup-kagenti.sh..."
+    ./scripts/kind/cleanup-kagenti.sh --cluster-name "$CLUSTER_NAME" || {
         log_error "Kagenti uninstall failed (non-fatal)"
     }
 else
--- a/charts/kagenti-deps/templates/mlflow.yaml
+++ b/charts/kagenti-deps/templates/mlflow.yaml
@@ -159,6 +159,7 @@ spec:
           {{- end }}
           export MLFLOW_SERVER_ALLOWED_HOSTS="$ALLOWED_HOSTS" && \
           {{- if .Values.mlflow.auth.enabled }}
+          export MLFLOW_SERVER_ALLOWED_HOSTS="$ALLOWED_HOSTS" && \
           # Set MLflow config via environment variables (used by mlflow_oidc_auth.app)
           # Both BACKEND_STORE_URI and TRACKING_URI needed for otel_router to use PostgreSQL
           export MLFLOW_BACKEND_STORE_URI="postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@postgres:5432/mlflow" && \
--- a/charts/kagenti-deps/templates/mlflow.yaml
+++ b/charts/kagenti-deps/templates/mlflow.yaml
@@ -181,14 +181,6 @@ spec:
           " && \
           mlflow db upgrade "$MLFLOW_BACKEND_STORE_URI" && \
           echo "MLflow database migrations complete" && \
-          #
-          # WORKAROUND: mlflow-oidc-auth doesn't include MLflow's otel_router
-          # which provides the /v1/traces OTLP endpoint for trace ingestion.
-          # We patch it at runtime to:
-          # 1. Add the missing otel_router for /v1/traces OTLP endpoint
-          # 2. Exclude /v1/traces from OIDC auth (mTLS via Istio handles auth)
-          # TODO: Remove this workaround when mlflow-oidc-auth includes otel_router
-          #
           python -c '
           # IMPORTANT: Patch BEFORE importing app (app is created at import time)
           from mlflow_oidc_auth.middleware.auth_middleware import AuthMiddleware
@@ -248,7 +240,6 @@ spec:
 
           # NOW import the app (which will use our patched functions)
           from mlflow_oidc_auth.app import app
-          print("App created with /v1/traces endpoint")
 
           # Gate /v1/traces with 503 until database init completes
           # This prevents OTEL collector from permanently dropping traces
--- a/charts/kagenti-deps/templates/mlflow.yaml
+++ b/charts/kagenti-deps/templates/mlflow.yaml
@@ -152,14 +152,20 @@ spec:
           ALLOWED_HOSTS="mlflow,mlflow:5000,mlflow.{{ .Release.Namespace }},mlflow.{{ .Release.Namespace }}:5000,mlflow.{{ .Release.Namespace }}.svc.cluster.local,mlflow.{{ .Release.Namespace }}.svc.cluster.local:5000,localhost,127.0.0.1"
           if [ -n "$ROUTE_HOST" ]; then
             ALLOWED_HOSTS="$ROUTE_HOST,$ALLOWED_HOSTS"
+            CORS_ORIGIN="https://$ROUTE_HOST"
+          else
+            CORS_ORIGIN="https://mlflow-{{ .Release.Namespace }}.apps.{{ .Values.domain | default "cluster.local" }}"
           fi
           {{- else }}
           ALLOWED_HOSTS="mlflow.{{ .Values.domain | default "localtest.me" }}:*,mlflow:*,mlflow.{{ .Release.Namespace }}:5000,mlflow.{{ .Release.Namespace }}.svc.cluster.local:5000,localhost:*,127.0.0.1:*"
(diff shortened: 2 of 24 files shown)


pdettori:
## fix(auth): add EXPECTED_AUDIENCE to authbridge-config ConfigMap

**Root cause:** The `authbridge-unified` ext_proc sidecar (envoy-proxy) reads `EXPECTED_AUDIENCE` from the `authbridge-config` ConfigMap for inbound JWT audience validation. The ConfigMap only had `JWT_AUDIENCE` (consumed by client-registration), so `EXPECTED_AUDIENCE` was unset — causing the ext_proc to reject all inbound JWTs with `"aud" not satisfied`.

**Symptom:** UI chat client received no response from agents. Agent card discovery (GET `/.well-known/agent-card.json`) worked because it's in the `bypass.inbound_paths` list, but chat POST requests were blocked at the sidecar.

**Fix:** Added `EXPECTED_AUDIENCE` key to the `authbridge-config` ConfigMap in `charts/kagenti/templates/agent-namespaces.yaml`, with the same value as `JWT_AUDIENCE` (`{{ keycloak.publicUrl }}/realms/{{ keycloak.realm }}`).

**Verified:** Streaming chat request to weather-service in team1 returns successful responses with no JWT errors after the fix.

.github/scripts/local-setup/kind-full-test.sh
```diff
@@ -240,14 +240,14 @@ if [ "$RUN_INSTALL" = "true" ]; then
 
     if [ "$CLEAN_KAGENTI" = "true" ]; then
         log_step "Uninstalling Kagenti (--clean-kagenti)..."
-        ./deployments/ansible/cleanup-install.sh || true
+        ./scripts/kind/cleanup-kagenti.sh || true
     fi
 
     log_step "Creating secrets..."
     ./.github/scripts/common/20-create-secrets.sh
```
huang195, line 247: **suggestion**: Ciao Paolo — the CI workflows were updated to call `20-create-secrets-bash.sh`, but this local dev script still calls the old Ansible-format `20-create-secrets.sh`. Since the installer is now `setup-kagenti.sh`, this should probably be `20-create-secrets-bash.sh` for consistency, no? Otherwise local devs running `kind-full-test.sh` get secrets in the wrong format. Piccolo fix! 🍝

charts/kagenti-deps/templates/mlflow.yaml
```diff
@@ -152,13 +152,20 @@ spec:
...
+            CORS_ORIGIN="https://mlflow-{{ .Release.Namespace }}.apps.{{ .Values.domain | default "cluster.local" }}"
           fi
           {{- else }}
           ALLOWED_HOSTS="mlflow.{{ .Values.domain | default "localtest.me" }}:*,mlflow:*,mlflow.{{ .Release.Namespace }}:5000,mlflow.{{ .Release.Namespace }}.svc.cluster.local:5000,localhost:*,127.0.0.1:*"
           ALLOWED_HOSTS="$ALLOWED_HOSTS,mlflow.{{ .Values.domain | default "localtest.me" }},mlflow,mlflow.{{ .Release.Namespace }},mlflow.{{ .Release.Namespace }}.svc.cluster.local,localhost,127.0.0.1"
+          CORS_ORIGIN="{{ .Values.mlflow.corsOrigin | default (printf "http://mlflow.%s:%s" (.Values.domain | default "localtest.me") (toString (.Values.ingressPort | default 8080))) }}"
           {{- end }}
           export MLFLOW_SERVER_ALLOWED_HOSTS="$ALLOWED_HOSTS" && \
           {{- if .Values.mlflow.auth.enabled }}
+          export MLFLOW_SERVER_ALLOWED_HOSTS="$ALLOWED_HOSTS" && \
```
huang195, line 166: **nit**: Duplicate `export MLFLOW_SERVER_ALLOWED_HOSTS` — it's already exported on line 164 (outside the auth conditional block), then exported again here inside `{{- if .Values.mlflow.auth.enabled }}`. Looks like a rebase artifact. Harmless at runtime, but the duplication might confuse the next reader.

charts/kagenti-deps/templates/mlflow.yaml
```diff
@@ -248,7 +247,25 @@ spec:
...
+          # CORS: the stock MLflow UI fetch() requires Access-Control-Allow-Origin
+          # when mlflow-oidc-auth is enabled (the auth redirect on expired sessions
+          # triggers opaque-redirect handling in the browser Fetch spec).
+          from starlette.middleware.cors import CORSMiddleware
+          import os as _os
+          _allowed_origins = [
+              o.strip() for o in
+              _os.environ.get("MLFLOW_CORS_ALLOWED_ORIGINS", "").split(",")
+              if o.strip()
+          ] or ["*"]
```
huang195, line 260: **suggestion**: The `or ["*"]` fallback combined with `allow_credentials=True` (line 264) violates the CORS spec — browsers reject `Access-Control-Allow-Origin: *` when credentials are included. If `MLFLOW_CORS_ALLOWED_ORIGINS` is unset, this silently breaks rather than failing loudly.

Consider either:
- Raising an error if the env var is empty (fail fast)
- Defaulting to a safe origin (e.g. `http://localhost:5000`)

The startup script *should* always set it, but defensive coding here would save a confusing debugging session if something upstream changes.

kagenti/tests/e2e/common/test_kind_installer.py
```diff
@@ -0,0 +1,734 @@
...
+@pytest.mark.kind_only
+class TestMLflow:
+    """Components installed by --with-mlflow."""
+
+    def test_mlflow_deployment_ready(self, installed_flags, k8s_apps_client):
+        """mlflow deployment must be ready when --with-mlflow."""
+        _skip_unless_flag(installed_flags, "mlflow")
+        _assert_deployment_ready(k8s_apps_client, "mlflow", "kagenti-system")
+
+    def test_mlflow_oauth_secret_exists(self, installed_flags, k8s_client):
```
huang195, line 550: **nit**: I see from the commit message that this skip is intentional (CI deploys with `mlflow.auth` disabled, so the Helm hook never creates the secret). Makes sense! But a reader without that context might wonder why a 404 on a required secret is a skip, not a failure. A brief inline comment like `# CI may deploy --with-mlflow but without mlflow.auth.enabled` would save them a trip to `git log`. Una piccola nota per il prossimo lettore!

kagenti/backend/app/routers/chat.py
```diff
@@ -490,7 +490,12 @@ async def _stream_a2a_response(
 
     except httpx.HTTPStatusError as e:
         error_msg = f"Agent error: {e.response.status_code}"
-        logger.error(f"{error_msg}: {e.response.text[:500]}")
+        try:
+            await e.response.aread()
```
huang195, line 494: **praise**: Bravo! This properly handles the `httpx.ResponseNotRead` issue — calling `await e.response.aread()` before accessing `.text` on a streaming response, with a clean fallback to `str(e)`. This was masking the real auth error, and the fix is both correct and minimal. Come una ricetta perfetta — nothing extra, nothing missing.

charts/kagenti/values.yaml
```diff
@@ -70,6 +70,7 @@ ui:
   url: http://kagenti-ui.localtest.me:8080
   # Frontend (UI v2) configuration
   frontend:
+    enabled: true
```
huang195, line 73: **nit**: New toggle without documentation — a one-line comment like `# Set to false for backend-only deploys (no frontend)` would help users scanning `values.yaml` understand when to flip this.

scripts/kind/setup-kagenti.sh
```diff
@@ -0,0 +1,1136 @@
+#!/usr/bin/env bash
```
huang195, line 1: **suggestion / general**: The `--dry-run` flag is a great feature, but approximately half of the ~40 `kubectl` mutating calls (apply, create, label, patch, wait) run directly instead of through `run_cmd`. In dry-run mode these will still mutate the cluster. Consider wrapping all mutating commands consistently, or adding a note in `--help` that dry-run coverage is partial.

**praise**: That said — the overall composable architecture here is fantastico. `set -euo pipefail` everywhere, pinned versions for every dependency, idempotent operations, auto-dependency resolution between flags. This is a massive upgrade from the Ansible monolith. Complimenti, chef! 👨‍🍳

huang195 (commented):
Mamma mia, Paolo — this is a *massive* rework and it's bellissimo! Replacing the Ansible installer with composable bash scripts is like swapping a five-course prix fixe for a build-your-own antipasti bar. The `--with-*` flag architecture, the idempotent `helm upgrade --install` pattern, the auto-dependency resolution — perfetto.

The commit history tells a great story: each fix is atomic with an excellent commit message explaining the *why*. The Keycloak realm-init OIDC scope/audience wiring is correctly structured, the chat.py streaming fix is clean, and the cleanup script's robustness (finalizer stripping, `--ignore-not-found`) shows real production experience. The test matrix covering positive AND negative assertions per flag is also ottimo.

Two items worth a second look: (1) the CORS `["*"]` + credentials fallback that browsers will reject, and (2) the `--dry-run` coverage gap. Neither blocks merge, but both are worth addressing.

**Areas reviewed**: Shell, Helm/K8s, Python, CI/GitHub Actions, Tests, Docs
**Commits**: 35 commits, all signed-off ✅
**CI status**: All 24 checks passing ✅
**Security**: No hardcoded secrets (CI dummy values are properly stubbed), GitHub Actions pinned to SHAs, `.secrets.yaml` is in `.gitignore` ✅

**Verdict**: COMMENT — Suggestions only, no blockers. Author can decide on follow-ups.

huang195 (approved):
/lgtm

PR #1285, opened 2026-04-19
Fix: add clickable URL links

### LINKED ISSUE #1284: URL fields are not clickable
### Feature Description

I expected the URL links in the Kagenti UI to be clickable. It seems that an `<a>` tag would be more appropriate than a `<ClipboardCopy>` or `<code>` element to represent all read-only URL fields. 

### Proposed Solution

Update `ui-v2` elements to support clickable URL links.

### Want to contribute?

### Additional Context

_No response_

kabicin:
## Summary

Updates read-only URL links in `ui-v2` to use clickable `<a>` tags. 

Fixes https://github.com/kagenti/kagenti/issues/1284

### DIFF
--- a/kagenti/ui-v2/src/pages/AgentDetailPage.tsx
+++ b/kagenti/ui-v2/src/pages/AgentDetailPage.tsx
@@ -3,7 +3,6 @@
 
 import React from 'react';
 import { useParams, useNavigate } from 'react-router-dom';
-import { copyToClipboard } from '../utils/clipboard';
 import ReactMarkdown from 'react-markdown';
 import remarkGfm from 'remark-gfm';
 import {
@@ -32,7 +31,6 @@ import {
   Alert,
   Grid,
   GridItem,
-  ClipboardCopy,
   Split,
   SplitItem,
   Flex,
@@ -472,9 +470,9 @@ export const AgentDetailPage: React.FC = () => {
                       <DescriptionListGroup>
                         <DescriptionListTerm>Agent URL</DescriptionListTerm>
                         <DescriptionListDescription>
-                          <ClipboardCopy isReadOnly hoverTip="Copy" clickTip="Copied" onCopy={copyToClipboard}>
+                          <a href={agentUrl} target="_blank" rel="noopener noreferrer">
                             {agentUrl}
-                          </ClipboardCopy>
+                          </a>
                         </DescriptionListDescription>
                       </DescriptionListGroup>
                       {serviceInfo && (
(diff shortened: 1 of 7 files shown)


kagenti/ui-v2/src/pages/ToolDetailPage.tsx
```diff
@@ -848,9 +846,9 @@ export const ToolDetailPage: React.FC = () => {
                   <DescriptionListGroup>
                     <DescriptionListTerm>MCP Server URL (in-cluster)</DescriptionListTerm>
                     <DescriptionListDescription>
```
pdettori, line 848: **suggestion**: `mcpInClusterUrl` is a Kubernetes in-cluster URL (e.g. `http://svc.ns.svc.cluster.local:8080`) — not reachable from the user's browser. Making it a clickable `<a>` is misleading; the previous `ClipboardCopy` was more useful here since users copy it for in-cluster `curl`.

Consider keeping `ClipboardCopy` for in-cluster URLs while converting browser-accessible URLs to `<a>` tags.

kagenti/ui-v2/src/pages/AgentDetailPage.tsx
```diff
@@ -472,9 +470,9 @@ export const AgentDetailPage: React.FC = () => {
                       <DescriptionListGroup>
                         <DescriptionListTerm>Agent URL</DescriptionListTerm>
                         <DescriptionListDescription>
```
pdettori, line 472: **suggestion**: Replacing `ClipboardCopy` with a plain `<a>` drops the copy-to-clipboard UX. For URLs users frequently copy (agent URLs, MCP server URLs), consider combining both: a clickable link *with* a copy button.

PatternFly's `ClipboardCopy` supports custom children, or you could add a small copy icon next to the `<a>`. This applies to the same pattern in `ToolDetailPage.tsx` and `BuildProgressPage.tsx` as well.

kagenti/ui-v2/src/pages/AgentDetailPage.tsx
```diff
@@ -568,12 +566,6 @@ export const AgentDetailPage: React.FC = () => {
                                         </DescriptionListDescription>
```
pdettori, line 566: **praise**: Good cleanup — the removed `agentCard.url` section was a duplicate of the `agentUrl` field shown above (since `agentUrl` is derived from `agentCard?.url`).

pdettori (commented):
Small, focused UI PR that converts read-only URL displays to clickable links across 7 files. All `<a>` tags properly use `target="_blank" rel="noopener noreferrer"`. Unused imports are cleaned up correctly.

The main consideration is that copy-to-clipboard functionality is lost — worth evaluating whether both clicking and copying are needed for these URLs.

**Areas reviewed**: Frontend (React/TypeScript)
**CI status**: "Deploy & Test" failing (Keycloak auth infra issue, unrelated to this PR). All lint/security checks pass.

pdettori (approved):
/lgtm

PR #1295, opened 2026-04-20
:bug: Import Kagenti realm

evaline-ju:

## Summary

On a fresh Kind install, Keycloak was starting without importing the kagenti realm. The keycloak-realm-import ConfigMap (created by keycloak-realm-init.yaml) contained the full realm definition but was never consumed because the Keycloak pod had no volume mount for it. The fix mounts the ConfigMap into /opt/keycloak/data/import and adds `--import-realm` to the Keycloak startup args, so the realm is imported atomically before the pod becomes ready.

Fixes (no separate issue yet) - seeing `Failed to fetch URL: Required role(s): kagenti-operator` on env var import on agent deploy


evaline-ju:
Closing in favor of #1296

PR #1306, opened 2026-04-22
fix: Enable Kind image preloading for dev env

huang195:
## Summary
- Re-enable `kind_images_preload: true` in `dev_values.yaml` to avoid Docker Hub rate limiting (429) during Kind cluster installs
- Bump three stale image versions in `preload-images.txt` to match current releases:
  - `ui-v2`: v0.5.0-alpha.11 → v0.6.0-alpha.4
  - `kagenti-operator`: 0.2.0-alpha.19 → 0.2.0-alpha.30
  - `backend`: v0.5.0-alpha.11 → v0.6.0-alpha.4

## Context
The `kind_images_preload` flag was accidentally disabled in commit 013fd8c0 (collateral damage from ToolHive removal). Without preloading, Kind nodes pull Docker Hub images at runtime, frequently hitting rate limits — causing cascading failures in SPIRE OIDC Discovery Provider rollout and SPIFFE IdP setup jobs.

The flag is set only in `dev_values.yaml` so it only affects `--env dev` (Kind local) installs. Other environments (minimal, auth, ocp, ocp-minimal) are unaffected.

## Test plan
- [ ] Run `./run-install.sh --env dev` on a fresh Kind cluster and verify images are preloaded
- [ ] Confirm SPIRE OIDC Discovery Provider rolls out without ImagePullBackOff
- [ ] Verify no impact on `--env ocp` installs

### DIFF
--- a/deployments/ansible/kind/preload-images.txt
+++ b/deployments/ansible/kind/preload-images.txt
@@ -20,6 +20,6 @@ quay.io/jetstack/cert-manager-webhook:v1.17.2
 quay.io/keycloak/keycloak:26.5.2
 ghcr.io/shipwright-io/build/shipwright-build-controller:v0.14.0
 ghcr.io/shipwright-io/build/shipwright-build-webhook:v0.14.0
-ghcr.io/kagenti/kagenti/ui-v2:v0.5.0-alpha.11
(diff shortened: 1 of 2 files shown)


huang195:
As discussed with Paolo on Slack, we can use `--preload` flag if needed. No need to make it a default. Closing

PR #1323, opened 2026-04-23
fix: Pass empty secret_values_file when default is missing

### LINKED ISSUE #1321: Installer secret_values_file check inconsistent between wrapper and playbook
### Describe the bug

`run-install.sh` and the underlying ansible playbook disagree about whether `.secret_values.yaml` is required.

The wrapper script warns "continuing without secrets" when the file is absent, implying it's optional. But the playbook's `01_setup_vars.yaml` has a stricter check that fails hard:

```
Secret values file '.../envs/.secret_values.yaml' not found.
```

### Steps to Reproduce

1. Clone kagenti, do NOT create `.secret_values.yaml`
2. Run `run-install.sh --env ocp`
3. Wrapper prints a warning about missing secrets and says "continuing without secrets"
4. Playbook fails at `01_setup_vars.yaml` because the file doesn't exist

### Expected Behavior

Either:
- The playbook should tolerate a missing file (matching the wrapper's "continuing without" message), OR
- The wrapper should fail early with a clear error instead of saying "continuing without secrets" when the playbook will fail anyway

### Workaround

Copy `secret_values.yaml.example` to `.secret_values.yaml`:
```bash
cp envs/secret_values.yaml.example envs/.secret_values.yaml
```

### Additional Context

- Tested on OCP 4.20.18 (fresh RHPDS cluster)
- The workaround is trivial, but the UX is confusing — the wrapper says it will continue, then fails

rdwj:
## Problem

`run-install.sh` and the ansible playbook disagree about whether `.secret_values.yaml` is required:

1. **Wrapper** (`run-install.sh`): When the default `.secret_values.yaml` is missing, prints `WARNING: ...continuing without secrets` and does NOT pass `secret_values_file` to ansible.

2. **Playbook** (`01_setup_vars.yaml`): Falls through to `secrets.file` from `default_values.yaml` (which is hardcoded to `../envs/.secret_values.yaml`), resolves the path, stats it, and fails hard:
   ```
   Secret values file '.../envs/.secret_values.yaml' not found.
   ```

The user sees "continuing without secrets" followed immediately by a hard failure — the wrapper promises something the playbook doesn't deliver.

## Root Cause

`default_values.yaml` line 13 defines `secrets.file: "../envs/.secret_values.yaml"`. The playbook's variable resolution on line 6 of `01_setup_vars.yaml` is:

```yaml
secret_values_file: "{{ secret_values_file | default(secrets.file | default('')) }}"
```

When the wrapper doesn't pass `secret_values_file`, `secrets.file` kicks in as the default, pointing to a file that doesn't exist.

## Solution

When the default file is missing, the wrapper now explicitly passes `secret_values_file: ""` in the extra-vars. This overrides the `secrets.file` default. All downstream conditional blocks in `01_setup_vars.yaml` are gated on `secret_values_file | length > 0`, so they are cleanly skipped:

- Path resolution (line 17): `when: ... secret_values_file | length > 0` → **skipped**
- Stat check (line 23): `when: secret_values_file_resolved is defined ...` → **skipped** (never set)
- Fail task (line 28): same guard → **skipped**
- include_vars (line 34): same guard → **skipped**

## Verification

Traced the variable resolution through both paths:

**Before fix** (no `.secret_values.yaml`):
```
wrapper: WARNING, no secret_values_file passed
playbook: secret_values_file = "" | default("../envs/.secret_values.yaml") = "../envs/.secret_values.yaml"
playbook: stat("../envs/.secret_values.yaml") → not found → FAIL
```

**After fix** (no `.secret_values.yaml`):
```
wrapper: WARNING, passes secret_values_file=""
playbook: secret_values_file = "" | default(...) = "" (defined, so default doesn't trigger)
playbook: "" | length > 0 → false → all blocks skipped ✓
```

The explicit `--secret` path is unchanged: if a user passes `--secret <path>` and the file is missing, the wrapper still fails early with `exit 2`.

Fixes #1321

### DIFF
--- a/deployments/ansible/run-install.sh
+++ b/deployments/ansible/run-install.sh
@@ -190,7 +190,9 @@ if [[ -n "$SECRET_FILE" ]]; then
 
   # Check existence: if the user explicitly provided --secret and the file
   # is missing, fail early. If we're using the default and it's missing,
-  # warn and skip adding the secret (the playbook will behave accordingly).
+  # warn and skip. We must pass an empty secret_values_file to override the
+  # default in default_values.yaml (secrets.file), otherwise the playbook
+  # resolves that default path, stats it, and fails.
   if [[ -f "$secret_resolved" ]]; then
     JSON_ENTRIES+=("\"secret_values_file\": \"$secret_resolved\"")
   else
@@ -199,6 +201,7 @@ if [[ -n "$SECRET_FILE" ]]; then
       exit 2
     else
       echo "WARNING: default secret values file not found at $secret_resolved; continuing without secrets." >&2
+      JSON_ENTRIES+=("\"secret_values_file\": \"\"")
     fi
   fi
 fi


pdettori (approved):
Clean, well-scoped fix. Verified against `upstream/main`:

- `default_values.yaml` confirms `secrets.file: "../envs/.secret_values.yaml"` is the problematic default that triggers the hard failure
- `01_setup_vars.yaml` uses `secret_values_file | default(secrets.file | default(''))` — an explicit empty string is "defined" in Jinja2, so the `default()` filter won't fall through to `secrets.file`
- All downstream guards (`path resolution`, `stat`, `fail`, `include_vars`) check `secret_values_file | length > 0` — empty string cleanly skips them all

Commit is signed-off, uses correct `Assisted-By` attribution, and CI is green.

**Areas reviewed**: Shell script change, Ansible playbook variable resolution flow

PR #1345, opened 2026-04-26
chore(skills): pin image tags for all release types

mrsabath:
## Summary
- Require image tag pinning for **all** release types, including alphas — previously alphas allowed unpinned `tag: latest`
- Add full pinning checklist to Phase 2.1 (Alpha), matching what RC/GA already require
- Update quick reference table so every release type shows "Yes — no `latest`"

## Test plan
- [ ] Review the updated skill file for consistency across all four release types
- [ ] Run `/release status` to confirm the skill loads correctly

### DIFF
--- a/.claude/skills/release/SKILL.md
+++ b/.claude/skills/release/SKILL.md
@@ -86,7 +86,7 @@ grep -A2 'name: kagenti-' charts/kagenti/Chart.yaml
 grep -n 'tag:' charts/kagenti/values.yaml
 ```
 
-Flag any `tag: latest` entries — these must be pinned before RC or GA.
+Flag any `tag: latest` entries — these must be pinned before any release (including alphas).
 
 ### 1.4 Determine release type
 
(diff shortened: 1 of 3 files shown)


pdettori (approved):
Two independent, well-motivated changes. The Dockerfile fix correctly separates dependency installation from app source to avoid stale Docker layer cache bugs — `uv pip install -r pyproject.toml` is valid uv syntax and the pre-push smoke test gates the push on a successful import check. The release skill update is straightforward: alphas now follow the same tag-pinning discipline as RC/GA, which avoids a reproducibility gap between release types. CI is fully green.

**Areas reviewed**: Skills/Markdown, CI/GitHub Actions, Dockerfile
**Commits**: 2 commits, both signed-off ✓
**CI status**: All passing (E2E HyperShift requires `/run-e2e` from maintainer — expected)

---

**nit**: The smoke-test build step (`.github/workflows/build.yaml`) has no `cache-from`/`cache-to`. The main build step likely uses GHA cache. Adding `cache-from: type=gha` / `cache-to: type=gha,mode=min` to the smoke step would avoid rebuilding all layers from scratch for the smoke run. Non-blocking.

PR #1346, opened 2026-04-27
feat: add Sandbox workload type to agent UI pages

cwiklik:
## Summary

- **ImportAgentPage**: Add Sandbox as 4th workload type option in the dropdown, gated behind `agentSandbox` feature flag. Uses array `.map()` pattern for PatternFly `FormSelect` compatibility (conditional JSX children don't work with native `<select>`)
- **AgentCatalogPage**: Render purple label for sandbox workload type in the agents table
- **AgentDetailPage**: Purple label + correct `agents.x-k8s.io/v1alpha1` apiVersion and `Sandbox` kind in YAML dump
- **Feature flags endpoint**: Make `GET /api/v1/config/features` public (no auth required) so feature flags load before login — flags are non-sensitive UI configuration
- **useFeatureFlags hook**: Remove auth dependency, fetch flags unconditionally on mount

Part of #1155

## Test plan

- [ ] Verify Sandbox appears as 4th option in Import Agent workload dropdown when `agentSandbox` flag is enabled
- [ ] Verify Sandbox does NOT appear when flag is disabled
- [ ] Verify purple "Sandbox" label renders in Agent Catalog table
- [ ] Verify Agent Detail page shows correct apiVersion/kind for sandbox agents
- [ ] Verify `/api/v1/config/features` returns flags without auth token
- [ ] Verify other workload types (Deployment, StatefulSet, Job) are unaffected

Assisted-By: Claude (Anthropic AI)

### DIFF
--- a/kagenti/ui-v2/src/pages/AgentCatalogPage.tsx
+++ b/kagenti/ui-v2/src/pages/AgentCatalogPage.tsx
@@ -3,6 +3,7 @@
 
 import React, { useState } from 'react';
 import { useNavigate } from 'react-router-dom';
+import { workloadTypeColor } from '@/utils/workloadType';
 import {
   PageSection,
   Title,
@@ -121,15 +122,7 @@ export const AgentCatalogPage: React.FC = () => {
   const renderWorkloadType = (workloadType: string | undefined) => {
     const type = workloadType || 'deployment';
     const label = type.charAt(0).toUpperCase() + type.slice(1);
-    let color: 'grey' | 'orange' | 'gold' | 'purple' = 'grey';
-    if (type === 'job') {
-      color = 'orange';
-    } else if (type === 'statefulset') {
-      color = 'gold';
-    } else if (type === 'sandbox') {
-      color = 'purple';
-    }
-    return <Label color={color} isCompact>{label}</Label>;
+    return <Label color={workloadTypeColor(type)} isCompact>{label}</Label>;
   };
 
   const renderStatusBadge = (status: string) => {
--- a/kagenti/ui-v2/src/pages/AgentCatalogPage.tsx
+++ b/kagenti/ui-v2/src/pages/AgentCatalogPage.tsx
@@ -121,11 +121,13 @@ export const AgentCatalogPage: React.FC = () => {
   const renderWorkloadType = (workloadType: string | undefined) => {
     const type = workloadType || 'deployment';
     const label = type.charAt(0).toUpperCase() + type.slice(1);
-    let color: 'grey' | 'orange' | 'gold' = 'grey';
+    let color: 'grey' | 'orange' | 'gold' | 'purple' = 'grey';
     if (type === 'job') {
       color = 'orange';
     } else if (type === 'statefulset') {
       color = 'gold';
+    } else if (type === 'sandbox') {
+      color = 'purple';
     }
     return <Label color={color} isCompact>{label}</Label>;
   };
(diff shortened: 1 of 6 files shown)


kagenti/ui-v2/src/pages/ImportAgentPage.tsx
```diff
@@ -163,7 +165,16 @@ export const ImportAgentPage: React.FC = () => {
   const [showImportModal, setShowImportModal] = useState(false);
```
huang195, line 166: **suggestion**: When `agentSandbox` is enabled, the default becomes `'sandbox'` but the dropdown still labels Deployment as `"(Recommended)"`. These two signals conflict — consider conditionally labeling Sandbox as recommended when the flag is active, or keeping `'deployment'` as the default regardless of the flag.

kagenti/ui-v2/src/pages/AgentDetailPage.tsx
```diff
@@ -989,8 +998,8 @@ export const AgentDetailPage: React.FC = () => {
                 >
                   {yaml.dump(
                     {
-                      apiVersion: agent.workloadType === 'statefulset' ? 'apps/v1' : agent.workloadType === 'job' ? 'batch/v1' : 'apps/v1',
-                      kind: agent.workloadType === 'statefulset' ? 'StatefulSet' : agent.workloadType === 'job' ? 'Job' : 'Deployment',
+                      apiVersion: agent.workloadType === 'sandbox' ? 'agents.x-k8s.io/v1alpha1' : agent.workloadType === 'statefulset' ? 'apps/v1' : agent.workloadType === 'job' ? 'batch/v1' : 'apps/v1',
```
huang195: **nit**: The nested ternary chains for `apiVersion` and `kind` are getting long (4 cases each). A lookup map — same pattern as `workloadTypeColor` above — would improve readability:
```ts
const WORKLOAD_META: Record<string, { apiVersion: string; kind: string }> = {
  sandbox: { apiVersion: 'agents.x-k8s.io/v1alpha1', kind: 'Sandbox' },
  statefulset: { apiVersion: 'apps/v1', kind: 'StatefulSet' },
  job: { apiVersion: 'batch/v1', kind: 'Job' },
  deployment: { apiVersion: 'apps/v1', kind: 'Deployment' },
};
```

kagenti/ui-v2/src/pages/AgentCatalogPage.tsx
```diff
@@ -121,11 +121,13 @@ export const AgentCatalogPage: React.FC = () => {
   const renderWorkloadType = (workloadType: string | undefined) => {
     const type = workloadType || 'deployment';
     const label = type.charAt(0).toUpperCase() + type.slice(1);
-    let color: 'grey' | 'orange' | 'gold' = 'grey';
+    let color: 'grey' | 'orange' | 'gold' | 'purple' = 'grey';
```
huang195: **nit**: This color logic duplicates `workloadTypeColor` in AgentDetailPage (switch statement). Could extract to a shared utility (e.g. `utils/workloadType.ts`) to keep the two pages in sync as more workload types are added.

huang195 (approved):
Clean, well-structured PR that correctly gates the sandbox workload type behind a feature flag. The security change (making `/api/v1/config/features` public) is reasonable — the endpoint returns only boolean UI configuration flags.

**Areas reviewed**: Python (backend), Frontend (React)
**Commits**: 3 commits, all signed-off ✓
**CI status**: All lint/security/tests passing

Issue #1366, opened 2026-04-27
feature: AuthBridge API in support of user interface


esnible:
### Feature Description

This API will be the mechanism for implementing #1310

The kagenti-backend will offer two new endpoints:
/{namespace}/{name}/identity-config
/{namespace}/{name}/identity-status

The initial implementation return the JSON from https://github.com/kagenti/kagenti-extensions/issues/339

The initial implementation will return the JSON from the first reachable pod.  We may add more functionality (handling scaling) in a follow-up, once we have verified the information is useful.

### Proposed Solution

Modify kagenti/backend/app/routers/agents.py to support the two new endpoints.

We will require `ROLE_OPERATOR` to get the information.

### Want to contribute?

- [x] I would like to work on this issue.

esnible:
Completed by https://github.com/kagenti/kagenti/pull/1367

PR #1367, opened 2026-04-28
Feat: Provide API access to AuthBridge config and statistics

esnible:
## Summary

For #1366

Introduces two new kagenti-backend endpoints:
/{namespace}/{name}/identity-config
/{namespace}/{name}/identity-status

Deploy Kagenti.  Enable this feature with

```bash
oc -n kagenti-system set env deployment/kagenti-backend KAGENTI_FEATURE_FLAG_AUTHBRIDGE_API=TRUE
```

Deploy an Agent and then curl the Kagenti API:

```bash
curl -v -H "Authorization: Bearer ${KAGENTI_TOKEN}" 'http://localhost:8000/api/v1/agents/team1/my-weather-service-with-authbridge/identity-config'
curl -v -H "Authorization: Bearer ${KAGENTI_TOKEN}" 'http://localhost:8000/api/v1/agents/team1/my-weather-service-with-authbridge/identity-status'
```

(Note that getting a Kagenti token has been difficult.  I go into Keycloak, Kagenti realm, and Kagenti client, and check Allow Direct Access Grants.  Then I do

```bash
KAGENTI_TOKEN=$(curl -s -X POST \
    http://keycloak.localtest.me:8080/realms/kagenti/protocol/openid-connect/token \
    -d "grant_type=password&client_id=kagenti&username=admin&password=${KAGENTI_UI_PW}" | jq --raw-output ".access_token")
echo KAGENTI_TOKEN is $KAGENTI_TOKEN
```
)

### DIFF
--- a/kagenti/backend/app/core/constants.py
+++ b/kagenti/backend/app/core/constants.py
@@ -199,6 +199,14 @@ static_resources:
         "@type": type.googleapis.com/envoy.extensions.filters.listener.tls_inspector.v3.TlsInspector
     filter_chains:
     - filter_chain_match:
+        destination_port: 9093
+      filters:
+      - name: envoy.filters.network.tcp_proxy
+        typed_config:
+          "@type": type.googleapis.com/envoy.extensions.filters.network.tcp_proxy.v3.TcpProxy
+          stat_prefix: outbound_passthrough_9093
+          cluster: original_destination
+    - filter_chain_match:
         transport_protocol: tls
       filters:
       - name: envoy.filters.network.tcp_proxy
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -14,7 +14,6 @@ import ipaddress
 from datetime import datetime, timezone
 from typing import Any, Dict, List, Literal, Optional
 from urllib.parse import urlparse
-from urllib.request import urlopen
 
 import httpx
 import yaml
@@ -3866,7 +3865,9 @@ async def fetch_env_from_url(request: FetchEnvUrlRequest) -> FetchEnvUrlResponse
         raise HTTPException(status_code=500, detail=f"Unexpected error: {str(e)}")
 
 
-@router.get("/{namespace}/{name}/identity-config", dependencies=[Depends(require_roles(ROLE_OPERATOR))])
+@router.get(
+    "/{namespace}/{name}/identity-config", dependencies=[Depends(require_roles(ROLE_OPERATOR))]
+)
 async def get_agent_identity_config(
     namespace: str,
     name: str,
@@ -3890,14 +3891,25 @@ async def get_agent_identity_config(
     for endpoint_slice in endpoint_slices.get("items", []):
         for endpoint in endpoint_slice.get("endpoints", []):
             for address in endpoint.get("addresses", []):
-                attempts = attempts+1
+                attempts += 1
                 # AuthBridge serves config and status on port 9093
                 url = f"http://{address}:9093/config"
                 try:
-                    r = urlopen(url, timeout=10)
-                    data = json.loads(r.read())
-                    data["AuthBridge"] = True
-                    return data
+                    async with httpx.AsyncClient(timeout=10.0, follow_redirects=True) as client:
+                        logger.debug(f"Making HTTP request to {url}")
+                        response = await client.get(url)
+                        response.raise_for_status()
+
+                        # Validate content isn't too large (max 1MB)
+                        content = response.text
+                        if len(content) > 1024 * 1024:
+                            raise HTTPException(
+                                status_code=413, detail="File content too large (max 1MB)"
+                            )
+
+                        data = json.loads(content)
+                        data["AuthBridge"] = True
+                        return data
                 except Exception as e:
                     # Ignore exceptions -- it could mean a pod is not ready,
                     # Check the next endpoint.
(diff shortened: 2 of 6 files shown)


kagenti/backend/app/routers/agents.py
```diff
@@ -3863,3 +3863,124 @@ async def fetch_env_from_url(request: FetchEnvUrlRequest) -> FetchEnvUrlResponse
...
+
+        content = response.text
+        if len(content) > 1024 * 1024:
+            raise HTTPException(status_code=502, detail="File content too large (max 1MB)")
+
+        data = json.loads(content)
+        if not isinstance(data, dict):
+            raise HTTPException(status_code=502, detail="File content not AuthBridge JSON")
+
+        return data
```
cwiklik, line 3987: **must-fix**: `_get_service_endpoints` declares `kube: KubernetesService = Depends(get_kubernetes_service)` but is called as a plain function (not a FastAPI route). `Depends()` only resolves inside FastAPI's dependency injection framework — in a plain function call without passing `kube`, the parameter will be the raw `Depends` sentinel object, not a `KubernetesService` instance. This will crash at runtime with `AttributeError`.

Fix: remove the `Depends` default and pass `kube` explicitly from the calling endpoints:

```python
def _get_service_endpoints(namespace: str, name: str, kube: KubernetesService) -> List[str]:
```

and at the call sites:

```python
addresses = _get_service_endpoints(namespace=namespace, name=name, kube=kube)
```

kagenti/backend/app/routers/agents.py
```diff
@@ -3863,3 +3863,124 @@ async def fetch_env_from_url(request: FetchEnvUrlRequest) -> FetchEnvUrlResponse
...
+        raise HTTPException(status_code=502, detail=str(e.reason))
+
+    attempts = 0
+    for address in addresses:
+        attempts += 1
+        # AuthBridge serves config and status on port 9093
+        url = f"http://{address}:9093/stats"
+        try:
+            data = await _fetch_authbridge_json(url)
+            data["AuthBridge"] = True
```
cwiklik: **nit**: Redundant `pass` after `logger.info(...)` — the `continue` on the next for-loop iteration is implicit. Same at line 3975.

kagenti/backend/app/core/constants.py
```diff
@@ -198,6 +198,14 @@
       typed_config:
         "@type": type.googleapis.com/envoy.extensions.filters.listener.tls_inspector.v3.TlsInspector
     filter_chains:
+    - filter_chain_match:
+        destination_port: 9093
+      filters:
+      - name: envoy.filters.network.tcp_proxy
+        typed_config:
```
cwiklik, line 205: **suggestion**: The 9093 passthrough is added to `DEFAULT_ENVOY_YAML` (used by UI-created agents) but not to `charts/kagenti/templates/agent-namespaces.yaml` envoy-config ConfigMap (used by Helm-deployed agents). This means the identity endpoints will only work for agents created through the UI, not those deployed via Helm. Should both envoy configs be updated for consistency?

cwiklik (dismissed):
## Review Summary

Adds two new endpoints (`identity-config`, `identity-status`) that proxy to AuthBridge sidecars on port 9093 via EndpointSlice discovery. Architecture is sound — EndpointSlices give direct pod IPs, avoiding kube-proxy. However, `_get_service_endpoints` uses `Depends()` as a default parameter in a non-route function, which will crash at runtime since FastAPI's DI doesn't resolve defaults in plain function calls.

**Areas reviewed**: Python (router, service, constants), Helm/K8s (RBAC, envoy config)
**Commits**: 6 commits, all signed-off: yes
**CI status**: all passing (E2E HyperShift running)

kagenti/backend/app/routers/agents.py
```diff
@@ -3863,3 +3863,125 @@ async def fetch_env_from_url(request: FetchEnvUrlRequest) -> FetchEnvUrlResponse
...
+
+
+async def _fetch_authbridge_json(url: str) -> dict:
+    """
+    Fetch JSON from an AuthBridge sidecar endpoint.
+
+    Raises on HTTP errors, oversized responses, or non-dict payloads.
+    """
+
+    async with httpx.AsyncClient(timeout=10.0, follow_redirects=True) as client:
```
pdettori, line 3974: **suggestion**: `follow_redirects=True` means the httpx client will chase redirects from the pod IP to any URL. Since these target internal pod IPs, risk is low — but `follow_redirects=False` (the default) would eliminate a mild SSRF vector from a compromised AuthBridge sidecar.

kagenti/backend/app/routers/agents.py
```diff
@@ -3863,3 +3863,125 @@ async def fetch_env_from_url(request: FetchEnvUrlRequest) -> FetchEnvUrlResponse
...
+        logger.debug("Making HTTP request to %s", url)
+        response = await client.get(url)
+        response.raise_for_status()
+
+        content = response.text
+        if len(content) > 1024 * 1024:
```
pdettori, line 3980: **nit**: The full response body is loaded into memory via `response.text` before the 1MB size check. For a proper guard, either check the `Content-Length` header before reading, or use `response.stream()` with a running byte counter. The 10s timeout mitigates the worst case, so this is minor.

pdettori (approved):
## Review Summary

All three prior review comments from @cwiklik are addressed:

1. **must-fix** (`Depends()` in plain function) — Fixed. `_get_service_endpoints` now takes `kube` as an explicit parameter, called correctly from both route handlers.
2. **nit** (redundant `pass`) — Fixed. No stray `pass` statements remain.
3. **suggestion** (9093 passthrough missing from Helm) — Fixed. Added to `charts/kagenti/templates/agent-namespaces.yaml`, matching `DEFAULT_ENVOY_YAML`.

**Feature gate**: Correct. `kagenti_feature_flag_authbridge_api` added to `config.py` (off by default), routes conditionally registered at module level.

**CodeQL Log Injection findings**: The code applies `sanitize_log()` (strips `\n`, `\r`, `\x00`) and uses parameterized `%s` logging throughout — these are standard CWE-117 mitigations. CodeQL still flags these because its static analysis can't trace through custom sanitizer functions. These are false positives given the applied defenses, though the CodeQL CI check is red.

**Areas reviewed**: Python (router, service, constants, config), Helm/K8s (RBAC, envoy config)
**Commits**: 9 commits, all signed-off: yes
**CI status**: All green except CodeQL summary check (false positives) and E2E HyperShift (pending)

PR #1379, opened 2026-04-28
fix(otel): use RHOAI MLflow gateway URL (status.url) for trace export instead of internal service URL

ChristianZaccaria:
## Summary
When deploying Kagenti on OpenShift with RHOAI MLflow, the OTel Collector was targeting MLflow's internal service URL, requiring the service-serving CA certificate to be mounted into the collector pod. This PR switches the collector to use the MLflow CR's `status.url` gateway URL — the same URL the kagenti-operator already uses for agent `MLFLOW_TRACKING_URI` injection — removing the need for custom CA machinery. TLS is verified via the container's system CA pool (Let's Encrypt trusted by default on ROSA).

## Changes
- `setup-kagenti.sh` — resolve `MLFLOW_TRACES_ENDPOINT` from MLflow CR `status.url` (mirrors operator's `resolveTrackingURI` logic) instead of hard-coding the internal service URL; the return 0 early-exit is replaced with break so that if the MLflow pod isn't ready after 5m, the script falls through to the status.url polling loop (which exits 1 after another 5m) — catching real deployment failures instead of silently exporting to a potentially wrong URL (worst-case wait increases from ~6m to ~11m).
- `_helpers.tpl` — replace `insecure: true` TLS config with `tls: {}` in the RHOAI MLflow exporter block so HTTPS is verified via the container's system CA pool
- `otel-collector.yaml` — remove `otel-service-ca` ConfigMap, its volume, and volume mount (only needed for the now-removed internal service path).

## Test Plan
- [ ] Prompt an agent and confirm traces appear in the MLflow UI under the target workspace and experiment
- [ ] Verify OTel collector logs show no TLS or authentication errors
- [ ] Confirm `setup-kagenti.sh` correctly reads `status.url` from the MLflow CR and populates `MLFLOW_TRACES_ENDPOINT`

### DIFF
--- a/scripts/ocp/setup-kagenti.sh
+++ b/scripts/ocp/setup-kagenti.sh
@@ -308,8 +308,8 @@ EOF
 
 _mlflow_wait_ready() {
   if $DRY_RUN; then
-    MLFLOW_TRACES_ENDPOINT="https://${MLFLOW_INSTANCE_NAME}.${MLFLOW_NAMESPACE}.svc.cluster.local:8443/v1/traces"
-    echo "  [dry-run] would wait for MLflow Service and pod in $MLFLOW_NAMESPACE"
+    MLFLOW_TRACES_ENDPOINT="https://mlflow-gateway.${DOMAIN}/v1/traces"
+    echo "  [dry-run] would wait for MLflow Service, pod, and gateway URL in $MLFLOW_NAMESPACE"
     return 0
   fi
 
@@ -326,9 +326,6 @@ _mlflow_wait_ready() {
   done
   log_success "MLflow Service found"
 
-  MLFLOW_TRACES_ENDPOINT="https://${MLFLOW_INSTANCE_NAME}.${MLFLOW_NAMESPACE}.svc.cluster.local:8443/v1/traces"
-  log_success "MLflow traces endpoint: $MLFLOW_TRACES_ENDPOINT"
-
   log_info "Waiting for MLflow pod to be Running..."
   tries=0
   while ! $KUBECTL get pods -n "$MLFLOW_NAMESPACE" \
@@ -337,7 +334,7 @@ _mlflow_wait_ready() {
     tries=$((tries + 1))
     if [ $tries -ge 60 ]; then
       log_warn "MLflow pod not Running after 5m — proceeding anyway (OTEL will retry)"
-      return 0
+      break
     fi
     sleep 5
   done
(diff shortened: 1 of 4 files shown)


scripts/ocp/setup-kagenti.sh
```diff
@@ -337,7 +334,7 @@ _mlflow_wait_ready() {
     tries=$((tries + 1))
     if [ $tries -ge 60 ]; then
       log_warn "MLflow pod not Running after 5m — proceeding anyway (OTEL will retry)"
-      return 0
+      break
```
pdettori, line 337: **suggestion**: The `return 0` → `break` change here and at line 349 makes the function stricter. Previously, if the pod wasn't ready after 5m, the script proceeded with the hard-coded internal URL ("OTEL will retry"). Now it falls through to the gateway-URL polling loop, which `exit 1`s after another 5m if `status.url` is never populated.

This is arguably better — it catches real deployment failures instead of silently exporting to a possibly-wrong URL. But the worst-case wait increases from ~6m to ~11m. Worth a brief note in the PR description about this intentional behavior change.

scripts/ocp/setup-kagenti.sh
```diff
@@ -349,11 +346,30 @@ _mlflow_wait_ready() {
     tries=$((tries + 1))
```
pdettori, line 346: **nit** (PR description only): The description says "revert `config-trusted-cabundle` condition to managed-MLflow-only path" but `config-trusted-cabundle` is untouched in this diff — only `otel-service-ca` is removed. Minor wording fix to avoid confusion for future readers.

pdettori (approved):
Reviewed with focus on Kind MLflow integration — **no impact on Kind**. All removed code is gated by `$.Values.openshift` (Kind sets `openshift=false`). The default `traces_endpoint: http://mlflow:5000/v1/traces` in `values.yaml` is unchanged.

The OCP/RHOAI path looks clean: resolving from `status.url` mirrors the operator's `resolveTrackingURI` logic and using the system CA pool removes the service-serving CA machinery.

Two minor inline notes below — nothing blocking.

ChristianZaccaria:
I've updated the PR description and rebased. Thank you for reviewing @pdettori !

evaline-ju:
direct e2e hypershift failure looks to be fixed on main https://github.com/kagenti/kagenti/commit/049c1e122206da7372ed389e569a63570bbb3a5a so this should only require a rebase for tests to pass 🤞

ChristianZaccaria:
@evaline-ju, thank you for reviewing and looking into the test failure. I've rebased, hopefully this will pass now :)

pdettori:
The E2E HyperShift check failure is **unrelated to this PR** — it failed at the "Create HyperShift cluster" infrastructure provisioning step, before any of this PR's code (OTel collector config, MLflow gateway URL changes) is deployed or exercised.

All other checks pass: lint, security scans, DCO, Helm lint, Kind Deploy & Test, backend/chart tests, and CodeQL. Safe to merge.

PR #1380, opened 2026-04-28
fix: replace misleading 'Agent CRD not found' error message

cwiklik:
## Summary

- Replace the legacy "Agent CRD not found" error message in the create-agent 404 handler
- The old message referenced a non-existent Agent CRD (removed during migration)
- New message names the workload type and relevant controllers (Shipwright for source builds, agent-sandbox controller for sandbox workloads)

Refs: https://github.com/kagenti/kagenti/issues/1155

### DIFF
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -3150,7 +3150,12 @@ async def create_agent(
         if e.status == 404:
             raise HTTPException(
                 status_code=404,
-                detail="Agent CRD not found. Is the kagenti-operator installed?",
+                detail=(
+                    f"Required CRD or resource not found for workload type "
+                    f"'{request.workloadType}'. Ensure the necessary controllers "
+                    f"are installed (e.g. Shipwright for source builds, "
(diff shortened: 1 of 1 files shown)


kagenti/backend/app/routers/agents.py
```diff
@@ -3150,7 +3150,12 @@ async def create_agent(
         if e.status == 404:
             raise HTTPException(
                 status_code=404,
-                detail="Agent CRD not found. Is the kagenti-operator installed?",
+                detail=(
```
huang195, line 3153: **nit**: Three of the four string parts have `f` prefix but no `{…}` interpolation — only the second line needs it. Ruff `F541` would flag this. Plain strings concatenate the same way. Take it or leave it.

huang195 (approved):
Dobra robota, Jarek. Clean one-commit fix — replaces a stale "Agent CRD not found" error (referencing infrastructure that no longer exists) with a message that names the actual workload type and relevant controllers. Straightforward and correct.

**Areas reviewed**: Python
**CI status**: all passing

No blocking issues. One nit inline about redundant f-string prefixes — take it or leave it.

PR #1383, opened 2026-04-28
CI: Add PR title verifier workflow

rubambiza:
## Summary
- Adds a caller workflow that invokes the org-level reusable PR title verifier from `kagenti/.github`
- Enforces PR title prefix conventions (Feat, Fix, Docs, etc.) on all pull requests
- No changes to existing workflows or repo configuration

### DIFF
--- a/.github/workflows/pr-verifier.yml
+++ b/.github/workflows/pr-verifier.yml
@@ -6,5 +6,5 @@ on:
 
 jobs:
   verify:
-    uses: kagenti/.github/.github/workflows/pr-verifier-required.yml@ae31c56d344b1efc6ef174338c488b1e355649bc
+    uses: kagenti/.github/.github/workflows/pr-verifier-required.yml@fccbcfaaadf24f180b4a754377ceff6d9e000297
 
--- a/.github/workflows/pr-verifier.yml
+++ b/.github/workflows/pr-verifier.yml
@@ -1,0 +1,10 @@
+name: PR Verifier
+
+on:
+  pull_request:
+    types: [opened, edited, synchronize, reopened]
+
+jobs:
+  verify:
+    uses: kagenti/.github/.github/workflows/pr-verifier-required.yml@main
+
--- a/.github/workflows/pr-verifier.yml
+++ b/.github/workflows/pr-verifier.yml
@@ -4,6 +4,8 @@ on:
   pull_request:
     types: [opened, edited, synchronize, reopened]
 
+permissions: {}
+
 jobs:
   verify-pr-title:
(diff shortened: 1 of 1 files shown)


rubambiza:
**Update:** The reusable workflow in `kagenti/.github` is now live ([PR #66](https://github.com/kagenti/.github/pull/66) merged), so this caller workflow will fire correctly as soon as this PR is merged.

However, the pinned SHA in this workflow will need a final bump once [kagenti/.github#67](https://github.com/kagenti/.github/pull/67) merges. That PR adds `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true` to suppress a Node.js 20 deprecation warning from the `action-pr-title` action (GitHub is removing Node 20 runner support on September 16, 2026). The action itself is safe for Node 24 (only uses `@actions/core` and `@actions/github`).

Once #67 merges, I'll push an updated SHA pin to this branch.

mrsabath (approved):
## Review Summary

Clean, minimal 11-line workflow that delegates PR title verification to the org-level reusable workflow in `kagenti/.github`. 

**Highlights:**
- Reusable workflow properly pinned to SHA `4e535f2` (verified — merge commit: "Force Node.js 24 for action-pr-title")
- `permissions: {}` follows least-privilege
- Trigger types correct: `opened, edited, synchronize, reopened`
- All 6 commits signed-off

LGTM — no issues found.

.github/workflows/pr-verifier.yml
```diff
@@ -0,0 +1,11 @@
+name: PR Verifier
```
ChristianZaccaria, line 1: Great addition, thanks! :)

I have one nit. The workflow name PR Verifier and filename pr-verifier.yml are a bit generic; this could mean the workflow is checking a number of things on the PR, while we are just verifying that the title has the right prefix. Perhaps it's worth naming it `pr-title-verifier` instead, and/or include a comment to explain what this workflow does.

rubambiza: @ChristianZaccaria That's a fair nit, and I'd love to address it. Unfortunately, it would require replication and opening PRs across 12+ other repos that I'd have to track for review and approval. Ofc, we can also work together if you're willing to approve the list of other PRs. These are still the human costs of our automation 😅

ChristianZaccaria: @rubambiza that's understandable. I could approve the list of PRs but don't have the merging powers. - This is a nit so we can merge as is or just adding a comment is fine.

ChristianZaccaria (approved):
/lgtm !!

pdettori:
I think we should just merge it at this point.

PR #1397, opened 2026-04-29
feat: Add release pin-validation script and E2E dispatch workflow

### LINKED ISSUE #1349: release: add deterministic pin-validation script and manual E2E dispatch for tagged releases
## Problem

The release process (`.claude/skills/release/SKILL.md`, `docs/releasing.md`) has too many non-deterministic steps that rely on human memory:

1. **No automated pin check.** Detecting `tag: latest` entries in `charts/kagenti/values.yaml` and `:latest` in templates is a manual grep. As of today there are 7+ `tag: latest` lines in `values.yaml`. A maintainer can push a bad tag and only discover the issue during testing — or never.

2. **No CI gate for release tags.** The existing `e2e-kind.yaml` and `e2e-hypershift.yaml` workflows trigger on push to `main` and `workflow_dispatch`, but they test the branch tip, not a specific pinned release tag. There is no built-in way to say "run the full E2E suite against `v0.6.0-alpha.5` with all the pinned versions from that tag" without manually editing environment variables.

3. **The release skill is advisory, not enforcing.** The checklist in the skill is helpful but doesn't stop a release with broken pins from being cut.

### 1. `scripts/check-release-pins.sh` — deterministic validation script

Add a script that exits non-zero if any release-breaking condition is found. It should check:

- [ ] No `tag: latest` in `charts/kagenti/values.yaml`
- [ ] No `:latest` anywhere in `charts/kagenti/templates/`
- [ ] All `dependencies[].version` in `charts/kagenti/Chart.yaml` are pinned (no `*`, no semver ranges)
- [ ] `Chart.lock` is present and not stale (its digests match what `helm dependency update` would produce)
- [ ] `ui.tag` and `backend.tag` in `values.yaml` are not empty and not `latest`
- [ ] (Optional) Each pinned image actually exists in `ghcr.io` — a `docker manifest inspect` smoke-check

Example output:
```
[FAIL] charts/kagenti/values.yaml:75  tag: latest   (uiOAuthSecret)
[FAIL] charts/kagenti/values.yaml:89  tag: latest   (agentOAuthSecret)
[OK]   charts/kagenti/Chart.yaml — all dependency versions pinned
[FAIL] charts/kagenti/templates/foo.yaml:12  :latest
2 error(s) found — release is NOT ready to tag
```

The script can be run locally and in CI. The release skill should call it at the start of Phase 2 (Prepare) and block on failure.

### 2. `e2e-release-validation.yaml` — manual-dispatch E2E workflow for tagged releases

Add a new `workflow_dispatch`-only workflow that:

- Accepts a `version` input (e.g., `v0.6.0-alpha.5`)  
- Checks out that tag (`git checkout $version`)
- Runs `scripts/check-release-pins.sh` first (fail fast before spinning up clusters)
- Runs the full Kind E2E suite against the checked-out tag
- Optionally (second input: `run_hypershift: true`) also runs the HyperShift E2E suite

This gives maintainers a **single button** to validate any alpha/rc/GA tag in CI before doing manual testing or announcing the release.

```yaml
on:
  workflow_dispatch:
    inputs:
      version:
        description: 'Release tag to validate (e.g. v0.6.0-alpha.5)'
        required: true
      run_hypershift:
        description: 'Also run HyperShift E2E'
        type: boolean
        default: false
```

### 3. Update the release skill and `docs/releasing.md`

- In Phase 2 (Prepare), add a mandatory step: `bash scripts/check-release-pins.sh` — must pass with zero errors before continuing.
- Add a step in Phase 4 (Verify) to trigger `e2e-release-validation.yaml` and link to the workflow run.
- Update the RC and GA checklists to include the workflow run URL as evidence.

## Acceptance Criteria

- [ ] `scripts/check-release-pins.sh` exists, is executable, and exits 1 if any `tag: latest` or `:latest` is found; exits 0 if all pins are clean
- [ ] The script catches the current `values.yaml` issues (7+ `latest` tags) and reports them clearly
- [ ] `.github/workflows/e2e-release-validation.yaml` exists with `workflow_dispatch` trigger accepting `version` and `run_hypershift` inputs
- [ ] The workflow checks out the specified tag, runs the pin-validation script, then runs the Kind E2E suite
- [ ] Release skill Phase 2 calls the pin script; Phase 4 links to the dispatch workflow
- [ ] `docs/releasing.md` updated to document both new tools

## Why This Matters

Each alpha and RC already burns significant testing time. Flagging bad pins before cluster creation is cheap. Making E2E runs against specific release tags a one-click operation removes a whole class of "it worked on main but broke at the tag" surprises and builds confidence before manual testing begins.

## Related

- `.claude/skills/release/SKILL.md`
- `docs/releasing.md`
- `.github/workflows/e2e-kind.yaml`
- `.github/workflows/e2e-hypershift.yaml`

rubambiza:
## Summary

Closes #1349

- Adds `scripts/check-release-pins.sh` — deterministic validation script that exits non-zero if any `tag: latest` or unpinned dependency is found. Supports `--json` (for CI summaries) and `--verify-images` (opt-in GHCR manifest check).
- Adds `e2e-release-validation.yaml` — `workflow_dispatch` workflow that validates a specific release tag by running pin checks then the full Kind E2E suite (with optional HyperShift).
- Adds `ci-release-pins.yaml` — PR gate that runs the pin script on any PR touching `charts/**`, keeping `main` release-ready.
- Pins the three `:latest` template defaults (`bitnami/kubectl`, `ose-cli`) by adding `common.oseCLIImage` to `values.yaml` and removing stale fallback defaults from templates.
- Updates the release skill (Phase 2 mandatory pin check, Phase 4 E2E dispatch) and `docs/releasing.md`.

## Test plan

- [x] `ci-release-pins.yaml` should self-trigger on this PR (touches `charts/**`) — expect it to report the 8 known `tag: latest` lines in `values.yaml` (these are component tags pinned during release prep, not a bug)
- [x] Run `bash scripts/check-release-pins.sh` locally — should report 8 errors (values.yaml) and 0 template errors
- [x] Run `bash scripts/check-release-pins.sh --json | python3 -m json.tool` — valid JSON
- [ ] `e2e-release-validation.yaml` dispatch workflow testable post-merge (workflow_dispatch requires workflow on default branch)

## Note on E2E workflow structure

We considered refactoring E2E workflows into a reusable `workflow_call` pattern but are deferring that to a follow-up PR to keep this change focused. See [comment](https://github.com/kagenti/kagenti/issues/1349#issuecomment-4346950674).

### DIFF
--- a/.github/workflows/ci-release-pins.yaml
+++ b/.github/workflows/ci-release-pins.yaml
@@ -1,0 +1,50 @@
+name: Release Pin Validation
+
+on:
+  pull_request:
+    paths:
+      - 'charts/**'
+
+permissions:
+  contents: read
+
+jobs:
+  check-pins:
+    name: Check Release Pins
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6
+
+      - name: Run pin-validation script
+        id: pins
+        run: |
+          bash scripts/check-release-pins.sh --json > pin-results.json
+          bash scripts/check-release-pins.sh
+
+      - name: Write summary
+        if: always()
+        run: |
+          echo "## Release Pin Validation" >> "$GITHUB_STEP_SUMMARY"
+          echo "" >> "$GITHUB_STEP_SUMMARY"
+
+          error_count=$(jq '.summary.errors' pin-results.json)
+          warning_count=$(jq '.summary.warnings' pin-results.json)
+
+          if [[ "$error_count" -gt 0 ]]; then
+            echo "| File | Line | Issue |" >> "$GITHUB_STEP_SUMMARY"
+            echo "|------|------|-------|" >> "$GITHUB_STEP_SUMMARY"
+            jq -r '.errors[] | "| \(.file) | \(.line) | \(.issue) |"' pin-results.json >> "$GITHUB_STEP_SUMMARY"
+            echo "" >> "$GITHUB_STEP_SUMMARY"
+            echo "**${error_count} error(s) found — not release-ready**" >> "$GITHUB_STEP_SUMMARY"
+          else
+            echo "All pins validated. Release is ready." >> "$GITHUB_STEP_SUMMARY"
+          fi
+
+          if [[ "$warning_count" -gt 0 ]]; then
+            echo "" >> "$GITHUB_STEP_SUMMARY"
+            echo "### Warnings" >> "$GITHUB_STEP_SUMMARY"
+            jq -r '.warnings[] | "- \(.file): \(.issue)"' pin-results.json >> "$GITHUB_STEP_SUMMARY"
+          fi
--- a/.github/workflows/ci-release-pins.yaml
+++ b/.github/workflows/ci-release-pins.yaml
@@ -18,11 +18,8 @@ jobs:
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6
 
-      - name: Run pin-validation script
-        id: pins
-        run: |
-          bash scripts/check-release-pins.sh --json > pin-results.json
-          bash scripts/check-release-pins.sh
+      - name: Capture pin-validation results
+        run: bash scripts/check-release-pins.sh --json > pin-results.json || true
 
       - name: Write summary
         if: always()
@@ -48,3 +45,6 @@ jobs:
             echo "### Warnings" >> "$GITHUB_STEP_SUMMARY"
             jq -r '.warnings[] | "- \(.file): \(.issue)"' pin-results.json >> "$GITHUB_STEP_SUMMARY"
           fi
+
+      - name: Run pin-validation check
+        run: bash scripts/check-release-pins.sh
--- a/.github/workflows/ci-release-pins.yaml
+++ b/.github/workflows/ci-release-pins.yaml
@@ -4,6 +4,9 @@ on:
   pull_request:
     paths:
       - 'charts/**'
+  # Allow maintainers to run the pin check manually on any branch — useful
+  # during release prep before a release/* branch exists.
+  workflow_dispatch:
 
 permissions:
   contents: read
@@ -46,5 +49,15 @@ jobs:
             jq -r '.warnings[] | "- \(.file): \(.issue)"' pin-results.json >> "$GITHUB_STEP_SUMMARY"
           fi
 
-      - name: Run pin-validation check
+      # Hard gate — only enforce on PRs targeting release/* branches and on
+      # manual dispatches. During normal development `tag: latest` in
+      # values.yaml is intentional (tags are pinned during release prep),
+      # so blocking every charts PR on main would be noise.
+      #
+      # The summary step above still runs on every charts PR, giving
+      # informational visibility without blocking merge.
+      - name: Run pin-validation check (release branches and manual dispatch only)
+        if: >-
+          github.event_name == 'workflow_dispatch' ||
+          (github.event_name == 'pull_request' && startsWith(github.base_ref, 'release/'))
         run: bash scripts/check-release-pins.sh
--- a/.github/workflows/e2e-release-validation.yaml
+++ b/.github/workflows/e2e-release-validation.yaml
@@ -1,0 +1,233 @@
+name: E2E Release Validation
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Release tag to validate (e.g. v0.6.0-alpha.5)'
+        required: true
+        type: string
+      run_hypershift:
+        description: 'Also run HyperShift E2E'
+        type: boolean
+        default: false
+
+# Only allow one release validation at a time
+concurrency:
+  group: e2e-release-validation
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+
+env:
+  K8S_VERSION: '1.35.0'
+  KAGENTI_EXTENSIONS_GIT_REF: 'main'
+
+jobs:
+  # -----------------------------------------------------------------------
+  # Job 1: Validate release pins (fast, no cluster needed)
+  # -----------------------------------------------------------------------
+  validate-pins:
+    name: Validate Release Pins
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+
+    steps:
+      - name: Checkout release tag
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6
+        with:
+          ref: ${{ inputs.version }}
+
+      - name: Run pin-validation script
+        id: pins
+        run: |
+          bash scripts/check-release-pins.sh --json > pin-results.json
+          cat pin-results.json
+
+      - name: Write pin-check summary
+        if: always()
+        run: |
+          echo "## Release Pin Validation — \`${{ inputs.version }}\`" >> "$GITHUB_STEP_SUMMARY"
+          echo "" >> "$GITHUB_STEP_SUMMARY"
+
+          error_count=$(jq '.summary.errors' pin-results.json)
+          warning_count=$(jq '.summary.warnings' pin-results.json)
+
+          if [[ "$error_count" -gt 0 ]]; then
+            echo "| File | Line | Issue |" >> "$GITHUB_STEP_SUMMARY"
+            echo "|------|------|-------|" >> "$GITHUB_STEP_SUMMARY"
+            jq -r '.errors[] | "| \(.file) | \(.line) | \(.issue) |"' pin-results.json >> "$GITHUB_STEP_SUMMARY"
+            echo "" >> "$GITHUB_STEP_SUMMARY"
+            echo "**${error_count} error(s) found — not release-ready**" >> "$GITHUB_STEP_SUMMARY"
+          else
+            echo "All pins validated. Release is ready." >> "$GITHUB_STEP_SUMMARY"
+          fi
+
+          if [[ "$warning_count" -gt 0 ]]; then
+            echo "" >> "$GITHUB_STEP_SUMMARY"
+            echo "### Warnings" >> "$GITHUB_STEP_SUMMARY"
+            jq -r '.warnings[] | "- \(.file): \(.issue)"' pin-results.json >> "$GITHUB_STEP_SUMMARY"
+          fi
+
+  # -----------------------------------------------------------------------
+  # Job 2: Kind E2E suite (runs after pin validation passes)
+  # -----------------------------------------------------------------------
+  e2e-kind:
+    name: Kind E2E (${{ inputs.version }})
+    needs: validate-pins
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+
+    steps:
+      - name: Checkout release tag
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6
+        with:
+          ref: ${{ inputs.version }}
+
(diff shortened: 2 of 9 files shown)


rubambiza:
**For reviewers**: The failing [`Release Pin Validation` CI](https://github.com/kagenti/kagenti/actions/runs/25168997265) indicates that the PR is **working as intended**!

.github/workflows/e2e-release-validation.yaml
```diff
@@ -0,0 +1,233 @@
...
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6
+        with:
+          ref: ${{ inputs.version }}
+
+      - name: Placeholder — HyperShift E2E
+        run: |
+          echo "HyperShift E2E for ${{ inputs.version }} would run here."
+          echo "This job delegates to the HyperShift workflow steps."
+          echo "TODO: duplicate or call reusable HyperShift E2E job."
+          exit 1
```
evaline-ju, line 240: This means run_hypershift: true will always fail which seems misleading, if this is a placeholder I would expect this to default to exit 0?

scripts/check-release-pins.sh
```diff
@@ -0,0 +1,291 @@
...
+    local file="$1"
+    local issue="$2"
+
+    warning_count=$((warning_count + 1))
+    warning_lines="${warning_lines}WARN|${file}|${issue}"$'\n'
+
+    if [[ -n "$json_warnings" ]]; then
+        json_warnings="${json_warnings},"
+    fi
+    json_warnings="${json_warnings}{\"file\": \"${file}\", \"issue\": \"${issue}\"}"
```
evaline-ju: There's a concern on JSON injection via unescaped string interpolation (lines 72–74, 80–83, 91–94)

Specifically for something like
```
  json_errors="${json_errors}{\"file\": \"${file}\", \"line\": ${line}, \"issue\":
  \"${issue}\"}"
```

$file, $line, and $issue are interpolated directly into the JSON string without escaping. A file path containing a " or \ breaks the JSON. More concretely, $issue contains the matched grep text — e.g. ":latest" — which already has a colon. If a template file path contained a quote or backslash, jq would fail to parse the output. The --json flag exists specifically for CI consumption via jq, so this is the path that matters.

Fix:
```
  json_errors="${json_errors}$(jq -n --arg file "$file" --argjson line "$line" --arg
  issue "$issue" \
    '{"file": $file, "line": $line, "issue": $issue}')"
```

scripts/check-release-pins.sh
```diff
@@ -0,0 +1,291 @@
...
+
+# ---------------------------------------------------------------------------
+# Check 3: All Chart.yaml dependency versions are pinned
+# ---------------------------------------------------------------------------
+
+if [[ ! -f "$CHART_FILE" ]]; then
+    add_error "charts/kagenti/Chart.yaml" 0 "file not found"
+else
+    # Extract dependency version lines (lines that follow "- name:" lines)
+    dep_versions=$(grep -A1 '^\- name:' "$CHART_FILE" | grep 'version:' || true)
```
evaline-ju: this could be a bit brittle since `version` just needs to be on the same level as `name` but there could be other fields like `repository` or others included between `name` and `version`. Agent suggestion:  `Fix: grep version: directly from the full dependencies block, or use yq '.dependencies[].version' Chart.yaml.`

.github/workflows/e2e-release-validation.yaml
```diff
@@ -0,0 +1,233 @@
...
+concurrency:
+  group: e2e-release-validation
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+
+env:
+  K8S_VERSION: '1.35.0'
+  KAGENTI_EXTENSIONS_GIT_REF: 'main'
```
evaline-ju: This variable doesn't seem to be directly used? 

If it is indirectly used, it might be good to leave a comment as to where. And since we may usually want to test against a pinned version of kagenti-extensions rather than main, perhaps this should be also update-able via `workflow_dispatch`?

scripts/check-release-pins.sh
```diff
@@ -0,0 +1,291 @@
...
+                is_unpinned=true
+            fi
+            if echo "$version" | grep -qE '[><~^]'; then
+                is_unpinned=true
+            fi
+
+            if [[ "$is_unpinned" == "true" ]]; then
+                deps_ok=false
+                line_number=$(grep -n "version:.*${version}" "$CHART_FILE" | head -1 | cut -d: -f1)
+                add_error "charts/kagenti/Chart.yaml" "${line_number:-0}" "unpinned dependency version: ${version}"
```
evaline-ju: direct agent suggestion: $version is extracted from Chart.yaml via cut/tr and goes directly into the issue field of the JSON. A dependency version string containing " or \ — unlikely in semver but possible in a malformed or adversarial Chart.yaml — would break JSON. More relevantly, $version also flows into a subsequent grep -n "version:.*${version}" call, which is a grep regex injection: a version string like 1.0.0+build or >=1.0 would be interpreted as a regex pattern rather than a literal string. Should be grep -Fn with a fixed string or grep -n "version:.*${version//./\\.}".

evaline-ju (commented):
Thanks for adding this! A few comments - mainly the exit 1 is of concern

.github/workflows/e2e-release-validation.yaml
rubambiza: Agreed on the misleading part. The original intention was for the `exit 1` to indicate to the user that the Hypershift E2E did not actually run (still a placeholder), rather than a false sense of security that all went well. Will fix!

.github/workflows/e2e-release-validation.yaml
rubambiza: It is used by the `run-authbridge-weather-e2e` script invoked further down. It has been turned into an [input](https://github.com/kagenti/kagenti/pull/1397/changes/c6fa7ced604b675bb2076c54e6c5781de46f9679#diff-506c4845efc72a70226af78ce73cd6a32500e4eaafe120b8b73073446dffb0c2R10) with default = main.

scripts/check-release-pins.sh
rubambiza: Used `grep -Fn` throughout now.

scripts/check-release-pins.sh
rubambiza: Addressed in https://github.com/kagenti/kagenti/commit/c6fa7ced604b675bb2076c54e6c5781de46f9679

scripts/check-release-pins.sh
rubambiza: Addressed in https://github.com/kagenti/kagenti/commit/c6fa7ced604b675bb2076c54e6c5781de46f9679

.github/workflows/ci-release-pins.yaml
```diff
@@ -0,0 +1,50 @@
...
+          fi
+
+          if [[ "$warning_count" -gt 0 ]]; then
+            echo "" >> "$GITHUB_STEP_SUMMARY"
+            echo "### Warnings" >> "$GITHUB_STEP_SUMMARY"
+            jq -r '.warnings[] | "- \(.file): \(.issue)"' pin-results.json >> "$GITHUB_STEP_SUMMARY"
+          fi
+
+      - name: Run pin-validation check
+        run: bash scripts/check-release-pins.sh
```
pdettori, line 63: **must-fix**: This step hard-fails on every PR touching `charts/**`, but `tag: latest` is intentional during normal development — it only gets pinned during release prep.

Suggested: gate this step on the target branch so it only enforces on release branches:

```yaml
      - name: Enforce pin check (release branches only)
        if: startsWith(github.base_ref, 'release/')
        run: bash scripts/check-release-pins.sh
```

The summary step above already gives informational visibility on every PR, which is great. The hard gate just shouldn't block non-release work.

.github/workflows/ci-release-pins.yaml
```diff
@@ -0,0 +1,50 @@
+name: Release Pin Validation
+
+on:
+  pull_request:
+    paths:
+      - 'charts/**'
```
pdettori, line 6: **suggestion**: Consider adding `workflow_dispatch: {}` as a trigger here too, so maintainers can manually run the pin check on any branch without needing the E2E validation workflow. Useful during release prep before the release branch exists.

pdettori (changes requested):
## Release Pin Validation — CI gate should not block regular PRs

The script, summary, and E2E dispatch workflow all look solid — the JSON injection and grep regex fixes from the third commit address Evaline's feedback well.

The one blocking issue: **`ci-release-pins.yaml` hard-fails on every PR touching `charts/**`**, but during normal development `tag: latest` in `values.yaml` is intentional (tags get pinned during release prep). The PR's own test plan acknowledges the 8 expected `latest` errors. This means the check will block every future charts PR on `main`.

**Suggested fix**: keep the summary (informational) on all charts PRs, but only enforce the hard gate on release branches:

```yaml
      - name: Enforce pin check (release branches only)
        if: startsWith(github.base_ref, 'release/')
        run: bash scripts/check-release-pins.sh
```

This way developers always see pinning status in the job summary, but the check only blocks merge when the target branch is a release branch (alpha, rc, stable). The `e2e-release-validation.yaml` (workflow_dispatch) already covers hard enforcement for tagged releases.

**Areas reviewed**: CI/GitHub Actions, Shell, Docs
**Commits**: 3 commits, all signed-off ✓
**CI status**: Check Release Pins failing (expected — this is the issue above), all other checks pass

rubambiza:
Thanks @pdettori — addressed in 9a55998:

- The summary step still runs on every charts PR (informational visibility preserved)
- The hard `check-release-pins.sh` enforcement now only fires when `github.base_ref` starts with `release/` or when the workflow is manually dispatched
- Added `workflow_dispatch:` trigger per your suggestion, so maintainers can run the check on any branch during release prep before a release branch exists
- Updated `docs/releasing.md` to describe the new gating behavior

The `Check Release Pins` job is expected to pass on this PR now (targets `main`, so only the summary runs; no hard gate). Hard enforcement is preserved via `e2e-release-validation.yaml` for tagged releases and `ci-release-pins.yaml` when targeting `release/*` branches.

charts/kagenti/templates/otel-collector-restart-job.yaml
```diff
@@ -83,7 +83,7 @@ spec:
       restartPolicy: Never
       containers:
       - name: restarter
-        image: {{ .Values.common.kubectlImage | default "bitnami/kubectl:latest" }}
+        image: {{ .Values.common.kubectlImage }}
```
pdettori, line 86: **suggestion**: Removing the `| default "bitnami/kubectl:latest"` fallback means that if `common.kubectlImage` is unset in values, the template will render an empty image string. Consider adding a pinned default here (like you did for `oseCLIImage` in values.yaml) — e.g. adding `kubectlImage: "bitnami/kubectl:1.32.2"` to values.yaml — so fresh installs with minimal value overrides don't break.

Non-blocking — the existing `values.yaml` already has `kubectlImage` set, so this only matters for users providing their own values without inheriting defaults.

pdettori (approved):
## Release Pin Validation — LGTM

The latest commit correctly addresses the previous review feedback: the CI gate now only hard-fails on `release/*` branches and `workflow_dispatch`, while the summary step remains informational on all charts PRs.

**Areas reviewed**: CI/GitHub Actions, Shell, Helm/K8s, Docs, Skills
**Commits**: 4 commits, all signed-off ✓
**CI status**: All checks passing ✓

One minor non-blocking suggestion inline.

charts/kagenti/templates/otel-collector-restart-job.yaml
rubambiza: This will be addressed in a separate minor issue/PR. I want to make sure this new functionality starts getting used ASAP. Thanks @pdettori

PR #1406, opened 2026-04-30
fix: update weather agent injection label

### LINKED ISSUE #1403: Update weather agent example manifests to use current injection label
# Update Weather Agent Example Manifests with Current Injection Label

## Problem

The weather agent example manifests use an outdated injection label that the AuthBridge webhook doesn't recognize:

**Current (outdated)**:
```yaml
kagenti.io/client-registration-inject: "true"
```

**Should be**:
```yaml
kagenti.io/inject: "enabled"
```

Without the correct label, the webhook skips SPIFFE component injection, preventing the weather agent from authenticating with the weather-tool via federated-jwt.

### Webhook Logs
```
INFO pod-mutator Skipping mutation: kagenti.io/inject not set to enabled
{"namespace": "team1", "crName": "weather-service", "value": ""}
```

The webhook checks for `kagenti.io/inject == "enabled"` and skips injection if not present.

### What Gets Injected
When `kagenti.io/inject: "enabled"` is present, the webhook injects:
```
INFO pod-mutator injection decision {"sidecar": "envoy-proxy", "inject": true}
INFO pod-mutator injection decision {"sidecar": "proxy-init", "inject": true}
INFO pod-mutator injection decision {"sidecar": "spiffe-helper", "inject": true}
INFO pod-mutator injection decision {"sidecar": "client-registration", "inject": true}
```

These sidecars enable:
- **envoy-proxy** - Token exchange for outbound calls to weather-tool
- **spiffe-helper** - SPIFFE workload identity (JWT-SVIDs)
- **client-registration** - Keycloak OAuth client registration

## Impact

Users following the weather agent example will:
1. Deploy the agent with the outdated label
2. Get no AuthBridge sidecars injected
3. Experience authentication failures when the agent tries to call the weather-tool
4. Have no SPIFFE identity for federated-jwt token exchange

## Why Only Weather Agent?

Other agent examples (recipe, sandbox variants) don't have any injection label and work fine without injection because they don't call external tools. Only the weather agent needs injection since it makes authenticated calls to the weather-tool.

## Affected Files

- `kagenti/examples/agents/weather_service_deployment.yaml` (line 32)
- `kagenti/examples/agents/weather_service_deployment_ocp.yaml` (line 31)

## Required Fix

In both files, **replace** the outdated label:

```yaml
# REMOVE:
kagenti.io/client-registration-inject: "true"

# ADD:
kagenti.io/inject: "enabled"
```

## Reproduction

```bash
# Deploy weather agent with outdated label
kubectl apply -f kagenti/examples/agents/weather_service_deployment.yaml

# Check webhook logs
kubectl logs -n kagenti-webhook-system deployment/kagenti-webhook-controller-manager | grep "Skipping mutation"

# Output:
# INFO pod-mutator Skipping mutation: kagenti.io/inject not set to enabled

# Verify no sidecars were injected
kubectl get pod -n team1 -l app.kubernetes.io/name=weather-service -o jsonpath='{.items[0].spec.containers[*].name}'
# Output: agent (should show: agent envoy-proxy spiffe-helper kagenti-client-registration)
```

## Related

- Issue discovered while testing SPIFFE federated-jwt integration (#1234)
- Blocks users from running the weather agent demo with tool authentication

Alan-Cha:
Replace `kagenti.io/client-registration-inject: "true"` with `kagenti.io/inject: "enabled"` in weather agent manifests.

The AuthBridge webhook now requires the `kagenti.io/inject: enabled` label to inject sidecars. The old label is no longer recognized.

Fixes #1403

### DIFF
--- a/kagenti/examples/agents/weather_service_deployment.yaml
+++ b/kagenti/examples/agents/weather_service_deployment.yaml
@@ -29,7 +29,7 @@ spec:
         kagenti.io/type: agent
         protocol.kagenti.io/a2a: ""
         kagenti.io/framework: LangGraph
-        kagenti.io/client-registration-inject: "true"
+        kagenti.io/inject: "enabled"
         app.kubernetes.io/name: weather-service
     spec:
       serviceAccountName: weather-service
--- a/kagenti/examples/agents/weather_service_deployment.yaml
+++ b/kagenti/examples/agents/weather_service_deployment.yaml
@@ -29,7 +29,6 @@ spec:
         kagenti.io/type: agent
         protocol.kagenti.io/a2a: ""
         kagenti.io/framework: LangGraph
-        kagenti.io/inject: "enabled"
         app.kubernetes.io/name: weather-service
     spec:
       serviceAccountName: weather-service
--- a/kagenti/examples/agents/weather_service_deployment_ocp.yaml
+++ b/kagenti/examples/agents/weather_service_deployment_ocp.yaml
@@ -28,7 +28,7 @@ spec:
         kagenti.io/type: agent
         protocol.kagenti.io/a2a: ""
         kagenti.io/framework: LangGraph
-        kagenti.io/client-registration-inject: "true"
+        kagenti.io/inject: "enabled"
         app.kubernetes.io/name: weather-service
     spec:
       serviceAccountName: weather-service
--- a/kagenti/examples/agents/weather_service_deployment_ocp.yaml
+++ b/kagenti/examples/agents/weather_service_deployment_ocp.yaml
@@ -28,7 +28,6 @@ spec:
         kagenti.io/type: agent
         protocol.kagenti.io/a2a: ""
         kagenti.io/framework: LangGraph
-        kagenti.io/inject: "enabled"
         app.kubernetes.io/name: weather-service
     spec:
       serviceAccountName: weather-service


esnible (approved):

ChristianZaccaria (changes requested):
Thanks for having a look at this. This PR is in the right direction, but with a few nits. I believe the injection mechanism is to opt-out, not opt-in. 

To trigger this side car we have two requirements:
1. `kagenti.io/type: agent` label on the deployment/pod.
2. The `kagenti-enabled: "true"` label set at the namespace level. This is already done automatically for `team1` and `team2` namespaces when deploying Kagenti.

The `kagenti.io/inject: enabled` is not required, the MutatingWebhook `objectSelector` confirms this: https://github.com/kagenti/kagenti-operator/blob/main/charts/kagenti-operator/templates/webhook/webhooks.yaml#L37-L47 

Removing just the `kagenti.io/client-registration-inject: "true"` as you did here is the right change, and removing the added `kagenti.io/inject: enabled`. This _should_ be sufficient to trigger the side car for the weather service.

Alan-Cha:
@ChristianZaccaria You're right! I might the corrections. Thanks for pointing it out.

ChristianZaccaria (approved):
/lgtm thanks!!

Ladas (commented):
Reviewed from PR #1395 (OpenShell test matrix).

This label change affects our test matrix — we have credential security tests (`test_credentials__*`) that parametrize across `weather-agent-supervised`. If the injection label changes from `kagenti.io/client-registration-inject` to `kagenti.io/inject`, our NemoClaw smoke tests (`test_T0_4_infra_nemoclaw.py`) which check `kagenti.io/inject: disabled` should still pass since they test a different label value.

No conflicts with our test renames, but the weather agent manifest change should be tested with our capability matrix after merge to verify T4.1 HITL: Network tests still pass (they exec into the weather agent pod).

Alan-Cha:
@Ladas How do we want to proceed with this PR?

Ladas (approved):
👍 I'll do more testing with openshell

PR #1322, opened 2026-04-23
fix: Use HTTPS for SPIFFE IdP setup job bundle endpoint

rdwj:
## Problem

The `kagenti-spiffe-idp-setup-job` connects to `spire-spiffe-oidc-discovery-provider` using HTTP on port 80, but the Service only exposes HTTPS on port 443. The job loops through all 30 retry attempts and times out with:

```
Connection to spire-spiffe-oidc-discovery-provider...svc.cluster.local timed out. (connect timeout=5)
```

This is non-blocking (the ansible playbook continues past it), but produces alarming-looking errors in the logs.

## Solution

Change the `SPIFFE_BUNDLE_ENDPOINT` env var URL scheme from `http://` to `https://` in the job definition in `deployments/ansible/roles/kagenti_installer/tasks/main.yml`. The default HTTPS port (443) matches the service definition.

## Verification

Tested on OCP 4.20.18 (fresh RHPDS cluster, us-east-2):

1. **Service only serves HTTPS** — confirmed by inspecting the `spire-spiffe-oidc-discovery-provider` Service, which exposes only port 443.

2. **HTTPS endpoint works** — verified from within the job pod:
   ```python
   >>> urllib.request.urlopen('https://spire-spiffe-oidc-discovery-provider.zero-trust-workload-identity-manager.svc.cluster.local:443/keys')
   <http.client.HTTPResponse ...>  # 200 OK, valid JWKS
   ```

3. **The OIDC Discovery Provider is configured for HTTPS** — the `SpireOIDCDiscoveryProvider` operand CR uses `jwtIssuer: "https://..."` (see `charts/kagenti-deps/templates/spire-operand.yaml`).

Fixes #1320

PR #1407, opened 2026-04-30
fix: use HTTP for SPIFFE bundle endpoint on Kind clusters

pdettori:
## Summary

- PR #1322 changed the Ansible installer's `SPIFFE_BUNDLE_ENDPOINT` from `http://` to `https://`, which is correct for OpenShift (port 443/HTTPS) but breaks Kind clusters where the SPIRE OIDC discovery provider service only exposes port 80 (HTTP)
- Makes the URL scheme conditional using the existing `enable_openshift` flag, matching the pattern already used for `SPIFFE_TLS_VERIFY`
- On OpenShift: `https://` + `SPIFFE_TLS_VERIFY=false`
- On Kind: `http://` + `SPIFFE_TLS_VERIFY=true` (irrelevant for HTTP)

## Root Cause

The `spire-spiffe-oidc-discovery-provider` service is configured differently per platform:
- **OpenShift**: exposes port 443 (HTTPS via Route/Ingress TLS termination)
- **Kind**: exposes port 80 → nginx sidecar on 8080 (plain HTTP, no TLS)

After PR #1322, the Ansible installer always uses `https://`, causing the job to fail on Kind with `BackoffLimitExceeded` after 10 retries.

## Test plan

- [ ] Deploy Kind cluster using Ansible installer, verify `kagenti-spiffe-idp-setup-job` completes successfully
- [ ] Verify no regression on OpenShift deployments (HTTPS still used)

### DIFF
--- a/deployments/ansible/roles/kagenti_installer/tasks/main.yml
+++ b/deployments/ansible/roles/kagenti_installer/tasks/main.yml
@@ -1170,7 +1170,7 @@
                       - name: SPIFFE_TRUST_DOMAIN
                         value: "spiffe://{{ kagenti_deps_values.domain }}"
                       - name: SPIFFE_BUNDLE_ENDPOINT
-                        value: "https://spire-spiffe-oidc-discovery-provider.{{ charts.spire.server_namespace }}.svc.cluster.local/keys"
+                        value: "{{ 'https' if (enable_openshift | default(false)) else 'http' }}://spire-spiffe-oidc-discovery-provider.{{ charts.spire.server_namespace }}.svc.cluster.local/keys"
                       - name: SPIFFE_IDP_ALIAS
                         value: "{{ (kagenti_deps_values.authBridge | default({})).spiffeIdpAlias | default('spire-spiffe') }}"
                       - name: SPIFFE_TLS_VERIFY


Alan-Cha (approved):
## Review Summary

Excellent targeted fix for the Kind vs OpenShift platform difference in SPIFFE OIDC discovery endpoint configuration. The change correctly mirrors the existing `SPIFFE_TLS_VERIFY` pattern.

**Areas reviewed:** YAML/Ansible, commit conventions, PR format, security

**Commits:** 1 commit, properly signed-off with `Assisted-By` (not `Co-Authored-By`) ✅

**CI status:** All checks passing (E2E HyperShift pending manual trigger)

**Code quality:**
- Consistent with existing `enable_openshift` flag usage
- Minimal, surgical change
- Clear root cause documentation in PR body

✅ **Approved** - No blocking issues found.

---

**Suggestion (non-blocking):** Consider adding an explicit test case to the E2E test suite that verifies the `kagenti-spiffe-idp-setup-job` completes successfully on both Kind and OpenShift. The current test plan in the PR body is good, but codifying this as an automated check would prevent regression.

For example, in `kagenti/tests/e2e/` you could add a check that waits for this job to reach `status.succeeded=1` within a timeout.

Alan-Cha (approved):

esnible (approved):

PR #1416, opened 2026-04-30
fix: use server-side apply for Gateway API CRDs

pdettori:
## Summary
- Switch Gateway API experimental CRD installation from client-side `kubectl apply` to `--server-side --force-conflicts`
- Fixes `metadata.annotations: Too long: may not be more than 262144 bytes` error on `httproutes.gateway.networking.k8s.io`
- The HTTPRoute CRD in v1.4.0 experimental bundle exceeds the annotation size limit when using client-side apply

## Test plan
- [x] Verified `kubectl apply --server-side --force-conflicts` successfully installs all Gateway API v1.4.0 experimental CRDs on a Kind cluster
- [ ] Re-run `scripts/openshell/deploy-shared.sh` end-to-end


pdettori:
Superseded by #1417 which includes this fix plus additional fixes (OIDC discovery, sandbox namespace, service env var collision, TLS certificate SAN).

PR #1417, opened 2026-05-01
fix(openshell): resolve gateway deployment failures on Kind

pdettori:
## Summary

- Fix `CreateContainerConfigError`: add `runAsUser: 1000` so Kubernetes can verify the non-numeric `openshell` user satisfies `runAsNonRoot`
- Fix OIDC discovery failure: inject `hostAliases` mapping `keycloak.localtest.me` to the Keycloak ClusterIP (`localtest.me` resolves to `127.0.0.1` inside the pod)
- Fix wrong sandbox namespace: pass `--sandbox-namespace` to the gateway so it watches the tenant namespace instead of defaulting to `default`
- Fix service env var collision: set `enableServiceLinks: false` to prevent Kubernetes-injected `OPENSHELL_SERVER_PORT` from conflicting with the gateway's `OPENSHELL_` config prefix
- Fix TLS certificate SAN: include the ingress hostname in the server certificate so the CLI can verify the cert during TLS passthrough
- Fix Gateway API CRD apply: use `--server-side --force-conflicts` to avoid annotation size limits on re-runs

## Test plan

- [x] `OPENSHELL_IMAGE_TAG=latest scripts/openshell/deploy-tenant.sh team1` — pod reaches 3/3 Running
- [x] `openshell status` — shows `Connected`
- [x] Gateway logs confirm OIDC JWKS loaded and sandbox namespace is correct
- [ ] Verify on fresh Kind cluster (destroy and recreate)

### DIFF
--- a/charts/openshell/templates/certificate.yaml
+++ b/charts/openshell/templates/certificate.yaml
@@ -15,6 +15,9 @@ spec:
     - openshell-server.{{ .Values.tenant }}.svc
     - openshell-server.{{ .Values.tenant }}.svc.cluster.local
     - localhost
+    {{- if .Values.ingress.host }}
+    - {{ .Values.ingress.host }}
+    {{- end }}
   issuerRef:
     name: {{ .Values.tls.issuerRef }}
     kind: {{ .Values.tls.issuerKind }}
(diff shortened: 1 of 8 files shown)


Ladas (approved):
@pdettori excellent, just started hitting this debugging the gateway tests, looks great!

PR #1238, opened 2026-04-13
build(deps): bump python base image digest across all Dockerfiles

pdettori:
## Summary

- Updates `python:3.14-slim` digest from `fb83750` to `bc389f7` in all 5 Dockerfiles
- Bundles identical Dependabot PRs #1203, #1205, #1206, #1207, #1208 into a single atomic update

## Files changed

| Dockerfile | Component |
|------------|-----------|
| `kagenti/backend/Dockerfile` | Backend (2 FROM lines) |
| `kagenti/auth/ui-oauth-secret/Dockerfile` | UI OAuth secret |
| `kagenti/auth/agent-oauth-secret/Dockerfile` | Agent OAuth secret |
| `kagenti/auth/api-oauth-secret/Dockerfile` | API OAuth secret |
| `kagenti/auth/mlflow-oauth-secret/Dockerfile` | MLflow OAuth secret |

## Supersedes

Once merged, close the following individual Dependabot PRs:
- #1203
- #1205
- #1206
- #1207
- #1208

## Test plan

- [x] CI passes on this PR
- [x] Container images build successfully

Assisted-By: Claude (Anthropic AI) <noreply@anthropic.com>

PR #1420, opened 2026-05-01
fix(kind): fix arm64 Kind cluster install (image arch, SPIRE OIDC, credentials)

huang195:
## Summary

- Pin `python:3.14-slim` to multi-arch manifest digest (`sha256:5b3879b6`) replacing the amd64-only digest from PR #1238
- Cap `cryptography<47` in all auth job requirements — v47.0.0 arm64 wheels crash with SIGILL under Docker Desktop virtualization
- Build `spiffe-idp-setup` in Step 7 before job creation (was built too late in the `_BUILD_IMAGES` block)
- Add `ui-oauth-secret` to the `--build-images` list
- Pass `--build-images` in `kind-full-test.sh` to match CI behavior
- Patch SPIRE OIDC ConfigMap with `set_key_use: true` in the Ansible installer (chart v0.27.0 silently ignores the Helm value)
- Fix test user password: update password in Keycloak when user already exists so the `kagenti-test-user` secret matches the actual credential

## Root Cause

Three issues combined to break arm64 Kind installs:

1. **Image arch**: PR #1238 pinned `python:3.14-slim` to an amd64-only digest, and `cryptography` 47.0.0 ships arm64 wheels with CPU instructions unsupported by Docker Desktop — both cause SIGILL (exit 132)
2. **SPIRE OIDC**: The Ansible installer trusted the Helm value `spiffe-oidc-discovery-provider.config.set_key_use=true` but SPIRE chart v0.27.0 doesn't render it into the ConfigMap, leaving JWKS keys without the `use` field
3. **Credentials**: The `agent-oauth-secret` job skipped password update when the realm-imported user already existed, causing the secret to store a random password that didn't match Keycloak

## Test plan

- [x] Run `kind-full-test.sh --skip-cluster-destroy` on an arm64 Mac
- [x] Verify `kagenti-spiffe-idp-setup-job` and `kagenti-ui-oauth-secret-job` both succeed
- [x] Confirm UI loads at http://kagenti-ui.localtest.me:8080 with credentials from `show-services.sh`
- [x] Verify `from keycloak import KeycloakOpenIDConnection` works inside the container image

### DIFF
--- a/deployments/ansible/roles/kagenti_installer/tasks/main.yml
+++ b/deployments/ansible/roles/kagenti_installer/tasks/main.yml
@@ -838,8 +838,9 @@
 
 #
 # SPIFFE IdP setup job - creates Keycloak SPIFFE identity provider
-# Note: The SPIRE chart (v0.28.4+) now supports setKeyUse configuration natively,
-# so no ConfigMap patching is needed. The set_key_use: true config is set via Helm values.
+# Note: SPIRE chart v0.27.0 does NOT template config.set_key_use into the ConfigMap
+# despite accepting it as a Helm value. We patch the ConfigMap directly before
+# creating the job.
 - name: Run SPIFFE IdP setup job
   block:
     - name: Get kagenti-deps release values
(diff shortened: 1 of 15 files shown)


deployments/ansible/roles/kagenti_installer/tasks/main.yml
```diff
@@ -981,6 +982,32 @@
...
+        kubectl get configmap "$CM" -n "$NS" -o json | \
+          python3 -c "import sys,json; cm=json.load(sys.stdin); cm['data']['oidc-discovery-provider.conf']=r'''$PATCHED'''; json.dump(cm,sys.stdout)" | \
+          kubectl replace -f -
+        kubectl rollout restart deployment/spire-spiffe-oidc-discovery-provider -n "$NS"
+        kubectl rollout status deployment/spire-spiffe-oidc-discovery-provider -n "$NS" --timeout=120s
+        echo "OIDC ConfigMap patched with set_key_use: true"
+      args:
+        executable: /bin/bash
+      changed_when: "'patched' in patch_oidc_result.stdout"
+      register: patch_oidc_result
```
esnible, line 1009: **suggestion**: The `$PATCHED` variable is embedded in a Python raw string `r'''$PATCHED'''`. If the ConfigMap JSON ever contains triple-quotes (`'''`), this will break the Python parser.

Safer alternative using env var:
```bash
PATCHED="$PATCHED" python3 -c "import os,sys,json; cm=json.load(sys.stdin); cm['data']['oidc-discovery-provider.conf']=os.environ['PATCHED']; json.dump(cm,sys.stdout)"
```

Non-blocking — the current content is just JSON so it works today.

kagenti/auth/agent-oauth-secret/agent_oauth_secret.py
```diff
@@ -334,10 +334,14 @@ def create_user(self, username, password: Optional[str] = None):
             )
             typer.echo(f'Created user "{username}" (id: {user_id})')
         except KeycloakPostError:
```
esnible, line 336: **nit**: This `except KeycloakPostError` catches all post errors, not just 409 (user exists). If a different error occurs (e.g., 500), it silently treats it as "user already exists" and attempts password update. Consider checking `e.response_code == 409` like the realm creation pattern at line 290.

esnible (approved):
Well-structured fix addressing three distinct arm64 regression areas. All changes are consistent — digest applied uniformly across all 7 Dockerfiles, `cryptography<47` pinned in all 5 auth requirements files, and build ordering fix is sound. CI is fully green including Deploy & Test.

Two non-blocking observations below.

PR #1424, opened 2026-05-01
fix(e2e): fix HyperShift E2E test failures across multiple layers

huang195:
## Summary

Fix a cascade of E2E test failures on HyperShift (OCP 4.20) ephemeral CI clusters. Each fix exposed the next layer of pre-existing issues that were masked by earlier failures.

### 1. Skip MLflow RBAC when RHOAI is absent
Set `SKIP_MLFLOW=true` when the RHOAI CRD is not detected in `setup-kagenti.sh`, so the downstream `_mlflow_grant_otel_rbac` call and Helm `mlflow.enable` value are both consistently skipped.

### 2. Handle MCP CRD domain migration
The MCP gateway v0.6.0 renamed CRDs from `mcp.kuadrant.io` to `mcp.kagenti.com`. Kind CI still uses the old version; HyperShift uses the new. Updated `41-wait-crds.sh` to detect which domain is present with a 60s retry loop instead of a single instant check.

### 3. Add privileged PSA labels to agent namespaces
SPIFFE CSI volumes require `pod-security.kubernetes.io/enforce: privileged` on agent namespaces. Added the label to `agent-namespaces.yaml`.

### 4. Wait for ZTWIM CRDs before applying SPIRE operands
`_apply_operand_crs()` was applying SPIRE operand CRs (SpiffeCSIDriver, SpireServer, etc.) before the ZTWIM operator had registered its CRDs. The errors were silently swallowed by `|| true`. Added a wait loop for `spiffecsidrivers.operator.openshift.io` (up to 10 min).

### 5. Create CI-specific config and fix LLM endpoint
- Created `ocp_ci_values.yaml` that disables features not installed on ephemeral clusters (RHOAI, MLflow, Shipwright, Kiali) so tests are correctly skipped via `@pytest.mark.requires_features`.
- Patched weather agent LLM config on OpenShift when Shipwright is unavailable — the Kind manifest points to `dockerhost:11434` (local Ollama) which doesn't exist on OCP.

### 6. Scope health checks and fix test markers
- Scoped `test_no_failed_pods`, `test_no_crashlooping_pods`, and `test_all_jobs_completed` to Kagenti-managed namespaces only. On OpenShift, scanning all namespaces causes false positives from unrelated system pods.
- Changed Shipwright test markers from `requires_features(["tekton"])` to `requires_features(["shipwright"])`.
- Added missing `requires_features` marker to `TestToolShipwrightBuildIntegration`.
- Updated both HyperShift workflow files to use `ocp_ci_values.yaml`.

## Test plan

- [ ] E2E HyperShift passes on cluster without RHOAI/MLflow/Shipwright
- [ ] Deploy & Test (Kind) continues to pass
- [ ] RHOAI/MLflow/Shipwright/Kiali tests are skipped (not failed) on CI
- [ ] Platform health checks don't flag OpenShift system pods
- [ ] Weather agent can reach LLM endpoint on OpenShift

### DIFF
--- a/.github/scripts/hypershift/ci/80-run-e2e-tests.sh
+++ b/.github/scripts/hypershift/ci/80-run-e2e-tests.sh
@@ -23,6 +23,11 @@ uv sync --extra test
 # Use hypershift-full-test.sh with whitelist mode (--include-test)
 # hypershift-full-test.sh handles AGENT_URL detection from route and calls 90-run-e2e-tests.sh
 # Note: CLUSTER_SUFFIX is set by the workflow (e.g., pr594), don't override it
+#
+# Use CI-specific config that disables features not installed on ephemeral clusters
+# (RHOAI, MLflow, Shipwright, Kiali) so tests are correctly skipped.
+export KAGENTI_CONFIG_FILE="deployments/envs/ocp_ci_values.yaml"
+
 exec "$REPO_ROOT/.github/scripts/local-setup/hypershift-full-test.sh" \
     --include-test \
     --env ocp
--- a/.github/scripts/hypershift/create-cluster.sh
+++ b/.github/scripts/hypershift/create-cluster.sh
@@ -655,7 +655,7 @@ oc get nodes
 ./.github/scripts/kagenti-operator/74-deploy-weather-agent.sh
 
 export AGENT_URL="https://\$(oc get route -n team1 weather-service -o jsonpath='{.spec.host}')"
-export KAGENTI_CONFIG_FILE=deployments/envs/ocp_values.yaml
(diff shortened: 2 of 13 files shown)


Ladas (approved):
Reviewed from PR #1395.

This fixes the MLflow ClusterRole check that blocked our CI HyperShift E2E run (run 25205839906 — "E2E tests failed" at Deploy Kagenti step). The `_mlflow_grant_otel_rbac()` function in `setup-kagenti.sh` would fail hard when the RHOAI MLflow operator wasn't installed, even though MLflow integration is optional.

We saw this exact failure in our graph-loop iteration 7 and tracked it as "not our problem" — glad someone is fixing it properly with `--skip-mlflow` detection.

PR #1427, opened 2026-05-02
fix: Align PORT env var with servicePorts targetPort in tool deployments

huang195:
## Summary

- `_build_tool_env_vars` now accepts `service_ports` and overrides `PORT` to match `targetPort`
- Call sites reordered so `service_ports` is computed before `env_vars`
- Fixes connection resets when UI sets non-default tool ports (e.g. 9090)

## Root cause

PR #1175 changed the UI default tool port from 8000 to 9090, but the backend always injected `PORT=8000` from `DEFAULT_ENV_VARS`. The container listened on 8000 while the K8s Service routed to 9090.

## Test plan

- [x] Existing `test_tool_workloads.py` tests pass (37/37)
- [x] New tests for PORT override behavior pass (3/3)
- [x] Verified end-to-end: weather-tool with targetPort 9090 now gets `PORT=9090`

### DIFF
--- a/kagenti/backend/app/routers/tools.py
+++ b/kagenti/backend/app/routers/tools.py
@@ -1889,7 +1889,9 @@ async def finalize_tool_shipwright_build(
         if request.envVars:
             env_vars = _build_tool_env_vars(request.envVars, service_ports=service_ports)
         elif tool_config_dict.get("envVars"):
-            env_vars = _build_tool_env_vars([EnvVar(**ev) for ev in tool_config_dict["envVars"]], service_ports=service_ports)
+            env_vars = _build_tool_env_vars(
+                [EnvVar(**ev) for ev in tool_config_dict["envVars"]], service_ports=service_ports
+            )
         else:
             env_vars = _build_tool_env_vars(service_ports=service_ports)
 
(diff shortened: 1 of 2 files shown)


mrsabath (approved):
## Review Summary

Clean, well-scoped bug fix for the PORT mismatch between the UI default (9090 since PR #1175) and the backend's hardcoded `PORT=8000`. The fix correctly reorders call sites so `service_ports` is computed before `env_vars`, then overrides PORT to match `targetPort`. Both call paths (`create_tool` and `finalize_tool_shipwright_build`) are updated consistently.

### Suggestions (non-blocking)

1. **Test stub vs real function** — The new tests define a local stub of `_build_tool_env_vars` rather than importing the real function from `tools.py`. This means the tests validate the stub's behavior, not the production code. This is consistent with the existing file's stub-based pattern, but for these new tests consider importing the real function — the `sys.path` insert at the top already adds the backend to the path.

2. **Duplicate PORT edge case** — If a user explicitly provides `PORT` in `env_var_list`, it gets appended after the `service_ports` override, creating a duplicate entry. K8s uses the first occurrence, so the user's explicit value would be silently ignored. Low-risk (users rarely set PORT manually), but worth a dedup step or comment in a follow-up.

---

**Areas reviewed**: Python, Tests, Security
**Commits**: 3, all DCO signed-off ✓
**CI status**: Passing (CodeQL/Deploy&Test pending, E2E awaiting `/run-e2e`)

LGTM — nice root-cause analysis and fix.

Issue #1009, opened 2026-03-17
:bug: Kagenti self certificates expire after a few days


yoavkatz:
### Describe the bug

I have a local Kagenti install on my Mac.  Every few days it seems that certificates expire.   Sometime I see it in the GUI as “Invalid token: Signature has expired.”  and sometime when I interact with an installed  agent.    So far, I used

‘for ns in $(kubectl get ns -o jsonpath=‘{.items[*].metadata.name}‘); do\n kubectl rollout restart deployment -n $ns\n kubectl rollout restart statefulset -n $ns\n kubectl rollout restart daemonset -n $ns\ndone’

### Steps To Reproduce

Keep a Kagenti kind server up for a few days.

### Expected Behavior

Kagenti could be up for days without such errors. Alternatively, it should require one to explictly login again.

PR #1430, opened 2026-05-02
fix(ui): Retry streaming chat on 401 with force-refreshed token

huang195:
## Summary

- Adds 401 retry with force-refresh to the streaming chat path in AgentChat
- Exposes `forceRefreshToken` from AuthContext for component-level use

## Root cause

When a user creates a new agent, client-registration adds the agent's audience scope to Keycloak as a default scope. But the user's existing token was issued before the scope existed, so it lacks the `aud` claim. The non-streaming chat path already handles this via `apiFetch`'s 401 retry logic (issue #1009), but the streaming path used raw `fetch()` without any retry — resulting in a persistent 401 that required a full re-login to resolve.

## Changes

- **AuthContext.tsx**: Added `forceRefreshToken` to the exported `AuthContextType` interface and context value
- **AgentChat.tsx**: On 401 from the streaming endpoint, force-refreshes the token and retries the request once

## Test plan

- [x] TypeScript compiles cleanly (`tsc --noEmit`)
- [x] No new lint errors introduced (pre-existing only)
- [ ] Manual: create agent → immediately chat → should auto-recover without re-login

### DIFF
--- a/kagenti/ui-v2/src/components/AgentChat.tsx
+++ b/kagenti/ui-v2/src/components/AgentChat.tsx
@@ -89,7 +89,7 @@ export const AgentChat: React.FC<AgentChatProps> = ({ namespace, name }) => {
   const [showAgentCard, setShowAgentCard] = useState(false);
   const messagesEndRef = useRef<HTMLDivElement>(null);
   const abortControllerRef = useRef<AbortController | null>(null);
-  const { getToken, user } = useAuth();
+  const { getToken, forceRefreshToken, user } = useAuth();
   const currentUsername = user?.username || 'you';
 
   // Fetch agent card to check capabilities
@@ -147,7 +147,7 @@ export const AgentChat: React.FC<AgentChatProps> = ({ namespace, name }) => {
 
       try {
         // Get auth token if available
-        const token = await getToken();
+        let token = await getToken();
         const headers: Record<string, string> = {
           'Content-Type': 'application/json',
         };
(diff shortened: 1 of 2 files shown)


kagenti/ui-v2/src/components/AgentChat.tsx
```diff
@@ -158,18 +158,33 @@ export const AgentChat: React.FC<AgentChatProps> = ({ namespace, name }) => {
...
+          signal: controller.signal,
+        });
+
+        // On 401, force-refresh token and retry once (scope may have been added after login)
+        if (response.status === 401) {
+          await response.body?.cancel();
+          token = await forceRefreshToken();
+          if (token) {
+            headers['Authorization'] = `Bearer ${token}`;
+            response = await fetch(streamUrl, {
```
esnible, line 180: **suggestion**: The retry reuses the same `AbortController`. If the user clicks stop between the 401 and the retry completing, `response.body?.cancel()` could throw on an already-aborted stream. Consider wrapping the cancel in a try/catch or checking `controller.signal.aborted` before retrying:

```ts
if (response.status === 401 && !controller.signal.aborted) {
  try { await response.body?.cancel(); } catch { /* aborted */ }
  ...
}
```

kagenti/ui-v2/src/components/AgentChat.tsx
```diff
@@ -316,21 +331,26 @@ export const AgentChat: React.FC<AgentChatProps> = ({ namespace, name }) => {
...
+
+      let response = await fetch(hitlUrl, { method: 'POST', headers, body: hitlBody });
+
+      if (response.status === 401) {
+        await response.body?.cancel();
+        token = await forceRefreshToken();
+        if (token) {
+          headers['Authorization'] = `Bearer ${token}`;
+          response = await fetch(hitlUrl, { method: 'POST', headers, body: hitlBody });
         }
```
esnible, line 352: **nit**: The HITL retry fires but the response is never checked. Pre-existing (original code fire-and-forgot too), but now that you're inspecting `response.status` for 401, it'd be consistent to also check the retry response:

```ts
if (!response.ok) {
  console.error(`[AgentChat] HITL ${action} failed after retry: ${response.status}`);
}
```

esnible (approved):
Correct fix for a real usability bug — the streaming and HITL paths now match `apiFetch`'s existing 401 retry pattern.

**Areas reviewed**: Frontend (TypeScript/React)
**Commits**: 2, all DCO signed-off
**CI**: All passing (E2E HyperShift pending manual trigger — expected)

PR #1447, opened 2026-05-04
fix(ui): Propagate agent 401 as HTTP 401 for transparent token refresh

huang195:
## Summary

- Streaming chat endpoint now returns HTTP 401 (instead of HTTP 200 + SSE error) when the agent rejects a token due to audience mismatch
- This enables the existing frontend retry logic (PR #1430) to fire correctly: force-refresh the token, which picks up the new agent's audience scope from Keycloak, then retry transparently
- Removes dead `_stream_a2a_response` function replaced by `_stream_from_response` (owns an already-opened connection)

## Problem

When a user deploys a new agent:
1. `client-registration` creates an audience scope in Keycloak with the agent's SPIFFE ID
2. The user's existing token (issued before the scope existed) lacks the agent's audience
3. AuthBridge rejects with 401
4. Backend caught this as `HTTPStatusError`, yielded `data: {"error": "Agent error: 401"}` inside an HTTP 200 stream
5. Frontend retry logic checks `response.status === 401` but sees 200 — never retries

## Fix

Open the agent connection in the endpoint function **before** committing to `StreamingResponse`. If the agent returns 401, raise `HTTPException(401)` directly. The frontend's existing `forceRefreshToken()` fires, gets a new token with the correct audience, and retries — all transparent to the user.

## Test plan

- [x] New unit tests: `test_chat_stream_401.py` (agent 401 → HTTP 401, connection error → HTTP 503)
- [x] Existing test suite passes (256 tests)
- [ ] Manual: deploy weather-agent on fresh cluster, verify first chat attempt succeeds without re-login

### DIFF
--- a/kagenti/backend/app/routers/chat.py
+++ b/kagenti/backend/app/routers/chat.py
@@ -301,9 +301,7 @@ async def _stream_from_response(
                 detail = response.text[:500]
             except Exception:
                 detail = str(response.status_code)
-            logger.error(
-                "Agent error: %d: %s", response.status_code, detail.replace("\n", " ")
-            )
+            logger.error("Agent error: %d: %s", response.status_code, detail.replace("\n", " "))
             payload = {"error": f"Agent error: {response.status_code}", "session_id": session_id}
             yield f"data: {json.dumps(payload)}\n\n"
             return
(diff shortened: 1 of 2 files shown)


kagenti/backend/app/routers/chat.py
```diff
@@ -282,221 +282,148 @@ def _extract_text_from_parts(parts: list) -> str:
...
-                        _sidecar_mgr = get_sidecar_manager()
-                    except ImportError:
-                        pass
+        if response.status_code >= 400:
+            try:
+                await response.aread()
+                detail = response.text[:500]
+            except Exception:
+                detail = str(response.status_code)
+            logger.error(f"Agent error: {response.status_code}: {detail}")
```
esnible: Consider sanitizing detail, or not using f-strings, to avoid log injection.

esnible (approved):

PR #1465, opened 2026-05-05
feat(ocp): add --with-kiali, --with-builds, --with-kuadrant flags to installer

### LINKED ISSUE #1304: feat(ocp): add --with-kiali, --with-builds, --with-kuadrant flags to OCP installer
## Summary

The OCP installer (`scripts/ocp/setup-kagenti.sh`) hardcodes several components to disabled with no toggle, creating a feature gap vs the Kind installer (`scripts/kind/setup-kagenti.sh`) which supports all kagenti-deps components via composable `--with-*` flags.

## Missing flags

| Flag | Chart values | Chart template support |
|------|-------------|----------------------|
| `--with-kiali` | `components.kiali.enabled=true` | `kiali-operator.yaml` + `kiali-operand.yaml` (OLM) — already exists |
| `--with-builds` | `components.tekton.enabled=true` + `components.shipwright.enabled=true` | `tekton-operator.yaml` + `openshift-builds-operator.yaml` + `openshift-builds-operand.yaml` (OLM) — already exists |
| `--with-kuadrant` | N/A (no chart template) | Direct Helm install: `helm upgrade --install kuadrant-operator` + Kuadrant CR (same pattern as Kind script) |

## Context

- The Kind script supports `--with-all` which enables every component. The OCP script has no equivalent.
- All three components have existing Helm chart templates gated on `openshift: true` (except Kuadrant which is installed directly via Helm in both scripts).
- OpenShift Builds is the OCP distribution of Shipwright — same CRDs (`builds.shipwright.io`, `buildruns.shipwright.io`, etc.), used by the Kagenti backend build API.
- These are independent of RHOAI — RHOAI does not install OpenShift Builds or Kiali.

## Implementation notes

- Add boolean flags defaulting to `false` (matching Kind's additive `--with-*` pattern)
- Kiali: the existing `--no-hooks` workaround handles the `cluster-monitoring-config` conflict — no extra handling needed
- Builds: Tekton is a prerequisite for Shipwright on OCP (the OpenShift Builds operator requires it). The chart already handles this via OLM subscriptions.
- Kuadrant: follow the Kind script pattern — `helm upgrade --install kuadrant-operator` from `https://kuadrant.io/helm-charts/` + create Kuadrant CR
- Consider adding `--with-all` flag for parity with Kind script

## Acceptance criteria

- [ ] `--with-kiali` enables Kiali + Prometheus on OCP
- [ ] `--with-builds` enables Tekton + OpenShift Builds operator
- [ ] `--with-kuadrant` installs Kuadrant operator + CR (auto-enables MCP Gateway)
- [ ] `--with-all` enables all optional components
- [ ] Flags follow same dependency resolution as Kind (e.g., `--with-kuadrant` implies MCP Gateway)
- [ ] Help text (`-h`) documents the new flags

pdettori:
## Summary

- Adds composable `--with-kiali`, `--with-builds`, `--with-kuadrant`, and `--with-all` flags to the OCP installer (`scripts/ocp/setup-kagenti.sh`)
- All flags default to `false` (additive pattern matching the Kind installer)
- `--with-kuadrant` auto-enables MCP Gateway (overrides `--skip-mcp-gateway`)

## Changes

| Flag | Effect |
|------|--------|
| `--with-kiali` | Sets `components.kiali.enabled=true` in kagenti-deps chart |
| `--with-builds` | Sets `components.tekton.enabled=true` + `components.shipwright.enabled=true` |
| `--with-kuadrant` | Installs Kuadrant operator via Helm + creates Kuadrant CR |
| `--with-all` | Enables all three above |

## Test plan

- [x] `bash -n` syntax check passes
- [x] `--help` shows new flags correctly
- [x] Pre-commit hooks pass (lint, secret detection)
- [ ] `--dry-run --with-all` on OCP cluster shows correct helm flags
- [ ] `--with-kiali` on OCP enables Kiali operator + Prometheus
- [ ] `--with-builds` on OCP enables Tekton + OpenShift Builds
- [ ] `--with-kuadrant` on OCP installs Kuadrant + creates CR

Closes #1304

Assisted-By: Claude Code

### DIFF
--- a/scripts/ocp/setup-kagenti.sh
+++ b/scripts/ocp/setup-kagenti.sh
@@ -4,8 +4,8 @@
 # ============================================================================
 # Installs the Kagenti stack (SPIRE, cert-manager, Keycloak, operator, webhook,
 # MCP Gateway) on an OpenShift cluster. Run this BEFORE setup.sh --with-a2a.
-# Prometheus/Kiali are disabled. UI/backend installed by default (use --skip-ui to disable).
-# Shipwright disabled by default.
+# Optional layers enabled via --with-* flags (Kiali, Builds, Kuadrant).
+# UI/backend installed by default (use --skip-ui to disable).
 #
 # MLflow: provisions an MLflow instance via RHOAI's DSC mlflowoperator
 # and wires the OTEL collector to export traces to it.
@@ -15,6 +15,10 @@
 #   ./scripts/ocp/setup-kagenti.sh --kagenti-repo /path/to/kagenti  # Use local clone
 #   ./scripts/ocp/setup-kagenti.sh --kagenti-repo https://github.com/org/kagenti.git  # Clone from URL
 #   ./scripts/ocp/setup-kagenti.sh --realm nerc                 # Custom Keycloak realm (default: kagenti)
+#   ./scripts/ocp/setup-kagenti.sh --with-kiali                  # Enable Kiali + Prometheus
+#   ./scripts/ocp/setup-kagenti.sh --with-builds                # Enable Tekton + OpenShift Builds
+#   ./scripts/ocp/setup-kagenti.sh --with-kuadrant              # Enable Kuadrant (auto-enables MCP Gateway)
+#   ./scripts/ocp/setup-kagenti.sh --with-all                   # Enable all optional components
 #   ./scripts/ocp/setup-kagenti.sh --skip-ovn-patch             # Skip OVN gateway patch
 #   ./scripts/ocp/setup-kagenti.sh --skip-mcp-gateway           # Skip MCP Gateway install
 #   ./scripts/ocp/setup-kagenti.sh --skip-mlflow                # Disable Kagenti-Operator <-> MLflow integration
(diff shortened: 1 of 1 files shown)


scripts/ocp/setup-kagenti.sh
```diff
@@ -1029,6 +1053,43 @@ else
 fi
 echo ""
 
+# ============================================================================
+# Step 5b: Install Kuadrant operator (optional, --with-kuadrant)
```
ChristianZaccaria: **Suggestion from Claude:** 
In this PR Kuadrant is installed after MCP Gateway. In the Kind installer, Kuadrant is installed before MCP Gateway (Step 8b → Step 9), because Kuadrant provides AuthPolicy CRDs that MCP Gateway policies may depend on. In this OCP installer, MCP Gateway (Step 5) runs before Kuadrant (Step 5b). It diverges from the Kind install order and could become a latent issue if the chart ever includes Kuadrant policy resources. Consider swapping the order so Kuadrant installs first (Step 5 = Kuadrant, Step 5b = MCP Gateway), consistent with Kind.

ChristianZaccaria (dismissed):
Thanks for the PR! No breaking changes. All flags default to `false` and are fully isolated. I left just one suggestion, mainly to assess whether it's worth switching the step ordering between Kuadrant and MCP Gateway.

cwiklik (approved):
Clean PR — single commit, CI green, flag logic is correct. The ordering suggestion from @ChristianZaccaria is worth considering (install Kuadrant before MCP Gateway for CRD availability) but non-blocking since MCP Gateway chart doesn't currently create AuthPolicy resources.

**Areas reviewed**: Shell
**Commits**: 1, signed-off
**CI**: all passing

scripts/ocp/setup-kagenti.sh
pdettori: done! thanks you for the suggestion @ChristianZaccaria !

PR #1466, opened 2026-05-05
Feat: Render UI version in the UI's title bar

### LINKED ISSUE #1305: feature: Show Kagenti release info
### Feature Description

Kagenti GUI does not currently show release info, which is useful for bug reporting and installation issue resolution.

### Proposed Solution

We can include the release info in the GUI via an "About" menu item.

### Want to contribute?

### Additional Context

_No response_

esnible:
## Summary

Resolves #1305

If you have a running Kagenti and K8s API access:

```bash
make build-load-ui-frontend
UI_TAG=$(docker images | grep kagenti-ui-v2 | tail -n 1 | awk 'END {print $2}')
echo UI_TAG is $UI_TAG
oc -n kagenti-system set image deployment/kagenti-ui frontend=ghcr.io/kagenti/kagenti-ui-v2:${UI_TAG}
```

cc @webchang 

Getting the version from the _project.json_ file is typical.  In this implementation we calculate it while building the container image.  Kiali does something similar when a new release is made on GitHub.  That approach is an alternative.  (See https://github.com/kiali/kiali/blob/master/.github/workflows/bump-release-version.yml#L102 )

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -117,6 +117,8 @@ jobs:
           file: ${{ matrix.image_config.context }}/${{ matrix.image_config.dockerfile }}
           load: true
           tags: ${{ matrix.image_config.name }}:smoke-test
+          build-args: |
+            RELEASE_TAG=${{ github.ref_name }}
 
       - name: Run smoke test for ${{ matrix.image_config.name }}
         if: matrix.image_config.smoke_cmd
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -133,6 +133,8 @@ jobs:
           platforms: linux/amd64,linux/arm64
           tags: ${{ steps.meta_retry.outputs.tags || steps.meta.outputs.tags }}
           labels: ${{ steps.meta_retry.outputs.labels || steps.meta.outputs.labels }}
+          build-args: |
+            RELEASE_TAG=${{ github.ref_name }}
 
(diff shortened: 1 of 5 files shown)


mrsabath (approved):
## Review Summary

Clean implementation of version display in the UI title bar. The approach is sound: pass `RELEASE_TAG` as a Docker build-arg, `sed` it into `package.json` before `npm run build`, then import the version from the React component. Both CI (`build.yaml`) and local dev (`Makefile`) paths are covered.

**Verified:**
- `tsconfig.json` already has `resolveJsonModule: true` — JSON import works
- `sed -r` works on Alpine BusyBox (Deploy & Test confirms this)
- `github.ref_name` is safe: workflow triggers on tag push (`v*`) or branch push (`main`) — no `/` in either case
- Makefile uses `git describe --tags --abbrev=0` for local builds — falls back to `unknown` gracefully
- Version placed correctly before `npm run build` in the Dockerfile

**Minor nit (non-blocking):**
- `sed -i -r` → `sed -i -E` would be more POSIX-portable, though irrelevant for this Alpine-only Dockerfile.

---

**Areas reviewed**: Dockerfile, CI/GitHub Actions, Frontend (React/CSS), Makefile, Security
**Commits**: 4, all DCO signed-off ✓
**CI status**: All checks pass (E2E pending `/run-e2e`)

LGTM — nice feature, minimal footprint.

PR #1467, opened 2026-05-05
fix: add show-services.sh hint after deploy and credential setup

pdettori:
## Summary

- Adds a log line in `87-setup-test-credentials.sh` pointing users to `show-services.sh` to find login credentials
- Adds a `show-services.sh` hint in `kind-full-test.sh` when the cluster is kept (`--skip-cluster-destroy`)

The test credential setup script resets the `admin` user password to a random value (stored in `kagenti-test-user` secret). Users who expect `admin/admin` from the realm import are surprised when login fails. These hints surface the right command to run.

## Test plan

- [ ] Run `kind-full-test.sh --skip-cluster-destroy` and verify the `show-services.sh` hint appears in the output
- [ ] Run `87-setup-test-credentials.sh` directly and verify the log line about `show-services.sh` appears

Assisted-By: Claude Code

### DIFF
--- a/.github/scripts/common/87-setup-test-credentials.sh
+++ b/.github/scripts/common/87-setup-test-credentials.sh
@@ -277,3 +277,4 @@ fi
 log_success "Test credentials ready"
 log_info "  Test user: $TEST_USER (realm: $REALM)"
 log_info "  Service account: $E2E_CLIENT_ID"
+log_info "  Run ./.github/scripts/local-setup/show-services.sh to see login credentials"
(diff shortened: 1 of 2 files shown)


mrsabath (approved):
## Review Summary

Trivial UX improvement — adds log hints pointing users to `show-services.sh` for login credentials after deploy and credential setup. Addresses the common confusion when test setup resets the admin password to a random value.

No issues found. DCO signed, Shell Script Lint passes.

LGTM.

PR #1481, opened 2026-05-06
fix(chat): Forward A2A contextId on every turn, adopt once

huang195:
## Summary

The backend was synthesizing its own `session_id` and never forwarding it to the agent as `params.contextId`. Every HTTP call from the UI created a fresh A2A context server-side, which also caused the authbridge session telemetry viewer (`abctl`) to split a single conversation across many session buckets.

Two behaviors changed in `/kagenti/backend/app/routers/chat.py`:

1. **Forward the client's session_id as `params.contextId`** on every A2A send. Applies to both `/send` (non-streaming) and `/stream` (SSE). Multi-turn conversations now stay in one A2A context chain as the spec requires.

2. **Adopt the agent-assigned `contextId` from the response only on the first turn** (when the caller had no `session_id` yet). Some A2A SDKs mint a fresh `contextId` on every response even when the client supplied one — adopting that every turn was churning the UI's `session_id` state and causing telemetry to split across a new bucket per message.

Adds a `caller_supplied_session_id` flag to `_stream_from_response` so the adoption logic can see whether the caller is already on a known conversation, and one INFO-level log line tracing the outbound `contextId` and any first-turn adoption.

## Test plan

- [x] Manual: send 2-3 messages in one conversation via the UI, confirm `abctl` shows one session bucket with event count climbing by ~23 per message (instead of one new bucket per message).
- [x] Manual: start a fresh second conversation, confirm it creates exactly one new bucket and stays there across all its turns.
- [x] Confirmed the `[a2a-debug] adopting agent contextId (first turn): …` log line appears only on the first message of each conversation, never on follow-ups.
- [ ] Automated test coverage could follow — the current test suite in `backend/tests/` doesn't cover the A2A wire-format construction.

### DIFF
--- a/kagenti/backend/app/routers/chat.py
+++ b/kagenti/backend/app/routers/chat.py
@@ -157,18 +157,25 @@ async def send_message(
     agent_url = resolve_agent_url(name, namespace, kube)
     session_id = request.session_id or uuid4().hex
 
-    # Build A2A message payload
+    # Build A2A message payload. When the frontend supplied a session_id
+    # (subsequent turns of the same conversation), forward it as
+    # params.contextId so the agent joins the existing A2A task chain
+    # instead of spinning up a fresh context per message.
+    params: dict = {
+        "message": {
+            "role": "user",
+            "parts": [{"kind": "text", "text": request.message}],
+            "messageId": uuid4().hex,
+        },
+    }
+    if request.session_id:
+        params["contextId"] = request.session_id
+
     message_payload = {
         "jsonrpc": "2.0",
         "id": str(uuid4()),
         "method": "message/send",
-        "params": {
-            "message": {
-                "role": "user",
-                "parts": [{"kind": "text", "text": request.message}],
-                "messageId": uuid4().hex,
-            },
-        },
+        "params": params,
     }
 
     # Prepare headers with optional Authorization
@@ -192,6 +199,15 @@ async def send_message(
             content = ""
             if "result" in result:
                 result_data = result["result"]
+                # Adopt the agent-assigned contextId only on the very first
+                # turn (when the caller had no session_id yet). Subsequent
+                # turns must reuse the caller's session_id — some agent SDKs
+                # mint a fresh contextId every response, and adopting that
+                # would churn the UI state and split events across buckets.
+                if not request.session_id:
+                    agent_ctx = result_data.get("contextId") or result_data.get("sessionId")
+                    if agent_ctx:
+                        session_id = agent_ctx
                 # Handle Task response
                 if "status" in result_data and "message" in result_data.get("status", {}):
                     parts = result_data["status"]["message"].get("parts", [])
@@ -287,6 +303,7 @@ async def _stream_from_response(
     response: httpx.Response,
     session_id: str,
     username: Optional[str] = None,
+    caller_supplied_session_id: bool = False,
 ):
     """Stream SSE events from an already-connected agent response.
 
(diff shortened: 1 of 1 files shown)


kagenti/backend/app/routers/chat.py
```diff
@@ -192,6 +199,15 @@ async def send_message(
             content = ""
             if "result" in result:
                 result_data = result["result"]
+                # Adopt the agent-assigned contextId only on the very first
+                # turn (when the caller had no session_id yet). Subsequent
+                # turns must reuse the caller's session_id — some agent SDKs
+                # mint a fresh contextId every response, and adopting that
+                # would churn the UI state and split events across buckets.
+                if not request.session_id:
```
esnible, line 207: **nit**: The `/send` adoption guard is `if agent_ctx:` while the `/stream` path (added below in `_stream_from_response`) uses `if agent_ctx and agent_ctx != session_id:`. The extra check is harmless here since on a first turn `session_id` is a freshly-minted `uuid4().hex` (near-zero chance of colliding with the agent's contextId), but the two blocks read as if they should match. Consider adding `and agent_ctx != session_id` here for consistency.

esnible, line 207: **nit**: The PR description says "Adds one INFO log line tracing the outbound contextId and any first-turn adoption" and the test plan references a `[a2a-debug] adopting agent contextId (first turn)` log line. The diff contains no added log lines — it only removes an existing `logger.debug(...)`. Worth updating the PR description to match the actual change.

esnible (approved):
Single-file Python change that correctly fixes A2A `contextId` forwarding and first-turn adoption for both `/send` (non-streaming) and `/stream` (SSE) paths. The core logic is sound: forward the client's `session_id` as `params.contextId` on outbound A2A calls, and only adopt the agent-assigned `contextId` on the very first turn. The `ChatResponse` already returns `session_id`, so the adopted ID propagates back to the UI correctly.

All CI checks pass. The E2E HyperShift failure is an infrastructure teardown timeout unrelated to this change (Ansible reported `failed=0` in the final play recap — the cluster was successfully destroyed).

Assisted-By: Claude Code

PS: I would love to see a diagram, Mermaid or ASCII art, showing the flow of authorization and session headers and cookies from the browser to the ui-backend to authbridge to the workload, then to authbridge again and then to a tool and back.  https://github.com/kagenti/kagenti/blob/main/docs/tech-details.md gives the networking layer but as a new developer on the team it was difficult to understand what changes this PR introduced.

I was also unable to understand if these changes would have an effect if the end user was using the A2A Inspector instead of Kagenti to talk to the agents.

huang195:
/run-e2e

Previous run failed in `agent-oauth-secret-1` OpenShift build with `Could not resolve host: github.com` — transient DNS issue in the HyperShift CI env, unrelated to this PR. Retrying.

huang195:
/run-e2e

Prior run failed in `Deploy Kagenti` with `registry.redhat.io` returning 503/504 (cert-manager, servicemesh-operator, keycloak all stuck in ImagePullBackOff). Registry outage, unrelated to this PR. Retrying.

huang195:
E2E on hypershift failed 3 times not related to this PR:

- 1st (25413706469) │ Backend E2E tests                     │ HTTP 401 from agent — pre-existing Keycloak/token fixture issue 
- 2nd (25434841506) │ OpenShift build of agent-oauth-secret │ Could not resolve host: github.com — DNS flake                     
- 3rd (25437081400) │ Deploy Kagenti                        │ registry.redhat.io returning 503/504 — registry outage

Going to run e2e again

PR #1484, opened 2026-05-06
fix: Use oc delete for HyperShift cluster cleanup instead of ansible

RyanJenkins99:
## Problem                                                                                                                                                                                                                                                    
   
  The automated cleanup workflow (`.github/workflows/cleanup-stale-hypershift-clusters.yaml`) was detecting stale HyperShift clusters correctly but **failing to delete them** in CI.                                                                           
                  
  **Root cause:** The cleanup script called `hypershift-full-test.sh` which requires `hypershift-automation` (ansible playbooks). This dependency is not available in GitHub Actions runners, causing all deletions to fail with:                               
  Error: hypershift-automation not found
                                                                                                                                                                                                                                                                
  Additionally, the script reported "✓ Deleted X cluster(s)" even when deletions failed, making failures invisible.
                                                                                                                                                                                                                                                                
  ## Solution     
                                                                                                                                                                                                                                                                
  Replace ansible-based deletion with direct `oc delete hostedcluster` command:                                                                                                                                                                                 
  - Works in both CI and local environments (no external dependencies)
  - HostedCluster deletion triggers cascading cleanup of AWS resources                                                                                                                                                                                          
  - Add accurate deletion tracking with `DELETED_COUNT` and `FAILED_COUNT`                                                                                                                                                                                      
  - Fix misleading success messages when deletions fail                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                
  ## Testing                                                                                                                                                                                                                                                    
                  
  Created test cluster with auto-cleanup labels:                                                                                                                                                                                                                
  ```bash
  oc label hostedcluster kagenti-hypershift-ci-ztest \                                                                                                                                                                                                          
    kagenti.io/auto-cleanup=enabled \                                                                                                                                                                                                                           
    kagenti.io/ttl-hours=0
                                                                                                                                                                                                                                                                
  Verified:       
  - ✅ Workflow detects stale cluster                                                                                                                                                                                                                           
  - ✅ Script successfully deletes cluster with oc delete                                                                                                                                                                                                       
  - ✅ Accurate reporting of deletion results
                                                                                                                                                                                                                                                                
  Impact          
                                                                                                                                                                                                                                                                
  The automated cleanup workflow will now actually delete stale clusters every 3 hours instead of just reporting them.                                                                                                                                          
  ```

### DIFF
--- a/.github/scripts/hypershift/cleanup-stale-clusters.sh
+++ b/.github/scripts/hypershift/cleanup-stale-clusters.sh
@@ -156,6 +156,8 @@ NOW_EPOCH=$(date +%s)
 STALE_COUNT=0
 PROTECTED_COUNT=0
 OK_COUNT=0
+DELETED_COUNT=0
+FAILED_COUNT=0
 
 for CLUSTER_NAME in $CLUSTERS; do
     # Pattern filtering
@@ -229,12 +231,15 @@ for CLUSTER_NAME in $CLUSTERS; do
             echo "       Would delete (use --apply to execute)"
         else
             echo "       Deleting cluster..."
-            # Extract cluster suffix from full name
-            CLUSTER_SUFFIX="${CLUSTER_NAME##*-}"
-            "$REPO_ROOT/.github/scripts/local-setup/hypershift-full-test.sh" \
-                --include-cluster-destroy "$CLUSTER_SUFFIX" 2>&1 | tee "/tmp/cleanup-${CLUSTER_NAME}.log" || {
-                    log_error "Failed to delete $CLUSTER_NAME (see /tmp/cleanup-${CLUSTER_NAME}.log)"
-                }
+            # Use direct oc delete for reliable cleanup in CI and local environments
+            # The HostedCluster deletion triggers cascading cleanup of AWS resources
+            if oc delete hostedcluster "$CLUSTER_NAME" -n clusters 2>&1 | tee "/tmp/cleanup-${CLUSTER_NAME}.log"; then
+                log_success "Successfully deleted $CLUSTER_NAME"
+                DELETED_COUNT=$((DELETED_COUNT + 1))
+            else
+                log_error "Failed to delete $CLUSTER_NAME (see /tmp/cleanup-${CLUSTER_NAME}.log)"
+                FAILED_COUNT=$((FAILED_COUNT + 1))
+            fi
         fi
         echo ""
         STALE_COUNT=$((STALE_COUNT + 1))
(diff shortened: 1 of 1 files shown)


Ladas (approved):
## Security & Code Review

**Scope**: 1 file, +21/-8 — replaces ansible-based HyperShift cluster deletion with direct `oc delete`.

### Security Assessment: PASS

- **Command injection**: `$CLUSTER_NAME` comes from the Kubernetes API (`oc get ... -o jsonpath`), constrained to RFC 1123 DNS names (no shell metacharacters possible). Properly double-quoted in both `oc delete` and log path.
- **Privilege**: Same `oc` CLI with same `KUBECONFIG` — no privilege escalation.
- **Deletion scope**: Unchanged — filtered by label `kagenti.io/auto-cleanup=enabled` with TTL validation and protection checks.
- **Path traversal**: `/tmp/cleanup-${CLUSTER_NAME}.log` is safe since Kubernetes names cannot contain `..` or `/`.

### Code Review: PASS

- Root cause is solid: `hypershift-full-test.sh` requires `hypershift-automation` (ansible), which isn't available in GHA runners. Direct `oc delete hostedcluster` is the right fix — it triggers cascading AWS cleanup without external dependencies.
- Accurate deletion tracking with `DELETED_COUNT`/`FAILED_COUNT` fixes the misleading "Deleted X cluster(s)" message.
- The `tee` to per-cluster log files is a nice touch for debugging failed deletions.

No security vulnerabilities found. No functional issues.

Assisted-By: Claude Code

.github/scripts/hypershift/cleanup-stale-clusters.sh
```diff
@@ -265,8 +270,16 @@ if [ "$DRY_RUN" = "true" ] && [ "$STALE_COUNT" -gt 0 ]; then
...
+elif [ "$DRY_RUN" = "false" ]; then
+    if [ "$DELETED_COUNT" -gt 0 ]; then
+        log_success "Successfully deleted $DELETED_COUNT cluster(s)"
+    fi
+    if [ "$FAILED_COUNT" -gt 0 ]; then
+        log_error "Failed to delete $FAILED_COUNT cluster(s)"
+    fi
+    if [ "$STALE_COUNT" -eq 0 ]; then
+        log_success "No stale clusters found"
+    fi
```
pdettori, line 282: nit: This `else` branch is unreachable — `DRY_RUN` is only ever `"true"` (default) or `"false"` (via `--apply`). Not harmful, but dead code that could be removed for clarity.

pdettori (approved):
Clean fix — replaces an ansible dependency that doesn't exist in CI with direct `oc delete`. Deletion tracking with separate counters is a good improvement. All checks pass; one minor nit inline.

PR #1480, opened 2026-05-06
feat(test): multi-tenant E2E tests — MVP validation criteria #1-4, #6-8

pdettori:
## Summary

- Add `test_12_tenant_isolation.py` — multi-tenant isolation tests (criteria #6-8)
  - JWT audience-based auth rejection across tenants
  - ServiceAccount RBAC namespace scoping
  - Credential/secret isolation between tenant namespaces
- Add `test_13_sandbox_connectivity.py` — sandbox interactive access (criterion #2)
  - Gateway pod/service/endpoint health
  - kubectl exec into sandbox pods (terminal session capability)
- Tag existing tests with `mvp` marker (criteria #1, #3, #4 already covered)
- Register `mvp` pytest marker in conftest.py

## Test plan

- [x] `pytest -m mvp` collects all MVP-tagged tests
- [x] `pytest kagenti/tests/e2e/openshell/test_12_tenant_isolation.py --collect-only` shows 20 tests
- [x] CI workflow `e2e-openshell-kind.yaml` runs with 2 tenants deployed
- [x] Tenant isolation auth tests pass with OIDC-enabled gateways
- [x] RBAC tests verify namespace scoping via `kubectl auth can-i`

Closes #1362
Deferred criteria (#5, #9, #10, #11) tracked in #1479

Assisted-By: Claude Code

PR #1491, opened 2026-05-06
fix(e2e): Enable operator-managed client registration + gate fixture on audience scope

huang195:
## Summary

Fixes the intermittent HyperShift E2E 401 failure that surfaced on #1480 / #1481. The incident has two root causes working together; both are addressed here.

1. **Operator-managed client registration was never running.** The kagenti-operator chart passed `--enable-client-registration=true` (legacy webhook sidecar gate) but not `--enable-operator-client-registration` (the new controller's gate — defaults to `false`). Meanwhile the webhook's default precedence refuses to inject the legacy sidecar unless the workload opts in with `kagenti.io/client-registration-inject=true`. Result: neither path registered the agent in Keycloak; no `agent-<ns>-<workload>-aud` scope; test token had no `aud`; AuthBridge returned 401.
2. **The test fixture had no signal for (1).** `87-setup-test-credentials.sh` attached whatever agent audience scopes happened to exist at the moment it ran. When no agent was registered it silently logged *"No agent audience scopes found (agents may not be deployed yet)"* and moved on — the test then failed deep inside the agent call with a cryptic 401.

## Root cause (evidence from red run `25439863506` + run `25464732405`)

- Fixture at `14:24:38`: `"No agent audience scopes found (agents may not be deployed yet)"`
- AuthBridge at `14:24:47`: `JWT validation failed error="audience is required (prevents confused deputy attacks)"`
- Mutator at `14:17:57`: `deliveryPaths:"skip", keycloakClientCredentialsSecretName:"", injectClientRegistrationSidecar:false`
- Weather-service pod age `7m20s, RESTARTS=0` — Deployment never patched by the reconciler, so scope step never completed
- Operator setup log: `"Operator-managed client registration controller enabled"` **missing** — the controller was never started because the CLI flag was unset

### `charts/kagenti/values.yaml`

Flip the operator args so the deploy config commits to exactly one active client-registration path:

```diff
-      - "--enable-client-registration=true"
+      - "--enable-client-registration=false"
+      - "--enable-operator-client-registration=true"
```

- Disable `--enable-client-registration` (legacy webhook sidecar gate).
- Enable `--enable-operator-client-registration` (controller that registers the Keycloak client, creates the credentials Secret, and annotates the Deployment template; the webhook then mounts the Secret via its `ApplyKeycloakClientCredentialsSecretVolumes` path).

Legacy sidecar code remains in the operator binary as a fallback. Removing it is tracked as a follow-up once operator-managed registration is proven green in E2E.

### `87-setup-test-credentials.sh`

When agent deployments exist in `team1`, poll the realm for at least one `agent-*-aud` scope before attaching. Refresh the admin token each iteration (master-realm admin tokens have ~60s lifetime; a 5-min wait would otherwise fire `PUT` calls with an expired bearer).

```bash
AGENT_COUNT=$(kubectl get deployment -n team1 -l kagenti.io/type=agent -o name | wc -l)
if [ "$AGENT_COUNT" -gt 0 ]; then
    # poll Keycloak realm default-default-client-scopes, break on first agent-*-aud
fi
```

Defaults: 300s timeout, 5s poll interval, `SCOPE_WAIT_TIMEOUT` env var to override. Happy path (scope already exists) returns immediately — no measurable slowdown. On timeout, dumps reconciler diagnostics (pod-mutator annotations, authbridge-config ConfigMap, operator logs, Keycloak realm client list, realm default scopes) so the next maintainer can debug without a second CI run.

### `tests/conftest.py`

Extract token acquisition into `_acquire_agent_token` and add a `keycloak_agent_token_refresh` fixture returning a zero-arg callable. This lets a future test retry on 401 by minting a freshly-scoped token without touching fixture machinery. Not wired into tests in this PR.

## Why this is the right shape

The failure was a compound: an async scope creator that wasn't actually running, and a sync fixture that silently tolerated its absence. The correct fix is to **make the producer run by default** (chart flip) **and** to **make the consumer wait on the producer's observable output** (fixture gate). Nothing else has to change in the operator, authbridge, or workload YAML. The fixture gate stays even once operator-managed registration is fully trusted — it's defensive diagnostics for any future regression in the reconciler.

## Test plan

- [x] Shell lint (`shellcheck` in CI)
- [x] Existing E2E harness unit tests (none touched)
- [ ] `/run-e2e` on this PR. Expected log sequence:
      ```
      Operator-managed client registration controller enabled
      operator client registration applied workload=weather-service namespace=team1 secret=kagenti-keycloak-client-credentials-<hash>
      Waiting up to 300s for agent-*-aud scopes to appear in realm kagenti...
        Agent audience scope(s) detected after Ns
      Attaching agent audience scopes to kagenti-e2e-tests...
        Added scope 'agent-team1-weather-service-aud' to kagenti-e2e-tests
      ```
- [ ] `test_agent_simple_query`, `test_agent_multiturn_conversation` pass with no 401 from `weather-service`.
- [ ] Regression check on an already-healthy cluster: detection within 0–5s.

## Not in scope (tracked separately)

- Removing the legacy webhook-injected `client-registration` sidecar code from kagenti-operator (and the `kagenti-extensions/client-registration` image). Deferred until operator-managed registration is proven green across multiple E2E runs.
- Wiring the `keycloak_agent_token_refresh` fixture into test retry logic — left as follow-up pending a real-world 401 after this lands.

Refs: kagenti/kagenti-extensions#375 (complementary in-pod credential polling resilience).

### DIFF
--- a/kagenti/examples/agents/weather_service_deployment_ocp.yaml
+++ b/kagenti/examples/agents/weather_service_deployment_ocp.yaml
@@ -29,6 +29,14 @@ spec:
         protocol.kagenti.io/a2a: ""
         kagenti.io/framework: LangGraph
         app.kubernetes.io/name: weather-service
+        # Opt this E2E agent into the legacy client-registration sidecar
+        # (rather than the default operator-managed registration). The
+        # legacy path creates the Keycloak client + agent-*-aud scope
+        # synchronously at pod startup, so the test fixture's scope poll
+        # reliably finds it. The operator-managed reconciler has been
+        # observed stuck on HyperShift CI; see fix/e2e-audience-scope-gate.
+        # Remove this label once the operator-managed path is reliable.
+        kagenti.io/client-registration-inject: "true"
     spec:
       serviceAccountName: weather-service
       containers:
--- a/.github/scripts/common/87-setup-test-credentials.sh
+++ b/.github/scripts/common/87-setup-test-credentials.sh
@@ -57,10 +57,17 @@ kc_api() {
 # Get admin token (master realm)
 # ============================================================================
 
-ADMIN_TOKEN=$(curl -sk -X POST \
-    "$KEYCLOAK_URL/realms/master/protocol/openid-connect/token" \
-    -d "grant_type=password&client_id=admin-cli&username=$ADMIN_USER&password=$ADMIN_PASS" \
-    | python3 -c "import sys,json; print(json.load(sys.stdin)['access_token'])" 2>/dev/null || echo "")
+# Factored out so the scope-existence poll below can refresh the admin
+# token. The master-realm admin access token lives ~60s; a 5-minute wait
+# loop would otherwise fire API calls with an expired bearer.
+refresh_admin_token() {
+    ADMIN_TOKEN=$(curl -sk -X POST \
+        "$KEYCLOAK_URL/realms/master/protocol/openid-connect/token" \
+        -d "grant_type=password&client_id=admin-cli&username=$ADMIN_USER&password=$ADMIN_PASS" \
+        | python3 -c "import sys,json; print(json.load(sys.stdin)['access_token'])" 2>/dev/null || echo "")
+}
+
+refresh_admin_token
 
 if [ -z "$ADMIN_TOKEN" ]; then
     log_error "Failed to get Keycloak admin token"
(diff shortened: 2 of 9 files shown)


huang195:
/run-e2e

Third commit just landed:
- Opts the E2E weather-service into the legacy `client-registration` sidecar path via `kagenti.io/client-registration-inject: "true"` label. Sidecar creates Keycloak client + agent-*-aud scope synchronously at pod startup, so the fixture's scope poll should find it within seconds instead of timing out against the stuck operator reconciler.
- Scope-gate timeouts now dump comprehensive reconciler diagnostics inline (agent deployments, team1 ConfigMaps/Secrets, operator pod status, broader operator-log grep, Keycloak clients, realm default scopes) — making any future failure self-diagnostic without a second CI run.

Expected happy path:
```
Waiting up to 300s for agent-*-aud scopes to appear in realm kagenti...
  Agent audience scope(s) detected after ~15s
Attaching agent audience scopes to kagenti-e2e-tests...
  Added scope 'agent-team1-weather-service-aud' to kagenti-e2e-tests
...
test_agent_simple_query PASSED
test_agent_multiturn_conversation PASSED
```

kagenti/examples/agents/weather_service_deployment_ocp.yaml
```diff
@@ -29,6 +29,14 @@ spec:
         protocol.kagenti.io/a2a: ""
         kagenti.io/framework: LangGraph
         app.kubernetes.io/name: weather-service
+        # Opt this E2E agent into the legacy client-registration sidecar
+        # (rather than the default operator-managed registration). The
+        # legacy path creates the Keycloak client + agent-*-aud scope
+        # synchronously at pod startup, so the test fixture's scope poll
+        # reliably finds it. The operator-managed reconciler has been
```
pdettori, line 36: suggestion: The workaround label is well-documented in the comment — consider opening a tracking issue for its removal (once operator-managed registration is reliable on HyperShift) so it doesn't become permanent tech debt.

pdettori (approved):
LGTM — well-structured fix for the compound E2E 401 failure. Chart flag flip, scope-gate polling with token refresh, SPIRE trust domain plumbing, UI default alignment, and helm-unittest CI unblock all look correct. All 8 commits signed-off, CI green including E2E HyperShift.

One minor suggestion inline (tracking issue for the workaround label).

PR #1496, opened 2026-05-07
fix: show actual Keycloak credentials at end of Kind installer

### LINKED ISSUE #1495: Kind installer shows stale Keycloak credentials (admin/admin from keycloak-initial-admin)
## Problem

The Kind installer script `scripts/kind/setup-kagenti.sh` shows incorrect Keycloak credentials at the end of installation (line 1254):

```
Keycloak credentials:
  kubectl get secret keycloak-initial-admin -n keycloak -o go-template='User: {{.data.username | base64decode}}  Pass: {{.data.password | base64decode}}'
```

This returns `admin/admin`, but that is NOT the password users need to log into the Kagenti UI. The correct credentials are shown by `.github/scripts/local-setup/show-services.sh`.

## Root Cause

The `kagenti` Helm chart includes a Job (`agent-oauth-secret-job.yaml`) that runs automatically during installation. This Job:

1. Reads the Keycloak admin creds from `keycloak-initial-admin` (master realm, `admin/admin`) to authenticate to the Keycloak API
2. Creates/updates an `admin` user in the **kagenti realm** with a **randomly generated password** (`secrets.token_urlsafe(16)`)
3. Stores the random password in a K8s secret called `kagenti-test-user` in the `keycloak` namespace

Source: `kagenti/auth/agent-oauth-secret/agent_oauth_secret.py` (lines 437-482)

So the installation flow is:
1. `kagenti-deps` chart deploys Keycloak with realm import → kagenti realm user `admin` has password `admin`
2. `kagenti` chart deploys → `agent-oauth-secret` Job **resets** the `admin` user password to a random token
3. `setup-kagenti.sh` prints `keycloak-initial-admin` → shows `admin/admin` (the master realm bootstrap credential, NOT the kagenti realm user password)

The `keycloak-initial-admin` secret is the **master realm admin** credential (used by Keycloak's bootstrap process and the Job itself). The password users need for the Kagenti UI is in the `kagenti-test-user` secret, which contains the randomly generated password set by the Job.

## Two distinct credentials

| What | Secret | Realm | Used for |
|------|--------|-------|----------|
| Keycloak Admin Console | `keycloak-initial-admin` | master | Admin console at /admin |
| Kagenti UI / MLflow | `kagenti-test-user` | kagenti | Application login |

The setup script only shows the first one, but users typically need the second.

## Proposed Fix

The end-of-script message in `setup-kagenti.sh` should either:

**Option A** (recommended): Tell users to run `show-services.sh` for credentials:
```bash
echo "  Credentials: ./.github/scripts/local-setup/show-services.sh"
```

**Option B**: Read both secrets and display them with context:
```bash
KC_ADMIN_PASS=$(kubectl get secret keycloak-initial-admin -n keycloak -o jsonpath='{.data.password}' | base64 -d)
APP_PASS=$(kubectl get secret kagenti-test-user -n keycloak -o jsonpath='{.data.password}' | base64 -d 2>/dev/null || echo "pending (Job still running)")
echo "  Keycloak Admin Console: admin / $KC_ADMIN_PASS"
echo "  Kagenti UI (kagenti realm): admin / $APP_PASS"
```

## Files Involved

- `scripts/kind/setup-kagenti.sh` (line 1254) — prints misleading credentials
- `kagenti/auth/agent-oauth-secret/agent_oauth_secret.py` (lines 437-482) — Job that generates random password and stores in `kagenti-test-user`
- `charts/kagenti/templates/agent-oauth-secret-job.yaml` — Job manifest with `CREATE_KEYCLOAK_TEST_USER=true`
- `charts/kagenti-deps/templates/keycloak-k8s.yaml` (lines 174-185) — hardcoded `admin/admin` bootstrap secret
- `.github/scripts/local-setup/show-services.sh` (lines 144-151) — reads correct secrets

Assisted-By: Claude Code

pdettori:
## Summary

- Replaces the static `kubectl get secret keycloak-initial-admin` command with actual credential reads at runtime
- Displays **Keycloak admin console** password (from `keycloak-initial-admin` secret)
- Displays **Kagenti UI login** password (from `kagenti-test-user` secret) only when `--with-ui` is enabled
- Handles the race condition where the `agent-oauth-secret` Job hasn't created the secret yet

## Context

The `agent-oauth-secret` Helm Job resets the kagenti realm `admin` user password to a random token during chart install. The old message showed `admin/admin` from the bootstrap secret which doesn't match the actual UI login credentials.

Closes #1495

## Test plan

- [x] Run `scripts/kind/setup-kagenti.sh --with-all` and verify both credentials are shown correctly
- [x] Run `scripts/kind/setup-kagenti.sh --with-istio` (no `--with-ui`) and verify UI password line is not shown
- [x] Verify the displayed UI password matches what `show-services.sh` reports

Assisted-By: Claude Code

### DIFF
--- a/scripts/kind/setup-kagenti.sh
+++ b/scripts/kind/setup-kagenti.sh
@@ -1250,8 +1250,19 @@ if $WITH_SPIRE; then
   echo "  Tornjak:      http://spire-tornjak-ui.${DOMAIN}:8080"
 fi
 echo ""
-echo "  Keycloak credentials:"
-echo "    kubectl get secret keycloak-initial-admin -n keycloak -o go-template='User: {{.data.username | base64decode}}  Pass: {{.data.password | base64decode}}'"
+echo "  Credentials:"
+KC_ADMIN_USER=$(kubectl get secret keycloak-initial-admin -n keycloak -o jsonpath='{.data.username}' 2>/dev/null | base64 -d 2>/dev/null || echo "N/A")
+KC_ADMIN_PASS=$(kubectl get secret keycloak-initial-admin -n keycloak -o jsonpath='{.data.password}' 2>/dev/null | base64 -d 2>/dev/null || echo "N/A")
+echo "    Keycloak admin console: ${KC_ADMIN_USER} / ${KC_ADMIN_PASS}"
+if $WITH_UI; then
+  UI_USER=$(kubectl get secret kagenti-test-user -n keycloak -o jsonpath='{.data.username}' 2>/dev/null | base64 -d 2>/dev/null || echo "")
+  UI_PASS=$(kubectl get secret kagenti-test-user -n keycloak -o jsonpath='{.data.password}' 2>/dev/null | base64 -d 2>/dev/null || echo "")
+  if [ -n "$UI_PASS" ]; then
+    echo "    Kagenti UI login:       ${UI_USER} / ${UI_PASS}"
+  else
+    echo "    Kagenti UI login:       (pending — run show-services.sh once platform is ready)"
+  fi
+fi
 echo ""
 
 ELAPSED=$(( SECONDS - START_SECONDS ))
--- a/scripts/kind/setup-kagenti.sh
+++ b/scripts/kind/setup-kagenti.sh
@@ -1251,9 +1251,13 @@ if $WITH_SPIRE; then
 fi
 echo ""
 echo "  Credentials:"
-KC_ADMIN_USER=$(kubectl get secret keycloak-initial-admin -n keycloak -o jsonpath='{.data.username}' 2>/dev/null | base64 -d 2>/dev/null || echo "N/A")
-KC_ADMIN_PASS=$(kubectl get secret keycloak-initial-admin -n keycloak -o jsonpath='{.data.password}' 2>/dev/null | base64 -d 2>/dev/null || echo "N/A")
-echo "    Keycloak admin console: ${KC_ADMIN_USER} / ${KC_ADMIN_PASS}"
+KC_ADMIN_USER=$(kubectl get secret keycloak-initial-admin -n keycloak -o jsonpath='{.data.username}' 2>/dev/null | base64 -d 2>/dev/null)
+KC_ADMIN_PASS=$(kubectl get secret keycloak-initial-admin -n keycloak -o jsonpath='{.data.password}' 2>/dev/null | base64 -d 2>/dev/null)
+if [ -n "$KC_ADMIN_PASS" ]; then
+  echo "    Keycloak admin console: ${KC_ADMIN_USER} / ${KC_ADMIN_PASS}"
+else
+  echo "    Keycloak admin console: (pending — secret keycloak-initial-admin not ready)"
+fi
 if $WITH_UI; then
   UI_USER=$(kubectl get secret kagenti-test-user -n keycloak -o jsonpath='{.data.username}' 2>/dev/null | base64 -d 2>/dev/null || echo "")
   UI_PASS=$(kubectl get secret kagenti-test-user -n keycloak -o jsonpath='{.data.password}' 2>/dev/null | base64 -d 2>/dev/null || echo "")


Ladas (approved):
## Security & Code Review

**Scope**: 1 file, +13/-2 — replaces static `kubectl get secret` hint with actual runtime credential reads.

### Security Assessment: PASS

- **Credential handling**: Keycloak admin + UI test user passwords are read via `kubectl` with hardcoded secret names/namespaces/jsonpath — no untrusted input interpolation.
- **Exposure scope**: Credentials print to the developer's terminal only — not exported, not written to files. This is expected behavior for a local Kind installer.
- **Race condition handling**: Proper fallback with `2>/dev/null || echo "N/A"` and `[ -n "$UI_PASS" ]` guard before printing.
- **Variable scoping**: Credentials stored in local shell variables, exist only for the script's lifetime.

### Code Review: PASS

- Clean, minimal change that improves UX by showing actual credentials instead of a copy-paste command.
- The `--with-ui` conditional for UI password display is correct — avoids confusion when UI isn't deployed.
- Fallback message when `kagenti-test-user` secret isn't ready is helpful.

No security vulnerabilities found. No functional issues.

Assisted-By: Claude Code

scripts/kind/setup-kagenti.sh
```diff
@@ -1250,8 +1250,19 @@ if $WITH_SPIRE; then
   echo "  Tornjak:      http://spire-tornjak-ui.${DOMAIN}:8080"
 fi
 echo ""
-echo "  Keycloak credentials:"
-echo "    kubectl get secret keycloak-initial-admin -n keycloak -o go-template='User: {{.data.username | base64decode}}  Pass: {{.data.password | base64decode}}'"
+echo "  Credentials:"
+KC_ADMIN_USER=$(kubectl get secret keycloak-initial-admin -n keycloak -o jsonpath='{.data.username}' 2>/dev/null | base64 -d 2>/dev/null || echo "N/A")
+KC_ADMIN_PASS=$(kubectl get secret keycloak-initial-admin -n keycloak -o jsonpath='{.data.password}' 2>/dev/null | base64 -d 2>/dev/null || echo "N/A")
```
huang195: Paolo, a piccolo problemino 🇮🇹 — the `|| echo "N/A"` fallback on this line (and the username line above) never actually fires. If the secret is missing, `kubectl 2>/dev/null` produces empty stdout with exit 1, but `base64 -d 2>/dev/null` happily decodes empty input with exit 0, so the `||` branch is skipped and `KC_ADMIN_USER` / `KC_ADMIN_PASS` end up as empty strings — the output becomes `Keycloak admin console:  / ` instead of `N/A / N/A`.

In practice the bootstrap secret is always present by the time this line runs, but the UI block just below already handles this correctly with an explicit `-z "$UI_PASS"` check and a clear 'pending' message. Would be nice to mirror that pattern here for consistency (or just drop the `|| echo "N/A"` since it's dead code). Non-blocking — just inconsistent with the UI branch right below it.

scripts/kind/setup-kagenti.sh
```diff
@@ -1250,8 +1250,19 @@ if $WITH_SPIRE; then
...
+echo "    Keycloak admin console: ${KC_ADMIN_USER} / ${KC_ADMIN_PASS}"
+if $WITH_UI; then
+  UI_USER=$(kubectl get secret kagenti-test-user -n keycloak -o jsonpath='{.data.username}' 2>/dev/null | base64 -d 2>/dev/null || echo "")
+  UI_PASS=$(kubectl get secret kagenti-test-user -n keycloak -o jsonpath='{.data.password}' 2>/dev/null | base64 -d 2>/dev/null || echo "")
+  if [ -n "$UI_PASS" ]; then
+    echo "    Kagenti UI login:       ${UI_USER} / ${UI_PASS}"
+  else
+    echo "    Kagenti UI login:       (pending — run show-services.sh once platform is ready)"
+  fi
+fi
```
huang195, line 1269: Nit / follow-up: this credential-printing logic now lives in three places — here, `.github/scripts/local-setup/show-services.sh`, and `.github/scripts/kind/access-ui.sh`. When the secret name or realm ever changes (and it has before), three files have to move in lock-step. Not a refactor for this PR, but worth a TODO or follow-up issue to extract something like `scripts/common/print-creds.sh`. Bellissimo lavoro either way 😄

huang195 (approved):
### Summary

Piccolo PR, scoped, fixes a real papercut: users no longer get told `admin/admin` for a UI that doesn't accept `admin/admin`. Bellissimo. All claims check out against the codebase, CI is verde, and the sign-off and `Assisted-By` trailer are both correct per repo policy. One subtle bug in the `N/A` fallback (cosmetic — won't trigger in normal runs), and a note on duplicated credential-printing logic across three scripts. Neither blocks merge.

**Areas reviewed**: Shell, secret/config references, commit convention, cross-script consistency
**Commits**: 1, signed-off ✅
**CI status**: all passing

### Verification

| Claim | Verified |
|---|---|
| `kagenti-test-user` secret exists in `keycloak` ns | ✅ `charts/kagenti/templates/agent-oauth-secret-job.yaml` + `conftest.py` |
| `agent-oauth-secret` Job generates the password | ✅ template comment confirms |
| `keycloak-initial-admin` is the bootstrap-admin secret | ✅ consistent across `show-services.sh`, docs |
| `$WITH_UI` defined and set by `--with-ui` / `--with-all` | ✅ lines 42, 86, 95 |

Verdict: **APPROVE** — perfetto. The two inline comments below are non-blocking; follow-up or land as-is, your call Paolo.

scripts/kind/setup-kagenti.sh
pdettori: Good catch! Fixed in c8bea9bd — now uses the same `-z` guard pattern as the UI block.

scripts/kind/setup-kagenti.sh
pdettori: Agreed — opened #1497 to track extracting a shared helper.

scripts/kind/setup-kagenti.sh
pdettori: Grazie !

PR #1512, opened 2026-05-08
Bug fix: Force CRD apply in case helm missed applying new CRDs

### LINKED ISSUE #1335: bug: Helm upgrade does not install new CRDs from kagenti-operator subchart
## Problem

When upgrading the `kagenti` Helm release, new CRDs introduced by the `kagenti-operator-chart` OCI subchart are not installed. This is due to a [known Helm limitation](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#some-caveats-and-explanations) where CRDs from subcharts are not updated on `helm upgrade`.

During an upgrade from operator `0.2.0-alpha.22` to `0.2.0-alpha.30`, the new `agentruntimes.agent.kagenti.dev` CRD was not installed. This caused the controller-manager to crash-loop, which cascaded into the webhook having no healthy endpoints (see cascade in Issue 4 below).

## Steps to reproduce

1. Install Kagenti with an operator version that does not ship `AgentRuntime` CRD (e.g. `0.2.0-alpha.22`)
2. Run the Ansible installer (`run-install.sh --env ocp`) targeting a newer operator version (e.g. `0.2.0-alpha.30`) that introduces `agentruntimes.agent.kagenti.dev`
3. Observe that the CRD is not created and the controller-manager crashes with a missing resource error

## Root cause

- The `kagenti-operator-chart` is declared as an OCI dependency in `charts/kagenti/Chart.yaml` (line 9-12)
- There is no `charts/kagenti/crds/` directory
- There is no Ansible task in `deployments/ansible/roles/kagenti_installer/tasks/main.yml` that explicitly applies Kagenti CRDs
- The project already handles this correctly for SPIRE (separate `spire-crds` Helm release) but not for its own operator CRDs

## Suggested fix

Add an explicit CRD application step to the installer after `helm dependency update` (around line 1876 in `main.yml`) that extracts CRDs from the operator subchart tarball and applies them via `kubectl apply -f`, similar to the `spire-crds` pattern.

Alternatively, vendor the operator CRDs into `charts/kagenti/crds/` at release time.

## Workaround

Manually create the CRD:
```bash
oc apply -f - <<CRDEOF
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: agentruntimes.agent.kagenti.dev
spec:
  group: agent.kagenti.dev
  versions:
    - name: v1alpha1
      served: true
      storage: true
      schema:
        openAPIV3Schema:
          type: object
          x-kubernetes-preserve-unknown-fields: true
  scope: Namespaced
  names:
    plural: agentruntimes
    singular: agentruntime
    kind: AgentRuntime
CRDEOF
```

## Environment

- Platform: OpenShift 4.20.14
- Previous operator: `0.2.0-alpha.22`
- Target operator: `0.2.0-alpha.30`
- Installer: Ansible playbook via `run-install.sh --env ocp`

esnible:
## Summary

Resolves #1335

First, install a back-level version of Kagenti by editing _charts/kagenti/Chart.yaml_ and setting the Kagenti operator version to 0.22:

```YAML
dependencies:
- name: kagenti-operator-chart
  version: 0.2.0-alpha.22
```

Then, install Kagenti: `scripts/kind/setup-kagenti.sh --with-istio --with-spire --with-ui --with-backend`

Verify that only one CRD is returned by `oc get crd | grep agent` (the agentruntimes CRD was not used by alpha 0.22)

```
agentcards.agent.kagenti.dev                   2026-05-07T18:04:34Z
```

Then, replace the kagenti-operator-chart in _charts/kagenti/Chart.yaml_ setting the version to `0.2.0-rc.1` and re-install Kagenti: `scripts/kind/setup-kagenti.sh --with-istio --with-spire --with-ui --with-backend`

Verify that only two CRDs are returned by `oc get crd | grep agent`

```
agentcards.agent.kagenti.dev                   2026-05-07T18:04:34Z
agentruntimes.agent.kagenti.dev                2026-05-08T12:01:26Z
```

Fixes #

#1335

### DIFF
--- a/scripts/ocp/setup-kagenti.sh
+++ b/scripts/ocp/setup-kagenti.sh
@@ -1044,17 +1044,14 @@ run_cmd helm upgrade --install kagenti "$KAGENTI_REPO/charts/kagenti/" \
   --set "keycloak.realm=${KC_REALM}" \
   --set "kagenti-operator-chart.mlflow.enable=$([ "$SKIP_MLFLOW" = true ] && echo false || echo true)"
 
-# Helm upgrade can miss additional CRDs, install them now
-CRD_SRC_DIR="$KAGENTI_REPO/charts/kagenti/charts/kagenti-operator-chart/crds"
-if [ ! -d "$CRD_SRC_DIR" ]; then
-  OPERATOR_TGZ_MATCHES=( "$KAGENTI_REPO"/charts/kagenti/charts/kagenti-operator-chart-*.tgz )
-  if [[ ${#OPERATOR_TGZ_MATCHES[@]} -ne 1 || ! -f "${OPERATOR_TGZ_MATCHES[0]}" ]]; then
-    log_error "..."
-    exit 1
-  fi
-  run_cmd tar -C "${CRD_TMP_DIR}" -xzf "${OPERATOR_TGZ_MATCHES[0]}" kagenti-operator-chart/crds
-  CRD_SRC_DIR="${CRD_TMP_DIR}/kagenti-operator-chart/crds"
-fi
+# Helm upgrade can miss additional CRDs, install them now.
+# There are three ways this script is configured with chart location
+# - No --kagenti-repo given: always clone fresh from upstream main
+# - --kagenti-repo is GitHub/git URL (cloned into cache)
+# - Local path provided — "use as-is"
+# In all three cases, KAGENTI_CACHE_DIR should point to downloaded Kagenti source
+KAGENTI_CACHE_DIR="${HOME}/.cache/kagenti"
+CRD_SRC_DIR="$KAGENTI_CACHE_DIR/charts/kagenti/charts/kagenti-operator-chart/crds"
 run_cmd kubectl apply --server-side --force-conflicts -f "${CRD_SRC_DIR}/"
 
 log_success "Kagenti installed"
--- a/scripts/ocp/setup-kagenti.sh
+++ b/scripts/ocp/setup-kagenti.sh
@@ -1044,6 +1044,18 @@ run_cmd helm upgrade --install kagenti "$KAGENTI_REPO/charts/kagenti/" \
   --set "keycloak.realm=${KC_REALM}" \
   --set "kagenti-operator-chart.mlflow.enable=$([ "$SKIP_MLFLOW" = true ] && echo false || echo true)"
 
+# Helm upgrade can miss additional CRDs, install them now
+NEW_TMP_DIR=$(mktemp -d)
+trap "rm -rf ${NEW_TMP_DIR}" EXIT
+log_info "Ensuring kagenti CRDs are updated"
+OPERATOR_TGZ_MATCHES=( "$KAGENTI_REPO"/charts/kagenti/charts/kagenti-operator-chart-*.tgz )
+if [[ ${#OPERATOR_TGZ_MATCHES[@]} -ne 1 || ! -f "${OPERATOR_TGZ_MATCHES[0]}" ]]; then
+  log_error "Expected exactly one kagenti-operator-chart-*.tgz in $KAGENTI_REPO/charts/kagenti/charts/, found ${#OPERATOR_TGZ_MATCHES[@]}: ${OPERATOR_TGZ_MATCHES[*]}"
+  exit 1
+fi
+run_cmd tar -C "${NEW_TMP_DIR}" -xzf "${OPERATOR_TGZ_MATCHES[0]}" kagenti-operator-chart/crds
+run_cmd kubectl apply -f "${NEW_TMP_DIR}/kagenti-operator-chart/crds/"
+
 log_success "Kagenti installed"
 
 # Grant otel-collector SA MLflow RBAC in agent namespaces (created by kagenti chart above)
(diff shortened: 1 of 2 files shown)


scripts/ocp/setup-kagenti.sh
```diff
@@ -1044,6 +1044,18 @@ run_cmd helm upgrade --install kagenti "$KAGENTI_REPO/charts/kagenti/" \
   --set "keycloak.realm=${KC_REALM}" \
   --set "kagenti-operator-chart.mlflow.enable=$([ "$SKIP_MLFLOW" = true ] && echo false || echo true)"
 
+# Helm upgrade can miss additional CRDs, install them now
+CRD_TMP_DIR=$(mktemp -d)
+log_info "Ensuring kagenti CRDs are updated"
+OPERATOR_TGZ_MATCHES=( "$KAGENTI_REPO"/charts/kagenti/charts/kagenti-operator-chart-*.tgz )
+if [[ ${#OPERATOR_TGZ_MATCHES[@]} -ne 1 || ! -f "${OPERATOR_TGZ_MATCHES[0]}" ]]; then
```
huang195: **must-fix.** Regression for `--operator-repo` dev mode. When that flag is set, line 976 removes `kagenti-operator-chart-*.tgz` and lines 977-978 copy the local operator chart as a directory at `$KAGENTI_REPO/charts/kagenti/charts/kagenti-operator-chart/` instead. The new glob here finds zero tgz matches → `exit 1`, immediately after what looks like a successful helm install.

Fix suggestion — prefer the extracted directory when present, fall back to tgz:

```bash
CRD_SRC_DIR="$KAGENTI_REPO/charts/kagenti/charts/kagenti-operator-chart/crds"
if [ ! -d "$CRD_SRC_DIR" ]; then
  OPERATOR_TGZ_MATCHES=( "$KAGENTI_REPO"/charts/kagenti/charts/kagenti-operator-chart-*.tgz )
  if [[ ${#OPERATOR_TGZ_MATCHES[@]} -ne 1 || ! -f "${OPERATOR_TGZ_MATCHES[0]}" ]]; then
    log_error "..."
    exit 1
  fi
  run_cmd tar -C "${CRD_TMP_DIR}" -xzf "${OPERATOR_TGZ_MATCHES[0]}" kagenti-operator-chart/crds
  CRD_SRC_DIR="${CRD_TMP_DIR}/kagenti-operator-chart/crds"
fi
run_cmd kubectl apply -f "${CRD_SRC_DIR}/"
```

Kind script doesn't have `--operator-repo`, so only OCP is affected.

scripts/ocp/setup-kagenti.sh
```diff
@@ -1044,6 +1044,18 @@ run_cmd helm upgrade --install kagenti "$KAGENTI_REPO/charts/kagenti/" \
...
+  log_error "Expected exactly one kagenti-operator-chart-*.tgz in $KAGENTI_REPO/charts/kagenti/charts/, found ${#OPERATOR_TGZ_MATCHES[@]}: ${OPERATOR_TGZ_MATCHES[*]}"
+  exit 1
+fi
+run_cmd tar -C "${CRD_TMP_DIR}" -xzf "${OPERATOR_TGZ_MATCHES[0]}" kagenti-operator-chart/crds
+run_cmd kubectl apply -f "${CRD_TMP_DIR}/kagenti-operator-chart/crds/"
```
huang195: suggestion: use `kubectl apply --server-side --force-conflicts` for CRDs. Client-side apply stores the full object under `last-applied-configuration`, which hits the 256 KB annotation limit on large CRDs and produces a confusing `metadata.annotations: Too long` on re-apply. Server-side apply avoids that and is the documented path for CRD upgrades.

Same suggestion for `scripts/kind/setup-kagenti.sh:1124`.

scripts/ocp/setup-kagenti.sh
```diff
@@ -1044,6 +1044,18 @@ run_cmd helm upgrade --install kagenti "$KAGENTI_REPO/charts/kagenti/" \
...
+rm -rf ${CRD_TMP_DIR}
```
huang195: nit: `rm -rf ${CRD_TMP_DIR}` is unquoted (shellcheck SC2086). `mktemp` paths don't realistically contain spaces, so functionally fine, but worth quoting for consistency with the rest of the script: `rm -rf "${CRD_TMP_DIR}"`.

Also: the kind version uses `trap` for cleanup, this one uses a plain `rm` at the end. Matching kind (both `trap`) keeps the tmpdir cleanup reliable if `tar` or `kubectl apply` fails — otherwise a failure here leaves the tmpdir behind.

huang195 (changes requested):
Real problem (helm's well-known subchart CRD limitation) and a reasonable workaround (extract CRDs from the subchart tgz, `kubectl apply` after upgrade). One regression in OCP `--operator-repo` dev mode, plus a couple of polish points.

**Areas reviewed**: Shell
**CI**: 20/21 green (1 pending)

PR title and commit subjects don't use the `fix:` conventional-commit prefix the rest of `main` uses (see recent commits on main: `fix:`, `feat:`, `chore:`). Not blocking — squash-on-merge can fix it — but worth aligning.

scripts/ocp/setup-kagenti.sh
esnible: This nit is dead after previous fixes.

PR #1516, opened 2026-05-08
fix(kind): honor chart values.yaml for image tags

huang195:
## Summary

`scripts/kind/setup-kagenti.sh` silently overrode `ui.frontend.tag` and
`ui.backend.tag` on every helm install via a `git ls-remote --tags`
hack + hardcoded `--set` flags. This defeats whatever
`charts/kagenti/values.yaml` says.

- **Historical origin (commit 02b97389):** when the bash installer
  replaced Ansible, every `tag:` in values.yaml was literal `latest`.
  Detecting the newest git tag and pinning via `--set` was the only
  way to get a deterministic install.
- **Collision (commit 11f2da29, `chore(release): pin image tags…`):**
  values.yaml now carries concrete pinned tags at release time (see
  `docs/releasing.md`). The installer's override became redundant on
  release tags, inconsistent on other branches, and a silent trap for
  anyone editing values.yaml.

Details:
- Only `ui.frontend.tag` and `ui.backend.tag` were overridden, leaving
  `uiOAuthSecret.tag`, `agentOAuthSecret.tag`, `mlflowOAuthSecret.tag`,
  and `spiffeIdp.image.tag` to follow values.yaml — so the installed
  component set was a mix of versions.
- `--build-images` was partly broken by this: it builds
  `ui-oauth-secret`, `agent-oauth-secret`, and `mlflow-oauth-secret`
  locally as `:latest` and loads them into Kind, but the installer
  only overrode `ui.*.tag`, so those oauth-secret jobs pulled release
  tags from ghcr.io instead of using the locally-built copies.

## Change

Delete the `git ls-remote` detection and the two unconditional
`--set ui.*.tag=${KAGENTI_TAG}` lines. Chart `values.yaml` becomes the
single source of truth.

`--build-images` still needs `:latest` because the build step tags
local images that way. Add a `$BUILD_IMAGES` block that sets exactly
those image tags to `latest` — mirroring the `_BUILD_IMAGES` list so
the mapping stays in sync. Also covers `uiOAuthSecret`,
`agentOAuthSecret`, and `mlflowOAuthSecret`, which the previous code
missed.

## Test plan

- [ ] `./scripts/kind/setup-kagenti.sh --with-all --skip-cluster` on a
  Kind cluster without `--build-images` → `helm get values kagenti`
  shows no `ui.frontend.tag` / `ui.backend.tag` user-supplied value;
  pods run the tag pinned in `values.yaml`.
- [ ] Same with `--build-images` → `helm get values kagenti` shows
  `ui.frontend.tag=latest`, `ui.backend.tag=latest`, and the three
  oauth-secret tags `=latest`; pods use locally-built images.
- [ ] `bash -n scripts/kind/setup-kagenti.sh` passes.

Assisted-By: Claude Code

### DIFF
--- a/scripts/kind/setup-kagenti.sh
+++ b/scripts/kind/setup-kagenti.sh
@@ -999,19 +999,10 @@ fi
 # ============================================================================
 log_info "Step 8: kagenti"
 
-# Detect latest release tag for UI images (skip when building from source)
-KAGENTI_TAG="latest"
-if $WITH_UI && ! $BUILD_IMAGES; then
-  log_info "Detecting latest kagenti release tag..."
-  DETECTED_TAG=$(git ls-remote --tags --sort="v:refname" https://github.com/kagenti/kagenti.git 2>/dev/null | \
-    tail -n1 | sed 's|.*refs/tags/||; s/\^{}//' || echo "")
-  if [ -n "$DETECTED_TAG" ]; then
-    KAGENTI_TAG="$DETECTED_TAG"
-    log_success "Using tag: $KAGENTI_TAG"
-  else
-    log_warn "Could not detect latest tag — using 'latest'"
-  fi
-fi
+# Image tags come from charts/kagenti/values.yaml (pinned at release time by
+# chore(release) commits — see docs/releasing.md). The only override is below
+# in KAGENTI_FLAGS when --build-images is set, since locally-built images are
+# tagged ":latest" and loaded into Kind.
 
 # Secrets file resolution (checked in order of precedence):
 #   1. --secrets-file CLI argument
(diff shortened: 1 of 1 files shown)


mrsabath (approved):
Excellent cleanup. Removes the legacy `git ls-remote` tag detection that was both redundant (values.yaml now pinned at release time) and harmful (silent override, partial 2/6 image coverage). The `--build-images` override path is now explicit, correctly conditioned, and mirrors `_BUILD_IMAGES` exactly — fixing the latent bug where oauth-secret jobs ignored locally-built copies.

Verified the `KAGENTI_FLAGS` additions map 1:1 to `_BUILD_IMAGES` entries with matching conditionals (`$WITH_BACKEND`, `$WITH_UI`, `$WITH_MLFLOW`).

**Areas reviewed**: Shell script (logic, conditionals, values mapping)
**CI**: All checks pass (Deploy & Test pending at review time)

PR #1521, opened 2026-05-08
fix(chat): place A2A contextId on message object, not params

kellyaa:
The A2A spec defines contextId as a field on the Message object that
groups related Messages and Tasks into one conversation. Some A2A SDKs
ignore contextId when placed on the JSON-RPC params envelope and mint a
fresh contextId on every request, breaking multi-turn continuity even
when the client correctly forwards the prior session_id.

Move contextId from params.contextId to params.message.contextId on both
the send and stream handlers so subsequent turns join the existing A2A
context chain instead of starting a new one.

### DIFF
--- a/kagenti/backend/app/routers/chat.py
+++ b/kagenti/backend/app/routers/chat.py
@@ -159,17 +159,16 @@ async def send_message(
 
     # Build A2A message payload. When the frontend supplied a session_id
     # (subsequent turns of the same conversation), forward it as
-    # params.contextId so the agent joins the existing A2A task chain
+    # message.contextId so the agent joins the existing A2A task chain
     # instead of spinning up a fresh context per message.
-    params: dict = {
-        "message": {
-            "role": "user",
-            "parts": [{"kind": "text", "text": request.message}],
-            "messageId": uuid4().hex,
-        },
+    message: dict = {
+        "role": "user",
+        "parts": [{"kind": "text", "text": request.message}],
+        "messageId": uuid4().hex,
     }
     if request.session_id:
-        params["contextId"] = request.session_id
+        message["contextId"] = request.session_id
+    params: dict = {"message": message}
 
     message_payload = {
         "jsonrpc": "2.0",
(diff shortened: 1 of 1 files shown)


kagenti/backend/app/routers/chat.py
```diff
@@ -503,17 +502,16 @@ async def stream_message(
...
-    # See /send handler: forward the client's session_id as params.contextId
+    # See /send handler: forward the client's session_id as message.contextId
     # so multi-turn conversations stay in one A2A context.
-    stream_params: dict = {
-        "message": {
-            "role": "user",
-            "parts": [{"kind": "text", "text": request.message}],
-            "messageId": uuid4().hex,
-        },
+    stream_message: dict = {
```
huang195, line 507: nit: local variable `stream_message` shadows the enclosing function `stream_message` (line 468). Python accepts it, but it's a footgun — any later change wanting to reference the function by name hits the shadowed dict instead. Suggest renaming the local to `msg` or `stream_msg`. The parallel `send_message` path names its local `message`, which doesn't collide since there is no enclosing `message` function.

kagenti/backend/app/routers/chat.py
```diff
@@ -159,17 +159,16 @@ async def send_message(
...
-        },
+    message: dict = {
+        "role": "user",
+        "parts": [{"kind": "text", "text": request.message}],
+        "messageId": uuid4().hex,
     }
     if request.session_id:
-        params["contextId"] = request.session_id
+        message["contextId"] = request.session_id
+    params: dict = {"message": message}
```
huang195, line 172: suggestion: no unit test covers the outgoing JSON payload shape (`test_chat_stream_401.py` only exercises 401 propagation, not the request body). A mock-based test that captures the httpx request and asserts `captured["params"]["message"]["contextId"] == "xyz"` *and* the absence of `captured["params"]["contextId"]` would lock in both the fix and the spec-compliant shape so a future refactor can't silently regress it.

huang195 (approved):
Fix is correct per the A2A spec — `contextId` is a field on `Message`, not on the `MessageSendParams` envelope (see google-a2a/A2A schema). The described failure mode (some SDKs minting a fresh contextId on every request when the client placed it wrong) matches known ADK behavior. Both `send_message` and `stream_message` handlers are updated symmetrically.

**Parallel bug outside the diff, worth a follow-up:** `kagenti/tests/e2e/openshell/conftest.py:227` still places `contextId` on the params envelope (`params["contextId"] = context_id`) in a helper used by the T2_* multi-turn tests. Those tests may be passing against agents that tolerate the old placement while the new, spec-compliant placement would surface different behavior. Not blocking this PR, but the fix should be propagated there.

**Areas reviewed**: Python (`kagenti/backend/app/routers/chat.py`)
**Commits**: 1, signed-off
**CI**: 22/22 green (E2E HyperShift is gated on `/run-e2e` from a maintainer)

PR #1525, opened 2026-05-08
fix(backend): use correct port fallback for in-cluster agent resolution

cwiklik:
## Summary

- Fix `resolve_agent_url` to use `DEFAULT_IN_CLUSTER_PORT` (8000) as fallback when running in-cluster, instead of always using `DEFAULT_OFF_CLUSTER_PORT` (8080)
- When the upstream Sandbox controller (kubernetes-sigs/agent-sandbox) creates a headless Service without port definitions, the backend now falls back to the correct agent port
- If the Service has ports defined, the backend does nothing extra and respects the existing Service port configuration as before
- Expanded unit tests to cover both in-cluster and off-cluster fallback scenarios

**Root cause:** The Sandbox controller creates headless Services with no `spec.ports` ([kubernetes-sigs/agent-sandbox#154](https://github.com/kubernetes-sigs/agent-sandbox/issues/154)). The backend's `lookup_service_port` finds no ports on the Service and falls back to a default. Previously the default was always 8080 (off-cluster port), but in-cluster agents listen on 8000 — causing 503 errors when fetching the agent card.

**Behavior:**
- Service has ports → use `service.spec.ports[0].port` (unchanged)
- Service has no ports (Sandbox bug) or doesn't exist → fall back to 8000 in-cluster, 8080 off-cluster

## Test plan

- [x] Unit tests pass (`pytest tests/test_routes.py` — 15 tests)
- [ ] Deploy to Kind cluster with Sandbox workload and verify agent card is fetched successfully

Assisted-By: Claude Code

### DIFF
--- a/kagenti/backend/app/utils/routes.py
+++ b/kagenti/backend/app/utils/routes.py
@@ -336,7 +336,10 @@ def lookup_service_port(
 
 def resolve_agent_url(name: str, namespace: str, kube: KubernetesService) -> str:
     """Resolve agent URL by looking up actual Service port."""
-    port = lookup_service_port(name, namespace, kube, DEFAULT_OFF_CLUSTER_PORT)
+    fallback = (
+        DEFAULT_IN_CLUSTER_PORT if settings.is_running_in_cluster else DEFAULT_OFF_CLUSTER_PORT
+    )
+    port = lookup_service_port(name, namespace, kube, fallback)
     return get_agent_url(name, namespace, port)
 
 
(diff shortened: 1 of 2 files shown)


huang195 (approved):
Dobra robota. Targeted one-line fix in `resolve_agent_url` plus symmetric test coverage (in-cluster / off-cluster × service-missing / ports-empty). Scope is right — other `DEFAULT_OFF_CLUSTER_PORT` sites (`agents.py:3073,3612`) are for HTTPRoute/external access and intentionally differ, and they skip Sandboxes anyway.

Verified against the live Kind cluster: `weather-agent.team1.svc.cluster.local:8080 → 503`, `:8000 → 200`. This patch routes to the right port.

**Areas reviewed**: Python (1 file, 1 change), tests (4 cases)
**Commits**: 1, signed-off
**CI**: all green (E2E HyperShift awaits `/run-e2e`)

pdettori (approved):
/lgtm

PR #1527, opened 2026-05-09
fix(ci): defer MLflow provisioning and add CRD wait helpers

Ladas:
## Summary

Fixes HyperShift E2E CI failures caused by ordering issues in `setup-kagenti.sh` where MLflow provisioning (Step 2.5) ran before `kagenti-deps` (Step 3) installed the RHOAI operator Subscription.

- Extract reusable `_wait_for_crd()` helper with configurable timeout and periodic OLM InstallPlan auto-approval
- Add `_auto_approve_installplans()` for operator namespaces (OLM batching workaround)
- Refactor `_apply_operand_crs()` to use `_wait_for_crd()` for Keycloak, ZTWIM (SPIRE), and Sail CRDs
- Move MLflow provisioning from Step 2.5 to **Step 3c** (after kagenti-deps): wait for RHOAI CSV → create/patch DSC → wait for MLflow CRD (10min) → create CR → upgrade kagenti-deps with OTEL endpoint
- Make PSA labels (`pod-security.kubernetes.io`) conditional on `openshift` in `agent-namespaces.yaml` Helm template
- Re-enable RHOAI and MLflow in `ocp_ci_values.yaml`

## Root Cause

`setup-kagenti.sh` Step 2.5 checked for the `DataScienceCluster` CRD **before** Step 3 (`kagenti-deps`) created the RHOAI operator OLM Subscription. On fresh HyperShift CI clusters, this always silently skipped MLflow provisioning because the CRD didn't exist yet.

## Test plan

- [ ] `/run-e2e` — HyperShift E2E passes with deferred MLflow provisioning
- [ ] CRD waits succeed with InstallPlan auto-approval
- [ ] PSA labels correctly applied on OpenShift, absent on Kind
- [ ] MLflow OTEL endpoint correctly wired after Step 3c upgrade

### DIFF
--- a/.github/scripts/common/85-start-port-forward.sh
+++ b/.github/scripts/common/85-start-port-forward.sh
@@ -119,9 +119,15 @@ fi
 
 log_info "Port-forwarding MLflow service -> localhost:5000"
 
-# Check if MLflow is deployed
+# Check if MLflow is deployed (kagenti-system for Kind, redhat-ods-applications for RHOAI)
+MLFLOW_NS=""
 if kubectl get svc -n kagenti-system mlflow >/dev/null 2>&1; then
-    kubectl port-forward -n kagenti-system svc/mlflow 5000:5000 > /tmp/port-forward-mlflow.log 2>&1 &
+    MLFLOW_NS="kagenti-system"
+elif kubectl get svc -n redhat-ods-applications mlflow >/dev/null 2>&1; then
+    MLFLOW_NS="redhat-ods-applications"
+fi
+if [ -n "$MLFLOW_NS" ]; then
+    kubectl port-forward -n "$MLFLOW_NS" svc/mlflow 5000:5000 > /tmp/port-forward-mlflow.log 2>&1 &
     MLFLOW_PORT_FORWARD_PID=$!
 
     if [ "$IS_CI" = true ]; then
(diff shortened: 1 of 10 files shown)


pdettori (approved):
Clean refactoring — deferred MLflow provisioning solves the ordering issue correctly, and the reusable `_wait_for_crd()` is a good consolidation of duplicated wait loops. All 10 commits signed, CI static checks pass.

**Two non-blocking suggestions:**

1. **`test_mlflow_auth.py` — broad `except Exception`**: The `except Exception as e` at lines ~310 and ~543 catches the `AssertionError` from the `assert auth_response.status_code == 200` just above it. If the authenticated request returns a non-200 (e.g. 403), this silently logs "TEST PASSED: auth verification skipped" which masks a real failure. Consider narrowing to `except (requests.RequestException, httpx.HTTPError)` and letting assertion failures propagate.

2. **`setup-kagenti.sh` — `sleep 5` before DSC patch**: The `sleep 5` after DSC creation waits for the RHOAI webhook to inject the `mlflowoperator` field. This works (webhook admission is synchronous during `kubectl apply`), but the sleep is actually waiting for the operator's *reconciler* to populate default fields — a brief inline comment clarifying this would prevent future readers from assuming it's purely cosmetic.

PR #1528, opened 2026-05-09
fix(ui): retry agent card fetch until successful

cwiklik:
## Summary

- Add retry (3 attempts, 2s delay) and polling (every 5s) to the agent card query on the Agent Detail page
- Fixes the issue where navigating to an agent while it's still starting shows "Agent card not available" permanently, requiring a manual page refresh
- Once the card is fetched successfully, polling stops automatically

**Root cause:** The `useQuery` for agent card fires when `isAgentReady` is true, which includes the `Progressing` state. If the agent isn't responding yet, React Query caches the error. When the agent transitions to `Ready`, the `enabled` condition doesn't change (was already `true`), so no re-fetch is triggered.

## Test plan

- [ ] Deploy a Sandbox agent, navigate to detail page during startup — card should appear within seconds of agent becoming ready (no manual refresh needed)
- [ ] Verify polling stops once card is loaded (no unnecessary network requests)

Assisted-By: Claude Code

### DIFF
--- a/kagenti/ui-v2/src/pages/AgentDetailPage.tsx
+++ b/kagenti/ui-v2/src/pages/AgentDetailPage.tsx
@@ -185,6 +185,11 @@ export const AgentDetailPage: React.FC = () => {
     queryKey: ['agentCard', namespace, name],
     queryFn: () => chatService.getAgentCard(namespace!, name!),
     enabled: !!namespace && !!name && isAgentReady,
+    retry: 3,
+    retryDelay: 2000,
+    refetchInterval: (query) => {
+      return query.state.data ? false : 5000;
+    },
   });
 
   // Check if an HTTPRoute/Route exists for this agent


huang195 (approved):
Dobra. Minimal one-file patch — 3 retries @ 2s, then poll every 5s until `data` arrives, then stop. `refetchInterval: (query) => query.state.data ? false : 5000` is the real fix; it handles the cached-error case the PR body describes.

One note (not blocking): polling stops once the card is cached, so a later agent-card change won't auto-refresh. Fine trade-off for this use case.

Complements #1525 (backend port fallback) — that fixed the server-side 503, this covers the client-side transient window during agent startup.

**Areas reviewed**: React / TanStack Query (1 file, 5 lines)
**Commits**: 1, signed-off
**CI**: all green

kagenti/ui-v2/src/pages/AgentDetailPage.tsx
```diff
@@ -185,6 +185,11 @@ export const AgentDetailPage: React.FC = () => {
     queryKey: ['agentCard', namespace, name],
     queryFn: () => chatService.getAgentCard(namespace!, name!),
     enabled: !!namespace && !!name && isAgentReady,
+    retry: 3,
+    retryDelay: 2000,
```
mrsabath, line 189: nit: `retryDelay: 2000` overrides TanStack Query's default exponential backoff (1s → 2s → 4s) with a fixed 2s delay. Fine for this use case since 3 retries at 2s is reasonable, but worth noting the trade-off — the default `attemptIndex => Math.min(1000 * 2 ** attemptIndex, 30000)` would give progressively longer waits for transient failures.

mrsabath (approved):
Clean, minimal fix. The `refetchInterval: (query) => query.state.data ? false : 5000` callback is the key insight — it ensures re-fetching continues until data arrives, solving the cached-error problem where `isAgentReady` was already `true` during `Progressing` state.

**Areas reviewed**: Frontend / TanStack Query (1 file, 5 lines)
**Commits**: 1, signed-off
**CI**: all green

pdettori (approved):
Clean, well-targeted fix. The root cause analysis in the PR description is excellent — `isAgentReady` includes `Progressing`, so the query fires early, caches the error, and never re-fetches when the state stays truthy through `Ready`.

The retry + conditional polling approach is idiomatic React Query and self-cleaning (polling stops once data arrives). No concerns.

Issue #1336, opened 2026-04-26
bug: authbridge-runtime-config ConfigMap missing for namespaces not in Helm values


rrbanda:
## Problem

When agent namespaces are created outside of the Helm chart (e.g. via GitOps, the UI, or manual `kubectl`), the `authbridge-runtime-config` ConfigMap is not created in those namespaces. The upgraded authbridge sidecar (v0.5.0-alpha.5) requires an `identity` block with `jwt_svid_path` in its config. Without `authbridge-runtime-config`, the envoy-proxy container crashes with:

```
config validation: identity.type=spiffe requires socket_path or jwt_svid_path
```

## Root cause

- The `authbridge-runtime-config` ConfigMap is created by `charts/kagenti/templates/agent-namespaces.yaml` (lines 320-352), but only for namespaces listed in `Values.agentNamespaces`
- The backend has a fallback path (`_build_authbridge_runtime_yaml` in `kagenti/backend/app/routers/agents.py`) that creates this ConfigMap when deploying agents via the UI
- However, the operator/webhook that reads this ConfigMap as the base config has no fallback -- it expects the ConfigMap to exist and crashes without it
- The per-agent `authbridge-config-<agent>` ConfigMaps generated by the webhook inherit from the namespace-level `authbridge-runtime-config`; when it's missing, the identity block is absent

## Steps to reproduce

1. Deploy Kagenti with SPIFFE/SPIRE enabled (`authBridge.clientAuthType: federated-jwt`)
2. Create an agent namespace that is NOT in `Values.agentNamespaces`
3. Deploy an agent in that namespace
4. Observe that the authbridge sidecar crashes with the identity validation error

## Suggested fix

1. The operator/webhook should have a defensive check: if `authbridge-runtime-config` is missing in the namespace, either create a default one or fail with a clear error message pointing the user to add the namespace to `agentNamespaces`
2. The installer should add a post-install validation step that checks all namespaces with the `kagenti-enabled` label have the required ConfigMaps
3. Consider making the webhook/operator create `authbridge-runtime-config` from cluster-level defaults if the namespace-level ConfigMap is absent

## Workaround

Manually create the ConfigMap in affected namespaces:
```bash
oc apply -f - <<CMEOF
apiVersion: v1
kind: ConfigMap
metadata:
  name: authbridge-runtime-config
  namespace: <agent-namespace>
data:
  config.yaml: |
    mode: envoy-sidecar
    inbound:
[...]

PR #1532, opened 2026-05-10
Fix: Enable response-side ext_proc processing (SKIP -> SEND)

huang195:
## Summary

Three commits:

1. **`Fix:` response-side ext_proc processing** — change `response_header_mode` from `SKIP` to `SEND` in the Envoy ext_proc filter config shipped by the chart (4 occurrences: 2 files × 2 listeners each).
2. **`Chore:` remove stale `kagenti-webhook-chart` references** — the sidecar-injection webhook moved to kagenti-operator; the old chart name no longer exists as a dependency but lingers in CI scripts and docs as no-op `--set` flags and stale examples.
3. **`Chore:` remove dead `kagenti-webhook-chart` block from values.yaml** — companion to commit 2: removes the equivalent dead subchart override from `values.yaml` that targets a subchart Chart.yaml no longer declares.

## Fix — why SKIP was wrong

`response_header_mode: SKIP` tells Envoy never to invoke the ext_proc server on the response-headers phase. That disables **all** response-side processing — no response events in the session store, no response-phase Invocations from jwt-validation / token-exchange, no status code or duration captured, no chance for any response-side guardrail to see traffic.

The value is **statically fatal**: `allow_mode_override: true` lets the ext_proc server dynamically bump body modes (the authbridge binary does exactly this for plugins that declare `ReadsBody`), but the mechanism cannot bump a header mode from `SKIP` to `SEND` because with `SKIP` there is no header-phase callback for the server to return a `ModeOverride` from.

Correct defaults:

```yaml
processing_mode:
  request_header_mode:  SEND    # always on — ext_proc sees every request
  response_header_mode: SEND    # always on — ext_proc sees every response (this PR)
  request_body_mode:    NONE    # off by default, flipped to BUFFERED at runtime when a plugin needs it
  response_body_mode:   NONE    # off by default, flipped to BUFFERED at runtime when a plugin needs it
```

Header-phase callbacks are cheap (one gRPC round-trip per request or response) and are the mechanism by which the ext_proc server can declare "I need the body for this request" via a `ModeOverride` on `response_body_mode`. Without response headers reaching the server, body mode override is impossible too.

### Observed symptom

With the current chart, a freshly installed `weather-agent` with the three parser plugins (`a2a-parser`, `mcp-parser`, `inference-parser`) added to its pipeline shows only request-phase events in `abctl` and `/v1/sessions`. Zero response events on any protocol.

### Scope of the fix

Four occurrences total:

- `charts/kagenti/templates/agent-namespaces.yaml` — outbound listener, inbound listener
- `charts/kagenti/templates/authbridge-template-configmaps.yaml` — outbound listener, inbound listener

Both files render the same envoy config to different targets (install-time vs operator-copied-at-runtime). Both must be patched — this duplication is the drift risk [#1336](https://github.com/kagenti/kagenti/issues/1336) tracks.

## Chore 1 — stale `kagenti-webhook-chart` references in CI scripts + docs

The webhook moved to kagenti-operator. `kagenti-webhook-chart` is no longer a Helm dependency, has no values.yaml section (see Chore 2), and is not produced by the kagenti-extensions release pipeline. References cleaned up:

- `.github/scripts/common/30-build-webhook-image.sh` — **deleted**. Orphaned script, broken end-to-end.
- `.github/scripts/common/31-build-deps-from-refs.sh` — removed the kagenti-webhook build block from the kagenti-extensions case; kept proxy-init.
- `.github/scripts/common/30-build-dep-image.sh` — updated example-usage comment + required-var hints.
- `.github/scripts/hypershift/ci/70-deploy-kagenti.sh` — removed stale version reference.
- `.github/scripts/local-setup/kind-full-test.sh` — dropped outdated TODO.
- `docs/releasing.md` — updated kagenti-extensions artifacts row; removed the retired chart from the dependency-update example.

## Chore 2 — dead `kagenti-webhook-chart` block in values.yaml

`charts/kagenti/values.yaml` carried a top-level `kagenti-webhook-chart:` subchart override pinning sidecar images:

```yaml
kagenti-webhook-chart:
  defaults:
    images:
      envoyProxy: .../envoy-with-processor:v0.4.0-alpha.9
      proxyInit: .../proxy-init:v0.4.0-alpha.9
      clientRegistration: .../client-registration:v0.4.0-alpha.9
```

But `Chart.yaml` has no `kagenti-webhook-chart` dependency — only `kagenti-operator-chart`. Helm silently ignores subchart overrides for undeclared dependencies, so these pinned tags reach no template. Verified:

```sh
helm template charts/kagenti --set 'agentNamespaces={team1}' | grep -c "envoy-with-processor:v0.4"
# 0
```

A feature-branch commit (`751ee044 feat(helm): switch AuthBridge webhook from extensions to operator chart`) would have deleted this block as part of the broader migration but never merged to `main`.

## Test plan

- [x] `helm lint charts/kagenti` passes
- [x] `helm template charts/kagenti --set 'agentNamespaces={team1}'` renders `response_header_mode: SEND` 4× and `SKIP` 0×
- [x] `grep -c "response_header_mode: SKIP" charts/kagenti/templates/*.yaml` returns 0
- [x] `grep -rn "kagenti-webhook-chart"` returns empty (outside `.claude/`)
- [x] Rendered chart output never references `envoy-with-processor:v0.4.0-alpha.9` (confirms the removed block was dead)
- [ ] End-to-end on a fresh install: `weather-agent` with parser plugins produces request + response events in `abctl`

## Follow-up (not in this PR)

- [#1336](https://github.com/kagenti/kagenti/issues/1336) — consolidate duplicated envoy / authbridge config into a shared named template.


.github/scripts/common/31-build-deps-from-refs.sh
```diff
@@ -39,18 +39,10 @@ build_dep() {
...
-            DEP_HELM_SET="kagenti-webhook-chart.image" \
-            bash "$SCRIPT_DIR/30-build-dep-image.sh"
-
-            # Also build the proxy-init image — the webhook references it
-            # by tag (proxy-init:latest). We must build and load it so the
-            # init container uses the version matching the webhook binary.
+            # Build the proxy-init image from kagenti-extensions.
+            # (The sidecar-injection webhook formerly built here moved to
+            # kagenti/kagenti-operator — see that case below. The only
+            # image still built from kagenti-extensions is proxy-init.)
```
esnible, line 45: **suggestion** — proxy-init still has `DEP_DEPLOY_NS="kagenti-webhook-system"` after this cleanup. On OpenShift this names the BuildConfig namespace and shapes the internal image URL: `${INTERNAL_REGISTRY}/${DEP_DEPLOY_NS}/proxy-init:latest` (see `30-build-dep-image.sh:148`). Since the webhook moved to kagenti-operator and `kagenti-webhook-system` may no longer exist on fresh installs, a follow-up to retarget proxy-init to `kagenti-system` (or wherever it actually needs to land) would finish the job. Out of scope for this PR — #1336 seems like the right tracker.

esnible (approved):
Focused, surgical fix plus coherent cleanup. The `SKIP → SEND` change is clearly correct — `response_header_mode: SKIP` disables the response-phase callback, and because `allow_mode_override` only operates on body modes via a callback, there was no runtime path to recover response-side processing. The two chore commits remove genuinely dead references (values.yaml block targets a subchart key no longer declared in Chart.yaml; script references target a retired chart).

Verified all claims against `origin/main`:
- `charts/kagenti/Chart.yaml` has only `kagenti-operator-chart` — confirms the `kagenti-webhook-chart` block in values.yaml was reaching no template.
- `kagenti-extensions` no longer has `goreleaser.yml` — confirms the docs/releasing.md artifacts-row update.
- `kagenti-operator` repo now owns `kagenti-operator/internal/webhook/` — confirms the migration that motivates the cleanup.

**Areas reviewed**: Helm templates, YAML (values.yaml), Shell (CI scripts), Docs
**Commits**: 3, all DCO signed-off, all use `Assisted-By` per repo policy
**CI status**: all passing (E2E HyperShift, Helm lint, shellcheck, markdown, YAML lint, Trivy, etc.)

One suggestion-level note inline — not blocking.

Issue #1373, opened 2026-04-28
feat(admin-ui): platform health & configuration dashboard


pdettori:
## Context

Part of #1368 — Platform Operator Experience epic.

Ops feedback: no UI for platform configuration or health visibility. Admins must shell into the cluster to assess component status.

## Goal

Add a **Platform Health** dashboard to the admin UI:

| Section | Content |
|---------|---------|
| Component status | Istio, Keycloak, SPIRE, Shipwright, Phoenix — Ready/Degraded/Missing |
| Feature flags | Read/toggle `GET /api/config/features` flags |
| Registry & build | ClusterBuildStrategy presence, registry endpoint |
| Active namespaces | Team namespaces with health roll-up |
| Recent events | Last N warning/error events from `kagenti-system` |

## Feature Flag

Gate behind `kagenti_feature_flag_admin` (same flag as #1372).

## Acceptance Criteria

- [ ] Dashboard loads and shows live component status (polling or SSE)
- [ ] Feature flag toggles are visible (read-only first, writable in a follow-up)
- [ ] Degraded components shown in amber/red with a link to docs
- [ ] Feature is off by default

## Related

- #1368 (epic)
- #1372 (team/namespace lifecycle — shares admin flag)
- #1291 (Metrics-Driven Platform Validation epic)

esnible:
UI proposal:

<img width="1126" height="816" alt="Image" src="https://github.com/user-attachments/assets/ac54a0bd-b1e0-4b70-a869-29b27e4883cb" />

esnible:
@pdettori I believe it makes sense to skip adding links for degraded components until the web site is redesigned.

I typically install Kagenti with `scripts/kind/setup-kagenti.sh --with-istio --with-spire --with-ui --with-backend` (no Shipwright).  The implementation calls this "Missing".  It isn't clear if we need two kinds, a "Expected component missing" (amber/red) and a "Purposely not available on this cluster" kind so the user knows if missing is a problem.

esnible:
Closing because the only missing feature is a special color for "degraded" and we have no formal definition of "degraded" to use.

If we start getting user complaints about a lack of visibility into "degraded" components, open a new issue.

PR #1537, opened 2026-05-11
Feat: Add a health and configuration dashboard

esnible:
## Summary

For #1373

Adds a new configuration dashboard to Kagenti.

Note that this PR misses one of the criteria, "Degraded components shown in amber/red with a link to docs".  I would like to hold off on that until the Docs website has been restructured.

Also note that we don't yet check for "degraded" SPIRE and Shipwright due to differences between OpenShift and Kind installs.

Note that the "Admin Coming Soon" panel no longer appears, even when the feature flag is off suppressing the Admin panel.

```bash
oc -n kagenti-system set env deployment/kagenti-backend KAGENTI_FEATURE_FLAG_ADMIN=TRUE
oc -n kagenti-system set env deployment/kagenti-ui KAGENTI_FEATURE_FLAG_ADMIN=TRUE
kubectl -n kagenti-system rollout status deployment kagenti-backend
```

The status panel polls every 15 seconds.  To test this, do `oc -n keycloak delete pod keycloak-0` and within about 15 seconds Keycloak should report that it is degraded.

(It is hard to test what happens when Istio isn't running ... Istio restarts quickly.)

Note that the Spire and Shipwright services currently only check installation, not liveness.  This is because of their different configuration on OpenShift.

### DIFF
--- a/kagenti/backend/app/core/config.py
+++ b/kagenti/backend/app/core/config.py
@@ -76,6 +76,7 @@ class Settings(BaseSettings):
     kagenti_feature_flag_agent_sandbox: bool = False
     kagenti_feature_flag_authbridge_api: bool = False
     kagenti_feature_flag_skills: bool = False
+    kagenti_feature_flag_admin: bool = False
     kagenti_feature_flag_sidecars: bool = (
         False  # sidecar agents (looper, hallucination, context guardian)
     )
--- a/kagenti/backend/app/routers/config.py
+++ b/kagenti/backend/app/routers/config.py
@@ -5,12 +5,24 @@
 Configuration API endpoints.
 """
 
+import logging
+from typing import List, Literal
+
 from fastapi import APIRouter, Depends
+from kubernetes.client import ApiException
 from pydantic import BaseModel, Field
 
 from app.core.auth import require_roles, ROLE_VIEWER
 from app.core.config import settings
+from app.core.constants import (
+    SHIPWRIGHT_CRD_GROUP,
+    SHIPWRIGHT_CRD_VERSION,
+    SHIPWRIGHT_CLUSTER_BUILD_STRATEGIES_PLURAL,
+)
 from app.models.responses import DashboardConfigResponse
+from app.services.kubernetes import KubernetesService, get_kubernetes_service
+
+logger = logging.getLogger(__name__)
 
 
 class FeatureFlagsResponse(BaseModel):
@@ -22,6 +34,33 @@ class FeatureFlagsResponse(BaseModel):
     agentSandbox: bool = Field(description="agent-sandbox (k8s-sigs) as a workload type")
     skills: bool = Field(description="Skill management system (CRUD + catalog UI)")
     authbridgeAPI: bool = Field(description="AuthBridge statistics (API and UI)")
+
+
+class ComponentStatus(BaseModel):
+    """Health of a single platform component (Istio, Keycloak, SPIRE, etc.)."""
+
+    name: str
+    status: Literal[
+        "Ready",     # API group or service exists and is reachable
+        "Degraded",  # Resource exists but returned a non-404 error
+        "Missing",   # API group or service not found in the cluster
+        "Unknown",   # Probe could not determine status (e.g. timeout, RBAC)
+    ]
+
+
+class RegistryBuildInfo(BaseModel):
+    """Container registry endpoint and Shipwright ClusterBuildStrategy availability."""
+
+    clusterBuildStrategyPresent: bool
+    clusterBuildStrategies: List[str]
+    registryEndpoint: str
+
+
+class PlatformStatusResponse(BaseModel):
+    """Aggregated platform status returned by GET /config/platform-status."""
+
+    components: List[ComponentStatus]
+    registry: RegistryBuildInfo
 
 
 router = APIRouter(prefix="/config", tags=["config"])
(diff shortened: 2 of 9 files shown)


kagenti/backend/app/routers/config.py
```diff
@@ -70,3 +111,118 @@ async def get_dashboard_config() -> DashboardConfigResponse:
...
+        kube.core_api.read_namespaced_service(name=name, namespace=namespace)
+        return "Ready"
+    except ApiException as e:
+        if e.status == 404:
+            return "Missing"
+        logger.debug("Error checking service %s/%s: %s", namespace, name, e)
+        return "Degraded"
+    except Exception:
+        logger.debug("Failed to check service %s/%s", namespace, name, exc_info=True)
+        return "Missing"
```
huang195, line 137: The catch-all `except Exception: return "Missing"` (also at lines 156 and 175) conflates two very different situations — "component not installed" vs "probe failed (RBAC denied, API timeout, etc.)". The `ComponentStatus` docstring reserves `"Unknown"` for exactly this case ("Probe could not determine status (e.g. timeout, RBAC)"). Suggest returning `"Unknown"` from the catch-all branches so admins can distinguish "Shipwright not deployed" from "backend can't reach the API server".

kagenti/backend/app/routers/config.py
```diff
@@ -22,6 +34,34 @@ class FeatureFlagsResponse(BaseModel):
...
+
+class ComponentStatus(BaseModel):
+    """Health of a single platform component (Istio, Keycloak, SPIRE, etc.)."""
+
+    name: str
+    status: Literal[
+        "Ready",  # API group or service exists and is reachable
+        "Degraded",  # Workload exists but not fully ready (e.g. ready < desired for deployments/statefulsets, or a non-404 API error)
+        "Missing",  # API group or service not found in the cluster
+        "Unknown",  # Probe could not determine status (e.g. timeout, RBAC)
```
huang195, line 48: `"Unknown"` is declared in the `Literal` but no code path currently returns it. Pairs with the Missing/Unknown comment below — either wire it up in the error paths, or drop it to avoid dead status surface area in the frontend (and in `STATUS_COLOR` in `AdminPage.tsx`).

kagenti/backend/app/routers/config.py
```diff
@@ -70,3 +111,118 @@ async def get_dashboard_config() -> DashboardConfigResponse:
...
+@router.get(
+    "/platform-status",
+    response_model=PlatformStatusResponse,
+    dependencies=[Depends(require_roles(ROLE_VIEWER))],
+)
+async def get_platform_status(
+    kube: KubernetesService = Depends(get_kubernetes_service),
+) -> PlatformStatusResponse:
+    """Return health status of platform components, build strategies, and registry info."""
+    if not settings.kagenti_feature_flag_admin:
```
huang195, line 187: Nit: the feature-flag check runs after `Depends(get_kubernetes_service)` resolves, so the kube client is built even when the flag is off. Cheap to reorder — e.g. make the flag check its own dependency, or fetch the client lazily — but not harmful today since the 404 still returns correctly.

kagenti/backend/app/core/config.py
```diff
@@ -76,6 +76,7 @@ def _add_domain_cors_origin(self) -> "Settings":
     kagenti_feature_flag_agent_sandbox: bool = False
     kagenti_feature_flag_authbridge_api: bool = False
     kagenti_feature_flag_skills: bool = False
+    kagenti_feature_flag_admin: bool = False
```
huang195, line 79: Nit on naming: `kagenti_feature_flag_admin` reads as gating "the admin page", but it actually only gates the Platform Status card (the /admin route and auth-status card are always visible). If it stays narrow, consider `kagenti_feature_flag_platform_status`; if it's meant to be a broader admin umbrella later, keep it and maybe note the intent in the CLAUDE.md entry. Renaming once it's exposed via env vars and values.yaml is costly.

kagenti/ui-v2/src/pages/AdminPage.tsx
```diff
@@ -254,24 +373,12 @@ export const AdminPage: React.FC = () => {
...
-                <Alert variant="info" title="Coming soon" isInline>
-                  Platform configuration settings (global agent settings,
-                  resource quotas, namespace management, etc.) will be available
-                  in a future release.
-                </Alert>
-              </CardBody>
-            </Card>
-          </GridItem>
+          {/* Platform Status */}
+          {featureFlags.admin && (
```
huang195, line 377: Nit: when `featureFlags.admin` is false, the `md={6}` grid slot becomes empty and the Authentication card sits alone at half-width. Previously the "Coming soon" card filled that slot. Acceptable if intentional — just flagging in case you'd prefer the auth card to span full width when the status card is hidden.

huang195 (approved):
Adds a `kagenti_feature_flag_admin` gated Platform Status dashboard with a new `GET /config/platform-status` endpoint (404 when flag off) and a `PlatformStatusCard` UI that polls every 15s showing component health (Istio, Keycloak, SPIRE, Shipwright, Phoenix), feature flag state, and registry/build info. Replaces the old "Coming soon" Platform Configuration card.

Feature-flag pattern correctly mirrors existing flags — default false in `values.yaml`, `config.py`, `useFeatureFlags.ts`, and documented in `CLAUDE.md`. Env var correctly added to the `kagenti-backend` deployment (confirmed by checking upstream `ui.yaml` — the feature flag block is on the backend deployment, not the UI).

**Areas reviewed**: Python backend, TypeScript frontend, Helm values/templates, CLAUDE.md
**Commits**: 4, all signed-off ✓
**CI**: 23/24 passing; E2E HyperShift pending `/run-e2e`

Main thing worth addressing before merge is the Missing-vs-Unknown semantics. Everything else is polish.

Issue #1524, opened 2026-05-08
feature: Support customized installation with setup-kagenti.sh


esnible:
### Feature Description

Previously installs were done with _deployments/ansible/run-install.sh_.  That script supported an `--env-file <filename>` flag allowing customization.

_scripts/kind/setup-kagenti.sh_ does not support environment customization.  We should add support for `--env-file` or a similar flag to keep that useful functionality.

### Proposed Solution

Add support for `--env-file` to _scripts/kind/setup-kagenti.sh_.

### Want to contribute?

- [x] I would like to work on this issue.

### Additional Context

It might be good to change the path.  _scripts/kind/setup-kagenti.sh_ can be used on any Kubernetes cluster, not just those created with `kind`.

Initial discussion took place at https://github.com/kagenti/kagenti/issues/1266#issuecomment-4383750349

esnible:
It might also be good to support a `--with-examples` argument to install a test workload.  That would make it easier to use for blog articles about deployed systems, etc.

pdettori:
@esnible - definitely useful - we should start with the kind installer and then possibly do the same for the openshift installer.

PR #1548, opened 2026-05-11
Feat: Add optional helm values file override to kind install script

esnible:
## Summary

Adds `--kagenti-deps-values <filename>` and `--kagenti-values <filename>` optional arguments to _scripts/kind/setup-kagenti.sh_.  The files are presumed to be _helm_ YAML value files and passed on to helm.

## Related issue(s)

For #1524

First, created some overrides files:

```bash
cat <<EOF > /tmp/enable-phoenix-v2.yaml
components:
  phoenix:
    enabled: true
EOF

cat <<EOF > /tmp/enable-flags-v2.yaml
featureFlags:
  sandbox: true
  integrations: true
  triggers: true
  # (agentSandbox overrides are ignored because setup-kagenti.sh explicitly uses --set)
  # agentSandbox: true
  skills: true
  authbridgeAPI: true
EOF
```

Then, install Kagenti.  I used

```bash
scripts/kind/setup-kagenti.sh --with-istio --with-spire --with-ui --with-backend --with-otel \
    --kagenti-deps-values /tmp/enable-phoenix-v2.yaml \
   --kagenti-values /tmp/enable-phoenix-v2.yaml \
   --kagenti-values /tmp/enable-flags-v2.yaml
```

... but you can use `--with-all` if you want builds and other features.

After running the install, verify that Phoenix is running in Kubernetes and present on Kagenti's observability tab.  Verify the "Triggers" tab is present in the Kagenti UI.

### DIFF
--- a/scripts/kind/setup-kagenti.sh
+++ b/scripts/kind/setup-kagenti.sh
@@ -137,6 +137,8 @@ while [[ $# -gt 0 ]]; do
       echo "                      openaiApiKey, slackBotToken, etc.)"
       echo "  --cluster-name NAME Kind cluster name (default: kagenti)"
       echo "  --domain DOMAIN     Domain for services (default: localtest.me)"
+      echo "  --kagenti-values FILE Helm override file to apply to Kagenti chart"
+      echo "  --kagenti-deps-values FILE Helm override file to apply to Kagenti-deps chart"
       echo "  --dry-run           Show commands without executing"
       echo "  -h, --help          Show this help"
       exit 0 ;;
--- a/scripts/kind/setup-kagenti.sh
+++ b/scripts/kind/setup-kagenti.sh
@@ -66,6 +66,9 @@ MCP_GATEWAY_VERSION="0.6.0"
 KUADRANT_VERSION="1.4.2"
 AGENT_SANDBOX_VERSION="v0.4.3"
 
+KAGENTI_DEPS_VALUES_FILES=()
+KAGENTI_VALUES_FILES=()
+
 # ── Colors & logging ────────────────────────────────────────────────────────
 RED='\033[0;31m'; GREEN='\033[0;32m'; YELLOW='\033[1;33m'; BLUE='\033[0;34m'; NC='\033[0m'
 log_info()    { echo -e "${BLUE}→${NC} $1"; }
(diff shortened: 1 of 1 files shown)


scripts/kind/setup-kagenti.sh
```diff
@@ -537,11 +548,11 @@ DEPS_FLAGS=(
   --set "components.tekton.enabled=false"
   --set "components.shipwright.enabled=false"
   --set "components.kiali.enabled=${WITH_KIALI}"
-  --set "components.phoenix.enabled=false"
   --set "components.mlflow.enabled=${WITH_MLFLOW}"
   --set "mlflow.auth.enabled=${WITH_MLFLOW}"
   --set "components.rhoai.enabled=false"
 )
+DEPS_FLAGS=( "${DEPS_FLAGS[@]}" "${KAGENTI_DEPS_VALUES_FILES[@]}" )
```
huang195, line 555: The removal of `--set "components.phoenix.enabled=false"` a few lines up (and the matching one at ~line 1106 for the kagenti chart) is a deliberate behavior change — previously Phoenix was force-disabled regardless of chart default, now it falls through to the chart's default and is overridable via `--kagenti-deps-values`. Worth leaving a one-line comment next to each removal so a future archaeologist / rebaser doesn't re-add the `--set` and silently break the override path. Something like: `# Phoenix default comes from the chart; do NOT re-add --set — it would shadow --kagenti-deps-values.`

huang195, line 555: Nit: `DEPS_FLAGS=( "${DEPS_FLAGS[@]}" "${KAGENTI_DEPS_VALUES_FILES[@]}" )` reads cleaner as `DEPS_FLAGS+=("${KAGENTI_DEPS_VALUES_FILES[@]}")`. Same for `KAGENTI_FLAGS` at line 1106. Functionally identical; the `+=` form makes the "append" intent obvious at a glance.

scripts/kind/setup-kagenti.sh
```diff
@@ -187,6 +196,8 @@ echo "    Agent Sandbox: $WITH_AGENT_SANDBOX"
 echo "    Skip cluster:  $SKIP_CLUSTER"
 echo "    Build images:  $BUILD_IMAGES"
 echo "    Preload imgs:  $PRELOAD_IMAGES"
+echo "    Kagenti helm --values overrides: ${KAGENTI_VALUES_FILES[*]}"
```
huang195, line 199: Nit: when no overrides are passed, this prints `Kagenti helm --values overrides:` with a bare empty trailing string. Readable but slightly noisy. Options: guard with `${KAGENTI_VALUES_FILES[*]:-(none)}`, or skip the line entirely when the array is empty (`[[ ${#KAGENTI_VALUES_FILES[@]} -gt 0 ]] && echo ...`).

huang195 (approved):
Adds `--kagenti-values` and `--kagenti-deps-values` repeatable CLI flags to `scripts/kind/setup-kagenti.sh`, each appending `--values <file>` pairs to a bash array concatenated onto `DEPS_FLAGS` / `KAGENTI_FLAGS` at `helm install` time. Closes the ergonomic gap where operators had to edit the script to enable optional components like Phoenix.

As part of enabling override, the PR drops the hardcoded `--set components.phoenix.enabled=false` from both charts — necessary since `--set` beats `--values` in Helm's precedence order, so leaving the force-disable in place would have made `--kagenti-deps-values enable-phoenix.yaml` silently no-op. Worth a comment on that removal so a future rebase doesn't re-add it (see #1 below).

**Commits**: 3, all signed-off ✓
**CI**: 21/21 passing

PR #1507, opened 2026-05-07
feat(authbridge): Cut over Helm chart + backend to per-plugin config

huang195:
## Summary

Completes the three-PR authbridge per-plugin-config cutover started in
[kagenti/kagenti-extensions#378](https://github.com/kagenti/kagenti-extensions/pull/378)
and continued in
[kagenti/kagenti-operator#335](https://github.com/kagenti/kagenti-operator/pull/335).
This PR flips the kagenti Helm chart and backend ConfigMap generator
onto the per-plugin schema and pins the paired authbridge
(`v0.5.0-alpha.12`) + kagenti-operator (`v0.2.0-alpha.35`) image
tags.

## Schema rewrite

The chart previously emitted `authbridge-runtime-config` with the
top-level `inbound:` / `outbound:` / `identity:` / `bypass:` blocks.
Authbridge `v0.5.0-alpha.12` rejects that shape (empty-pipeline guard
— kagenti-extensions#378's open-proxy mitigation). This PR emits
`pipeline.inbound.plugins[]` + `pipeline.outbound.plugins[]` with one
entry each (`jwt-validation`, `token-exchange`).

Plugin-level defaults (`audience_file`, `bypass_paths`, identity file
paths) are applied by the authbridge binary from its own convention
layer and are **not** emitted by the chart. See
[authbridge/authlib/plugins/CONVENTIONS.md](https://github.com/kagenti/kagenti-extensions/blob/main/authbridge/authlib/plugins/CONVENTIONS.md).

### Files

- `charts/kagenti/templates/authbridge-template-configmaps.yaml` —
  the release-namespace template the operator propagates to agent
  namespaces that didn't come from the Helm chart.
- `charts/kagenti/templates/agent-namespaces.yaml` — the
  per-agent-namespace copy written directly at install time.
- `kagenti/backend/app/routers/agents.py` —
  `_build_authbridge_runtime_yaml`, used when the API creates an
  agent namespace through the UI.
- `kagenti/backend/tests/test_authbridge_runtime_config.py` —
  assertions walk `pipeline.<direction>.plugins[<name>].config` via
  a small `_plugin_config` helper. Also asserts the absence of
  fields that the plugin now defaults on its own (`bypass_paths`,
  `audience_file`, identity file paths) so the chart→authbridge
  contract stays honest over time.

## Image pins (atomic cutover)

`charts/kagenti/values.yaml`:

| Pin | Before | After |
|---|---|---|
[...]

PR #1558, opened 2026-05-12
fix(backend): resolve agent port from Sandbox pod template, not a constant

huang195:
## Summary

Two previously-merged fixes disagree about which port Sandbox agents listen on:

- **PR #1507** (2026-05-08): `_build_sandbox_manifest` aligns the Sandbox pod's `PORT` env and `containerPort` with `service_port` (default `DEFAULT_OFF_CLUSTER_PORT` = **8080**) so `AGENT_ENDPOINT`, the agent card's advertised `url`, and the actual listener all agree.
- **PR #1525** (2026-05-09): `resolve_agent_url` falls back to `DEFAULT_IN_CLUSTER_PORT` (**8000**) when the Service has no ports and the backend is running in-cluster.

With both in effect, the backend calls `...:8000` while the Sandbox agent listens on `...:8080`. Envoy returns 503 and the UI surfaces "Agent card not available."

## What this PR does

Replaces the hardcoded in-cluster fallback with a read of the owning Sandbox's pod template:

1. If the Service exposes ports (Deployment/StatefulSet/Job path) -> use the first port. **Unchanged behavior.**
2. Else, if a Sandbox with the same name exists -> read `spec.podTemplate.spec.containers[0]` — `containerPort` first, then the `PORT` env var.
3. Else -> fall back to `DEFAULT_OFF_CLUSTER_PORT` (8080) — matches both the sandbox-builder default and the external port used by Gateways/Ingresses.

This also handles Sandboxes created with a custom `servicePorts[0].port` — neither prior fix covered that case.

The long-term fix is upstream [kubernetes-sigs/agent-sandbox#154](https://github.com/kubernetes-sigs/agent-sandbox/issues/154) (Sandbox controller populating its Service with ports); this is the in-tree workaround until then.

## Test plan

- [x] `uv run pytest kagenti/backend/tests/test_routes.py -v` — 15/15 pass locally
- [x] Replace the two `:8000` assertions from #1525 with Sandbox-aware tests (containerPort path + PORT-env path) and a final-fallback test
- [x] Verified end-to-end on a local Kind cluster: built the patched backend image, rolled out, deployed the weather agent via UI, confirmed `GET /.well-known/agent-card.json` returns 200 and the agent card renders in the chat tab
- [ ] CI green

## Scope check

| Path | Behavior | Changed? |
|---|---|---|
| Deployment / StatefulSet / Job | Service has ports → first lookup wins | No |
| Sandbox, default port (8080) | Service empty → Sandbox lookup returns 8080 | **Fixes current bug** |
| Sandbox, custom port | Service empty → Sandbox lookup returns custom port | **New: improvement** |
| Off-cluster, nothing found | Final fallback `:8080` with domain | Same (was 8080) |

Assisted-By: Claude (Anthropic AI) &lt;noreply@anthropic.com&gt;

### DIFF
--- a/kagenti/backend/app/utils/routes.py
+++ b/kagenti/backend/app/utils/routes.py
@@ -6,6 +6,7 @@ Utility functions for creating HTTPRoutes (Kubernetes) and Routes (OpenShift).
 """
 
 import logging
+from typing import Optional
 
 from kubernetes.client import ApiException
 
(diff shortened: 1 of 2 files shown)


kagenti/backend/app/utils/routes.py
```diff
@@ -334,13 +335,57 @@ def lookup_service_port(
...
+    if not containers:
+        return None
+    ports = containers[0].get("ports") or []
+    if ports and isinstance(ports[0].get("containerPort"), int):
+        return ports[0]["containerPort"]
+    for env in containers[0].get("env", []) or []:
+        if env.get("name") == "PORT":
+            try:
+                return int(env.get("value"))
+            except (TypeError, ValueError):
```
cwiklik, line 365: suggestion: The `default_port=0` sentinel relies on `if port:` treating 0 as falsy to fall through to the Sandbox lookup. Consider adding a brief comment (e.g. `# 0 = sentinel; Service had no ports`) since this pattern is non-obvious and fragile if someone later adds explicit port-0 handling.

kagenti/backend/app/utils/routes.py
```diff
@@ -334,13 +335,57 @@ def lookup_service_port(
...
+    try:
+        sandbox = kube.get_sandbox(namespace=namespace, name=name)
+    except ApiException:
+        return None
+    containers = (
+        sandbox.get("spec", {}).get("podTemplate", {}).get("spec", {}).get("containers", [])
     )
-    port = lookup_service_port(name, namespace, kube, fallback)
-    return get_agent_url(name, namespace, port)
+    if not containers:
```
cwiklik, line 356: nit: `containers[0].get("env", []) or []` — the `or []` is redundant since `.get(key, [])` already returns `[]` when the key is absent. It only guards against an explicit `None` value in the dict, which is unlikely from the K8s API.

cwiklik (approved):
Clean fix for the port-mismatch bug between PR #1507 and #1525. The 3-tier resolution (Service → Sandbox pod template → default) is well-structured and test coverage is adequate.

**Areas reviewed**: Python (logic, imports, security), Tests
**Commits**: 2 commits, all signed-off
**CI**: lint/tests/security pass; Deploy & Test pending

PR #1560, opened 2026-05-12
fix: use /health for MLflow probes to avoid OIDC 401

### LINKED ISSUE #1559: MLflow pod CrashLoopBackOff: liveness probe hits OIDC-protected /version endpoint
## Problem

The MLflow deployment enters CrashLoopBackOff when `mlflow-oidc-auth` is enabled. The pod starts correctly (startup probe passes via TCP), but the liveness and readiness probes send unauthenticated `GET /version` requests which return HTTP 401.

After 3 consecutive liveness failures (~90s), the kubelet kills the container, creating an infinite restart loop.

## Root Cause

`mlflow-oidc-auth` v7.2.0 protects all endpoints with OIDC session authentication. The `/version` endpoint is **not** in the `unprotected_prefixes` tuple in `AuthMiddleware._is_unprotected_route`, so unauthenticated probe requests are rejected with 401.

## Fix

Change the liveness and readiness probe path from `/version` to `/health`. The `/health` endpoint is natively listed in `mlflow-oidc-auth`'s `unprotected_prefixes` and does not require authentication.

## Affected File

- `charts/kagenti-deps/templates/mlflow.yaml` (readinessProbe and livenessProbe)

## Reproduction

1. Deploy with `mlflow.auth.enabled=true`
2. Observe `kubectl get pods -n kagenti-system -l app=mlflow` showing CrashLoopBackOff
3. `kubectl logs` shows: `Authentication failed for /version: No session authentication`

pdettori:
## Summary

- Change MLflow liveness/readiness probe endpoint from `/version` to `/health`
- `/health` is natively unprotected in `mlflow-oidc-auth`'s `AuthMiddleware`, so probes pass without authentication
- Fixes CrashLoopBackOff when `mlflow.auth.enabled=true`

Fixes #1559

## Test plan

- [x] Deploy to Kind with `mlflow.auth.enabled=true`
- [x] Verify pod reaches `1/1 Running` with 0 restarts
- [x] Confirm `/health` returns 200 without authentication

Assisted-By: Claude Code

### DIFF
--- a/charts/kagenti-deps/templates/mlflow.yaml
+++ b/charts/kagenti-deps/templates/mlflow.yaml
@@ -410,14 +410,14 @@ spec:
           failureThreshold: 30  # Allow up to 5 minutes for startup
         readinessProbe:
           httpGet:
-            path: /version
+            path: /health
             port: 5000
           initialDelaySeconds: 5
           periodSeconds: 10
           failureThreshold: 3
         livenessProbe:
           httpGet:
-            path: /version
+            path: /health
             port: 5000
           initialDelaySeconds: 10
           periodSeconds: 30


charts/kagenti-deps/templates/mlflow.yaml
```diff
@@ -410,14 +410,14 @@ spec:
           failureThreshold: 30  # Allow up to 5 minutes for startup
         readinessProbe:
           httpGet:
-            path: /version
+            path: /health
```
huang195, line 413: **suggestion** (non-blocking): The probe fix itself is correct. Worth flagging that this PR surfaces two pre-existing test failures in `kagenti/tests/e2e/common/test_mlflow_auth.py`:

- `test_mlflow_version_endpoint` (L278) and `test_mlflow_health_check` (L482) both hit `/version` with no token and assert the response is in `{200, 302, 307}`. With `mlflow.auth.enabled=true`, `mlflow-oidc-auth` returns a direct `401 {"detail":"Authentication required"}` — not a redirect — so both tests fail.

They were hidden on main because MLflow crashlooped during install and the tests never ran. Now that install succeeds, they surface. Options:

1. Update the tests to accept `401` as a valid "auth is enforced" signal (or switch the health-probe test to hit `/health` instead of `/version`).
2. Bundle the test fix into this PR.
3. Follow-up PR.

Happy with any of the three — doesn't block this fix from merging.

huang195 (approved):
## Summary

Minimal 2-line change to switch MLflow probes from `/version` → `/health`. Root-cause fix for #1559 — `/version` is OIDC-protected in `mlflow-oidc-auth`, causing CrashLoopBackOff when `mlflow.auth.enabled=true`.

Verified against upstream [`mlflow-oidc/mlflow-oidc-auth`](https://github.com/mlflow-oidc/mlflow-oidc-auth) source — `/health` is explicitly in the unprotected-routes list (`mlflow_oidc_auth/middleware/auth_middleware.py`), with a dedicated unit test asserting `_is_unprotected_route("/health") is True`.

`startupProbe` uses `tcpSocket` so not affected. Complete fix.

**Areas reviewed**: Helm, auth claims (cross-repo), CI
**Commits**: 1, signed-off (DCO pass)
**CI status**: `Deploy & Test` failing, but the 2 failures (`test_mlflow_version_endpoint`, `test_mlflow_health_check` in `kagenti/tests/e2e/common/test_mlflow_auth.py`) are a pre-existing test-design mismatch that this fix **exposes** (main branch was failing earlier — install timed out on the CrashLoop this PR resolves). See inline comment.

PR #1562, opened 2026-05-13
fix: remove deprecated sidecar client registration toggle from UI

Alan-Cha:
## Summary

Client registration is now exclusively handled by the kagenti-operator controller (see kagenti-operator#321). The UI toggle and backend label injection for sidecar-based client registration are no longer needed.

### UI (Frontend)
- ❌ Remove `clientRegistrationInject` toggle from ImportAgentPage and ImportToolPage
- ❌ Remove `clientRegistrationInject` field from API types
- 📝 Update AuthBridge descriptions to reflect operator-managed registration
- 📝 Change "three sidecars" → "two sidecars" (envoy-proxy + spiffe-helper, no more client-registration)

### Backend
- 🔒 Deprecate `clientRegistrationInject` parameter (still accepted for backward compatibility but ignored)
- ❌ Remove logic that sets `KAGENTI_CLIENT_REGISTRATION_INJECT_LABEL` when field is True
- 📝 Add comments explaining client registration is now operator-managed

## Context

Per @huang195's comment:
> One thing I noticed is that we completely removed client registration image from Operator's helm values. A problem could arise since our UI still has the toggle to allow a user to select client registration to be done in a sidecar. This would break the agent if this setting is enabled. We should either remove this toggle from the UI or add the client registration back to the Operator's helm values.

This PR removes the toggle since sidecar-based client registration is sunset.

## Testing

- [ ] Deploy Kagenti from origin/main to Kind cluster
- [ ] Import weather agent with AuthBridge enabled
- [ ] Verify operator handles client registration (check keycloak-admin-secret in kagenti-system only)
- [ ] Verify no client-registration sidecar injected
- [ ] Verify weather agent works correctly

## Related

- kagenti-operator#321 - Move keycloak-admin-secret to operator namespace
- #1422 - Security fix that moved credentials to kagenti-system

Assisted-By: Claude Code

### DIFF
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -2414,8 +2414,6 @@ def _build_common_labels(
         labels[KAGENTI_ENVOY_PROXY_INJECT_LABEL] = "false"
     if request.spiffeHelperInject is False:
         labels[KAGENTI_SPIFFE_HELPER_INJECT_LABEL] = "false"
-    # Note: Client registration is now handled by kagenti-operator, not sidecars.
-    # The clientRegistrationInject field is deprecated and ignored.
     return labels
 
 
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -2414,8 +2414,8 @@ def _build_common_labels(
         labels[KAGENTI_ENVOY_PROXY_INJECT_LABEL] = "false"
     if request.spiffeHelperInject is False:
         labels[KAGENTI_SPIFFE_HELPER_INJECT_LABEL] = "false"
-    if request.clientRegistrationInject is True:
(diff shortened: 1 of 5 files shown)


pdettori (approved):
Clean, well-scoped removal of the deprecated sidecar client registration toggle.

**Reviewed areas**: Frontend (React/TypeScript), Backend (Python), API types

**Findings**:
- UI checkbox and state removed correctly from both ImportAgentPage and ImportToolPage
- Backend label-setting logic removed; request field kept for backward compatibility (good deprecation approach)
- TypeScript interfaces updated — field was optional so no breaking change for existing callers
- Indentation fix in `api.ts` for adjacent fields is appropriate (pre-existing issue exposed by removal)
- Descriptions correctly updated: "three sidecars" → "two sidecars"
- Both commits signed-off ✓
- CI: Deploy & Test failed on unrelated MLflow auth tests; all lint/security/unit checks pass

No blocking issues found.

PR #1576, opened 2026-05-13
fix(kind): handle empty arrays under bash 3.2 set -u

cwiklik:
## Summary

- Fix `setup-kagenti.sh` crash on macOS default bash 3.2 when no `--kagenti-values` or `--kagenti-deps-values` are passed
- Bash 3.2 treats empty arrays as unbound under `set -u`; bash 5.x does not
- Uses `${ARR[*]:-}` for display and `${ARR[@]+"${ARR[@]}"}` for array expansion

## Root cause

Commit a183d4d4 ("Add optional helm values file override") introduced `KAGENTI_VALUES_FILES=()` with `${KAGENTI_VALUES_FILES[*]}` expansion under `set -euo pipefail`. This works on bash 5 but fails on macOS default bash 3.2.

## Test plan

- [x] Verified fix works on `/bin/bash` (3.2) and bash 5.x
- [ ] `scripts/kind/setup-kagenti.sh --with-ui --with-spire` runs past line 199

Assisted-By: Claude Code

### DIFF
--- a/scripts/kind/setup-kagenti.sh
+++ b/scripts/kind/setup-kagenti.sh
@@ -196,8 +196,8 @@ echo "    Agent Sandbox: $WITH_AGENT_SANDBOX"
 echo "    Skip cluster:  $SKIP_CLUSTER"
 echo "    Build images:  $BUILD_IMAGES"
 echo "    Preload imgs:  $PRELOAD_IMAGES"
-echo "    Kagenti helm --values overrides: ${KAGENTI_VALUES_FILES[*]}"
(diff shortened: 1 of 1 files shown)


pdettori (approved):
Clean, minimal fix for bash 3.2 compatibility under `set -u`.

Uses the canonical idioms:
- `${ARR[*]:-}` for display contexts (safe default)
- `${ARR[@]+"${ARR[@]}"}` for array expansion (conditional expansion)

CI passes including shell lint. No issues found.

PR #1301, opened 2026-04-21
feat: :wrench: Update MCP gateway to v0.6.0 and enable gateway tracing

evaline-ju:
## Summary

Update to use latest MCP gateway [v0.6.0](https://github.com/Kuadrant/mcp-gateway/releases/tag/v0.6.0) - the main change is the CRD name and the switch to the controller. This version also provides MCP gateway router tracing, where some of the detailed traces (like initialize, notifications, etc. are filtered out in the Otel collector config by default). Here, we patch the env vars (i.e. use user-provided env vars) to enable the traces to be collected.

Testing with the weather agent and updated weather tool (https://github.com/kagenti/agent-examples/pull/255) can provide a trace like the following
Screenshot 2026-04-21 at 1 21 20 PM

## Related issue(s)

Fixes #1135 

Assisted-By: Claude (Anthropic AI) [noreply@anthropic.com](mailto:noreply@anthropic.com) for finding what was necessary to update MCP gateway to this version

PR #1578, opened 2026-05-13
fix: :wrench: Update MCP gateway router tracing for MLFlow

evaline-ju:
## Summary

- Ports the MCP gateway otel env var patch from the Ansible installer (added in #1301) to the bash setup script setup-kagenti.sh. The MCP gateway chart  does not expose OTel config via Helm values, so kubectl set env is the only injection point.
- Adds an AuthorizationPolicy allowing any mesh workload to push traces to the OTel collector (previously blocked by  implicit deny from ambient mesh with a workload-scoped policy). Uses selector + rules: [{}] to work for any agent namespace without configuration.
- Updates the otel collecgtor configuration for MLFlow: the attributes["http.method"] != nil dropped all HTTP spans including MCP router tool call spans in MLFlow. Now we drop HTTP method verb spans (POST/GET/DELETE) and MCP gateway session handshake noise (initialize, notifications/*, tools/list, mcp-router.* GET/DELETE).

### DIFF
--- a/charts/kagenti-deps/templates/otel-collector.yaml
+++ b/charts/kagenti-deps/templates/otel-collector.yaml
@@ -169,8 +169,8 @@ data:
 {{- end }}
 {{- if and .Values.components.otel.enabled .Values.components.mcpGateway.enabled }}
 ---
-# Allow workloads to push traces to the OTel collector.
-# Covers MCP Gateway router (mcp-system) and agent namespaces (team1, team2, kagenti-system).
+# Allow any mesh workload to push traces to the OTel collector.
+# Open allow-all on the collector workload avoids hardcoding agent or MCP gateway namespaces.
 apiVersion: security.istio.io/v1
 kind: AuthorizationPolicy
 metadata:
--- a/charts/kagenti-deps/templates/otel-collector.yaml
+++ b/charts/kagenti-deps/templates/otel-collector.yaml
@@ -167,3 +167,22 @@ data:
   base.yaml: |
     {{- include "kagenti.otel.collectorConfig" . | nindent 4 }}
 {{- end }}
+{{- if and .Values.components.otel.enabled .Values.components.mcpGateway.enabled }}
+---
+# Allow workloads to push traces to the OTel collector.
+# Covers MCP Gateway router (mcp-system) and agent namespaces (team1, team2, kagenti-system).
+apiVersion: security.istio.io/v1
+kind: AuthorizationPolicy
+metadata:
+  name: otel-collector-from-mcp-system
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "kagenti.labels" . | nindent 4 }}
+spec:
+  selector:
+    matchLabels:
+      app: otel-collector
+  action: ALLOW
+  rules:
+  - {}
+{{- end }}
--- a/charts/kagenti-deps/templates/otel-collector.yaml
+++ b/charts/kagenti-deps/templates/otel-collector.yaml
@@ -167,7 +167,7 @@ data:
   base.yaml: |
     {{- include "kagenti.otel.collectorConfig" . | nindent 4 }}
 {{- end }}
-{{- if and .Values.components.otel.enabled .Values.components.mcpGateway.enabled }}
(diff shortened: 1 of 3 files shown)


charts/kagenti-deps/templates/otel-collector.yaml
```diff
@@ -167,3 +167,22 @@ data:
   base.yaml: |
     {{- include "kagenti.otel.collectorConfig" . | nindent 4 }}
 {{- end }}
+{{- if and .Values.components.otel.enabled .Values.components.mcpGateway.enabled }}
```
esnible: **must-fix**: This guard references `.Values.components.mcpGateway.enabled`, but `mcpGateway` is not defined under `components:` in `charts/kagenti-deps/values.yaml`. Helm fails templating with `nil pointer evaluating interface {}.enabled` — this is the cause of the failing `Deploy & Test` CI job.

Fix options:
1. Add `mcpGateway:\n  enabled: false` (or `true`) under `components:` in `values.yaml`, **or**
2. Make the guard nil-safe, e.g. `{{- if and .Values.components.otel.enabled (and .Values.components.mcpGateway .Values.components.mcpGateway.enabled) }}`.

esnible (changes requested):
### Summary

The new AuthorizationPolicy template references `.Values.components.mcpGateway.enabled`, but this key does not exist in `charts/kagenti-deps/values.yaml`. The failing `Deploy & Test` CI job confirms this: `nil pointer evaluating interface {}.enabled` at `otel-collector.yaml:170:50`.

**Areas reviewed**: Helm/K8s, Shell, YAML, CI
**CI status**: 1 failing (Deploy & Test — root caused to this PR)

Assisted-By: Claude Code

charts/kagenti-deps/templates/otel-collector.yaml
evaline-ju: updating to be more general with just the otel-collector check. The otel-collector needs to accept traces from all mesh workloads — agents, MCP gateway, future components.

evaline-ju:
Looks like the remaining test failure was already existing (401)

charts/kagenti-deps/values.yaml
```diff
@@ -295,7 +296,8 @@ otel:
           traces:
             span:
               - 'IsMatch(name, "^a2a\\..*")'
-              - 'attributes["http.method"] != nil'
+              - 'name == "POST" or name == "GET" or name == "DELETE"'
```
esnible: **suggestion**: Filtering by `name == "POST"` is fragile because span names depend on the OTel SDK / instrumentation library's HTTP semconv version. Some emit `"HTTP POST"`, some `"POST /path"`, some just `"POST"`. The previous rule (`attributes["http.method"] != nil`) was more portable. If the goal is to drop only the bare-method server spans (and keep MCP router spans even when they have `http.method`), consider something like:

```
- 'attributes["http.method"] != nil and IsMatch(name, "^(POST|GET|DELETE)$")'
```

This combines the attribute check with a regex on the span name, so it only drops spans where the name *is exactly* the verb — independent of which instrumentation produced it.

esnible (approved):
### Summary

The previously-flagged must-fix (nil-pointer in AuthorizationPolicy guard for `.Values.components.mcpGateway.enabled`) is resolved by 648f9d5 — the simpler `{{- if .Values.components.otel.enabled }}` guard avoids the missing values key entirely, and Helm install now succeeds in CI.

The remaining `Deploy & Test` CI failure (MLflow `test_mlflow_version_endpoint` and `test_mlflow_health_check` returning 401) is **pre-existing on main** — same two tests fail on the last 5 main runs (e.g., run 25816964245 from 2026-05-13). Not caused by this PR.

**Areas reviewed**: Helm/K8s, Shell, YAML, CI
**Commits**: 3, all signed-off
**CI status**: 1 failing (Deploy & Test — pre-existing flake on main)

Assisted-By: Claude Code

charts/kagenti-deps/values.yaml
evaline-ju: tested and incorporated the suggestion

PR #1581, opened 2026-05-14
fix: use backend-managed ClusterIP Service for Sandbox workloads

cwiklik:
## Summary

- Align Sandbox `containerPort` and `PORT` env to `8000` (`DEFAULT_IN_CLUSTER_PORT`), same as Deployments
- Backend creates its own ClusterIP Service for Sandbox workloads with `port: 8080 → targetPort: 8000`
- Service labels use `request.workloadType` instead of hardcoded `deployment`
- Do **not** set `spec.service: false` on the Sandbox CR (field not in released CRD)

## Why ClusterIP instead of headless Service?

The Sandbox controller auto-creates a headless Service (`ClusterIP: None`). A headless Service cannot do port translation — DNS resolves directly to pod IPs and kube-proxy is not in the path. A client connecting to `service:8080` reaches `podIP:8080`, not `podIP:8000`.

Since the platform addresses agents at port 8080 (`DEFAULT_OFF_CLUSTER_PORT`) via `AGENT_ENDPOINT`, and the container listens on 8000 (`DEFAULT_IN_CLUSTER_PORT`), we need a ClusterIP Service for kube-proxy to translate `8080 → 8000`.

Verified in Kind: a headless Service with `port: 8080, targetPort: 8000` fails (connection refused on 8080), while a ClusterIP Service with the same spec works correctly.

## Sandbox controller interaction

The controller still creates its headless Service, but it cannot overwrite our ClusterIP Service (ClusterIP is immutable). Routing works correctly through ours. The only side effect is a cosmetic "ClusterIP mismatch" warning in the Sandbox operator log.

agent-sandbox recently merged `spec.service` opt-out support ([PR #775](https://github.com/kubernetes-sigs/agent-sandbox/pull/775), merged 2026-05-13) but it is not yet included in any release (latest is v0.4.3, released 2026-04-28). The project follows a roughly weekly release cadence, so the next release will likely include it. Once available, we can set `service: false` on the Sandbox CR to suppress the warning.

## Test plan

- [x] 6 unit tests: containerPort, PORT env, no service field, custom ports, service labels
- [x] Kind: Sandbox agent card reachable via ClusterIP Service `8080 → 8000`
- [x] Kind: Deployment agent unaffected

Supersedes workaround PRs #1507, #1525, #1558

Assisted-By: Claude Code

### DIFF
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -2610,7 +2610,7 @@ def _build_service_manifest(request: "CreateAgentRequest") -> dict:
     Returns:
         Service manifest dictionary.
     """
-    labels = _build_common_labels(request, WORKLOAD_TYPE_DEPLOYMENT)
+    labels = _build_common_labels(request, request.workloadType)
     selector_labels = _build_selector_labels(request)
 
     # Build service ports
@@ -3064,8 +3064,8 @@ async def create_agent(
                 )
                 logger.info(f"Created Sandbox '{request.name}' in namespace '{request.namespace}'")
 
-            # Create Service (not needed for Jobs or Sandboxes)
-            if request.workloadType not in (WORKLOAD_TYPE_JOB, WORKLOAD_TYPE_SANDBOX):
+            # Create Service (not needed for Jobs — Sandbox uses backend-managed Service)
+            if request.workloadType != WORKLOAD_TYPE_JOB:
                 service_manifest = _build_service_manifest(request)
                 kube.create_service(
                     namespace=request.namespace,
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -2885,7 +2885,6 @@ def _build_sandbox_manifest(
         },
         "spec": {
             "replicas": 1,
-            "service": False,
             "podTemplate": {
                 "metadata": {
                     "labels": {
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -3066,11 +3066,27 @@ async def create_agent(
             # Create Service (not needed for Jobs — Sandbox uses backend-managed Service)
             if request.workloadType != WORKLOAD_TYPE_JOB:
                 service_manifest = _build_service_manifest(request)
-                kube.create_service(
-                    namespace=request.namespace,
-                    body=service_manifest,
-                )
-                logger.info(f"Created Service '{request.name}' in namespace '{request.namespace}'")
+                try:
+                    kube.create_service(
+                        namespace=request.namespace,
+                        body=service_manifest,
+                    )
+                    logger.info(
+                        f"Created Service '{request.name}' in namespace '{request.namespace}'"
+                    )
+                except ApiException as e:
+                    if e.status == 409 and request.workloadType == WORKLOAD_TYPE_SANDBOX:
+                        logger.warning(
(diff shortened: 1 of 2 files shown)


kagenti/backend/app/routers/agents.py
```diff
@@ -3072,8 +3063,8 @@ async def create_agent(
                 )
                 logger.info(f"Created Sandbox '{request.name}' in namespace '{request.namespace}'")
 
-            # Create Service (not needed for Jobs or Sandboxes)
-            if request.workloadType not in (WORKLOAD_TYPE_JOB, WORKLOAD_TYPE_SANDBOX):
+            # Create Service (not needed for Jobs — Sandbox uses backend-managed Service)
```
huang195, line 3066: Race with the Sandbox controller's headless Service: if the controller wins (creates its `ClusterIP: None` Service before this `kube.create_service()` runs), our call hits `AlreadyExists` and the workload ends up with the broken headless Service. Backend usually wins in practice (controller reconcile is async), but worth either: (a) catching `AlreadyExists` and patching/replacing, or (b) logging a clear error so it's diagnosable when it does flake.

kagenti/backend/tests/test_sandbox_manifest.py
```diff
@@ -0,0 +1,107 @@
+# Copyright 2025 IBM Corp.
+# Licensed under the Apache License, Version 2.0
+
+"""Unit tests for Sandbox manifest builder and Service creation."""
+
+from unittest.mock import MagicMock
```
huang195, line 6: nit: `MagicMock()` without `spec=CreateAgentRequest` silently returns a Mock for any attribute — if the schema gains a field that the manifest builder reads, tests pass while prod breaks. Switch to `MagicMock(spec=CreateAgentRequest)` to catch drift.

kagenti/backend/tests/test_sandbox_manifest.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def test_service_labels_use_request_workload_type(self):
+        """Service labels should reflect the actual workload type, not hardcoded deployment."""
+        from app.routers.agents import _build_service_manifest
+
+        request = _make_request(workloadType="sandbox")
+        manifest = _build_service_manifest(request)
+
+        labels = manifest["metadata"]["labels"]
+        assert labels.get("kagenti.io/workload-type") == "sandbox"
```
huang195, line 108: No test covers the control-flow change in `create_agent` (Sandbox now goes through `_build_service_manifest`). The unit tests exercise the builders in isolation, but the actual fix is the conditional on agents.py line ~3066. A small test asserting `kube.create_service` is called for `workloadType=sandbox` would lock in the behavior.

huang195 (approved):
### Summary

Aligns Sandbox `containerPort` to 8000 (matching Deployment), drops the broken `spec.service: false` opt-out, and adds a backend-managed ClusterIP Service so kube-proxy translates 8080→8000 (the auto-created headless Service can't do this). Diff is clean, well-scoped, mirrors the existing StatefulSet pattern. PR body explains the *why* thoroughly.

**Areas reviewed**: Python (backend), tests, CI status
**CI status**: 21 pass, 1 fail (`Deploy & Test` — unrelated Keycloak/Hibernate duplicate-key issue, not caused by this PR), E2E HyperShift pending manual trigger

No blockers. Dobra robota.

kagenti/backend/app/routers/agents.py
cwiklik: addressed

Issue #1468, opened 2026-05-05
epic: Agent Platform Auth UX


davidhadas:
**Executive summary.** Kagenti's value is hard to convey without showing, end-to-end, how a real application authenticates, discovers agents, and exchanges tasks with them. This epic builds a demo application layer that serves as a reference implementation of production-like integration — covering token acquisition, agent discovery, task submission, streaming responses, and lifecycle observation across multiple Kagenti use-case types. The outcome is a one-command-deployable, well-documented demo, integrated into the official Kagenti demos, that makes the platform compelling and easy to understand for customers and developers alike.

# Summary

To present and help customers value Kagenti we need to show E2E how users use agents in Kagenti.

Progress:
[X] Initial User App - https://github.com/kagenti/kagenti/pull/1583
[ ] - Add Token Exchange step before sending a task to an agent to ensure the agent receive token with correct aud
[ ] - Analyze how users may dynamically control the scopes they approve prior to accessing the agent platform. 
[ ] - Analyze how users may know what agents are available to them (rather than what agents are installed). 

### Problem

  
Build a full demo application layer that showcases how real-world applications interact with agents deployed via Kagenti on Kubernetes. This epic defines the scope, components, and deliverables needed to make Kagenti demos compelling, interactive, and easy to understand.

The demo application layer will demonstrate how users and systems:

- Obtain access tokens to authenticate with Kagenti

- Discover available agents

- Send tasks and payloads to agents

- Receive and stream responses

- Observe agent behavior and lifecycle

This layer will serve as a reference implementation for how applications can integrate with Kagenti in production-like scenarios.

The demo application will also illustrate how an application layer can be built for different Kagenti use case types, as defined here:
https://github.com/kagenti/kagenti/blob/main/docs/use-case-types.md

By integrating these use cases into the demo, users will see concrete examples of how Kagenti supports patterns such as conversational agents, workflow agents, tool-using agents, and more.

### Success criteria

- One‑step deployment works: Users can deploy Kagenti, demo agents, and the demo application layer with a single command and interact immediately.

- Authentication is demonstrated: The demo shows how a client obtains a token and uses it to call Kagenti.

- Multiple Kagenti use‑case types are showcased: At least two use‑case types from the official taxonomy are implemented with dedicated demo flows.

- Architecture is documented: Clear diagrams and explanations show how the demo app communicates with Kagenti, handles tokens, and sends/receives tasks.

[...]

PR #1583, opened 2026-05-15
Adding a Kagenti Demo App

### LINKED ISSUE #1582: feature: Create a first Kagenti App Demo using BFF design and basic RBAC
### Feature Description

Present how an App can be developed to work with Kagenti

- Demo a minimalist App
- Include User Login
- Include agent discovery - show only agents a user is allowed to use
- Use RBAC to control what a user is allowed to use

### Proposed Solution

In this initial demo, a user with kagenti_operator role is allowed to use all agents and a user without this role is not allowed to use any agent. 

### Want to contribute?

### Additional Context

_No response_

davidhadas:
## Summary
A demonstration of user application via which Kagenti users can get services from agents running in Kagenti. 

The user: 

1. Login to Kagenti using the _Kagenti IdP_ (An IdP Kagenti is configured to work with)
2. Choose an Agent from a list of available agents
3. Send a request to the agent 
4. Get a response

This demonstration also support Human In The Loop (a Kagenti feature under development). A user request may require one or more agents in Kagenti to access one or more Third Party Systems under a _Third Party IdP_ (not federated with the  _Kagenti IdP_ .  The Human In The Loop enable an OAuth 2 process against the _Third Party IdP_. If the request require accessing such Third Party Systems the user:

1. Login to Kagenti using the _Kagenti IdP_ (An IdP Kagenti is configured to work with)
2. Choose an Agent from a list of available agents
3. Send a request to the agent and get a response
5. Give Consent to use his _Third Party IdP_ credentials to support his request 
6. Login to the _Third Party IdP_
7. Get a response

The PR code showcases how an app can be developed and document two design patterns:
- Designing a simpler "Backend For Frontend (BFF)" between the Frontend and the existing Kagenti Backend (which is what was implemented in this PR) - we limited this demo as a single user demo. The BFF supports a single user at a time.
- Designing a full App Backend  that works with the Kagenti APIs - leaving it for a future demo if someone feel the urge :wink:

The PR adds a user role in KeyCloak to control which users can activate agents. 
Later we will work on improving this model to show Fine‑Grained Access Control

## Related issue(s)
Epic: https://github.com/kagenti/kagenti/issues/1468

Fixes #1582

### DIFF
--- a/kagenti/examples/app-demo/backend/app/routes/agents.py
+++ b/kagenti/examples/app-demo/backend/app/routes/agents.py
@@ -1,0 +1,32 @@
+from urllib.parse import quote
+
+from fastapi import APIRouter, Request
+from fastapi.responses import JSONResponse
+
+router = APIRouter(tags=["agents"])
+
+
+async def _proxy_get(request: Request, path: str) -> JSONResponse:
+    client = request.app.state.http_client
+    headers = {}
+    auth = request.headers.get("authorization")
+    if auth:
+        headers["Authorization"] = auth
+    resp = await client.get(path, headers=headers)
+    return JSONResponse(content=resp.json(), status_code=resp.status_code)
+
+
+@router.get("/namespaces")
+async def list_namespaces(request: Request):
+    return await _proxy_get(request, "/api/v1/namespaces")
+
+
+@router.get("/agents")
+async def list_agents(request: Request, namespace: str = ""):
+    """List all agents. Access is enforced at chat time."""
+    path = (
+        f"/api/v1/agents?namespace={quote(namespace, safe='')}"
+        if namespace
+        else "/api/v1/agents"
+    )
+    return await _proxy_get(request, path)
--- a/kagenti/examples/app-demo/backend/app/routes/chat.py
+++ b/kagenti/examples/app-demo/backend/app/routes/chat.py
@@ -1,0 +1,119 @@
+import json
+import logging
+from urllib.parse import quote
+
+import httpx
+from fastapi import APIRouter, Request
+from fastapi.responses import JSONResponse
+from pydantic import BaseModel
+
+from app.config import settings
+
+logger = logging.getLogger(__name__)
+
+router = APIRouter(tags=["chat"])
+
+
+class ChatRequest(BaseModel):
+    message: str
+    session_id: str | None = None
+
+
+@router.post("/chat/{namespace}/{name}/send")
+async def send_message(
+    namespace: str,
+    name: str,
+    body: ChatRequest,
+    request: Request,
+):
+    """Use the Kagenti streaming endpoint, accumulate all chunks,
+    and return a single JSON response."""
+    headers = {
+        "Content-Type": "application/json",
+        "Accept": "text/event-stream",
+    }
+    auth = request.headers.get("authorization")
+    if auth:
+        headers["Authorization"] = auth
+
+    upstream = httpx.AsyncClient(
+        base_url=settings.kagenti_api_url,
+        timeout=httpx.Timeout(120.0, connect=10.0),
+    )
+
+    try:
+        resp = await upstream.send(
+            upstream.build_request(
+                "POST",
+                f"/api/v1/chat/{quote(namespace, safe='')}/{quote(name, safe='')}/stream",
+                json=body.model_dump(exclude_none=True),
+                headers=headers,
+            ),
+            stream=True,
+        )
+    except httpx.RequestError as e:
+        await upstream.aclose()
+        logger.error("Cannot connect to Kagenti backend: %s", e)
+        return JSONResponse(
+            content={"detail": "Cannot connect to upstream"},
+            status_code=503,
+        )
+
+    if resp.status_code >= 400:
+        try:
+            await resp.aread()
+            detail = resp.text[:500]
+        except Exception:
+            detail = str(resp.status_code)
+        await resp.aclose()
+        await upstream.aclose()
+        return JSONResponse(
+            content={"detail": detail},
+            status_code=resp.status_code,
+        )
+
+    content_parts: list[str] = []
+    error_msg: str | None = None
+    session_id = body.session_id or ""
+
+    try:
+        async for line in resp.aiter_lines():
+            if not line or not line.startswith("data: "):
+                continue
+            data = line[6:]
+            if data == "[DONE]":
+                break
+            try:
+                chunk = json.loads(data)
+            except json.JSONDecodeError:
+                continue
+
+            if chunk.get("session_id"):
+                session_id = chunk["session_id"]
+
+            if chunk.get("error"):
+                error_msg = chunk["error"]
+                break
+
+            text = chunk.get("content", "")
+            if text:
+                content_parts.append(text)
+    finally:
+        await resp.aclose()
+        await upstream.aclose()
+
+    if error_msg:
+        return JSONResponse(
+            content={"detail": error_msg},
+            status_code=502,
+        )
+
+    return JSONResponse(
+        content={
+            "content": "\n".join(content_parts)
+            if content_parts
+            else "No response from agent",
+            "session_id": session_id,
+            "is_complete": True,
+        }
+    )
--- a/kagenti/examples/app-demo/backend/app/routes/token_broker.py
+++ b/kagenti/examples/app-demo/backend/app/routes/token_broker.py
@@ -1,0 +1,150 @@
+import asyncio
+import logging
+
+from fastapi import APIRouter, Request
+from fastapi.responses import JSONResponse, Response
+from pydantic import BaseModel
+
+from app.config import settings
+from app.token_broker import TokenBrokerClient
+
+logger = logging.getLogger(__name__)
+
+router = APIRouter(tags=["token-broker"])
+
+
+# IMPORTANT: Single-user demo limitation
+# These module-level globals create a process-wide singleton session.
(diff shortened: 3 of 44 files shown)


kagenti/examples/app-demo/backend/app/routes/token_broker.py
```diff
@@ -0,0 +1,133 @@
...
+        )
+
+    auth = request.headers.get("authorization", "")
+    if not auth.startswith("Bearer "):
+        return JSONResponse(content={"detail": "Unauthorized"}, status_code=401)
+    jwt = auth[7:]
+
+    await _cleanup()
+
+    logger.info(f"Creating Token Broker session with redirect_url: {body.redirect_url}")
```
mrsabath: This is the CodeQL log-injection alert (#1040). `body.redirect_url` is user-supplied (from the `SessionRequest` body) and logged via an f-string, so a CRLF in it could forge log lines. The alert shows as resolved, but that's an auto-resolve from the line shifting — the code path is unchanged. Suggest lazy logging so the logging layer handles it:

```python
logger.info("Creating Token Broker session with redirect_url: %s", body.redirect_url)
```

Minor / demo-appropriate, but it would clear the scanner finding too.

kagenti/examples/app-demo/backend/app/routes/token_broker.py
```diff
@@ -0,0 +1,133 @@
...
+
+def _get_client() -> TokenBrokerClient | None:
+    global _client
+    if _client is None and settings.token_broker_url:
+        _client = TokenBrokerClient(settings.token_broker_url)
+    return _client
+
+
+async def _polling_loop(jwt: str, queue: asyncio.Queue) -> None:
+    client = _get_client()
```
mrsabath, line 41: These module-level globals (`_jwt`, `_event_queue`, `_poll_task`) make the token-broker session effectively a process-wide singleton — a second concurrent user would clobber the first's polling session and JWT. That's fine for a single-user demo, but worth a one-line comment here so a future reader doesn't mistake it for multi-user-safe.

kagenti/examples/app-demo/backend/app/routes/chat.py
```diff
@@ -0,0 +1,118 @@
+import json
```
mrsabath, line 1: nit: `namespace`/`name` are interpolated into the upstream request path without URL-encoding. Kubernetes naming rules make this safe in practice, so no action needed — but `urllib.parse.quote(..., safe="")` would make it robust if those ever come from a less-constrained source.

mrsabath (commented):
Nice end-to-end demo — the BFF pattern and the HITL token-broker flow are clearly laid out, and the docs are thorough. All CI green; E2E HyperShift is just waiting on a maintainer `/run-e2e`.

I reviewed the security-sensitive surfaces (token broker, Keycloak admin scripts, manifests) against the head SHA and the repo's existing examples. Leaving a few comments inline — none are blockers. The one worth a look is the CodeQL log-injection alert on the user-supplied `redirect_url` (#1040): it auto-resolved on a line shift, but the f-string log is still present — a lazy `%s` closes it cleanly.

I deliberately did **not** flag the in-cluster HTTP-to-Keycloak, unpinned base images, or missing `securityContext` — those match every other `kagenti/examples/*` and are fine for demo code. One note outside the diff: the single `"Demo App"` commit is signed-off (DCO passes) but could use an emoji-prefixed, more descriptive subject per our commit conventions on a future squash.

Not approving yet — leaving room for other reviewers to weigh in first.

**Areas reviewed**: Python (backend BFF + token broker), Keycloak scripts, K8s manifests, Dockerfiles, docs, conventions, CI · **Commits**: 1, signed-off: yes · **CI**: passing (E2E gated on maintainer)

kagenti/examples/app-demo/backend/app/routes/token_broker.py
davidhadas: fixed

kagenti/examples/app-demo/backend/app/routes/token_broker.py
davidhadas: fixed

kagenti/examples/app-demo/backend/app/routes/chat.py
davidhadas: fixed

kagenti/examples/app-demo/backend/app/routes/token_broker.py
mrsabath: Verified fixed in `a1aab2f` — `safe_redirect_url` now strips CR/LF and the log uses lazy `%s`. Closes the log-injection vector (CodeQL #1040) properly, not just via the line-shift auto-resolve. Thanks!

kagenti/examples/app-demo/backend/app/routes/token_broker.py
mrsabath: Verified in `a1aab2f` — the new comment block above the globals clearly documents the single-user limitation and the per-user/Redis migration path. Exactly what was needed for a future reader. 👍

kagenti/examples/app-demo/backend/app/routes/chat.py
mrsabath: Verified in `a1aab2f` — `quote(namespace, safe="")` / `quote(name, safe="")` now applied to the path segments in `chat.py`. Robust regardless of source. (Minor: the `?namespace=` query value in `agents.py:list_agents` is still raw, but that is a query param to an in-cluster backend with k8s-constrained values — non-blocking.)

mrsabath (approved):
Re-review of `a1aab2f`. All three comments from my 2026-06-08 pass have been addressed and I verified each fix against the current diff:

1. **Log injection (`token_broker.py`)** — `redirect_url` is now CR/LF-stripped into `safe_redirect_url` and logged with lazy `%s`. This closes the CodeQL log-injection alert (#1040) at the code level, not just via the line-shift auto-resolve. ✅
2. **Process-wide singleton globals (`token_broker.py`)** — a clear comment block now documents the single-user limitation and the per-user / Redis migration path. ✅
3. **Path interpolation (`chat.py`)** — `quote(namespace, safe="")` / `quote(name, safe="")` applied to the upstream path segments. ✅

Since my last pass, **Deploy & Test** also went green (22m50s). No new issues introduced by the force-push: no `.claude/`/`.vscode/` config files, and all Keycloak credentials are read from K8s secrets (no hardcoded literals). The demo-acceptable items I called out last time (in-cluster HTTP to Keycloak, unpinned base images, missing `securityContext`) remain consistent with the other `kagenti/examples/*` and are fine for demo code.

Approving. Nice work on the BFF + HITL token-broker demo, David. Only remaining gate is **E2E HyperShift**, which still needs a maintainer `/run-e2e`.

**Areas reviewed**: Python (BFF backend, token broker, Keycloak scripts), K8s manifests, Dockerfiles, frontend, docs, conventions, CI · **Commits**: 1, signed-off: yes (DCO green) · **CI**: passing (E2E gated on maintainer)

kagenti/examples/app-demo/backend/app/routes/agents.py
```diff
@@ -0,0 +1,26 @@
...
+
+@router.get("/namespaces")
+async def list_namespaces(request: Request):
+    return await _proxy_get(request, "/api/v1/namespaces")
+
+
+@router.get("/agents")
+async def list_agents(request: Request, namespace: str = ""):
+    """List all agents. Access is enforced at chat time."""
+    path = f"/api/v1/agents?namespace={namespace}" if namespace else "/api/v1/agents"
```
pdettori: nit (non-blocking): `namespace` is interpolated raw into the query path here, while the frontend sends it via `encodeURIComponent`. The `chat.py` fix applied `urllib.parse.quote(..., safe="")` to its path segments — mirroring that here (`quote(namespace, safe="")`) would make this robust if `namespace` ever comes from a less-constrained source. Low risk today since the value is k8s-constrained and the call is in-cluster, so no action required for the demo.

kagenti/examples/app-demo/backend/app/routes/token_broker.py
```diff
@@ -0,0 +1,149 @@
...
+    jwt = auth[7:]
+
+    await _cleanup()
+
+    # Sanitize user input to prevent log injection (remove newlines/carriage returns)
+    safe_redirect_url = body.redirect_url.replace("\n", "").replace("\r", "")
+    logger.info(
+        "Creating Token Broker session with redirect_url: %s", safe_redirect_url
+    )
+    ok = await client.create_session(jwt, body.redirect_url)
```
pdettori, line 109: nit (non-blocking): this correctly passes the *original* `body.redirect_url` (not `safe_redirect_url`) — the CR/LF stripping was only for the log sink, and the real redirect should preserve the original value. A one-line comment noting this would prevent a future reader from "tidying" it to `safe_redirect_url` and silently changing redirect behavior.

pdettori (approved):
Approving — all prior review comments are addressed and verified (log-injection CodeQL #1040/#1048/#1049 resolved, single-user globals documented, `chat.py` path segments URL-encoded, all Dockerfiles SHA-pinned). Two minor, non-blocking follow-ups inline below.

**Must-fix before merge (housekeeping, not code):** the branch is currently `BEHIND main` (`mergeStateStatus: BEHIND`). No conflicts, but please update/rebase onto the latest `main` and re-run CI before merging so we merge against current HEAD.

_Assisted-By: Claude Code_

Issue #361, opened 2025-10-29
feature: Kagenti uses deprecated endpoint


esnible:
### Feature Description

Google's A2A examples log "WARNING - jsonrpc_app.py:556 - Deprecated agent card endpoint '/.well-known/agent.json' accessed. Please use '/.well-known/agent-card.json' instead. This endpoint will be removed in a future version."

Kagenti hard-codes the URL in https://github.com/kagenti/kagenti/blob/main/kagenti/ui/lib/constants.py#L85

### Proposed Solution

- Verify none of our examples depend on the old URL
- Update the constant

### Want to contribute?

- [ ] I would like to work on this issue.

### Additional Context

I tested this with https://github.com/a2aproject/a2a-samples/tree/main/samples/python/agents/adk_facts

PR #1584, opened 2026-05-15
chore: post-architecture follow-ups for kagenti-extensions#411 / kagenti-operator#361

huang195:
## Summary

Sweep through the kagenti repo to land all the consumer-side changes for the architecture moves that already merged upstream:

- **kagenti-extensions#411** — three-binary split (`authbridge-proxy` / `authbridge-envoy` / `authbridge-lite`) with one combined image per mode and `spiffe-helper` bundled inside. Dropped the standalone `authbridge-light`, `client-registration`, and `spiffe-helper` images from CI.
- **kagenti-operator#361** — operator-managed Keycloak client registration; per-workload mode resolution chain (`AgentRuntime.Spec.AuthBridgeMode` → namespace ConfigMap → deprecated annotation → cluster default = `proxy-sidecar`); the per-sidecar inject feature gates were removed.

Goal of this PR: every place in the kagenti repo that referenced the dropped sidecars / images / fields gets updated, and every place that needs to surface the new `authBridgeMode` selector picks it up. Single PR for reviewer coherence; commit-by-commit for narrow blast radius if any one slice needs revisiting.

## Commits (9)

| | Commit | Touch |
|---|---|---|
| 1 | `chore(chart): pin authbridge v0.6.0-alpha.1, drop dead image keys` | Chart.yaml subchart bump → operator chart 0.3.0-alpha.1; values.yaml drops authbridgeLight/clientRegistration/spiffeHelper, adds authbridge + authbridge-lite. |
| 2 | `chore(chart): drop hardcoded mode, refresh sidecar UID matrix` | Drop `mode: envoy-sidecar` in both template ConfigMaps so cluster default wins; refresh SCC + RoleBinding comments for the single-combined-sidecar shape. |
| 3 | `ci(token-exchange): rebuild image list for three-binary split` | TX e2e build script switches to authbridge / authbridge-envoy / authbridge-lite / proxy-init. |
| 4 | `chore(examples): drop legacy client-registration-inject label` | Weather-agent OCP fixture no longer opts into a sidecar image that no longer ships. |
| 5 | `test(e2e): replace dead client-registration / authbridge-light checks` | Replace `test_agent_has_client_registration` with an operator-managed-Secret check; drop the authbridge-light log fallback; pin `mode: envoy-sidecar` back into `50-enable-kagenti-ns.sh` (this fixture explicitly tests the Envoy ext_proc path). |
| 6 | `refactor(backend): replace dead inject fields with authBridgeMode` | Drop `envoyProxyInject` / `spiffeHelperInject` / `clientRegistrationInject` from `Create*Request` + `Finalize*Request` + helper-fn signatures + label-write blocks; introduce a single typed `authBridgeMode` field; agents.py plumbs it onto `AgentRuntime.Spec.AuthBridgeMode`; tools.py plumbs it onto the deprecated `kagenti.io/authbridge-mode` pod annotation; drop the `mode: envoy-sidecar` hardcode in the backend's config-builder. |
| 7 | `feat(ui): collapse advanced injection into envoy-mode toggle` | ImportAgentPage + ImportToolPage replace the "Advanced Injection Controls" + "Combined AuthBridge container" Alert with a single nested "Use envoy-sidecar mode" checkbox; `services/api.ts` + `types/index.ts` updated to the new shape. |
| 8 | `refactor(tui): replace dead inject fields with AuthBridgeMode` | Mirror of the backend change in `kagenti/tui/internal/api/types.go`. |
| 9 | `docs: refresh deployment guide and release inventory for #411 / #361` | Rewrite the deployment-modes table + minimal config example + reference table for the per-plugin schema; add the resolution-chain section; fix the kagenti-extensions image inventory in releasing.md. |

## Verification

- `helm template charts/kagenti` renders cleanly; `kagenti-platform-config` no longer leaks the three dropped image keys.
- `bash -n` on both updated TX-e2e scripts.
- `python3 -c "import ast; ast.parse(...)"` on every modified Python file.
- `go build ./... && go vet ./...` on `kagenti/tui`.
- `npx tsc --noEmit` on `kagenti/ui-v2` (exit 0).
- Backend-wide grep shows no lingering references to the dropped fields, parameters, or constants.

## Test plan

- [ ] `helm install kagenti charts/kagenti --dry-run` and inspect the rendered ConfigMaps + the operator subchart.
- [ ] Bring up a kind cluster off this branch + the matching kagenti-extensions release and walk through the weather-agent demo end-to-end to confirm the new UI checkbox path produces a working pod with the correct mode.
- [ ] Run a workload with `Spec.AuthBridgeMode: envoy-sidecar` and verify the operator picks the `authbridge-envoy` image; default workload should pick `authbridge`.
- [ ] If TX e2e is exercised: confirm the `mode: envoy-sidecar` pin in `50-enable-kagenti-ns.sh` keeps the test on its existing container-name expectations.

### DIFF
--- a/charts/kagenti/templates/kagenti-authbridge-scc.yaml
+++ b/charts/kagenti/templates/kagenti-authbridge-scc.yaml
@@ -2,17 +2,22 @@
 # ——————————————————————————————————————————————————————————————————————————————
 #  Custom SCC for AuthBridge sidecars in agent namespaces.
 #
-#  The kagenti-webhook injects an AuthBridge sidecar stack into agent pods:
-#    - proxy-init (init): iptables interception (needs NET_ADMIN/NET_RAW, root, non-privileged)
-#    - envoy-proxy (sidecar): Envoy for auth token injection (requests UID 1337)
-#    - authbridge-light (sidecar): OAuth token management (no hardcoded UID)
-#    - spiffe-helper (sidecar): SPIRE identity (UBI image, no hardcoded UID)
+#  The kagenti-webhook injects an AuthBridge sidecar stack into agent pods.
+#  After kagenti-extensions#411 there's a single combined sidecar per mode
+#  (spiffe-helper bundled inside, client-registration is operator-managed):
+#    - proxy-sidecar (default): authbridge container (UID 1337)
+#    - envoy-sidecar:
+#        - proxy-init (init): iptables interception
+#          (NET_ADMIN/NET_RAW, root, non-privileged)
+#        - authbridge-envoy (sidecar): Envoy + ext_proc + spiffe-helper
+#          (UID 1337)
 #
 #  This SCC grants the minimum permissions needed for AuthBridge sidecars:
 #    - proxy-init needs root + NET_ADMIN/NET_RAW capabilities (not privileged mode)
-#    - envoy-proxy requests UID 1337 (for iptables exclusion coordination)
-#    - authbridge-light and spiffe-helper run as image-default UIDs (no override)
-#  RunAsAny is required because proxy-init needs root and envoy needs UID 1337.
+#    - the combined authbridge / authbridge-envoy sidecar runs as UID 1337
+#      (matches Envoy's traditional UID for iptables exclusion coordination
+#      and is preserved by authbridge-proxy for image consistency)
+#  RunAsAny is required because proxy-init needs root and the sidecar needs UID 1337.
 #  Bound to agent namespaces via the `groups` field below.
 # ——————————————————————————————————————————————————————————————————————————————
 apiVersion: security.openshift.io/v1
--- a/charts/kagenti/templates/kagenti-authbridge-scc.yaml
+++ b/charts/kagenti/templates/kagenti-authbridge-scc.yaml
@@ -5,12 +5,13 @@
 #  The kagenti-webhook injects an AuthBridge sidecar stack into agent pods.
 #  After kagenti-extensions#411 there's a single combined sidecar per mode
 #  (spiffe-helper bundled inside, client-registration is operator-managed):
-#    - proxy-sidecar (default): authbridge container (UID 1337)
+#    - proxy-sidecar (default): container "authbridge-proxy"
+#      from the "authbridge" image (UID 1337)
 #    - envoy-sidecar:
 #        - proxy-init (init): iptables interception
 #          (NET_ADMIN/NET_RAW, root, non-privileged)
-#        - authbridge-envoy (sidecar): Envoy + ext_proc + spiffe-helper
-#          (UID 1337)
+#        - container "envoy-proxy" from the "authbridge-envoy" image:
+#          Envoy + ext_proc + spiffe-helper (UID 1337)
 #
 #  This SCC grants the minimum permissions needed for AuthBridge sidecars:
 #    - proxy-init needs root + NET_ADMIN/NET_RAW capabilities (not privileged mode)
--- a/kagenti/backend/app/core/constants.py
+++ b/kagenti/backend/app/core/constants.py
@@ -34,10 +34,13 @@ APP_KUBERNETES_IO_COMPONENT = "app.kubernetes.io/component"
 KAGENTI_SPIRE_LABEL = "kagenti.io/spire"
 KAGENTI_SPIRE_ENABLED_VALUE = "enabled"
 
-# Per-sidecar injection labels (matched by kagenti-webhook precedence evaluator)
-KAGENTI_ENVOY_PROXY_INJECT_LABEL = "kagenti.io/envoy-proxy-inject"
-KAGENTI_SPIFFE_HELPER_INJECT_LABEL = "kagenti.io/spiffe-helper-inject"
-KAGENTI_CLIENT_REGISTRATION_INJECT_LABEL = "kagenti.io/client-registration-inject"
+# Per-sidecar injection labels are gone after kagenti-extensions#411 /
+# kagenti-operator#361: the per-sidecar split (envoy-proxy +
+# spiffe-helper + client-registration) collapsed into one combined
+# authbridge container, spiffe-helper bundled and gated by
+# SPIRE_ENABLED, client-registration operator-managed. Per-workload
+# mode selection moved to AgentRuntime.Spec.AuthBridgeMode (label-based
+# selection no longer exists).
 
 # Port exclusion annotations (matched by kagenti-webhook init-iptables.sh)
 KAGENTI_OUTBOUND_PORTS_EXCLUDE = "kagenti.io/outbound-ports-exclude"
--- a/kagenti/backend/app/core/constants.py
+++ b/kagenti/backend/app/core/constants.py
@@ -34,13 +34,11 @@ APP_KUBERNETES_IO_COMPONENT = "app.kubernetes.io/component"
 KAGENTI_SPIRE_LABEL = "kagenti.io/spire"
 KAGENTI_SPIRE_ENABLED_VALUE = "enabled"
 
-# Per-sidecar injection labels are gone after kagenti-extensions#411 /
-# kagenti-operator#361: the per-sidecar split (envoy-proxy +
-# spiffe-helper + client-registration) collapsed into one combined
-# authbridge container, spiffe-helper bundled and gated by
-# SPIRE_ENABLED, client-registration operator-managed. Per-workload
-# mode selection moved to AgentRuntime.Spec.AuthBridgeMode (label-based
-# selection no longer exists).
+# Per-sidecar opt-out labels (envoy-proxy / spiffe-helper /
+# client-registration) are gone after kagenti-extensions#411 — they
+# referenced separate sidecars that no longer exist. The master enable
+# /disable trigger is still KAGENTI_INJECT_LABEL above; per-workload
+# mode now lives on AgentRuntime.Spec.AuthBridgeMode.
 
 # Port exclusion annotations (matched by kagenti-webhook init-iptables.sh)
 KAGENTI_OUTBOUND_PORTS_EXCLUDE = "kagenti.io/outbound-ports-exclude"
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -70,10 +70,6 @@ from app.core.constants import (
     # SPIRE identity constants
     KAGENTI_SPIRE_LABEL,
     KAGENTI_SPIRE_ENABLED_VALUE,
-    # Per-sidecar injection labels
-    KAGENTI_ENVOY_PROXY_INJECT_LABEL,
-    KAGENTI_SPIFFE_HELPER_INJECT_LABEL,
-    KAGENTI_CLIENT_REGISTRATION_INJECT_LABEL,
     # Port exclusion annotations
     KAGENTI_OUTBOUND_PORTS_EXCLUDE,
     KAGENTI_INBOUND_PORTS_EXCLUDE,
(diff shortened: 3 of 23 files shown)


kagenti/backend/app/core/constants.py
```diff
@@ -34,10 +34,13 @@
...
-KAGENTI_ENVOY_PROXY_INJECT_LABEL = "kagenti.io/envoy-proxy-inject"
-KAGENTI_SPIFFE_HELPER_INJECT_LABEL = "kagenti.io/spiffe-helper-inject"
-KAGENTI_CLIENT_REGISTRATION_INJECT_LABEL = "kagenti.io/client-registration-inject"
+# Per-sidecar injection labels are gone after kagenti-extensions#411 /
+# kagenti-operator#361: the per-sidecar split (envoy-proxy +
+# spiffe-helper + client-registration) collapsed into one combined
+# authbridge container, spiffe-helper bundled and gated by
+# SPIRE_ENABLED, client-registration operator-managed. Per-workload
+# mode selection moved to AgentRuntime.Spec.AuthBridgeMode (label-based
+# selection no longer exists).
```
cwiklik: **suggestion**: This 7-line comment overstates the removal. `constants.py` still exports `KAGENTI_AUTHBRIDGE_INJECT_LABEL` (the master enable/disable label), which is the primary injection trigger. The comment reads as if all label-based injection is gone, which could mislead a future reader into thinking the master label is dead code too.

Consider trimming to 2-3 lines: what was removed (the three per-sidecar labels) and why (collapsed into the combined container).

kagenti/backend/app/routers/agents.py
```diff
@@ -244,13 +240,15 @@ class CreateAgentRequest(BaseModel):
...
-    # Per-sidecar injection controls (None = use webhook defaults)
-    envoyProxyInject: Optional[bool] = None
-    spiffeHelperInject: Optional[bool] = None
-    clientRegistrationInject: Optional[bool] = None
+    # Per-workload AuthBridge mode override. Maps to
+    # AgentRuntime.Spec.AuthBridgeMode; when None the operator falls
+    # back through namespace ConfigMap → cluster default
+    # (proxy-sidecar). The lite/waypoint shapes are accepted by the
+    # operator but not surfaced through the UI today.
+    authBridgeMode: Optional[Literal["proxy-sidecar", "envoy-sidecar", "lite", "waypoint"]] = None
```
cwiklik, line 251: **suggestion**: `Optional[Literal[...]]` — Pydantic v1's `Optional` wrapping silently downgrades `Literal` to `Optional[str]` for validation, so invalid mode strings are accepted at the API boundary and only rejected later by the operator. If you want runtime rejection, use a `StrEnum` or add a `@validator`. Not blocking since the operator catches it, but the mismatch surfaces late.

kagenti/backend/app/routers/agents.py
```diff
@@ -2448,12 +2443,23 @@ def _build_agentruntime_manifest(
     namespace: str,
     workload_type: str,
     agent_type: str = RESOURCE_TYPE_AGENT,
+    auth_bridge_mode: Optional[str] = None,
 ) -> dict:
```
cwiklik, line 2447: **suggestion**: `auth_bridge_mode: Optional[str]` bypasses the `Literal` constraint on the model. Consider defining a type alias:
```python
AuthBridgeMode = Literal["proxy-sidecar", "envoy-sidecar", "lite", "waypoint"]
```
and reusing it here and on `_ensure_agentruntime`. Low priority — the function is internal — but it documents intent and catches typos.

kagenti/tui/internal/api/types.go
```diff
@@ -204,9 +204,11 @@ type CreateAgentRequest struct {
...
 	AuthBridgeEnabled bool         `json:"authBridgeEnabled"`
 	SpireEnabled     bool          `json:"spireEnabled"`
-	EnvoyProxyInject          *bool `json:"envoyProxyInject,omitempty"`
-	SpiffeHelperInject        *bool `json:"spiffeHelperInject,omitempty"`
-	ClientRegistrationInject  *bool `json:"clientRegistrationInject,omitempty"`
+	// AuthBridgeMode maps to AgentRuntime.Spec.AuthBridgeMode for
+	// agents (the deprecated kagenti.io/authbridge-mode annotation
+	// for tools). Valid values: "proxy-sidecar" (default),
+	// "envoy-sidecar", "lite", "waypoint". Empty = cluster default.
+	AuthBridgeMode string `json:"authBridgeMode,omitempty"`
```
cwiklik, line 211: **suggestion (Go semantics)**: `AuthBridgeMode string` with `omitempty` means Go's zero value `""` is omitted from JSON, which matches the backend's `None` semantics. Good. However, Go's `string` conflates "not set" and "explicitly empty" — if the backend ever needs to distinguish them, this would need `*string`. Not blocking, just flagging the semantic gap.

charts/kagenti/templates/kagenti-authbridge-scc.yaml
```diff
@@ -2,17 +2,22 @@
...
-#    - authbridge-light (sidecar): OAuth token management (no hardcoded UID)
-#    - spiffe-helper (sidecar): SPIRE identity (UBI image, no hardcoded UID)
+#  The kagenti-webhook injects an AuthBridge sidecar stack into agent pods.
+#  After kagenti-extensions#411 there's a single combined sidecar per mode
+#  (spiffe-helper bundled inside, client-registration is operator-managed):
+#    - proxy-sidecar (default): authbridge container (UID 1337)
+#    - envoy-sidecar:
+#        - proxy-init (init): iptables interception
+#          (NET_ADMIN/NET_RAW, root, non-privileged)
+#        - authbridge-envoy (sidecar): Envoy + ext_proc + spiffe-helper
```
cwiklik: **nit**: `authbridge container` — is this the container *name* (`authbridge`) or the *image* name (`authbridge-proxy`)? Consider using the actual names to avoid ambiguity, e.g.:
```
#    - proxy-sidecar (default): authbridge container / authbridge-proxy image (UID 1337)
```

kagenti/backend/tests/test_authbridge_runtime_config.py
```diff
@@ -35,7 +35,13 @@ def test_build_authbridge_runtime_yaml_client_secret():
...
     cfg = yaml.safe_load(result)
 
-    assert cfg["mode"] == "envoy-sidecar"
+    # The backend no longer emits a top-level `mode:` — the operator
+    # layers it on per workload from
+    #   AgentRuntime.Spec.AuthBridgeMode → namespace ConfigMap →
+    #   deprecated annotation → cluster default (proxy-sidecar).
+    # Locking absence here so a future revert that re-pins
+    # mode in _build_authbridge_runtime_yaml doesn't go unnoticed.
+    assert "mode" not in cfg
```
cwiklik, line 44: **suggestion**: Good negative assertion locking the absence of `mode`. Consider also asserting `pipeline` is still present — if a future change accidentally drops both, this test would still pass.
```python
assert "mode" not in cfg
assert "pipeline" in cfg  # structural sanity
```

cwiklik (changes requested):
## Review: chore: post-architecture follow-ups for kagenti-extensions#411 / kagenti-operator#361

**Areas reviewed**: Python (backend), Helm/K8s (charts), Go (TUI), TypeScript (UI), Tests, Docs
**CI status**: passing, mergeable

Clean removal of the per-sidecar injection labels and migration to `authBridgeMode`. One must-fix below, plus suggestions.

### Must-fix

**`kagenti/backend/tests/test_sandbox_manifest.py` not updated.** This file (not in the diff) references the dropped fields at lines 26-28:
```python
req.envoyProxyInject = overrides.get("envoyProxyInject", None)
req.spiffeHelperInject = overrides.get("spiffeHelperInject", None)
req.clientRegistrationInject = overrides.get("clientRegistrationInject", None)
```
Since `CreateAgentRequest` no longer has these attributes, this test will raise `AttributeError` at runtime. Replace with `req.authBridgeMode = overrides.get("authBridgeMode", None)` and add a test case that exercises the new field.

---

Inline comments below cover suggestions and nits on the files in the diff.

huang195:
Thanks for the thorough review. Pushed `4bf26a01` addressing three of the inline suggestions; rationale for the rest below.

### Top-level "must-fix" — `test_sandbox_manifest.py`

That file does not exist in the repo. `kagenti/backend/tests/` has no `test_sandbox_manifest.py`, and a repo-wide grep for `envoyProxyInject` / `spiffeHelperInject` / `clientRegistrationInject` returns zero hits. The actual sandbox-related test file is `deployments/sandbox/tests/test_sandbox_profile.py` and it doesn't reference the dropped fields either. No change needed; please double-check the source you used to find this.

### Inline comments — addressed

| File | What changed |
|---|---|
| `constants.py` | Trimmed the 7-line comment. You're right that `KAGENTI_INJECT_LABEL = "kagenti.io/inject"` is still the master trigger — wording now reflects "only the per-sidecar opt-out labels are gone." The constant name in your comment (`KAGENTI_AUTHBRIDGE_INJECT_LABEL`) doesn't exist — the real one is `KAGENTI_INJECT_LABEL` — but the underlying point stands and is now clearer in the comment. |
| `kagenti-authbridge-scc.yaml` | Disambiguated container vs image: "container `authbridge-proxy` from the `authbridge` image" / "container `envoy-proxy` from the `authbridge-envoy` image." |
| `test_authbridge_runtime_config.py` | Added `assert "pipeline" in cfg` alongside the negative `assert "mode" not in cfg`. |

### Inline comments — skipped (with rationale)

- **`agents.py:251` `Optional[Literal[...]]` downgrade**: the comment describes Pydantic v1 behavior. `kagenti/backend/pyproject.toml` pins `pydantic>=2.9.0`; v2 enforces `Optional[Literal[...]]` correctly at the API boundary, so invalid mode strings raise `ValidationError` before reaching the operator. No silent downgrade.
- **`agents.py:2447` type alias for `auth_bridge_mode` parameter**: low priority by your own note ("function is internal"). Leaving the helper's parameter as `Optional[str]` to keep the signature simple.
- **`tui/types.go:211` `string` vs `*string`**: flagged as "not blocking" — keeping `string` with `omitempty` since the backend's `Optional[str]` serializes `None` as omission, which Go's `omitempty` matches. No current consumer needs to distinguish "not set" from "explicitly empty."

kagenti/backend/app/routers/agents.py
cwiklik, line 251: **nit (follow-up)**: `kagenti/backend/tests/test_sandbox_manifest.py` (exists on `main` via PR #1581) still sets `envoyProxyInject`, `spiffeHelperInject`, `clientRegistrationInject` on its mock (lines 26-28). Harmless dead code now that `_build_common_labels` no longer reads them, but worth removing in a cleanup pass.

cwiklik (approved):
## Re-review after fix commit 4bf26a0

**Areas re-checked**: Python (backend), Helm (SCC), Tests
**CI status**: all passing

Fix commit addresses the three suggestions cleanly (constants.py comment trimmed, SCC disambiguated, pipeline assertion added). Rationale for the skipped items is sound — Pydantic v2 does enforce `Optional[Literal[...]]` correctly.

### Original must-fix — `test_sandbox_manifest.py`

Downgrading to nit. The file *does* exist on `main` at the PR's base SHA (merged via PR #1581, commit `0687fe8d`), but my concern about breakage was wrong: `MagicMock(spec=...)` allows explicit attribute assignment (`req.envoyProxyInject = ...`) even when the attribute is absent from the spec class — it only blocks auto-created attribute *access*. Since `_build_common_labels` no longer reads those fields post-PR, the tests pass fine. The mock just carries three dead attribute assignments (lines 26-28) that could be cleaned up in a follow-up.

LGTM.

huang195:
Also fixed MLFlow E2E failure, but it's hitting another problem. That needs to be fixed separately. Merging this one

huang195:
Update on Deploy & Test: the MLflow fix from `8a9b69b4` worked — that step is past. The next failure is **AuthBridge Weather (advanced) E2E**, root cause traced to a stale `kagenti.io/client-registration-inject: "true"` label on `weather-tool-advanced.yaml` / `weather-service-advanced.yaml` in **kagenti-extensions**. The label tells the operator's `ClientRegistrationReconciler` to skip (deferring to a legacy in-pod sidecar that was removed in #411), so registration never happens at all. Same exact bug as commit 4 in this PR (`weather_service_deployment_ocp.yaml`), just in a different repo's manifests.

Filed [kagenti-extensions#414](https://github.com/kagenti/kagenti-extensions/pull/414) with the fix. This kagenti PR's CI will stay red on Deploy & Test until that lands and is consumed by the e2e workflow (which clones extensions `main`). Not a regression introduced by this PR — it was latent behind the MLflow failures, surfaced once those were fixed.

PR #1586, opened 2026-05-15
docs: :memo: Update gateway use with Sandbox

evaline-ju:
## Summary

With Sandbox enabled, there is no Deployment to update the MCP_URL for. This PR updates the directions to update the agent pod through the Sandbox spec.

### DIFF
--- a/docs/gateway.md
+++ b/docs/gateway.md
@@ -101,11 +101,11 @@ kubectl set env deployment/weather-service -n <namespace> MCP_URL="http://mcp-ga
 
 When the sandbox feature flag is on, agents run as `Sandbox` resources rather than plain `Deployment`s. The `Sandbox` spec must be updated directly — `kubectl set env` does not apply to `Sandbox` objects.
 
-Patch the `Sandbox` spec and reapply (requires `MCP_URL` to already exist in the spec env array):
+Patch the `Sandbox` spec and reapply. This requires `MCP_URL` to already exist in the env array — if it is missing, add it first via `kubectl edit sandbox weather-service -n <namespace>`.
 
 ```bash
 kubectl get sandbox weather-service -n <namespace> -o json \
-  | jq '(.spec.podTemplate.spec.containers[0].env[] | select(.name == "MCP_URL")).value = "http://mcp-gateway-istio.gateway-system.svc.cluster.local:8080/mcp"' \
+  | jq '(.spec.podTemplate.spec.containers[] | select(.name == "agent").env[] | select(.name == "MCP_URL")).value = "http://mcp-gateway-istio.gateway-system.svc.cluster.local:8080/mcp"' \
   | kubectl apply -f -
 ```
 
--- a/docs/gateway.md
+++ b/docs/gateway.md
@@ -89,9 +89,33 @@ However, we need to define a new environment variable so the Agent can access
 various tools managed by the Gateway. Namely, we need to set `MCP_URL` to
 `http://mcp-gateway-istio.gateway-system.svc.cluster.local:8080/mcp`.
 
-The `weather-service` deployment can be edited manually or patched via a command like `kubectl set env deployment/weather-service -n default MCP_URL="http://mcp-gateway-istio.gateway-system.svc.cluster.local:8080/mcp"`, with the namespace adjusted as appropriate.
+#### Deployments (non-sandbox)
 
-Once the Gateway implementation has stabilized, `MCP_URL` can be set to this
+If the agent is running as a standard Kubernetes `Deployment`, patch the env var directly:
+
+```bash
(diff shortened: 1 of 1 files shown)


docs/gateway.md
```diff
@@ -89,9 +89,33 @@ However, we need to define a new environment variable so the Agent can access
...
+```bash
+kubectl set env deployment/weather-service -n <namespace> MCP_URL="http://mcp-gateway-istio.gateway-system.svc.cluster.local:8080/mcp"
+```
+
+#### Sandboxes (when sandbox feature flag is enabled)
+
+When the sandbox feature flag is on, agents run as `Sandbox` resources rather than plain `Deployment`s. The `Sandbox` spec must be updated directly — `kubectl set env` does not apply to `Sandbox` objects.
+
+Patch the `Sandbox` spec and reapply (requires `MCP_URL` to already exist in the spec env array):
+
```
huang195, line 105: **Suggestion (nit)**: "requires `MCP_URL` to already exist in the spec env array" — for a reader landing on this page cold, it'd help to add a one-line "if not present, edit the Sandbox spec via `kubectl edit sandbox …` or set it at import time" so they don't get a silently failing patch. The jq `select(.name == "MCP_URL")` produces no match (and no error) when the entry is missing — the pipeline succeeds with the original env unchanged.

docs/gateway.md
```diff
@@ -89,9 +89,33 @@ However, we need to define a new environment variable so the Agent can access
...
+```bash
+kubectl get sandbox weather-service -n <namespace> -o json \
+  | jq '(.spec.podTemplate.spec.containers[0].env[] | select(.name == "MCP_URL")).value = "http://mcp-gateway-istio.gateway-system.svc.cluster.local:8080/mcp"' \
+  | kubectl apply -f -
```
huang195, line 109: **Suggestion**: The jq expression assumes the agent is `containers[0]`. If a kagenti sandbox template ever has multiple user-defined containers (or a future revision injects something ahead of the agent into `podTemplate.spec.containers[]`), this silently mutates the wrong container's env — no error, just a confused reader.

Safer: select by container name, e.g.

```
jq '(.spec.podTemplate.spec.containers[] | select(.name == "weather-service").env[] | select(.name == "MCP_URL")).value = "…"'
```

docs/gateway.md
```diff
@@ -89,9 +89,33 @@ However, we need to define a new environment variable so the Agent can access
...
+```
+
+**Note:** Due to a known upstream limitation ([kubernetes-sigs/agent-sandbox#581](https://github.com/kubernetes-sigs/agent-sandbox/issues/581)), updating the `Sandbox` spec does not automatically restart running pods. After applying the spec change, manually delete the pod to force a restart:
+
+```bash
+kubectl delete pod -n <namespace> -l app.kubernetes.io/name=weather-service
+```
```
huang195, line 116: **Suggestion**: The pod selector `app.kubernetes.io/name=weather-service` assumes the sandbox controller propagates that label to the pod. Worth verifying once against an actual running sandbox pod — if the controller uses a different label (e.g. `agent-sandbox.x-k8s.io/sandbox-name=…`), this `kubectl delete pod -l …` is a silent no-op and the user concludes the docs are wrong. If unsure, fall back to `kubectl get pods -n <ns>` then `kubectl delete pod <name>`.

huang195 (approved):
## Summary

Useful docs split for `docs/gateway.md`: the existing `kubectl set env deployment/...` instructions get separated from a new procedure for `Sandbox` resources (jq-patch + manual pod restart, with a link to upstream agent-sandbox#581). Without this, users on the sandbox feature flag hit a non-actionable instruction. No must-fix issues — three suggestions on the patch commands' robustness, plus a PR-title note below.

**Areas reviewed**: Docs only. **CI**: all green. **Commit**: signed off.

### General comment (PR title)

Title is `docs: :memo: …`. The PR template embedded in the body says prefixes need to be Capitalized from a fixed list (`Docs`, not `docs`). I don't see a title-rule workflow in the checks output, so this may not be enforced today — but if Repository Rulesets enforce it, this title would be blocked. Also `docs:` + `:memo:` is doubly-prefixed; one or the other reads cleaner. Leaving as a nit; merge as-is is fine if title rules aren't enforced.

### Verdict

APPROVE — suggestions are about making the demo commands more robust against label/container-index assumptions, not blockers.

Assisted-By: Claude Code

docs/gateway.md
evaline-ju: ✔️ this matches what I have running in my kind cluster

docs/gateway.md
evaline-ju: sandbox makes the container name 'agent' so this has been updated

PR #1587, opened 2026-05-15
:sparkles: feat: add PVC support for Sandbox workloads

### LINKED ISSUE #1585: feat: support PersistentVolumeClaims for Sandbox workloads
## Summary

Sandbox workloads currently use only `emptyDir` volumes — all data is lost when the pod restarts. The upstream Sandbox CRD already supports `spec.volumeClaimTemplates` and the agent-sandbox controller handles PVC provisioning, but kagenti's backend doesn't populate that field.

## Motivation

Agents that need to persist state across restarts (caches, checkpoints, workspace files) have no durable storage option in sandbox mode today.

## What needs to change

1. **API model** — Add optional storage config to `CreateAgentRequest` (size, storageClass, mountPath)
2. **Backend manifest builder** — `_build_sandbox_manifest()` (`agents.py:2858`) needs to populate `spec.volumeClaimTemplates` and add matching `volumeMounts` to the container
3. **UI** — Optional form fields to configure persistent storage when creating a sandbox agent
4. **RBAC** — Verify the backend ServiceAccount can create PVCs in agent namespaces
5. **Cleanup** — PVCs from `volumeClaimTemplates` survive Sandbox deletion (same as StatefulSet). The delete flow (`agents.py:936`) needs to clean up orphaned PVCs

## Suggested approach

Start with steps 2 and 5 only — hardcode a default PVC template behind a feature flag, handle cleanup on delete. No UI or API model changes needed for a first pass.

## References

- Sandbox CRD `spec.volumeClaimTemplates`: `deployments/openshell/agent-sandbox-crd.yaml`
- Sandbox controller `reconcilePVCs()`: upstream `sandbox_controller.go:782`
- Current manifest builder: `kagenti/backend/app/routers/agents.py:2858`
- Current delete flow: `kagenti/backend/app/routers/agents.py:936`

cwiklik:
## Summary

- Add persistent storage (PVC) support for Sandbox workloads end-to-end: backend model, manifest builder, PVC cleanup on delete, UI form field, TUI types, and tests
- When enabled, replaces the `shared-data` emptyDir with a `volumeClaimTemplate` so data survives pod restarts
- Follows the existing StatefulSet PVC pattern from tools.py

## Details

**Backend** (`agents.py`, `kubernetes.py`):
- `PersistentStorageConfig` model with `enabled` (default false) and `size` (default 1Gi)
- `_build_sandbox_manifest()` emits `spec.volumeClaimTemplates` when enabled
- Agent delete flow cleans up PVCs via label-selector lookup
- New `list_persistent_volume_claims` / `delete_persistent_volume_claim` on KubernetesService

**UI** (`ImportAgentPage.tsx`, `api.ts`):
- "Persistent Volume Size" text input shown when workload type is `sandbox` or `statefulset`

**TUI** (`types.go`):
- `PersistentStorageConfig` struct and field on `CreateAgentRequest`

**Tests** (`test_sandbox_manifest.py`):
- 6 new tests: no PVC by default, PVC when enabled, emptyDir replacement, volume mount unchanged, disabled keeps emptyDir, default size

## Test plan

- [x] `uv run pytest tests/test_sandbox_manifest.py -v` — 13 tests passing
- [x] `ruff check` and `ruff format` clean
- [x] Local Kind cluster: deployed Sandbox agent with PVC via UI, confirmed `shared-data-<name>` PVC bound and pod running 3/3
- [ ] Verify PVC cleanup on agent deletion

Closes #1585

Assisted-By: Claude Code

### DIFF
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -204,6 +204,13 @@ class ServicePort(BaseModel):
     protocol: str = "TCP"
 
 
+class PersistentStorageConfig(BaseModel):
+    """Persistent storage configuration for Sandbox and StatefulSet agents."""
+
+    enabled: bool = False
+    size: str = "1Gi"
+
+
 class CreateAgentRequest(BaseModel):
     """Request to create a new agent."""
 
@@ -259,6 +266,9 @@ class CreateAgentRequest(BaseModel):
 
     # Outbound routing rules (authproxy-routes ConfigMap)
     outboundRoutes: Optional[List["OutboundRoute"]] = None
+
+    # Persistent storage (for Sandbox and StatefulSet workloads)
+    persistentStorage: Optional[PersistentStorageConfig] = None
 
     # Shipwright build configuration
     shipwrightConfig: Optional[ShipwrightBuildConfig] = None
@@ -931,7 +941,7 @@ async def delete_agent(
         else:
             logger.warning("Failed to delete Job '%s': %s", safe_name, e.reason)
 
-    # Delete the Sandbox (if exists)
+    # Delete the Sandbox (if exists) and its PVCs
     if settings.kagenti_feature_flag_agent_sandbox:
         try:
             kube.delete_sandbox(namespace=namespace, name=name)
@@ -941,6 +951,18 @@ async def delete_agent(
                 logger.debug("Sandbox '%s' not found (may be other workload type)", safe_name)
             else:
                 logger.warning("Failed to delete Sandbox '%s': %s", safe_name, e.reason)
+
+        try:
+            pvcs = kube.list_persistent_volume_claims(
+                namespace=namespace,
+                label_selector=f"app.kubernetes.io/name={name}",
+            )
+            for pvc_name in pvcs:
+                kube.delete_persistent_volume_claim(namespace=namespace, name=pvc_name)
+                messages.append(f"PVC '{pvc_name}' deleted")
+        except ApiException as e:
+            if e.status != 404:
+                logger.warning("Failed to clean up PVCs for '%s': %s", safe_name, e.reason)
 
     # Delete the Service
     try:
@@ -2286,6 +2308,8 @@ def _build_agent_shipwright_build_manifest(
         resource_config["inboundPortsExclude"] = request.inboundPortsExclude
     if request.defaultOutboundPolicy:
         resource_config["defaultOutboundPolicy"] = request.defaultOutboundPolicy
+    if request.persistentStorage:
+        resource_config["persistentStorage"] = request.persistentStorage.model_dump()
     # Add env vars if present
     if request.envVars:
         resource_config["envVars"] = [ev.model_dump(exclude_none=True) for ev in request.envVars]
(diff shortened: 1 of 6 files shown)


kagenti/backend/app/routers/agents.py
```diff
@@ -2922,13 +2946,28 @@ def _build_sandbox_manifest(
...
+                    ),
                 },
             },
         },
     }
 
+    if request.persistentStorage and request.persistentStorage.enabled:
+        manifest["spec"]["volumeClaimTemplates"] = [
+            {
+                "metadata": {"name": "shared-data"},
```
pdettori: **must-fix**: VCT metadata has no labels, but the PVC cleanup in `delete_agent` queries by `label_selector=f"app.kubernetes.io/name={name}"`. Kubernetes StatefulSet PVCs inherit labels from VCT metadata, not the parent StatefulSet — so created PVCs won't have that label and cleanup will silently return an empty list, leaving orphaned PVCs.

Fix:
```python
"metadata": {
    "name": "shared-data",
    "labels": {"app.kubernetes.io/name": request.name},
},
```

Note: the existing `tools.py` VCT pattern also omits labels, but it doesn't have PVC cleanup code so it's not broken there.

kagenti/backend/tests/test_sandbox_manifest.py
```diff
@@ -82,6 +83,85 @@ def test_custom_service_ports_override_container_port(self):
...
+        """volumeClaimTemplates present with correct size when persistentStorage enabled."""
+        from app.routers.agents import _build_sandbox_manifest
+
+        storage = PersistentStorageConfig(enabled=True, size="5Gi")
+        request = _make_request(persistentStorage=storage)
+        manifest = _build_sandbox_manifest(request=request, image="test:latest")
+
+        vct = manifest["spec"]["volumeClaimTemplates"]
+        assert len(vct) == 1
+        assert vct[0]["metadata"]["name"] == "shared-data"
```
pdettori, line 110: **suggestion**: This asserts the VCT metadata `name` but not the labels. Since `delete_agent` depends on `app.kubernetes.io/name` being present on PVCs (which inherit VCT metadata labels), add an assertion like:
```python
assert vct[0]["metadata"]["labels"]["app.kubernetes.io/name"] == "test-agent"
```
This would have caught the current label gap and will prevent regressions.

kagenti/ui-v2/src/pages/ImportAgentPage.tsx
```diff
@@ -1025,6 +1036,24 @@ export const ImportAgentPage: React.FC = () => {
                 </FormHelperText>
               </FormGroup>
 
+              {(workloadType === 'sandbox' || workloadType === 'statefulset') && (
+                <FormGroup label="Persistent Volume Size" fieldId="persistentStorageSize">
+                  <TextInput
```
pdettori: **suggestion**: No input validation on the storage size — values like `"foo"` or `""` will reach the K8s API and fail with an opaque error. A regex like `/^\d+(Gi|Mi|Ki|Ti|G|M|K|T)$/` on the frontend (or a `@validator` on the Pydantic model) would catch invalid values earlier.

kagenti/ui-v2/src/pages/ImportAgentPage.tsx
```diff
@@ -505,6 +508,10 @@ export const ImportAgentPage: React.FC = () => {
         outboundPortsExclude: authBridgeEnabled && outboundPortsExclude ? outboundPortsExclude : undefined,
         inboundPortsExclude: authBridgeEnabled && inboundPortsExclude ? inboundPortsExclude : undefined,
         defaultOutboundPolicy: authBridgeEnabled && defaultOutboundPolicy ? defaultOutboundPolicy : undefined,
+        // Persistent storage (Sandbox / StatefulSet)
+        persistentStorage: (workloadType === 'sandbox' || workloadType === 'statefulset')
```
pdettori: **suggestion**: The UI always sends `enabled: true` for sandbox/statefulset — there's no opt-out. Users who want ephemeral sandboxes (emptyDir) can't get them through the UI. Consider adding an "Enable Persistent Storage" checkbox; the backend model already supports `enabled: false`.

pdettori (changes requested):
Well-structured feature — adds PVC support end-to-end across backend, UI, TUI, and tests. The Shipwright build finalize path correctly round-trips the config. Types are consistent across Python/Go/TypeScript. CI is green (Deploy & Test still pending).

One blocking issue: VolumeClaimTemplate metadata is missing the `app.kubernetes.io/name` label that the PVC cleanup code depends on, which will silently leave orphaned PVCs on agent deletion.

**Areas reviewed**: Python (backend models, manifest builder, delete flow, K8s service), TypeScript (UI form, API types), Go (TUI types), Tests
**Commits**: 1, signed-off ✓
**CI**: 17 pass, 1 pending (Deploy & Test), 1 manual (E2E HyperShift)

pdettori (approved):
LGTM

pdettori:
need rebase

PR #1589, opened 2026-05-15
fix(ocp): guard Istio CRDs in Helm templates and add post-CRD reconciliation

pdettori:
## Summary

- **Fix fresh-install failure**: `helm install kagenti-deps` fails on a new OpenShift cluster because `AuthorizationPolicy` (security.istio.io/v1) is rendered before the Sail operator registers its CRDs. Added `.Capabilities.APIVersions.Has` guards to `otel-collector.yaml` and `istio-telemetry.yaml`.
- **Post-CRD reconciliation**: Added a `helm upgrade` pass after Step 3b (Istio + cert-manager ready) so CRD-gated resources are rendered in a single script run — no manual steps or re-runs needed.
- **Helm flag dedup**: Extracted common `--set` flags into `KAGENTI_DEPS_HELM_ARGS` array (was duplicated 3x).
- **MLflow preflight fix**: Guard against missing DSC instance (not just missing CRD) — avoids hard exit when RHOAI operator is installed but no `default-dsc` exists.
- **Agent sandbox support**: Added `--with-agent-sandbox` flag (included in `--with-all`) to install the `kubernetes-sigs/agent-sandbox` controller and set `featureFlags.agentSandbox` in the kagenti Helm chart.
- **UX**: Show `show-services.sh` hint in the final banner.

## Test plan

- [x] Fresh install on a clean OpenShift cluster: `./scripts/ocp/setup-kagenti.sh --kagenti-repo $(pwd) --with-builds`
- [x] Verify `AuthorizationPolicy` exists after install: `kubectl get authorizationpolicies -n kagenti-system`
- [x] Re-run on existing cluster (upgrade path) succeeds without errors
- [x] `--with-agent-sandbox` installs CRD and controller
- [x] Cluster without RHOAI DSC instance skips MLflow gracefully (no hard exit)

Assisted-By: Claude Code


scripts/ocp/setup-kagenti.sh
```diff
@@ -59,6 +60,8 @@ DRY_RUN=false
 WITH_KIALI=false
 WITH_BUILDS=false
```
cwiklik, line 61: **suggestion**: `AGENT_SANDBOX_VERSION="v0.4.3"` — the latest release is [v0.4.6](https://github.com/kubernetes-sigs/agent-sandbox/releases/tag/v0.4.6). Worth bumping unless there's a specific reason to pin to v0.4.3 (known regression, API break, etc.).

cwiklik (approved):
Clean fix for the fresh-install CRD ordering issue on OpenShift. The `.Capabilities.APIVersions.Has` guards and post-CRD reconciliation `helm upgrade` are the correct Helm pattern for resources that depend on operator-installed CRDs. Helm flag dedup into `KAGENTI_DEPS_HELM_ARGS` is a nice cleanup. MLflow DSC guard, agent-sandbox support, and `_adopt_for_helm` calls for cert-manager resources are all sound.

**Areas reviewed**: Helm/K8s (templates), Shell (installer), Security
**CI status**: all passing (Deploy & Test pending)

PR #1591, opened 2026-05-15
:bug: Show correct pod status for Sandbox workloads

### LINKED ISSUE #1590: :bug: No ready replicas on agents with sandbox workloads
### Describe the bug

Image

### Steps To Reproduce

1. Ran install on `scripts/kind/setup-kagenti.sh --with-all` at v0.6.0-rc.4
2. Deployed weather agent from source, using ollama settings

### Expected Behavior

No replicaset for the sandbox creating pods should either not refer to replicas, or provide the ready 'pods'

cwiklik:
## Summary

- Fix Sandbox agents always showing `0/1 ready` on the detail page regardless of actual health
- Show "Pods: 1/1 ready" for Sandbox (derived from Ready condition), hide replica row for Jobs, keep "Replicas" for Deployment/StatefulSet

## Problem

Sandbox CRs don't have `readyReplicas` in their status — only `replicas: 1` and a `conditions` array. The UI unconditionally read `status.readyReplicas` (defaulting to 0) and `spec.replicas` (defaulting to 1), always displaying `0/1 ready`.

## Test plan

- [ ] Sandbox agent detail page shows `1/1 pods ready` when pod is running
- [ ] Deployment agent detail page still shows `N/N ready` with "Replicas" label
- [ ] Job agent detail page does not show a replica/pod row

Fixes #1590

Assisted-By: Claude Code

### DIFF
--- a/kagenti/ui-v2/src/pages/AgentDetailPage.tsx
+++ b/kagenti/ui-v2/src/pages/AgentDetailPage.tsx
@@ -481,18 +481,27 @@ export const AgentDetailPage: React.FC = () => {
                           </Label>
                         </DescriptionListDescription>
                       </DescriptionListGroup>
-                      <DescriptionListGroup>
-                        <DescriptionListTerm>Replicas</DescriptionListTerm>
-                        <DescriptionListDescription>
-                          {readyReplicas}/{replicas} ready
-                          {availableReplicas > 0 && ` (${availableReplicas} available)`}
-                          {workloadType === 'statefulset' && updatedReplicas < replicas && (
-                            <Label color="blue" isCompact style={{ marginLeft: 8 }}>
-                              {updatedReplicas}/{replicas} updated
-                            </Label>
-                          )}
-                        </DescriptionListDescription>
-                      </DescriptionListGroup>
+                      {workloadType === 'sandbox' ? (
+                        <DescriptionListGroup>
+                          <DescriptionListTerm>Pods</DescriptionListTerm>
+                          <DescriptionListDescription>
+                            {isReady ? (status.replicas ?? 1) : 0}/{status.replicas ?? 1} ready
+                          </DescriptionListDescription>
+                        </DescriptionListGroup>
+                      ) : workloadType !== 'job' && (
+                        <DescriptionListGroup>
+                          <DescriptionListTerm>Replicas</DescriptionListTerm>
+                          <DescriptionListDescription>
+                            {readyReplicas}/{replicas} ready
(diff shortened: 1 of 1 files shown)


kagenti/ui-v2/src/pages/AgentDetailPage.tsx
```diff
@@ -481,18 +481,27 @@ export const AgentDetailPage: React.FC = () => {
...
-                              {updatedReplicas}/{replicas} updated
-                            </Label>
-                          )}
-                        </DescriptionListDescription>
-                      </DescriptionListGroup>
+                      {workloadType === 'sandbox' ? (
+                        <DescriptionListGroup>
+                          <DescriptionListTerm>Pods</DescriptionListTerm>
+                          <DescriptionListDescription>
+                            {isReady ? (status.replicas ?? 1) : 0}/{status.replicas ?? 1} ready
```
huang195, line 488: `isReady` is also `true` while `readyStatus === 'Progressing'` (see lines 326-329 on main: `isRunningOrReady = readyStatus === 'Ready' || readyStatus === 'Progressing'`). For a Sandbox that's still starting up, the page would show "1/1 pods ready" before the pod is actually Ready.

A stricter signal — just `conditions.some(c => (c.type === 'Ready' || c.type === 'Available') && c.status === 'True')` — would match real pod readiness. Not a regression vs. the existing Ready label (which has the same Progressing-counts-as-ready semantics), so non-blocking.

huang195 (approved):
Narrow UI fix for #1590. Three-way render branch on `workloadType` (sandbox / job / deployment+statefulset) is the right shape — variables it relies on (`isReady`, `status.replicas`, `workloadType`) are all defined upstream, no new state added.

**Areas reviewed**: Frontend (1 TSX file), kagenti workload-type semantics
**Commits**: 1, signed-off (DCO passing)
**CI**: 21 passing, `Deploy & Test` pending, `E2E HyperShift` gated on `/run-e2e`

One suggestion inline. One nit: PR title is `:bug: fix: ...` — kagenti convention (`repo:pr` / `git:commit`) is `<emoji> <description>`, no `fix:` prefix needed. Drop one.

Dobra robota.

kagenti/ui-v2/src/pages/AgentDetailPage.tsx
cwiklik: Fixed in 5f4fdfc — Sandbox pod count now checks conditions directly for Ready/Available, so it shows 0/1 during startup instead of 1/1.

PR #1593, opened 2026-05-16
fix(backend): create Service for Sandbox in source-build path too

huang195:
## Summary

Commit [`c2e656fc`](../commit/c2e656fc) (introduced by [#1581](../pull/1581), May 14) fixed Sandbox-Service creation in the **image-based** agent-deploy path so the operator's `AgentCardReconciler` can resolve a Service and fetch the card. That fix was applied to one of the two agent-creation code paths in `agents.py`. The other one — the **source-build / Shipwright finalize** flow at line ~3605 — was missed.

The missed gate still reads:

```python
if final_workload_type not in (WORKLOAD_TYPE_JOB, WORKLOAD_TYPE_SANDBOX):
    # … create Service
```

So when a user imports an agent that's **built from source** AND the workload type resolves to `sandbox`, no Service is created. The agent ends up with:

- A `Sandbox` CR, pod running with the AuthBridge sidecar injected, `.well-known/agent-card.json` reachable from inside the pod
- A `kagenti-operator`-managed `AgentCard` CR stuck on `Synced=False`, `reason=ServiceNotFound`:
  ```
  failed to get service <name>: Service "<name>" not found
  ```
- In the kagenti UI: agent card never loads

## Reproduction

Set up a kind cluster with `setup-kagenti.sh --with-ui --with-spire --with-agent-sandbox --with-builds`, then import an agent through the UI's "Import from source" flow. The Sandbox CR + pod are created, but no Service. `kubectl get agentcard -n team1` shows `Synced=False` indefinitely.

Manually applying a Service with the matching selector clears the AgentCardReconciler within one reconcile cycle (~30s).

### 1. Apply the same fix to the source-build path

Mirror `#1581`'s pattern from the image path to the source-build path:

| Change | Why |
|---|---|
| Drop `WORKLOAD_TYPE_SANDBOX` from the exclusion set | Service is needed for the operator's agent-card discovery |
| Wrap `create_service` in `try/except`, handle 409 by `delete_service` + retry | Sandbox controller can create a short-lived Service ahead of us; replace with backend-managed ClusterIP shape, same as the image-path's `#1581` handler |
| Update the comment to match line 3069 (`"Sandbox uses backend-managed Service"`) | Old comment was `"not needed for Jobs or Sandboxes"` — pre-`#1581` assumption |

### 2. Extract a shared helper + replace the tautological test

After commit 1 the two paths had **exact** duplicated try/except/log boilerplate. Extract `_create_or_replace_service(kube, namespace, name, manifest, workload_type)` next to `_build_service_manifest`; both call sites collapse to two-line invocations. Removes ~30 lines of duplication and ensures the paths can't drift again.

The pre-existing test `test_sandbox_not_excluded_from_service_creation` (added by #1581) was tautological — it asserted `WORKLOAD_TYPE_SANDBOX not in {WORKLOAD_TYPE_JOB}`, pure set theory, no application code. It would not have caught the source-build regression this PR fixes. Replace with a real test class exercising the helper with `MagicMock(KubernetesService)`:

- `test_job_workload_skips_service_creation`
- `test_sandbox_creates_service` ← would have caught the original `#1581`-era bug
- `test_deployment_creates_service`
- `test_sandbox_409_replaces_existing_service` ← Sandbox controller race recovery
- `test_deployment_409_propagates` ← Deployment must NOT silently replace
- `test_non_409_propagates_for_sandbox` ← only 409 triggers replace

## Test plan

- [x] `python3 -c "import ast; ast.parse(...)"` on the modified module + test file
- [x] Test discovery names 6 new functions in `TestCreateOrReplaceService`
- [ ] CI runs `pytest` against the test file (the new `_create_or_replace_service` helper is exercised end-to-end via the mocked `KubernetesService`)
- [ ] Re-run the source-build import flow on a kind cluster; verify Service is created and AgentCard reaches `Synced=True`

## Review thread addressed

- ✅ #1 — Corrected `kagenti#1424` → `kagenti#1581` in commit message + this PR body
- ⏸ #2 — Reviewer marked non-blocking; mirroring the image path's identical behavior is preferable to one-sided hardening
- ✅ #3 — Tautological test replaced with 6 real tests covering the helper's behavior matrix; the new test class would have caught the regression that motivated this PR

### DIFF
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -2627,22 +2627,25 @@ def _create_or_replace_service(
     """
     if workload_type == WORKLOAD_TYPE_JOB:
         return
-    # name/namespace are DNS-1123 (validated by Kubernetes API) so no real
-    # log-injection risk, but use lazy %-formatting to satisfy CodeQL's
-    # py/log-injection rule and match Python logging best practice.
+    # Strip CR/LF before logging — name and namespace come from the FastAPI
+    # request body. Kubernetes will reject non-DNS-1123 names so this is
+    # belt-and-suspenders, but the explicit sanitization satisfies CodeQL's
+    # py/log-injection taint analysis on the user-input → log-sink flow.
+    safe_name = name.replace("\n", "").replace("\r", "")
+    safe_namespace = namespace.replace("\n", "").replace("\r", "")
     try:
         kube.create_service(namespace=namespace, body=service_manifest)
-        logger.info("Created Service '%s' in namespace '%s'", name, namespace)
+        logger.info("Created Service '%s' in namespace '%s'", safe_name, safe_namespace)
     except ApiException as e:
         if e.status == 409 and workload_type == WORKLOAD_TYPE_SANDBOX:
             logger.warning(
                 "Service '%s' already exists (Sandbox controller race); "
                 "replacing with backend-managed ClusterIP Service",
-                name,
+                safe_name,
             )
             kube.delete_service(namespace=namespace, name=name)
             kube.create_service(namespace=namespace, body=service_manifest)
-            logger.info("Replaced Service '%s' in namespace '%s'", name, namespace)
+            logger.info("Replaced Service '%s' in namespace '%s'", safe_name, safe_namespace)
         else:
             raise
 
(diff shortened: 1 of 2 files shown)


esnible (approved):
Verified the bug and fix against `upstream/main`: line 3605 has the asymmetric gate (`final_workload_type not in (WORKLOAD_TYPE_JOB, WORKLOAD_TYPE_SANDBOX)`), while the image path's #1581 gate at line 3069 only excludes Job. Confirmed.

**What I checked**
- ✅ Both call sites collapse to two-line invocations of `_create_or_replace_service`; ~30 lines of duplication removed.
- ✅ Helper preserves all behavior from #1581: Job → skip, Sandbox 409 → delete+recreate, Deployment 409 → propagate, non-409 → propagate.
- ✅ New test class is non-tautological. `test_sandbox_creates_service` would have caught both the original pre-#1581 bug and the source-build regression this PR fixes.
- ✅ CodeQL log-injection mitigation (CR/LF strip on `name` / `namespace` before logging) is correct — Kubernetes DNS-1123 rejection makes this belt-and-suspenders, but explicit sanitization satisfies the taint analyzer.
- ✅ Adjacent gates correctly untouched: AgentRuntime CR (line 3094 / 3614) and HTTPRoute (line 3105 / 3631) still exclude Sandbox, as they should.
- ✅ All 5 commits signed-off, conventional emoji prefixes, imperative mood.
- ✅ All 22 CI checks pass (pending E2E HyperShift waits for `/run-e2e`).

Areas reviewed: Python (helper + tests), test design.

Assisted-By: Claude Code

PR #1596, opened 2026-05-17
fix(test): retry agent conversation on LLM connection errors

Ladas:
## Summary

The weather agent returns "LLM execution failed: Connection error" when the external MaaS endpoint (`litellm-prod.apps.maas.redhatworkshops.io`) is transiently unreachable from ephemeral HyperShift clusters.

The test already has retry logic for transient errors but the exact string "Connection error" (from `openai.APIConnectionError`) wasn't in the retry list — only "Cannot connect" was.

- Add "Connection error" to `_TRANSIENT_ERRORS` in `test_agent_simple_query`
- Add "Connection error" to the retry check in `test_agent_multiturn_conversation`

## Root Cause

PR #1527's green HyperShift run had these tests passing. Subsequent main branch runs fail because the ephemeral cluster intermittently can't reach the external MaaS LLM endpoint. The agent correctly reports the connection failure, but the test treats it as a hard failure instead of retrying.

## Test plan

- [ ] `/run-e2e` — HyperShift E2E passes with LLM retry on connection errors

### DIFF
--- a/.github/scripts/hypershift/ci/55-cleanup-existing-cluster.sh
+++ b/.github/scripts/hypershift/ci/55-cleanup-existing-cluster.sh
@@ -313,7 +313,7 @@ cleanup_orphaned_aws_resources() {
                 fi
 
                 echo "  - Attempting manual VPC delete:"
-                VPC_DELETE_OUTPUT=$(aws ec2 delete-vpc --region "$AWS_REGION" --vpc-id "$vpc" 2>&1)
+                VPC_DELETE_OUTPUT=$(aws ec2 delete-vpc --region "$AWS_REGION" --vpc-id "$vpc" 2>&1) || true
                 VPC_DELETE_EXIT=$?
 
                 if [ $VPC_DELETE_EXIT -eq 0 ]; then
--- a/.github/scripts/hypershift/ci/55-cleanup-existing-cluster.sh
+++ b/.github/scripts/hypershift/ci/55-cleanup-existing-cluster.sh
@@ -558,7 +558,10 @@ if [ "$HC_EXISTS" = "true" ] || [ "$NS_EXISTS" = "true" ]; then
 
     # Verify HostedCluster cleanup completed
     if oc get hostedcluster "$CLUSTER_NAME" -n clusters &>/dev/null; then
-        echo "::warning::HostedCluster still exists after cleanup"
+        echo "::warning::HostedCluster still exists after cleanup — force-deleting"
+        oc patch hostedcluster "$CLUSTER_NAME" -n clusters -p '{"metadata":{"finalizers":null}}' --type=merge 2>/dev/null || true
+        oc delete hostedcluster "$CLUSTER_NAME" -n clusters --wait=false 2>/dev/null || true
+        sleep 10
     fi
 
     if oc get ns "$CONTROL_PLANE_NS" &>/dev/null; then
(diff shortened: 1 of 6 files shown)


Ladas:
## Blocked: stale HostedCluster `pr1596` on CI management cluster

The HyperShift E2E is blocked by a stale `pr1596` HostedCluster that can't be deleted:
- Cleanup script runs but reports "HostedCluster still exists after cleanup"
- Manually triggered `Cleanup Stale HyperShift Clusters` workflow — succeeds in 14s but doesn't force-delete
- Likely stuck finalizers on the HostedCluster resource

**Need manual intervention** on the CI management cluster:
```bash
# Remove finalizers and delete
kubectl patch hostedcluster <MANAGED_BY_TAG>-pr1596 -n clusters --type=merge -p '{"metadata":{"finalizers":null}}'
kubectl delete hostedcluster <MANAGED_BY_TAG>-pr1596 -n clusters
```

The actual code changes (proxy-init `OUTBOUND_PORTS_EXCLUDE=8080,443` workaround) haven't been tested yet because cluster creation is blocked.

.github/scripts/hypershift/ci/55-cleanup-existing-cluster.sh
```diff
@@ -313,7 +313,7 @@ cleanup_orphaned_aws_resources() {
                 fi
 
                 echo "  - Attempting manual VPC delete:"
-                VPC_DELETE_OUTPUT=$(aws ec2 delete-vpc --region "$AWS_REGION" --vpc-id "$vpc" 2>&1)
+                VPC_DELETE_OUTPUT=$(aws ec2 delete-vpc --region "$AWS_REGION" --vpc-id "$vpc" 2>&1) || true
```
pdettori, line 316: nit: After `|| true`, `VPC_DELETE_EXIT=$?` will always be `0` — the subsequent success/failure branch on line 318 is now dead code. Fine if the intent is "always continue", but you could simplify by removing the exit-code check entirely.

.github/scripts/kagenti-operator/74-deploy-weather-agent.sh
```diff
@@ -205,5 +215,63 @@ EOF
             kubectl describe pods -n team1 -l app.kubernetes.io/name=weather-service 2>&1 | tail -30 || true
             exit 1
         fi
+
+        # Diagnostic: verify LLM endpoint is reachable from inside the pod
```
pdettori, line 219: suggestion: The inline Python diagnostic is ~45 lines embedded in a heredoc. Consider extracting to a standalone script (e.g. `scripts/diagnostics/test_llm_connectivity.py`) or at minimum adding a comment noting when this can be removed (once the connectivity issue is fully resolved upstream).

kagenti/tests/e2e/common/test_agent_conversation.py
```diff
@@ -465,12 +464,13 @@ async def test_agent_multiturn_conversation(
 
                 if last_result["task_failed"]:
                     error_text = last_result["full_response"][:_DIAG_ERROR_LIMIT]
-                    if (
-                        "Cannot connect" in error_text
-                        and attempt < _LLM_QUERY_MAX_ATTEMPTS
-                    ):
+                    is_transient = any(
```
pdettori, line 467: nit: The multiturn test uses a local tuple `("Cannot connect", "Connection error")` while the simple-query test reuses the module-level `_TRANSIENT_ERRORS` constant. Consider reusing `_TRANSIENT_ERRORS` here too for consistency.

pdettori (approved):
Solid debugging journey. The PR correctly identifies and fixes the root cause (authbridge `HTTPS_PROXY` blocking external LLM egress) and adds useful diagnostics. CI fixes for stuck HostedCluster cleanup and VPC deletion are straightforward robustness improvements.

**Areas reviewed**: Shell, Python, Security
**Commits**: 14, all DCO signed-off
**CI**: all passing (including E2E HyperShift)

PR #1620, opened 2026-05-20
fix(docs): remove hardcoded v0.5.1 tag from README quickstart

### LINKED ISSUE #1618: README quickstart references scripts/kind/setup-kagenti.sh which doesn't exist at recommended v0.5.1 tag
## Description

The Install section of `README.md` instructs users to:

```bash
git checkout v0.5.1
scripts/kind/setup-kagenti.sh --with-ui --with-spire --with-agent-sandbox --with-builds
```

`scripts/kind/setup-kagenti.sh` was added to `main` after the `v0.5.1` tag was cut, so users following these instructions get:

```
bash: scripts/kind/setup-kagenti.sh: No such file or directory
```

## Steps to Reproduce

```bash
git clone https://github.com/kagenti/kagenti.git
cd kagenti
git checkout v0.5.1
scripts/kind/setup-kagenti.sh --with-ui --with-spire --with-agent-sandbox --with-builds
# ERROR: bash: scripts/kind/setup-kagenti.sh: No such file or directory
```

## Root Cause

The README was updated (on `main`) to recommend the new `scripts/kind/setup-kagenti.sh` installer and to `git checkout v0.5.1`, but the `v0.5.1` tag predates the script. The script only exists on `main`.

## Fix Options

1. **Remove the `git checkout v0.5.1` step** — let users install from `main`
2. **Cut a new release** that includes the script and update the README to reference it
3. **Use a dynamic tag lookup** instead of hardcoding a version that may not contain the referenced scripts

Option 2 is probably the right call — a new release that includes the Kind installer script, with the README referencing that tag.

## Environment

- README.md on `main` branch as of 2026-05-20
- Tag `v0.5.1` (commit `8c92519e`)

jeremyeder:
## Summary

- Remove `git checkout v0.5.1` from the quickstart — `scripts/kind/setup-kagenti.sh` was added after that tag, so users get `No such file or directory`
- Remove the version lookup tip since users now just use `main`

Fixes #1618

## Test plan

- [ ] Follow the updated quickstart from a clean clone
- [ ] Verify `scripts/kind/setup-kagenti.sh` exists and runs

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,10 +119,6 @@ Kagenti provides a set of components and assets that make it easier to manage AI
 git clone https://github.com/kagenti/kagenti.git
 cd kagenti
 
-# Check out the latest stable release (recommended)
-# Find the current version at https://github.com/kagenti/kagenti/releases/latest
-git checkout v0.5.1
-
 # Copy and configure secrets
 cp deployments/envs/secret_values.yaml.example deployments/envs/.secret_values.yaml
 # Edit deployments/envs/.secret_values.yaml with your values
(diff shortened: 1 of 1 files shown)


pdettori (approved):
Clean fix. Verified that `scripts/kind/setup-kagenti.sh` does not exist at tag v0.5.1 (confirming the bug) and that it exists on `main` with `--help` support (confirming the fix). All CI checks pass, commit is properly signed off.

Assisted-By: Claude Code

PR #1263, opened 2026-04-16
ci: build platform images from source in E2E workflows

pdettori:
## Summary

E2E tests were running against pre-built `:latest` images from ghcr.io, not the actual code from the PR branch. Backend, UI, and agent-oauth-secret changes were invisible to CI — tests could pass while the assembled release broke.

- Add `26-build-platform-images.sh` — builds backend, ui-v2, and agent-oauth-secret from source before tests run
- Kind path: `docker build` + `kind load` + `rollout restart`
- OpenShift path: delegates to `37-build-platform-images.sh` (BuildConfig + ImageStream)
- Wire build step into all 4 E2E workflows (Kind PR, Kind push, HyperShift PR, HyperShift push)
- Extend `37-build-platform-images.sh` to also build agent-oauth-secret on OpenShift
- Fix pre-existing bug: `agent-oauth-secret/Dockerfile` was missing `COPY auth/shared_utils.py` — caught by the new build-from-source step on its first run

### Impact

| Metric | Before | After |
|--------|--------|-------|
| Images from source | 1 (ui-oauth-secret) | 4 (+ backend, ui-v2, agent-oauth-secret) |
| CI time (Kind) | ~30 min | ~35-37 min (+5-7 min build) |
| False negatives | Backend/UI changes invisible | All PR code tested |

This is **Phase 1** of the strategy in #1186. Follow-on phases:
- Phase 2: Cross-repo CI triggers (extensions/operator PRs trigger kagenti E2E)
- Phase 3: Nightly + release validation workflows

Part of PRs to address  #1186

## Test plan

- [x] Kind PR workflow builds all 3 images from source (check `26-build-platform-images.sh` logs)
- [x] `86-print-version-matrix.sh` shows locally-built images, not ghcr.io digests
- [x] Build-from-source caught a real latent bug on first run (`shared_utils` missing from agent-oauth-secret Dockerfile)
- [x] Total Kind workflow time stays under 45 minutes (14m53s)
- [x] All 22/23 CI checks pass (HyperShift E2E requires `/run-e2e` maintainer trigger)
- [x] HyperShift workflow builds via OpenShift BuildConfig (requires `/run-e2e`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1621, opened 2026-05-21
ci: add scheduled cleanup for SHA-tagged container images

pdettori:
## Summary

- Adds a daily GitHub Actions workflow that deletes container image versions from ghcr.io older than 7 days when tagged with the ephemeral `<branch>-<sha>` pattern
- Covers all 7 packages built by the `Build-Publish` workflow (ui-v2, backend, *-oauth-secret, spiffe-idp-setup)
- Also cleans up untagged manifests (orphaned multi-arch layers)
- Protected tags (`v*`, `latest`) are never touched

## Context

PR #1263 builds all platform images from source in E2E workflows, which means every merge to main creates 7 new tagged images. These accumulate indefinitely at https://github.com/orgs/kagenti/packages?repo_name=kagenti and need periodic cleanup.

## Design choices

| Choice | Rationale |
|--------|-----------|
| 7-day retention | Covers a full sprint for bisect/debugging; images are always rebuildable from git |
| Pattern: `<branch>-<hex7-12>` | Matches both `main-abc1234` and PR branch tags like `feature-xyz-abc1234` |
| `fail-fast: false` | One package error doesn't block cleanup of others |
| `workflow_dispatch` with dry_run | Safe manual preview before first real run |
| `actions/github-script` | Full control over pattern matching without external action dependencies |

## Test plan

- [ ] Trigger workflow manually with `dry_run: true` to verify correct identification of stale images
- [ ] Verify protected tags (`latest`, `v0.6.0`) are never listed for deletion
- [ ] Let one real run execute and confirm image count drops on the packages page

Assisted-By: Claude Code

### DIFF
--- a/.github/workflows/cleanup-container-images.yaml
+++ b/.github/workflows/cleanup-container-images.yaml
@@ -19,13 +19,14 @@ on:
         type: number
         default: 7
 
-permissions:
-  contents: read
-  packages: write
+permissions: {}
 
 jobs:
   cleanup:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/cleanup-container-images.yaml
+++ b/.github/workflows/cleanup-container-images.yaml
@@ -1,0 +1,139 @@
+name: Cleanup Container Images
+
+on:
+  # Daily cleanup at 03:17 UTC (off-peak, avoids :00/:30 contention)
+  schedule:
+    - cron: '17 3 * * *'
+
+  # Manual trigger with configurable options
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: 'Dry run (show what would be deleted without actually deleting)'
+        required: false
+        type: boolean
+        default: false
+      retention_days:
+        description: 'Delete SHA-tagged images older than this many days'
+        required: false
+        type: number
+        default: 7
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  cleanup:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        package:
+          - ui-v2
+          - backend
+          - ui-oauth-secret
+          - agent-oauth-secret
+          - api-oauth-secret
+          - mlflow-oauth-secret
+          - spiffe-idp-setup
+
+    steps:
+      - name: Delete stale SHA-tagged images for ${{ matrix.package }}
+        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea  # v7.0.1
+        with:
+          script: |
+            const package_name = `kagenti/${{ matrix.package }}`;
+            const retentionDays = ${{ inputs.retention_days || 7 }};
+            const dryRun = ${{ inputs.dry_run || false }};
+            const cutoffDate = new Date(Date.now() - retentionDays * 24 * 60 * 60 * 1000);
+
+            // Pattern: any branch prefix followed by a 7-12 char hex SHA
+            const shaTagPattern = /^.+-[0-9a-f]{7,12}$/;
+            // Tags to never delete
+            const protectedTagPattern = /^(v\d|latest$)/;
+
+            console.log(`Package: ${package_name}`);
+            console.log(`Retention: ${retentionDays} days (cutoff: ${cutoffDate.toISOString()})`);
+            console.log(`Dry run: ${dryRun}`);
+            console.log('---');
+
+            let deleted = 0;
+            let skipped = 0;
+            let page = 1;
+            const perPage = 100;
+
+            while (true) {
+              const versions = await github.rest.packages.getAllPackageVersionsForPackageOwnedByOrg({
+                package_type: 'container',
+                package_name: package_name,
+                org: 'kagenti',
+                page: page,
+                per_page: perPage,
+              });
+
+              if (versions.data.length === 0) break;
+
+              for (const version of versions.data) {
+                const tags = version.metadata?.container?.tags || [];
+                const createdAt = new Date(version.created_at);
+
+                // Skip if no tags (untagged manifests are handled separately)
+                if (tags.length === 0) {
+                  // Untagged versions older than cutoff — delete them too
+                  if (createdAt < cutoffDate) {
+                    if (dryRun) {
+                      console.log(`[DRY RUN] Would delete untagged version ${version.id} (${createdAt.toISOString()})`);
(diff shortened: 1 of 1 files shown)


.github/workflows/cleanup-container-images.yaml
```diff
@@ -0,0 +1,140 @@
...
+            let page = 1;
+            const perPage = 100;
+
+            while (true) {
+              const versions = await github.rest.packages.getAllPackageVersionsForPackageOwnedByOrg({
+                package_type: 'container',
+                package_name: package_name,
+                org: 'kagenti',
+                page: page,
+                per_page: perPage,
```
huang195, line 74: **Suggestion**: The loop fetches pages and deletes inline. Issue: deleting versions shifts page boundaries on the *server side*, so when the loop increments `page` and re-fetches, items that "moved up" into the just-processed page slot are skipped.

Concretely: 200 versions, `perPage=100`. Page 1 returns versions 1-100; if 40 are deleted; page 2 now returns versions 141-200 (since the list re-numbered), missing 101-140.

The cron runs daily, so missed items get caught on the next run — cleanup *eventually* converges, just not in one shot. Two clean fixes:

- **(a)** Collect all `package_version_id` values to delete in a first pass, then iterate the collected list calling `deletePackageVersionForOrg` — no pagination drift.
- **(b)** Always re-fetch page 1 after each iteration, which works because deletions move surviving items "up".

(a) is clearer code. Not blocking — daily cron + eventual consistency is acceptable for cleanup.

.github/workflows/cleanup-container-images.yaml
```diff
@@ -0,0 +1,140 @@
...
+            const cutoffDate = new Date(Date.now() - retentionDays * 24 * 60 * 60 * 1000);
+
+            // Pattern: any branch prefix followed by a 7-12 char hex SHA
+            const shaTagPattern = /^.+-[0-9a-f]{7,12}$/;
+            // Tags to never delete
+            const protectedTagPattern = /^(v\d|latest$)/;
+
+            console.log(`Package: ${package_name}`);
+            console.log(`Retention: ${retentionDays} days (cutoff: ${cutoffDate.toISOString()})`);
+            console.log(`Dry run: ${dryRun}`);
```
huang195, line 59: **Observation**: `^.+-[0-9a-f]{7,12}$` handles the typical 7-char short SHA `${GITHUB_SHA::7}` and lets you go up to 12. Solid choice for current `Build-Publish` tagging. If kagenti ever standardizes on full 40-char SHAs (unlikely), the upper bound would need bumping.

.github/workflows/cleanup-container-images.yaml
```diff
@@ -0,0 +1,140 @@
...
+                const tags = version.metadata?.container?.tags || [];
+                const createdAt = new Date(version.created_at);
+
+                // Skip if no tags (untagged manifests are handled separately)
+                if (tags.length === 0) {
+                  // Untagged versions older than cutoff — delete them too
+                  if (createdAt < cutoffDate) {
+                    if (dryRun) {
+                      console.log(`[DRY RUN] Would delete untagged version ${version.id} (${createdAt.toISOString()})`);
+                    } else {
```
huang195: **Observation**: Untagged manifests <7 days old are intentionally kept — the comment notes orphaned multi-arch leftovers, which is right. Worth flagging explicitly that this also keeps any in-progress multi-arch assemblies from being deleted mid-build, since the GHCR multi-arch flow briefly produces untagged "child" manifests that are temporarily orphan-shaped before the parent index references them. The 7-day window covers any realistic build, so this is fine — just calling out the implicit safety.

.github/workflows/cleanup-container-images.yaml
```diff
@@ -0,0 +1,140 @@
...
+
+    steps:
+      - name: Delete stale SHA-tagged images for ${{ matrix.package }}
+        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea  # v7.0.1
+        with:
+          script: |
+            const package_name = `kagenti/${{ matrix.package }}`;
+            const retentionDays = ${{ inputs.retention_days || 7 }};
+            const dryRun = ${{ inputs.dry_run || false }};
+            const cutoffDate = new Date(Date.now() - retentionDays * 24 * 60 * 60 * 1000);
```
huang195, line 50: **Nit**: `actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea  # v7.0.1` — perfetto, full SHA pin + readable version comment is exactly the pattern `Verify Action Pinning` wants.

huang195 (approved):
Daily scheduled GitHub Actions workflow to clean up SHA-tagged container image versions older than 7 days from `ghcr.io/kagenti/*`. Covers the seven packages produced by `Build-Publish` (ui-v2, backend, four `*-oauth-secret` images, spiffe-idp-setup). Protected tags (`v*`, `latest`) explicitly preserved. Untagged manifest leftovers (orphaned multi-arch layers) cleaned up too if past the retention window. Top-level `permissions: {}` with `packages: write` correctly scoped to the cleanup job, `actions/github-script` pinned to SHA + version comment, `fail-fast: false`, and `workflow_dispatch` with `dry_run` + `retention_days` inputs for safe first-run preview. Bellissimo packaging.

**Areas reviewed**: GitHub Actions workflow YAML, github-script JavaScript logic, Scorecard / pinning hygiene, CI failures
**Commits**: 2, both signed-off (DCO passing)

## Two CI failures — neither caused by this PR

| Check | Status | Cause |
|---|---|---|
| `Scorecard` | fail | Score 9: "jobLevel 'packages' permission set to 'write'". **Unavoidable** for a workflow that legitimately deletes packages. The PR already does the right thing — top-level `permissions: {}`, `packages: write` only at job level. Score 9 is the ceiling once any write permission is present; getting to 10 would require removing the write entirely, defeating the workflow's purpose. Worth a one-line note in the PR body noting this is the floor for a package-deletion workflow so the next reviewer doesn't chase it. |
| `Deploy & Test` | fail | Unrelated: `AgentCardReconciler` can't find `weather-tool-advanced` Service, Keycloak DB constraint violations (duplicate key in `DEFAULT_CLIENT_SCOPE`), missing `keycloak` deployment in the `keycloak` namespace. None of these are touched by a workflow-only PR. Pre-existing cluster-setup flake — recent PRs have hit the same. Maintainer can ignore for this PR. |

Four inline comments below — one suggestion (pagination convergence), three observations / nits.

**Verdict: APPROVE** — clean, well-scoped, well-pinned. Mamma mia, even the cron minute (`:17`) is well-chosen — avoids the on-the-hour herd. Nice work.

PR #1622, opened 2026-05-21
fix(deps): pin kubernetes<36 to fix RBAC regression

pdettori:
## Summary

- Pin `kubernetes>=31.0.0,!=36.*` in backend `pyproject.toml` to avoid a breaking bug in v36.0.0
- Regenerate `uv.lock` with uv 0.11.15 (preserves revision=3 format with upload-time metadata)

## Root Cause

The `kubernetes` Python client v36.0.0 introduced a breaking change in its auth handling:
- `load_incluster_config()` stores the bearer token under `api_key['authorization']`
- `auth_settings()` only checks for `api_key['BearerToken']`

This mismatch causes the `Authorization` header to never be sent, so all K8s API calls authenticate as `system:anonymous` and return 403 Forbidden.

Upstream issues: [#2582](https://github.com/kubernetes-client/python/issues/2582), [#2584](https://github.com/kubernetes-client/python/issues/2584)

## Impact (v0.6.0-rc.5)

1. **Agent listing fails** — UI shows "Error loading agents. Permission denied. Check RBAC configuration."
2. **Shipwright detection fails** — `api_group_exists("shipwright.io")` returns false, "Build from Source" option hidden
3. **All K8s operations from the backend are broken**

## Verification

Tested on Kind cluster:
- Before fix: `kubernetes v36.0.0` — all API calls return 403
- After fix: `kubernetes v35.0.0` — agent listing, namespace discovery, build strategies all work

## Test plan

- [ ] Deploy to Kind cluster, verify agent listing works
- [ ] Verify "Build from Source" option appears in Import Agent page
- [ ] Run E2E tests


kagenti/backend/uv.lock
```diff
@@ -1,5 +1,4 @@
 version = 1
```
huang195, line 1: **Suggestion**: The lock file lost its `revision = 2` line and **all 941 `upload-time` annotations** got stripped. This is a downgrade from v2 to v1 lock format, suggesting the lock was regenerated with an older `uv` version (v2 format requires a recent uv).

Upstream main currently has `revision = 2` on this file. Two paths forward:
- (a) Re-run `uv lock` with a current `uv` so v2 format is preserved (better reproducibility metadata).
- (b) Accept the v1 downgrade as intentional and note it in the commit message.

Otherwise the next person running `uv lock` on a v2-aware host will silently re-add the metadata, churning the lock back and forth. Bellissimo would be (a) before merge.

kagenti/backend/pyproject.toml
```diff
@@ -11,7 +11,10 @@ dependencies = [
     "uvicorn[standard]>=0.30.0",
     "pydantic>=2.9.0",
     "pydantic-settings>=2.5.0",
```
huang195, line 13: **Suggestion**: `kubernetes>=31.0.0,<36` works, but `<36` will block v37+ from being auto-resolved once the upstream fix lands — somebody will have to come back and bump the pin.

Two more forward-looking patterns:
- `kubernetes>=31.0.0,!=36.*` — excludes only the broken major
- Keep `<36` but add a follow-up note in the comment so it's clear when to revisit

Issues #2582/#2584 are still open as of today, so v37 is hypothetical, but `!=36.*` future-proofs the pin without losing today's safety.

huang195 (approved):
Pins `kubernetes` Python client to `>=31.0.0,<36` in the backend `pyproject.toml` to dodge the v36.0.0 BearerToken regression, plus the regenerated `uv.lock` resolving to v35.0.0. **Bug claim verified independently**: kubernetes-client/python issues [#2582](https://github.com/kubernetes-client/python/issues/2582) and [#2584](https://github.com/kubernetes-client/python/issues/2584) — both open — confirm `load_incluster_config()` writes `api_key['authorization']` while `auth_settings()` reads `api_key['BearerToken']`, dropping the auth header on every K8s API call. v36 shipped 2026-05-20, so this hit the moment uv resolved to it. Mamma mia, what timing.

**Areas reviewed**: Python deps (`pyproject.toml`), `uv.lock` regeneration, kagenti commit/PR conventions
**Commits**: 1, signed-off (DCO passing — twice over, two trailers from different addresses, both Paolo)
**CI status**: 21/23 green; `Deploy & Test` and `E2E HyperShift` pending

Two inline suggestions below — one on the lock-file format downgrade (worth applying before merge), one on the pin shape. Plus three meta nits worth knowing about:

1. **PR title**: `fix(deps): pin kubernetes<36 to fix RBAC regression` — kagenti `repo:pr` convention is `<emoji> <description>`, not conventional-commits text prefix. 🐛 up front would match the existing convention. Not blocking.
2. **PR body footer**: reads `Assisted-By: Claude Code`. Kagenti convention (root `CLAUDE.md` "PR Bodies" section) is `Assisted-By: Claude (Anthropic AI) <noreply@anthropic.com>`. The commit message uses the right form already — just the PR description footer is the legacy phrasing. One-line edit.
3. **Commit message trailers**: two `Signed-off-by` lines, one from `paolo@us.ibm.com` and one from `dettori@us.ibm.com`. DCO is happy with either; the duplicate is a small artifact of running `-s` twice or amending with two different `user.email` configs. Squash to a single trailer if you have to amend for the lock-file fix anyway.

**Verdict: APPROVE** — fix is correct, bug claim is independently verified against upstream issues, and the impact (UI showing "Permission denied" + Build-from-Source disappearing) is exactly what v0.6.0-rc.5 deployments would have hit. Perfetto otherwise.

kagenti/backend/uv.lock
pdettori: Done — regenerated with uv 0.11.15, lock file now has `revision = 3` with upload-time metadata preserved.

kagenti/backend/pyproject.toml
pdettori: Good call — switched to `!=36.*` so v37+ auto-resolves once upstream fixes land. Added the issue link in the comment too.

PR #1623, opened 2026-05-21
ci: mitigate API rate limiting in container image cleanup

pdettori:
## Summary

- Replaces parallel matrix jobs with a single sequential step processing all 7 packages — eliminates parallel budget contention on the shared 5000 req/hour GitHub API limit
- Adds rate-limit checking every 50 deletions — if remaining requests drop below 100, sleeps until the reset window
- Caps total deletions at 4000 per run to leave headroom for pagination GETs; backlog drains across multiple daily runs

## Context

The first run of #1621 hit the GitHub API rate limit after ~715 deletions on a single package. With 7 matrix jobs running in parallel, the shared 5000 req/hour budget was exhausted almost immediately given the months-long backlog of SHA-tagged images.

## Design

For the initial backlog (thousands of images), the workflow will take a few daily runs to fully catch up. Once caught up, daily runs only need to clean ~7 new images/day — well within budget.

## Test plan

- [ ] Trigger manually — verify rate-limit check fires and caps are respected
- [ ] Verify packages page shows reduced image count after a few runs
- [ ] Confirm Scorecard still passes at score 9 (job-level `packages: write` is required)

Assisted-By: Claude Code

### DIFF
--- a/.github/workflows/cleanup-container-images.yaml
+++ b/.github/workflows/cleanup-container-images.yaml
@@ -27,102 +27,148 @@ jobs:
     permissions:
       contents: read
       packages: write
-    strategy:
-      fail-fast: false
-      matrix:
-        package:
-          - ui-v2
-          - backend
-          - ui-oauth-secret
-          - agent-oauth-secret
-          - api-oauth-secret
-          - mlflow-oauth-secret
-          - spiffe-idp-setup
 
     steps:
-      - name: Delete stale SHA-tagged images for ${{ matrix.package }}
+      - name: Cleanup stale SHA-tagged container images
         uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea  # v7.0.1
         with:
           script: |
-            const package_name = `kagenti/${{ matrix.package }}`;
+            const packages = [
+              'ui-v2',
+              'backend',
+              'ui-oauth-secret',
+              'agent-oauth-secret',
+              'api-oauth-secret',
+              'mlflow-oauth-secret',
+              'spiffe-idp-setup',
+            ];
+
             const retentionDays = ${{ inputs.retention_days || 7 }};
             const dryRun = ${{ inputs.dry_run || false }};
             const cutoffDate = new Date(Date.now() - retentionDays * 24 * 60 * 60 * 1000);
+
+            // Cap deletions to stay well within the 5000 req/hour rate limit.
+            // Each package needs ~1 GET per 100 versions + 1 DELETE per stale version.
+            // Budget: ~4000 requests total (leave headroom for pagination GETs).
+            const maxDeletesPerRun = 4000;
 
             // Pattern: any branch prefix followed by a 7-12 char hex SHA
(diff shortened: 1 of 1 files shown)


.github/workflows/cleanup-container-images.yaml
```diff
@@ -27,102 +27,148 @@ jobs:
...
+              console.log(`Found ${toDelete.length} to delete, ${skipped} preserved`);
+              totalSkipped += skipped;
+
+              // Phase 2: delete with rate-limit awareness
+              for (const item of toDelete) {
+                if (totalDeleted >= maxDeletesPerRun) {
+                  hitCap = true;
+                  console.log(`Reached per-run cap (${maxDeletesPerRun}). Remaining cleanup deferred to next run.`);
+                  break;
+                }
```
mrsabath, line 159: **suggestion**: Phase 1 pagination GETs (`getAllPackageVersionsForPackageOwnedByOrg`) aren't covered by `checkRateLimit()` — only the delete loop in Phase 2 checks every 50 ops. With 7 packages × multiple pages each, listing alone consumes part of the budget before any delete runs. If a prior run already left the budget low and Phase 2 cap was hit, the next run could exhaust the remaining budget during listing before reaching `checkRateLimit()`. Consider calling `checkRateLimit()` once per package before Phase 1, or every N pagination calls.

.github/workflows/cleanup-container-images.yaml
```diff
@@ -27,102 +27,148 @@ jobs:
...
+                    continue;
+                  }
+
+                  if (!tags.every(tag => shaTagPattern.test(tag))) {
+                    skipped++;
+                    continue;
+                  }
+
+                  if (createdAt >= cutoffDate) {
+                    skipped++;
```
mrsabath, line 139: **suggestion**: Package processing is in fixed order, so when `hitCap` triggers mid-run, the same early packages (`ui-v2`, `backend`) get cleaned first every day until their backlog is empty, while later packages (`spiffe-idp-setup`) wait until earlier packages are caught up. For initial backlog drain this is acceptable, but consider rotating the start index (e.g. by day-of-year mod 7) or shuffling, so cap pressure is spread fairly across packages.

mrsabath (approved):
Sequentializing the matrix and adding rate-limit awareness is the right call — the parallel matrix sharing a single 5000 req/hour budget was the root cause of the original failure. Logic is correct: `hitCap` propagates across packages (outer-loop check before each package's Phase 1), `checkRateLimit()` waits with a 5s buffer past reset, and the 4000 cap leaves ~1000 requests of headroom for pagination GETs.

**Areas reviewed**: CI/GitHub Actions (single workflow file, +116/-70)
**Commits**: 1 commit, signed-off ✓, conventional `ci:` prefix ✓
**CI status**: All required checks passing; E2E HyperShift pending `/run-e2e` trigger (expected)
**Security**: `actions/github-script` pinned to SHA ✓, scoped permissions (`packages: write`) ✓, no secrets

Two non-blocking suggestions inline on Phase 1 budget coverage and package fairness across runs.

PR #1627, opened 2026-05-21
feat(ui): Add mtlsMode select to AgentRuntime create form

huang195:
## Summary

Surfaces the new `AgentRuntime.Spec.MTLSMode` field (added in [kagenti-operator PR #369](https://github.com/kagenti/kagenti-operator/pull/369)) through the kagenti backend + UI + TUI struct so users can pick a per-workload mTLS posture from the Import Agent form rather than hand-editing the CR or namespace ConfigMap.

The mTLS feature is fully implemented on the operator + authbridge side ([kagenti-extensions PR #424](https://github.com/kagenti/kagenti-extensions/pull/424) merged authbridge mTLS, kagenti-operator PR #369 merged the CR field + auto-SPIRE + validation). This PR is the user-facing wiring.

### Backend (`kagenti/backend/app/routers/agents.py`)

- `mtlsMode: Optional[Literal["disabled", "permissive", "strict"]]` on `CreateAgentRequest` AND `FinalizeShipwrightBuildRequest`.
- `@model_validator` on both that mirrors the operator's validating webhook check: rejects `mtlsMode != disabled` when `authBridgeMode == envoy-sidecar`. Returns 422 from FastAPI rather than letting the operator webhook deny later.
- `_build_agentruntime_manifest` emits `spec.mtlsMode` when set, omits the key when `None`.
- Both `_ensure_agentruntime` call sites updated (direct-create and Shipwright finalize).
- Shipwright finalize: the field is stashed on the BuildRun via `kagenti.io/agent-config` annotation at submit time and read back at finalize, so build-from-source agents inherit the form setting (parity with direct-image agents).

### UI (`kagenti/ui-v2/src/pages/ImportAgentPage.tsx`)

- New `FormSelect` "Sidecar mTLS" inside the existing **AuthBridge Advanced Configuration** ExpandableSection (sits next to `defaultOutboundPolicy`).
- Three options: `disabled` (default) / `permissive` / `strict`.
- Greys out and force-resets to `disabled` when `useEnvoyMode` is checked.
- Helper text covers two operator-relevant caveats:
  - Requires SPIRE in the cluster (the operator auto-enables SPIRE for workloads with `mtlsMode != disabled`).
  - Permissive falls back **silently** to plaintext on TLS handshake failure; use strict for production posture once rollout completes.

### Types

- `MtlsMode` type alias in `types/index.ts`.
- `mtlsMode` field on `agentService.create` and `shipwrightService.finalizeBuild` body types in `services/api.ts`.

### TUI (`kagenti/tui/internal/api/types.go`)

- `MTLSMode string` field on `CreateAgentRequest`. No form/CLI work — TUI's deploy form doesn't yet expose `authBridgeMode` either; bundling form work would widen parity gaps. Tracked as a follow-up.

### Tests

- New `kagenti/backend/tests/test_agentruntime_manifest.py` — 11 tests covering manifest builder, cross-field validator on both request models, default behavior, Pydantic enum enforcement.

## Defense in depth

The incompatible combo `mtlsMode != disabled` + `authBridgeMode = envoy-sidecar` is rejected at four layers:

| Layer | Mechanism |
|---|---|
| UI | `FormSelect` greyed out + force-reset to `disabled` when `useEnvoyMode` is checked |
| Backend | Pydantic `model_validator` on both `CreateAgentRequest` and `FinalizeShipwrightBuildRequest` |
| Operator | `AgentRuntimeValidator.checkMTLSCompatibleWithMode` (kagenti-operator PR #369) |
| Authbridge | (Implicit) envoy-sidecar mode doesn't read the `mtls:` block; the field has no effect there |

## Trade-off worth flagging

The form **always** sends an explicit `mtlsMode` value (default `disabled`). Combined with the operator's CR-pin semantic, this means **UI-created agents opt out of any namespace-level `mtls.mode` setting**. A cluster admin who sets `mtls.mode: strict` in `authbridge-runtime-config` expecting it to apply to all new agents in the namespace will be surprised when UI-created agents come up plaintext.

Reasoning: the form should reflect what the user picks; cluster admins should use admission policy (Kyverno / OPA) rather than ConfigMap defaults to enforce posture. Power users can still get namespace-default inheritance by `kubectl edit agentruntime <name>` to clear `Spec.MTLSMode`.

## Out of scope (intentional)

- **Tools (`kagenti/backend/app/routers/tools.py`)**: tools don't create AgentRuntime CRs (they use the deprecated `kagenti.io/authbridge-mode` pod annotation), so `mtlsMode` has no place to live. Clarifying comment added near tools' `authBridgeMode` field.
- **AgentDetailPage display**: showing the resolved `mtlsMode` for an existing agent would require a backend GET endpoint for the AgentRuntime CR. Deferred — agents are read-only post-creation today, and adding GET-only display is a separate concern.
- **Edit-after-creation**: kagenti has no edit-agent flow. Changing mtlsMode on an existing agent requires `kubectl edit agentruntime <name>`. Same constraint applies to `authBridgeMode`.
- **envoy-sidecar mTLS**: requires extending the kagenti meta-chart's envoy-config template to configure SDS. Comparable in size to the original PR #424.
- **TUI form parity**: the TUI doesn't expose `authBridgeMode` either; widening that form is a separate piece of work.

## Test plan

- [x] Backend: `pytest tests/test_agentruntime_manifest.py` — 11/11 pass
- [x] Backend: smoke-test via `python -c 'from app.routers.agents import ...'` — model validates, manifest emits, validator rejects the bad combo
- [x] UI: `npx tsc --noEmit` clean
- [x] UI: `npm run lint` adds no new errors (the two pre-existing `'id' unused-var` warnings on the `outboundRoutes` destructure shifted line numbers due to state additions; count unchanged at 2)
- [x] TUI: `go build ./...` clean
- [ ] CI passes
- [ ] End-to-end demo verification:
  1. Deploy operator from `main` (PR #369 merged)
  2. Deploy kagenti with this branch
  3. Open the UI, navigate to **Import Agent**, expand **AuthBridge Advanced Configuration**
  4. Pick `mtlsMode = strict`, submit
  5. Verify `kubectl get agentruntime <name> -o yaml` shows `spec.mtlsMode: strict`
  6. Verify the per-agent ConfigMap has `mtls: {mode: strict}` block
  7. Verify two agent pods perform TLS handshakes on the wire (tcpdump from a debug pod)
  8. Toggle the form to `useEnvoyMode = true`, observe `mtlsMode` greys out and resets to `disabled`
  9. Negative path: bypass the form (curl POST), confirm backend returns 422 with the forward-pointing error

## Files

- `kagenti/backend/app/routers/agents.py` — Pydantic field + validator + manifest plumbing + Shipwright finalize plumbing
- `kagenti/backend/app/routers/tools.py` — clarifying comment (no field added)
- `kagenti/backend/tests/test_agentruntime_manifest.py` — 11 new tests
- `kagenti/ui-v2/src/types/index.ts` — `MtlsMode` type alias
- `kagenti/ui-v2/src/services/api.ts` — request body types
- `kagenti/ui-v2/src/pages/ImportAgentPage.tsx` — state + FormSelect + submit-mapping
- `kagenti/tui/internal/api/types.go` — Go struct field

Net: ~+308 / -4 LOC across 7 files.

### DIFF
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -3619,9 +3619,7 @@ async def finalize_shipwright_build(
         # and we read it back here so build-from-source agents inherit
         # the same setting as direct-image agents).
         final_mtls_mode = (
-            request.mtlsMode
-            if request.mtlsMode is not None
-            else stored_config.get("mtlsMode")
+            request.mtlsMode if request.mtlsMode is not None else stored_config.get("mtlsMode")
         )
 
         # Persistent storage
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -20,7 +20,7 @@ import yaml
 from fastapi import APIRouter, Depends, HTTPException, Query
 import kubernetes.client
 from kubernetes.client import ApiException
-from pydantic import BaseModel, Field, field_validator
+from pydantic import BaseModel, Field, field_validator, model_validator
 
 from app.core.auth import ROLE_OPERATOR, ROLE_VIEWER, require_roles
 from app.utils.routes import get_agent_url
@@ -257,6 +257,18 @@ class CreateAgentRequest(BaseModel):
     # operator but not surfaced through the UI today.
     authBridgeMode: Optional[Literal["proxy-sidecar", "envoy-sidecar", "lite", "waypoint"]] = None
 
+    # Per-workload mTLS posture between AuthBridge sidecars. Maps to
+    # AgentRuntime.Spec.MTLSMode. The kagenti UI sends an explicit
+    # value (default "disabled") so users always see what they get;
+    # this means UI-created agents opt out of any namespace-level
+    # mtls.mode setting (CR-pin semantic in the operator). The
+    # operator's validating webhook rejects the envoy-sidecar +
+    # non-disabled combo because Envoy SDS isn't currently configured
+    # by the kagenti envoy-config — we mirror that check below as a
+    # model_validator so the form gets a fast 422 instead of a
+    # webhook denial after the manifest is built.
+    mtlsMode: Optional[Literal["disabled", "permissive", "strict"]] = None
+
     # Port exclusion annotations
     outboundPortsExclude: Optional[str] = None
     inboundPortsExclude: Optional[str] = None
@@ -283,6 +295,29 @@ class CreateAgentRequest(BaseModel):
                 f"Supported types: {', '.join(SUPPORTED_WORKLOAD_TYPES)}"
             )
         return v
+
+    @model_validator(mode="after")
+    def _check_mtls_compatible_with_mode(self) -> "CreateAgentRequest":
+        """Mirror the operator's AgentRuntime validating webhook check.
+
+        envoy-sidecar mode does not currently support mTLS — the kagenti
+        envoy-config doesn't configure SDS, so a strict/permissive value
+        here would silently produce a workload running plaintext while
+        the user believed mTLS was on. The operator webhook will reject
+        this combo at admission; we reject it earlier so the form gets
+        a clean 422 instead of a webhook denial after the manifest has
+        been constructed.
+        """
+        if self.authBridgeMode == "envoy-sidecar" and self.mtlsMode not in (
+            None,
+            "disabled",
+        ):
+            raise ValueError(
+                "mtlsMode must be 'disabled' when authBridgeMode is "
+                "'envoy-sidecar' (envoy-sidecar mTLS is tracked as a "
+                "follow-up; use proxy-sidecar or lite for mTLS today)"
+            )
+        return self
 
 
 class CreateAgentResponse(BaseModel):
@@ -2308,6 +2343,8 @@ def _build_agent_shipwright_build_manifest(
         resource_config["inboundPortsExclude"] = request.inboundPortsExclude
     if request.defaultOutboundPolicy:
         resource_config["defaultOutboundPolicy"] = request.defaultOutboundPolicy
+    if request.mtlsMode:
+        resource_config["mtlsMode"] = request.mtlsMode
     if request.persistentStorage:
         resource_config["persistentStorage"] = request.persistentStorage.model_dump()
     # Add env vars if present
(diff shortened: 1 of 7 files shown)


kagenti/ui-v2/src/pages/ImportAgentPage.tsx
```diff
@@ -1207,6 +1221,30 @@ export const ImportAgentPage: React.FC = () => {
                     <FormSelectOption key="exchange" value="exchange" label="exchange — require token exchange for all outbound traffic" />
                   </FormSelect>
```
cwiklik, line 1222: suggestion: The `MtlsMode` type is defined in `types/index.ts` but the state here is typed inline as a string union literal. Consider using the type alias for consistency:
```tsx
const [mtlsMode, setMtlsMode] = useState<MtlsMode>('disabled');
```

kagenti/backend/app/routers/agents.py
```diff
@@ -2308,6 +2343,8 @@ def _build_agent_shipwright_build_manifest(
         resource_config["inboundPortsExclude"] = request.inboundPortsExclude
     if request.defaultOutboundPolicy:
```
cwiklik, line 2344: nit: `if request.mtlsMode:` works here because `"disabled"` is truthy and you want to stash it. But it reads oddly compared to the `is not None` pattern used in the finalize path (line ~3614). Consider `if request.mtlsMode is not None:` for consistency with the rest of the file's None-check pattern.

kagenti/tui/internal/api/types.go
```diff
@@ -211,6 +211,13 @@ type CreateAgentRequest struct {
 	AuthBridgeEnabled bool                     `json:"authBridgeEnabled"`
 	SpireEnabled      bool                     `json:"spireEnabled"`
 	AuthBridgeMode    string                   `json:"authBridgeMode,omitempty"`
+	// MTLSMode maps to AgentRuntime.Spec.MTLSMode. Backend rejects
+	// non-disabled values when AuthBridgeMode is "envoy-sidecar"
+	// (Envoy SDS not currently configured by the kagenti chart).
+	// Empty string sends nothing (operator default applies); the TUI
```
cwiklik, line 217: nit: minor formatting — `MTLSMode` has extra padding to align with `PersistentStorage` below, but `AuthBridgeMode` on the line above doesn't follow the same alignment. Not blocking, just inconsistent.

cwiklik (approved):
Well-structured PR surfacing `AgentRuntime.Spec.MTLSMode` through the full stack. Defense-in-depth validation at 4 layers (UI, backend, operator webhook, implicit envoy). Good test coverage (11 tests). The CR-pin trade-off (always sending explicit `disabled`) is documented and deliberate.

**Areas reviewed**: Python, Frontend, TUI (Go), Security
**Commits**: 2 commits, all signed-off
**CI status**: passing (Deploy & Test pending)

Assisted-By: Claude Code

PR #1629, opened 2026-05-21
fix(ocp): add --reset-values to helm upgrade to prevent stale overrides

pdettori:
## Summary

- Add `--reset-values` to the main `helm upgrade --install kagenti` command in the OCP installer

## Problem

Without `--reset-values`, `helm upgrade` carries forward user-supplied values from the previous release. This causes version switches to be silently ignored — e.g. when downgrading from `v0.7.0-alpha.1` to `v0.6.0-rc.5`, the old `ui.frontend.tag: v0.7.0-alpha.1` override persists because helm merges it on top of the new chart defaults.

## Why this is safe

The installer script already passes all required configuration explicitly:
- `-f .secrets.yaml` (credentials)
- `--set` flags for keycloak URL, realm, SA cert, SPIFFE prefix, feature flags, operator image, etc.

There is nothing legitimate that `--reuse-values` would preserve that isn't already declared in the script. Any ad-hoc `helm upgrade --set` done between script runs is an anti-pattern for a declarative installer.

## Test plan

- [ ] Run `setup-kagenti.sh --kagenti-repo .` on a cluster that previously had a different version
- [ ] Verify deployed image tags match the chart's `values.yaml` defaults

### DIFF
--- a/scripts/ocp/setup-kagenti.sh
+++ b/scripts/ocp/setup-kagenti.sh
@@ -1202,6 +1202,7 @@ run_cmd $KUBECTL create namespace mcp-system --dry-run=client -o yaml | $KUBECTL
 
 run_cmd helm upgrade --install kagenti "$KAGENTI_REPO/charts/kagenti/" \
   -n kagenti-system --create-namespace \
+  --reset-values \
   -f "$SECRETS_FILE" \
   ${KAGENTI_UI_FLAGS[@]+"${KAGENTI_UI_FLAGS[@]}"} \
   ${OPERATOR_IMAGE_FLAGS[@]+"${OPERATOR_IMAGE_FLAGS[@]}"} \
--- a/scripts/ocp/setup-kagenti.sh
+++ b/scripts/ocp/setup-kagenti.sh
@@ -661,6 +661,7 @@ EOF
     # cluster-monitoring-config is managed by another operator)
     log_info "kagenti-deps already installed — upgrading (hooks skipped)"
     run_cmd helm upgrade kagenti-deps "$KAGENTI_REPO/charts/kagenti-deps/" \
+      --reset-values \
       "${KAGENTI_DEPS_HELM_ARGS[@]}"
     # Apply operand CRs on upgrade too — catches CRs missed by a previous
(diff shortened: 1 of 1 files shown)


scripts/ocp/setup-kagenti.sh
```diff
@@ -1202,6 +1202,7 @@ run_cmd $KUBECTL create namespace mcp-system --dry-run=client -o yaml | $KUBECTL
 
 run_cmd helm upgrade --install kagenti "$KAGENTI_REPO/charts/kagenti/" \
   -n kagenti-system --create-namespace \
+  --reset-values \
```
mrsabath, line 1207: **suggestion** (out of scope, follow-up OK): the `kagenti-deps` upgrade calls earlier in this script (around lines 663 and 930 on `main`) have the same `helm upgrade ... "${KAGENTI_DEPS_HELM_ARGS[@]}"` pattern and could carry forward stale user-supplied values across re-runs in the same way the main chart does. They may benefit from the same `--reset-values` treatment.

⚠️ Important: the deferred MLflow augmentation around line 1088 *intentionally* uses `--reuse-values` because it adds MLflow config on top of the existing release — that one must stay as-is.

Happy to land this PR as-is and address `kagenti-deps` in a follow-up.

mrsabath (approved):
Targeted 1-line fix. `--reset-values` is the correct primitive to break the stale-override carryover described in the PR body, and the safety argument holds: the script passes all required overrides explicitly via `-f .secrets.yaml` and `--set` flags, so chart defaults as the baseline is the desired behavior.

One non-blocking observation inline about the sibling `kagenti-deps` upgrades — same pattern, possibly the same issue — but out of scope for this PR.

**Areas reviewed**: Shell (1 file, +1/-0)
**Commits**: 1 commit, signed-off ✓, conventional `fix(ocp):` prefix ✓
**CI status**: All required checks passing; Deploy & Test pending (expected gating)

scripts/ocp/setup-kagenti.sh
pdettori: addressed

PR #1631, opened 2026-05-21
fix(ocp): enable Build from Source with OpenShift internal registry

pdettori:
## Summary

- Wire the existing `shipwright_default_strategy` config setting into `select_build_strategy()` so OCP's TLS-enabled internal registry uses the standard `buildah` ClusterBuildStrategy (instead of the Kind-only `buildah-insecure-push` which doesn't exist on OCP)
- OCP installer (`--with-builds`) now automatically configures the internal registry URL and build strategy
- Documentation rewritten to show that internal registry is used automatically with `--with-builds`

## Problem

On OpenShift, "Build from Source" failed because:
1. The installer never set `defaultRegistryUrl` for the backend
2. `select_build_strategy()` hardcoded `buildah-insecure-push` for internal registries, but that ClusterBuildStrategy only exists on Kind — OCP's internal registry has TLS and needs the standard `buildah` strategy

## Changes

| File | Change |
|------|--------|
| `kagenti/backend/app/services/shipwright.py` | Use `settings.shipwright_default_strategy` instead of hardcoded constant |
| `charts/kagenti/templates/ui.yaml` | Pass `SHIPWRIGHT_DEFAULT_STRATEGY` env var |
| `charts/kagenti/values.yaml` | Add `shipwrightDefaultStrategy` field |
| `scripts/ocp/setup-kagenti.sh` | Add `BUILD_FLAGS` block for `--with-builds` |
| `deployments/envs/ocp_*.yaml` | Add `shipwrightDefaultStrategy: "buildah"` |
| `docs/ocp/openshift-install.md` | Rewrite Build from Source section |

## Test plan

- [ ] On Kind: verify no behavior change (strategy stays `buildah-insecure-push`)
- [ ] On OCP with `--with-builds`: verify backend receives correct env vars
- [ ] On OCP: Build from Source in UI creates Shipwright Build with `buildah` strategy targeting internal registry
- [ ] On OCP: built image is pushed to internal registry without push secret

Assisted-By: Claude Code

### DIFF
--- a/charts/kagenti/templates/ui.yaml
+++ b/charts/kagenti/templates/ui.yaml
@@ -123,6 +123,10 @@ spec:
             - name: DEFAULT_REGISTRY_URL
               value: {{ .Values.ui.backend.defaultRegistryUrl | quote }}
             {{- end }}
+            {{- if .Values.ui.backend.shipwrightDefaultStrategy }}
+            - name: SHIPWRIGHT_DEFAULT_STRATEGY
+              value: {{ .Values.ui.backend.shipwrightDefaultStrategy | quote }}
+            {{- end }}
             # Feature flags
             - name: KAGENTI_FEATURE_FLAG_SANDBOX
               value: "{{ .Values.featureFlags.sandbox }}"
--- a/kagenti/backend/app/services/shipwright.py
+++ b/kagenti/backend/app/services/shipwright.py
@@ -67,14 +67,19 @@ def select_build_strategy(registry_url: str, requested_strategy: Optional[str] =
     (no TLS); on OpenShift it should be "buildah" (TLS-enabled internal registry).
     For external registries, always uses the secure "buildah" strategy.
 
+    Note: for internal registries, any requested_strategy that doesn't match the
+    configured internal strategy is overridden. This ensures the correct TLS
+    behavior regardless of what the caller requests.
+
     Args:
         registry_url: The registry URL to push images to
-        requested_strategy: Optional explicitly requested strategy
+        requested_strategy: Optional explicitly requested strategy (ignored for
+            internal registries if it doesn't match the configured strategy)
 
     Returns:
         The build strategy name to use
     """
-    from app.core.config import settings
+    from app.core.config import settings  # deferred: avoid circular dep with config→constants
 
     is_internal_registry = (
         registry_url == DEFAULT_INTERNAL_REGISTRY or "svc.cluster.local" in registry_url
--- a/kagenti/backend/app/services/shipwright.py
+++ b/kagenti/backend/app/services/shipwright.py
@@ -62,8 +62,10 @@ def select_build_strategy(registry_url: str, requested_strategy: Optional[str] =
     """
     Select the appropriate build strategy based on the registry.
 
-    For internal registries (svc.cluster.local), uses the insecure push strategy.
-    For external registries with TLS, uses the secure strategy.
+    For internal registries, uses the configured default strategy (configurable
+    via SHIPWRIGHT_DEFAULT_STRATEGY env var). On Kind this is "buildah-insecure-push"
+    (no TLS); on OpenShift it should be "buildah" (TLS-enabled internal registry).
+    For external registries, always uses the secure "buildah" strategy.
 
     Args:
         registry_url: The registry URL to push images to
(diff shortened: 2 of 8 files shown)


kagenti/backend/app/services/shipwright.py
```diff
@@ -72,22 +74,22 @@ def select_build_strategy(registry_url: str, requested_strategy: Optional[str] =
...
     is_internal_registry = (
         registry_url == DEFAULT_INTERNAL_REGISTRY or "svc.cluster.local" in registry_url
     )
+    internal_strategy = settings.shipwright_default_strategy
 
-    # If a strategy was explicitly requested, use it unless it's secure for internal registry
     if requested_strategy:
-        if is_internal_registry and requested_strategy == SHIPWRIGHT_STRATEGY_SECURE:
-            # Override to insecure for internal registries
+        if is_internal_registry and requested_strategy != internal_strategy:
```
huang195, line 90: **Observation**: Subtle semantic change in the override path.

**Old:** `if is_internal_registry and requested_strategy == SHIPWRIGHT_STRATEGY_SECURE: return SHIPWRIGHT_STRATEGY_INSECURE` — only forces SECURE → INSECURE on internal registries; any other custom strategy passes through.

**New:** `if is_internal_registry and requested_strategy != internal_strategy: return internal_strategy` — forces *any* mismatched strategy to `internal_strategy` on internal.

Effect: a power user who passes `requested_strategy="my-custom-strategy"` and points at an internal registry previously got `"my-custom-strategy"` through; now gets `internal_strategy` (which equals `INSECURE` on Kind, `"buildah"` on OCP).

Likely intentional — the new behavior is more consistent — but the PR description doesn't call this out explicitly. Worth one line in the description so future operators reading the diff know it's by design.

kagenti/backend/app/services/shipwright.py
```diff
@@ -72,22 +74,22 @@ def select_build_strategy(registry_url: str, requested_strategy: Optional[str] =
     Returns:
         The build strategy name to use
     """
```
huang195, line 81: **Nit**: `from app.core.config import settings` is imported inside the function rather than at module top. If this is to avoid circular imports, consider a one-line `# deferred import: avoid circular dep with X` comment so the next reader doesn't move it to the top and break things. If it's just stylistic, top-level is more conventional. Cheap clarification.

scripts/ocp/setup-kagenti.sh
```diff
@@ -1200,6 +1200,14 @@ if [ -n "$OPERATOR_IMAGE" ]; then
   log_info "Operator image: ${OPERATOR_IMAGE}"
 fi
 
+# Build-from-source flags: configure the OCP internal registry and build strategy
```
huang195, line 1203: **Observation**: `${BUILD_FLAGS[@]+"${BUILD_FLAGS[@]}"}` is the correct bash idiom for "expand only if set" — handles the `WITH_BUILDS=false` case cleanly without `set -u` tripping. Mirrors the surrounding `KAGENTI_UI_FLAGS` / `OPERATOR_IMAGE_FLAGS` pattern. Perfetto.

charts/kagenti/templates/ui.yaml
```diff
@@ -123,6 +123,10 @@ spec:
             - name: DEFAULT_REGISTRY_URL
               value: {{ .Values.ui.backend.defaultRegistryUrl | quote }}
             {{- end }}
+            {{- if .Values.ui.backend.shipwrightDefaultStrategy }}
```
huang195, line 126: **Observation**: `{{- if .Values.ui.backend.shipwrightDefaultStrategy }}` correctly only injects the env var when configured (empty string is the default in `values.yaml`). On Kind with no override, the env var is absent → backend reads `config.py` default `"buildah-insecure-push"` → preserves existing behavior. Backward-compat verified.

huang195 (approved):
Replaces the hardcoded `SHIPWRIGHT_STRATEGY_INSECURE` for internal registries with a configurable `settings.shipwright_default_strategy`, plumbed through Helm chart → backend env var → `config.py` → `select_build_strategy()`. OCP installer + env profiles set it to `"buildah"` so OpenShift's TLS-enabled internal registry works out of the box with `--with-builds`. Kind keeps the existing `"buildah-insecure-push"` default. Doc rewrite splits "Build from Source" into two clear paths (internal vs external registry) and tidies up some duplicated `## Running the demo` / `# Running the Demo` headings that had drifted in the old version. Bellissima cleanup.

**Areas reviewed**: Python (`shipwright.py`), Helm chart + values, OCP installer bash, OCP env YAMLs, `openshift-install.md` rewrite, kagenti commit/PR conventions
**Commits**: 1, signed-off (DCO passing); commit message uses correct `Assisted-By` form
**CI status**: 23/24 green; `Deploy & Test` pending

## Three meta items, easy to apply before merge:

1. **Test coverage gap (`kagenti/backend/tests/test_shipwright_shared.py`)**: Existing 7 `select_build_strategy` tests **all still pass with this change** — but only because the config default `shipwright_default_strategy = "buildah-insecure-push"` happens to equal the old `SHIPWRIGHT_STRATEGY_INSECURE` constant. The new plumbing isn't actually exercised by tests. Worth one new case that uses `monkeypatch.setattr(settings, "shipwright_default_strategy", "buildah")`, calls `select_build_strategy(DEFAULT_INTERNAL_REGISTRY)`, and asserts the result is `"buildah"`. Without this test, a future regression that hardcoded the strategy back to `INSECURE` would silently pass CI on Kind and only break on OCP.

2. **PR title**: lacks emoji prefix per kagenti `repo:pr` convention (`<emoji> <description>`, not conventional-commits text prefix). 🐛 (bug-fix) up front would match. Same pattern as recent #1622.

3. **PR body footer**: reads `Assisted-By: Claude Code`. Kagenti convention (root `CLAUDE.md` "PR Bodies" section) is `Assisted-By: Claude (Anthropic AI) <noreply@anthropic.com>`. The commit message uses the right form already; just the PR description footer is the legacy phrasing.

Four inline observations / suggestions below.

**Verdict: APPROVE** — clean fix that lets `--with-builds` actually work on OCP without operators having to know about the registry/strategy mismatch. The test coverage gap (#1 above) is the only thing worth catching pre-merge — five-line addition. Mamma mia, mostly polish — nice work.

PR #1639, opened 2026-05-22
fix(ocp): use local chart tag when --kagenti-repo is a local path

pdettori:
## Summary

- When the OCP installer is run with `--kagenti-repo .` (local path), it was still querying GitHub for the latest tag and overriding image tags. This defeated `--reset-values` and local chart pins.
- Now reads the tag from the local chart's `values.yaml` instead of querying GitHub when a local repo path is provided.

## Root Cause

The installer always ran `git ls-remote --tags` against GitHub to determine the image tag, regardless of whether a local repo was specified. The latest GitHub tag (`v0.7.0-alpha.1`) overrode the chart's pinned `v0.6.0-rc.5` tags.

## Test plan

- [ ] Run `./scripts/ocp/setup-kagenti.sh --kagenti-repo . --with-all --skip-mlflow` from a checkout on `release-0.6` — should use `v0.6.0-rc.5`
- [ ] Run without `--kagenti-repo` — should still fetch latest tag from GitHub (existing behavior)

Assisted-By: Claude Code

### DIFF
--- a/scripts/ocp/setup-kagenti.sh
+++ b/scripts/ocp/setup-kagenti.sh
@@ -1337,8 +1337,15 @@ elif [ -z "$KAGENTI_REPO_ORIGINAL" ]; then
   KAGENTI_UI_FLAGS+=(--set "ui.frontend.tag=v${LATEST_TAG}")
   KAGENTI_UI_FLAGS+=(--set "ui.backend.tag=v${LATEST_TAG}")
 else
-  # Local or explicit repo: trust the chart's pinned image tags
-  log_info "Using image tags from chart values (--kagenti-repo provided)"
+  # Local or explicit repo: read ui.frontend.tag from the chart (all component tags share the same release version)
+  LATEST_TAG=$(grep -A4 'frontend:' "$KAGENTI_REPO/charts/kagenti/values.yaml" | grep -m1 'tag:' | awk '{print $2}')
+  if [ -n "$LATEST_TAG" ]; then
+    log_success "Using tag from local chart: ${LATEST_TAG}"
+    KAGENTI_UI_FLAGS+=(--set "ui.frontend.tag=${LATEST_TAG}")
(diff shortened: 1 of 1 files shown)


scripts/ocp/setup-kagenti.sh
```diff
@@ -1124,15 +1124,24 @@ if $SKIP_UI; then
...
   KAGENTI_UI_FLAGS+=(--set components.ui.enabled=false)
 else
-  log_info "Detecting latest kagenti release tag..."
-  LATEST_TAG=$(git ls-remote --tags --sort="v:refname" https://github.com/kagenti/kagenti.git | tail -n1 | sed 's|.*refs/tags/v||; s/\^{}//')
-  if [ -z "$LATEST_TAG" ]; then
-    log_warn "Could not detect latest tag — using 'latest'"
-    LATEST_TAG="latest"
+  if [[ "$KAGENTI_SOURCE" == *"(local)"* ]]; then
+    # Local repo: use the tag from the chart's values.yaml (already pinned)
+    LATEST_TAG=$(grep -m1 'tag:' "$KAGENTI_REPO/charts/kagenti/values.yaml" | awk '{print $2}')
```
evaline-ju: Suggestion: grep -m1 'tag:' returns the first tag: line in the file — which today is ui.frontend (~line 79 of values.yaml). This works because both ui.frontend.tag and ui.backend.tag share the same release tag and frontend appears first.

The coupling is fragile: any new tag: entry inserted above the frontend section silently picks up the wrong value. Consider making intent explicit: LATEST_TAG=$(yq '.ui.frontend.tag' "$KAGENTI_REPO/charts/kagenti/values.yaml") or without yq: LATEST_TAG=$(grep -A2 'frontend:' "$KAGENTI_REPO/charts/kagenti/values.yaml" | grep -m1 'tag:' | awk '{print $2}')

A brief comment noting which tag is being read would help future devs reading.

evaline-ju (approved):
LGTM with suggestion

PR #1642, opened 2026-05-22
docs: restructure sandbox guide with unified steps

pdettori:
## Summary
- Reorganize sandbox guide so platform-specific instructions (Kind vs OpenShift) are isolated to Step 1 only — all subsequent steps (shared infra, tenant deploy, CLI config, provider, sandbox) are unified
- Move OpenShift self-signed certificates section to the end as a platform-specific appendix instead of nesting it under CLI configuration
- Fix `configure-cli.sh` idempotency bug: replace non-existent `openshell gateway destroy` with the correct `openshell gateway remove` command
- Add troubleshooting entry for the "Gateway already exists" error when re-running `configure-cli.sh`

## Test plan
- [ ] Follow guide on Kind: `scripts/kind/setup-kagenti.sh --with-agent-sandbox --with-spire`, then steps 2–8
- [ ] Verify `configure-cli.sh` handles re-registration cleanly (run twice in a row)
- [ ] Verify OpenShift section links correctly from the note in Step 1

Assisted-By: Claude Code

### DIFF
--- a/docs/sandbox-guide.md
+++ b/docs/sandbox-guide.md
@@ -141,7 +141,8 @@ preconfigured users:
 ## Step 6: Create a Provider
 
 Providers define how the sandbox connects to an LLM. You must be logged in as an
-admin user.
+admin user. The guide works with any OpenAI-compatible or Anthropic-compatible
+endpoint — you just supply your own base URL and API key.
 
 ```bash
 export ANTHROPIC_AUTH_TOKEN="your-api-key-here"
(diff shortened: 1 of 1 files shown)


docs/sandbox-guide.md
```diff
@@ -210,7 +153,7 @@ openshell provider create --name claude --type anthropic \
 # for example
 openshell provider create --name claude --type anthropic \
   --credential ANTHROPIC_AUTH_TOKEN \
-  --config ANTHROPIC_BASE_URL=https://ete-litellm.bx.cloud9.ibm.com 
+  --config ANTHROPIC_BASE_URL=https://ete-litellm.bx.cloud9.ibm.com
```
rubambiza: Does this imply that the guide is only useful for folks at IBM or is this plug-and-play if someone has another provider? If the latter, I'd recommend documenting what the process looks like for someone using other providers.

rubambiza (approved):
Clean documentation reorganization. Platform-specific instructions are now isolated to Step 1, all subsequent steps are unified, and the OpenShift self-signed certificates section is correctly moved to an appendix. Cross-references are valid (verified `#openshift-with-self-signed-certificates` anchor). The new troubleshooting entry for "Gateway already exists" is a useful addition.

**Areas reviewed**: Documentation structure, cross-references, command accuracy
**Commits**: 1, signed-off
**CI status**: All passing

Assisted-By: Claude Code

PR #1644, opened 2026-05-22
fix(ocp): MLflow integration gaps on RHOAI

### LINKED ISSUE #1637: fix(ocp): MLflow integration gaps on RHOAI
## Summary

The MLflow integration in the OCP installer (`scripts/ocp/setup-kagenti.sh`) has several gaps when running against the current RHOAI MLflow operator version.

### 1. `status.url` never populated — installer waits 5 minutes unnecessarily

The `_mlflow_wait_ready()` function waits up to 5 minutes for `status.url` to be set on the MLflow CR. However, the RHOAI MLflow operator does not create a Route/Ingress, so this field is never populated. The installer eventually falls back to the in-cluster endpoint (`http://mlflow.mlflow.svc.cluster.local:8080`), but the 5-minute wait is unnecessary.

**Suggested fix:** Detect earlier that the MLflow pod is Running + CR status is "Available" and skip the `status.url` wait, going directly to the in-cluster fallback.

### 2. ClusterRole name mismatch in `_mlflow_grant_otel_rbac()`

The function expects `mlflow-operator-mlflow-integration` but RHOAI provides:
- `mlflow-operator-mlflow-edit`
- `mlflow-operator-mlflow-view`

This causes the RBAC grant to fail silently (or with a warning).

**Suggested fix:** Check for the actual ClusterRole names provided by the installed operator version.

### 3. OTEL collector MLflow exporter not wired by installer

The kagenti-deps OTEL collector template conditionally includes the MLflow exporter based on `components.mlflow.enabled` and `mlflow.auth.enabled` Helm values, but the OCP installer never sets these values during `helm upgrade`. As a result, the OTEL collector only has the `debug` exporter active — no traces are forwarded to MLflow.

**Suggested fix:** When `--with-mlflow` (or `--with-all`) is used, pass the appropriate Helm values to enable the MLflow exporter in the OTEL collector config.

## Workaround

Use `--skip-mlflow` when running the installer to bypass MLflow provisioning entirely:

```bash
./scripts/ocp/setup-kagenti.sh --with-all --skip-mlflow
```

## Environment

- RHOAI MLflow operator (OpenShift AI)
- Kagenti v0.6.0-rc5 / release-0.6 branch

pdettori:
## Summary

Resolves multiple issues with the RHOAI MLflow integration on OpenShift:

- **Fix installer wait logic**: Detect `Available` condition on the MLflow CR early — RHOAI never populates `status.url` (no Route/Ingress created), eliminating a 5-minute timeout
- **Fix ClusterRole name mismatch**: Try `mlflow-operator-mlflow-edit` (actual RHOAI name) before falling back to `mlflow-operator-mlflow-integration`
- **Wire OTEL MLflow exporter**: Pass `otel.mlflow.enabled=true` and the resolved traces endpoint during `helm upgrade --reset-values`
- **Fix TLS verification**: Mount `service-ca.crt` for RHOAI MLflow TLS, add `bearertokenauth/mlflow` extension using projected SA token
- **Fix trace rendering**: Upgrade RHOAI subscription channel from `fast-3.x` (3.3.1, ships MLflow 3.6.x) to `stable-3.4` (3.4.0, ships MLflow 3.10.1) which includes the fix for the "Trace failed to render" UI bug ([MLflow PR #20596](https://github.com/mlflow/mlflow/pull/20596))
- **Expose MLflow UI**: Deploy OAuth proxy route for browser access to the MLflow dashboard
- **Add RHOAI MLflow docs**: Document the integration setup, authentication flow, version requirements, and OTEL collector configuration differences

### Root cause of trace rendering failure

RHOAI 3.3.x ships MLflow 3.6.x which has a client-side JavaScript bug: the trace detail view enters an infinite fetch loop when `num_spans` metadata (set by the OTLP ingestion path) does not match what the trace viewer JS expects. All API calls return 200 OK — the data is served correctly, but the JS fails to parse the span tree. Fixed in MLflow 3.10.0.

### OTEL collector to RHOAI MLflow authentication

The collector uses its projected ServiceAccount token with `bearertokenauth` extension. MLflow `kubernetes-auth` plugin validates the token via Kubernetes SubjectAccessReview. TLS is verified using the OpenShift `service-ca.crt` (auto-injected into pods). Compression is disabled (RHOAI incompatibility) and retry is disabled (RHOAI returns JSON body with `Content-Type: application/x-protobuf`, causing parse errors on retry).

Fixes #1637

## Test plan

- [x] Run installer with --with-mlflow on RHOAI cluster — no 5-min stall
- [x] OTEL collector configmap includes otlphttp/mlflow exporter with bearertokenauth
- [x] RoleBinding references mlflow-operator-mlflow-edit ClusterRole
- [x] Traces ingested via OTEL collector land in MLflow (POST /v1/traces 200 OK)
- [x] MLflow 3.10.1 running after RHOAI 3.4.0 upgrade
- [x] Trace detail view renders correctly in MLflow UI (user verification)

Assisted-By: Claude Code

### DIFF
--- a/scripts/ocp/setup-kagenti.sh
+++ b/scripts/ocp/setup-kagenti.sh
@@ -441,7 +441,6 @@ _mlflow_grant_otel_rbac() {
         break
       fi
     done
-    [ -n "$cr_name" ] && break
     tries=$((tries + 1))
     if [ $tries -ge 24 ]; then
       log_warn "No MLflow ClusterRole found (tried: ${candidates[*]}) after 2m — MLflow OTEL RBAC skipped"
@@ -1147,9 +1146,15 @@ EOF
   # SubjectAccessReview.
   log_info "Ensuring MLflow OAuth proxy exists for browser access..."
   if ! $KUBECTL get deployment mlflow-oauth-proxy -n "$MLFLOW_NAMESPACE" &>/dev/null; then
+    local oauth_proxy_image
+    oauth_proxy_image=$(oc adm release info --image-for=oauth-proxy 2>/dev/null)
+    if [ -z "$oauth_proxy_image" ]; then
+      log_warn "Could not resolve oauth-proxy image — skipping MLflow OAuth proxy"
+      return 0
+    fi
     # Cookie secret for oauth-proxy session encryption
     local cookie_secret
-    cookie_secret=$(openssl rand -base64 32 | tr -d '\n' | head -c 32 | base64)
+    cookie_secret=$(head -c 32 /dev/urandom | base64 | tr -d '\n')
     $KUBECTL apply -f - <<OAUTH_EOF
 apiVersion: v1
 kind: ServiceAccount
@@ -1186,7 +1191,7 @@ spec:
       serviceAccountName: mlflow-oauth-proxy
       containers:
       - name: oauth-proxy
-        image: $(oc adm release info --image-for=oauth-proxy 2>/dev/null)
+        image: ${oauth_proxy_image}
         args:
         - --https-address=:8443
         - --provider=openshift
@@ -1426,12 +1431,13 @@ else
     if [ -n "$_EXP_TOKEN" ]; then
       _EXP_RESP=$($KUBECTL run mlflow-exp-create --rm -i --restart=Never \
         --image=curlimages/curl -n kagenti-system \
-        -- curl -sk -X POST \
-        -H "Authorization: Bearer $_EXP_TOKEN" \
-        -H "x-mlflow-workspace: $_EXP_WS" \
+        --env="TOK=$_EXP_TOKEN" \
+        -- sh -c 'curl -sk -X POST \
+        -H "Authorization: Bearer $TOK" \
+        -H "x-mlflow-workspace: '"$_EXP_WS"'" \
         -H "Content-Type: application/json" \
-        -d "{\"name\":\"$_EXP_NAME\"}" \
-        "https://mlflow.${MLFLOW_NAMESPACE}.svc.cluster.local:8443/api/2.0/mlflow/experiments/create" 2>/dev/null)
+        -d "{\"name\":\"'"$_EXP_NAME"'\"}" \
+        "https://mlflow.'"${MLFLOW_NAMESPACE}"'.svc.cluster.local:8443/api/2.0/mlflow/experiments/create"' 2>/dev/null)
       if echo "$_EXP_RESP" | grep -q "experiment_id"; then
         _EXP_ID=$(echo "$_EXP_RESP" | python3 -c "import sys,json; print(json.load(sys.stdin)['experiment_id'])" 2>/dev/null)
         log_success "Created MLflow experiment '$_EXP_NAME' (id=$_EXP_ID) in workspace $_EXP_WS"
--- a/scripts/ocp/setup-kagenti.sh
+++ b/scripts/ocp/setup-kagenti.sh
@@ -1414,6 +1414,34 @@ if [ "$SKIP_MLFLOW" = true ]; then
   log_success "Skipping MLflow RBAC grant (--skip-mlflow)"
 else
   _mlflow_grant_otel_rbac
+
+  # Create default MLflow experiment in the workspace namespace.
+  # The otel-collector sends traces with x-mlflow-experiment-id header; the experiment
+  # must exist or MLflow will reject the traces.
+  # Uses kubectl run with a curl pod since the installer runs outside the cluster.
+  if ! $DRY_RUN; then
+    _EXP_WS="${MLFLOW_WORKSPACE:-team1}"
+    _EXP_NAME="kagenti-traces"
+    _EXP_TOKEN=$($KUBECTL create token otel-collector -n kagenti-system --duration=600s 2>/dev/null)
+    if [ -n "$_EXP_TOKEN" ]; then
+      _EXP_RESP=$($KUBECTL run mlflow-exp-create --rm -i --restart=Never \
+        --image=curlimages/curl -n kagenti-system \
+        -- curl -sk -X POST \
+        -H "Authorization: Bearer $_EXP_TOKEN" \
+        -H "x-mlflow-workspace: $_EXP_WS" \
+        -H "Content-Type: application/json" \
+        -d "{\"name\":\"$_EXP_NAME\"}" \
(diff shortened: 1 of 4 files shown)


scripts/ocp/setup-kagenti.sh
```diff
@@ -1088,12 +1134,145 @@ otel:
...
+
+  # Expose MLflow UI via an OpenShift OAuth proxy so browser users can
+  # authenticate through OpenShift SSO. The proxy injects the user's token
+  # as a bearer token which MLflow's kubernetes-auth plugin validates via
+  # SubjectAccessReview.
+  log_info "Ensuring MLflow OAuth proxy exists for browser access..."
+  if ! $KUBECTL get deployment mlflow-oauth-proxy -n "$MLFLOW_NAMESPACE" &>/dev/null; then
+    # Cookie secret for oauth-proxy session encryption
+    local cookie_secret
+    cookie_secret=$(openssl rand -base64 32 | tr -d '\n' | head -c 32 | base64)
```
huang195: Cookie secret generation is more convoluted than it needs to be: `openssl rand -base64 32 | tr -d '\n' | head -c 32 | base64` truncates to 32 base64-alphabet characters and re-encodes them — the cookie ends up with ~6 bits/char of entropy instead of 8. Functionally OK (oauth-proxy accepts 16/24/32-byte secrets), but simpler with full entropy:

```bash
cookie_secret=$(head -c 32 /dev/urandom | base64 | tr -d '\n')
```

scripts/ocp/setup-kagenti.sh
```diff
@@ -1088,12 +1134,145 @@ otel:
...
+      app: mlflow-oauth-proxy
+  template:
+    metadata:
+      labels:
+        app: mlflow-oauth-proxy
+    spec:
+      serviceAccountName: mlflow-oauth-proxy
+      containers:
+      - name: oauth-proxy
+        image: $(oc adm release info --image-for=oauth-proxy 2>/dev/null)
```
huang195: `image: $(oc adm release info --image-for=oauth-proxy 2>/dev/null)` is interpolated at heredoc expansion time, and stderr is silenced. If the lookup fails (disconnected env, auth issue, missing release manifest access) the YAML applied to the cluster has an empty `image:` field and `kubectl apply` returns a confusing schema error rather than the real cause.

Suggest resolving and validating before the heredoc:

```bash
local oauth_proxy_image
oauth_proxy_image=$(oc adm release info --image-for=oauth-proxy)
if [ -z "$oauth_proxy_image" ]; then
  log_warn "Could not resolve oauth-proxy image — skipping MLflow OAuth proxy"
  return 0
fi
```

scripts/ocp/setup-kagenti.sh
```diff
@@ -1235,6 +1414,34 @@ if [ "$SKIP_MLFLOW" = true ]; then
...
+  # Uses kubectl run with a curl pod since the installer runs outside the cluster.
+  if ! $DRY_RUN; then
+    _EXP_WS="${MLFLOW_WORKSPACE:-team1}"
+    _EXP_NAME="kagenti-traces"
+    _EXP_TOKEN=$($KUBECTL create token otel-collector -n kagenti-system --duration=600s 2>/dev/null)
+    if [ -n "$_EXP_TOKEN" ]; then
+      _EXP_RESP=$($KUBECTL run mlflow-exp-create --rm -i --restart=Never \
+        --image=curlimages/curl -n kagenti-system \
+        -- curl -sk -X POST \
+        -H "Authorization: Bearer $_EXP_TOKEN" \
```
huang195: The MLflow SA bearer token is interpolated into `kubectl run -- curl ... -H "Authorization: Bearer $_EXP_TOKEN"` as command-line args, so it's visible in the curl pod's `argv` (anyone with `pods/exec` or `pods/log` access in `kagenti-system`). The token is short-lived (600s) so risk is bounded, but consider passing via env or stdin to keep it out of process arguments — e.g. `--env=TOK=$_EXP_TOKEN` and reference `$TOK` inside `sh -c`.

scripts/ocp/setup-kagenti.sh
```diff
@@ -401,16 +427,24 @@ _mlflow_wait_ready() {
...
   local tries=0
-  while ! $KUBECTL get clusterrole "$cr_name" &>/dev/null; do
+  while [ -z "$cr_name" ]; do
+    for candidate in "${candidates[@]}"; do
+      if $KUBECTL get clusterrole "$candidate" &>/dev/null; then
+        cr_name="$candidate"
+        break
+      fi
+    done
+    [ -n "$cr_name" ] && break
```
huang195: Nit: `[ -n "$cr_name" ] && break` is redundant — the outer `while [ -z "$cr_name" ]` already terminates on the same condition once the inner `for` sets `cr_name`.

scripts/ocp/setup-kagenti.sh
```diff
@@ -1088,12 +1134,145 @@ otel:
...
+      break
+    fi
+    sleep 5
+  done
+
+  local mlflow_host=""
+  mlflow_host=$($KUBECTL get route mlflow -n "$MLFLOW_NAMESPACE" \
+    -o jsonpath='{.spec.host}' 2>/dev/null || echo "")
+  if [ -n "$mlflow_host" ]; then
+    local mlflow_url="https://${mlflow_host}/mlflow/"
```
huang195, line 1274: Nit: `mlflow_url="https://${mlflow_host}/mlflow/"` — confirm the `/mlflow/` subpath is correct for RHOAI MLflow. If the upstream serves at `/`, the dashboard link will 404. (Test plan checkbox suggests it works in the user's env, just flagging in case path varies by RHOAI version.)

huang195 (approved):
## Summary

Solid, targeted fix-up of the RHOAI MLflow integration. Correctly handles the RHOAI operator's no-`status.url` behavior, accommodates the renamed `mlflow-edit` ClusterRole, and documents the version cliff at MLflow 3.10.0.

**Areas reviewed**: Helm/values, Shell (setup-kagenti.sh), Docs
**Commits**: 5, all signed-off
**CI status**: passing (E2E HyperShift gated on `/run-e2e` — normal)

Approving — the inline notes are cleanup suggestions, none blocking.

Assisted-By: Claude Code

scripts/ocp/setup-kagenti.sh
pdettori: yes - this is confirmed by manual test to work

PR #1646, opened 2026-05-22
fix(openshell): use correct CLI subcommand to remove gateway registration

### LINKED ISSUE #1645: configure-cli.sh uses non-existent 'gateway destroy' subcommand, fails on re-run
## Bug

`scripts/openshell/configure-cli.sh` fails when the gateway is already registered with the local CLI (e.g. on second run):

```
→ Registering gateway with openshell CLI
⚠ Gateway openshell-team1 already exists — removing to re-register
Error:   × Gateway 'openshell-team1' already exists.
  │ Remove it first with: openshell gateway remove openshell-team1
  │ Or choose a different name with: --name <name>
```

## Root Cause

Line 147 uses `openshell gateway destroy --name "$GATEWAY_NAME"` which is not a valid CLI subcommand. The correct command is `openshell gateway remove <name>` (positional argument, no `--name` flag).

The `|| true` suppresses the error from the invalid `destroy` command, so the gateway is never actually removed. The subsequent `gateway add` then fails because it still exists.

## Expected Behavior

Running `configure-cli.sh` multiple times should be idempotent — it removes the existing registration and re-adds it with fresh parameters and certs.

## Fix

Change:
```bash
openshell gateway destroy --name "$GATEWAY_NAME" 2>/dev/null || true
```
To:
```bash
openshell gateway remove "$GATEWAY_NAME" 2>/dev/null || true
```

Assisted-By: Claude Code

pdettori:
## Summary

- Fix `configure-cli.sh` using non-existent `openshell gateway destroy --name` subcommand
- Replace with correct `openshell gateway remove <name>` (positional argument)
- Script is now idempotent on re-runs as originally intended

## Root Cause

The removal command `openshell gateway destroy --name "$GATEWAY_NAME"` silently failed (suppressed by `|| true`), so the gateway was never removed. The subsequent `gateway add` then failed with "Gateway already exists."

## Test plan

- [x] Run `scripts/openshell/configure-cli.sh team1` on a fresh CLI state — gateway registered successfully
- [x] Run it again — existing gateway removed and re-registered without error
- [x] Verified `openshell gateway remove --help` confirms the correct syntax

Fixes #1645

Assisted-By: Claude Code

### DIFF
--- a/scripts/openshell/configure-cli.sh
+++ b/scripts/openshell/configure-cli.sh
@@ -144,7 +144,7 @@ if $DRY_RUN; then
 else
   if openshell gateway info --gateway "$GATEWAY_NAME" &>/dev/null; then
     log_warn "Gateway $GATEWAY_NAME already exists — removing to re-register"
-    openshell gateway destroy --name "$GATEWAY_NAME" 2>/dev/null || true
+    openshell gateway remove "$GATEWAY_NAME" 2>/dev/null || true
   fi
   openshell "${ADD_ARGS[@]}"
   log_success "Gateway registered"


huang195 (approved):
Surgical 1-line fix: `openshell gateway destroy --name "$NAME"` (non-existent subcommand, silently swallowed by `|| true`) → `openshell gateway remove "$NAME"` (positional). Restores the script's intended idempotence on re-runs.

21 CI checks passing; OpenShell PoC and Deploy & Test still pending — those exercise this very script, so worth confirming they go green before merge.

PR #1649, opened 2026-05-22
docs: fix OpenShift section in sandbox guide and add troubleshooting

pdettori:
## Summary

- Rewrites the OpenShift section to provide explicit `deploy-shared.sh` and `deploy-tenant.sh` steps instead of incorrectly claiming the OCP installer handles OpenShell deployment
- Adds troubleshooting entries for two common OCP failures:
  - `deploy-tenant.sh` hanging because the `openshell` Keycloak realm doesn't exist
  - `deploy-shared.sh` hanging due to RHBK bootstrap admin credential mismatch after pod recreation
- Adds a detailed recovery procedure for the Keycloak credential mismatch

## Test plan

- [x] Verified end-to-end on live OCP cluster (kagenti1.kubestellar.org)
- [x] Confirmed `deploy-shared.sh` → `deploy-tenant.sh` → `configure-cli.sh` → `sandbox create` flow works after fixes

Assisted-By: Claude Code

### DIFF
--- a/docs/sandbox-guide.md
+++ b/docs/sandbox-guide.md
@@ -288,24 +288,35 @@ credentials. This causes `deploy-shared.sh` to hang at the kcadm login step.
 To fix, reset the admin password via the database:
 
 ```bash
-# Generate a new password hash matching the secret
+# Read username and password from the secret (avoids hardcoding env-specific values)
+KC_USER=$(kubectl get secret keycloak-initial-admin -n keycloak \
+  -o jsonpath='{.data.username}' | base64 -d)
 KC_PASS=$(kubectl get secret keycloak-initial-admin -n keycloak \
   -o jsonpath='{.data.password}' | base64 -d)
 
+# Find the admin user ID
+ADMIN_ID=$(kubectl exec -n keycloak postgres-kc-0 -- psql -U postgres -d postgres -tA -c \
+  "SELECT id FROM user_entity ue JOIN realm r ON ue.realm_id=r.id WHERE r.name='master' AND ue.username='${KC_USER}';")
+
+# Read hash iterations from the existing credential (varies by Keycloak version)
+ITERS=$(kubectl exec -n keycloak postgres-kc-0 -- psql -U postgres -d postgres -tA -c \
+  "SELECT credential_data::json->>'hashIterations' FROM credential WHERE user_id='${ADMIN_ID}';")
+
+# Generate a new password hash matching the secret
 python3 -c "
 import hashlib, base64, os, json
-dk = hashlib.pbkdf2_hmac('sha256', '$KC_PASS'.encode(), (salt := os.urandom(16)), 27500)
+dk = hashlib.pbkdf2_hmac('sha256', '${KC_PASS}'.encode(), (salt := os.urandom(16)), ${ITERS})
 print(json.dumps({'value': base64.b64encode(dk).decode(), 'salt': base64.b64encode(salt).decode(), 'additionalParameters': {}}))
 " > /tmp/kc-secret-data.json
+chmod 600 /tmp/kc-secret-data.json
 
-# Find the admin user ID and update the credential
-ADMIN_ID=$(kubectl exec -n keycloak postgres-kc-0 -- psql -U postgres -d postgres -t -c \
-  "SELECT id FROM user_entity ue JOIN realm r ON ue.realm_id=r.id WHERE r.name='master' AND ue.username='temp-admin';")
-
+# Update the credential in the database
 kubectl exec -n keycloak postgres-kc-0 -- psql -U postgres -d postgres -c \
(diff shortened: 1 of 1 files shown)


docs/sandbox-guide.md
```diff
@@ -248,13 +268,48 @@ openshell sandbox exec -n <sandbox-name> -- claude --print "Hello"
...
+
+python3 -c "
+import hashlib, base64, os, json
+dk = hashlib.pbkdf2_hmac('sha256', '$KC_PASS'.encode(), (salt := os.urandom(16)), 27500)
+print(json.dumps({'value': base64.b64encode(dk).decode(), 'salt': base64.b64encode(salt).decode(), 'additionalParameters': {}}))
+" > /tmp/kc-secret-data.json
+
+# Find the admin user ID and update the credential
+ADMIN_ID=$(kubectl exec -n keycloak postgres-kc-0 -- psql -U postgres -d postgres -t -c \
+  "SELECT id FROM user_entity ue JOIN realm r ON ue.realm_id=r.id WHERE r.name='master' AND ue.username='temp-admin';")
```
huang195: Both `temp-admin` and `hashIterations=27500` (line 297, also embedded in `credential_data` on line 307) are environment-dependent and may drift between RHBK / upstream Keycloak versions. Suggest fetching them dynamically so the recipe survives version changes:

```bash
# Username from the secret rather than hardcoding 'temp-admin'
KC_USER=$(kubectl get secret keycloak-initial-admin -n keycloak \
  -o jsonpath='{.data.username}' | base64 -d)

# Read iterations from the existing credential row instead of hardcoding 27500
ITERS=$(kubectl exec -n keycloak postgres-kc-0 -- psql -U postgres -d postgres -tA -c \
  "SELECT credential_data::json->>'hashIterations' FROM credential WHERE user_id='${ADMIN_ID}';")
```

Then interpolate `$KC_USER` and `$ITERS` into the SELECT/UPDATE. Modern Keycloak defaults are much higher (600,000+); 27500 will silently produce a hash the runtime rejects on those builds.

docs/sandbox-guide.md
```diff
@@ -248,13 +268,48 @@ openshell sandbox exec -n <sandbox-name> -- claude --print "Hello"
...
+```bash
+# Generate a new password hash matching the secret
+KC_PASS=$(kubectl get secret keycloak-initial-admin -n keycloak \
+  -o jsonpath='{.data.password}' | base64 -d)
+
+python3 -c "
+import hashlib, base64, os, json
+dk = hashlib.pbkdf2_hmac('sha256', '$KC_PASS'.encode(), (salt := os.urandom(16)), 27500)
+print(json.dumps({'value': base64.b64encode(dk).decode(), 'salt': base64.b64encode(salt).decode(), 'additionalParameters': {}}))
+" > /tmp/kc-secret-data.json
```
huang195, line 310: `/tmp/kc-secret-data.json` contains a hashed admin credential and is never cleaned up. Add a trailing `rm -f /tmp/kc-secret-data.json` after the UPDATE so it doesn't sit on shared bastions / dev laptops. (Also worth a `umask 077` or `chmod 600` if you don't switch to passing the JSON via stdin.)

docs/sandbox-guide.md
```diff
@@ -248,13 +268,48 @@ openshell sandbox exec -n <sandbox-name> -- claude --print "Hello"
...
+
+# Find the admin user ID and update the credential
+ADMIN_ID=$(kubectl exec -n keycloak postgres-kc-0 -- psql -U postgres -d postgres -t -c \
+  "SELECT id FROM user_entity ue JOIN realm r ON ue.realm_id=r.id WHERE r.name='master' AND ue.username='temp-admin';")
+
+kubectl exec -n keycloak postgres-kc-0 -- psql -U postgres -d postgres -c \
+  "UPDATE credential SET secret_data='$(cat /tmp/kc-secret-data.json)',
+   credential_data='{\"hashIterations\":27500,\"algorithm\":\"pbkdf2-sha256\",\"additionalParameters\":{}}'
+   WHERE user_id='${ADMIN_ID// /}';"
```
huang195: Nit: `${ADMIN_ID// /}` only strips spaces — `psql -t` also emits a trailing newline and possibly a leading tab depending on locale. `$(... | tr -d '[:space:]')` (or using `psql -tA`) is more robust and reads more clearly as "strip whitespace".

huang195 (approved):
## Summary

Useful docs fix — replaces the inaccurate "installer handles OpenShell" line with explicit `deploy-shared.sh` / `deploy-tenant.sh` steps, and the new troubleshooting entries match common failure modes (gateway crashloop on missing realm, RHBK admin credential mismatch).

**Areas reviewed**: Docs, embedded shell/SQL recovery recipe
**Commits**: 1, signed-off
**CI status**: passing

Approving — the inline notes are about making the Keycloak recovery procedure less brittle for slightly-different environments. None blocking.

Assisted-By: Claude Code

PR #1651, opened 2026-05-22
fix(ci): protect referenced child manifests from cleanup deletion

pdettori:
## Summary

- Fixes the container image cleanup workflow that was deleting platform-specific child manifests (amd64, arm64, attestations) referenced by protected tagged images
- Before deleting untagged manifests, the workflow now fetches each protected tag's OCI manifest list via the GHCR registry API and collects all child digests into a "do not delete" set
- If the registry token exchange fails, all untagged deletions are skipped (fail-safe)

## Root Cause

Multi-arch images are OCI Image Indexes. Their platform-specific manifests appear as **untagged** versions in the GitHub Packages API. The original workflow deleted all untagged versions older than 7 days, which broke `docker pull` for every release image older than 7 days (e.g., `v0.6.0-rc.1`).

## Test plan

- [ ] Trigger workflow manually with `dry_run: true` — verify no protected child manifests appear in "Would delete" output
- [ ] Confirm `docker manifest inspect ghcr.io/kagenti/kagenti/spiffe-idp-setup:latest` still resolves after a real run
- [ ] Check that SHA-tagged orphaned untagged manifests ARE still cleaned up
- [ ] Re-push affected release tags to rebuild broken images after merge

Assisted-By: Claude Code

### DIFF
--- a/.github/workflows/cleanup-container-images.yaml
+++ b/.github/workflows/cleanup-container-images.yaml
@@ -48,8 +48,6 @@ jobs:
             const cutoffDate = new Date(Date.now() - retentionDays * 24 * 60 * 60 * 1000);
 
             // Cap deletions to stay well within the 5000 req/hour rate limit.
-            // Each package needs ~1 GET per 100 versions + 1 DELETE per stale version.
-            // Budget: ~4000 requests total (leave headroom for pagination GETs).
             const maxDeletesPerRun = 4000;
 
             // Pattern: any branch prefix followed by a 7-12 char hex SHA
@@ -81,6 +79,49 @@ jobs:
               }
             }
 
+            // Resolve child digests referenced by protected manifest lists.
+            // Multi-arch images are OCI Image Indexes whose platform-specific
+            // child manifests appear as untagged versions in the Packages API.
+            // We must not delete those or the tagged image breaks.
+            async function getProtectedDigests(pkg, protectedTags) {
+              const digests = new Set();
+              if (protectedTags.length === 0) return digests;
+
+              const tokenRes = await fetch(
+                `https://ghcr.io/token?service=ghcr.io&scope=repository:kagenti/kagenti/${pkg}:pull`,
+                { headers: { 'Authorization': `Basic ${Buffer.from(`x:${process.env.GITHUB_TOKEN}`).toString('base64')}` } }
+              );
+              if (!tokenRes.ok) {
+                console.log(`Warning: could not get registry token for ${pkg}, skipping untagged cleanup`);
+                return null;
+              }
+              const { token } = await tokenRes.json();
+
+              for (const tag of protectedTags) {
+                try {
+                  const res = await fetch(
+                    `https://ghcr.io/v2/kagenti/kagenti/${pkg}/manifests/${tag}`,
+                    {
+                      headers: {
+                        'Authorization': `Bearer ${token}`,
+                        'Accept': 'application/vnd.oci.image.index.v1+json, application/vnd.docker.distribution.manifest.list.v2+json'
+                      }
+                    }
+                  );
+                  if (!res.ok) continue;
+                  const manifest = await res.json();
+                  if (manifest.manifests) {
+                    for (const m of manifest.manifests) {
+                      digests.add(m.digest);
+                    }
+                  }
+                } catch (e) {
+                  console.log(`Warning: could not resolve manifest for ${tag}: ${e.message}`);
+                }
+              }
+              return digests;
+            }
+
             // Rotate package order by day-of-year so cap pressure is spread
             // fairly across all packages during backlog drain.
             const dayOfYear = Math.floor((Date.now() - new Date(new Date().getFullYear(), 0, 0).getTime()) / 86400000);
@@ -94,11 +135,12 @@ jobs:
               console.log(`\nPackage: ${package_name}`);
               console.log('---');
 
-              // Check rate limit before listing pages for this package
               await checkRateLimit();
 
-              // Phase 1: collect versions to delete
-              const toDelete = [];
+              // Phase 1: iterate versions, classify tagged vs untagged
+              const taggedToDelete = [];
+              const untaggedCandidates = [];
+              const protectedTags = [];
               let skipped = 0;
               let page = 1;
               const perPage = 100;
@@ -120,12 +162,13 @@ jobs:
 
                   if (tags.length === 0) {
                     if (createdAt < cutoffDate) {
-                      toDelete.push({ id: version.id, label: `untagged ${version.id}`, createdAt });
+                      untaggedCandidates.push({ id: version.id, name: version.name, label: `untagged ${version.id}`, createdAt });
                     }
                     continue;
                   }
 
                   if (tags.some(tag => protectedTagPattern.test(tag))) {
+                    protectedTags.push(...tags.filter(tag => protectedTagPattern.test(tag)));
                     skipped++;
                     continue;
                   }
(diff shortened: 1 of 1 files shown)


.github/workflows/cleanup-container-images.yaml
```diff
@@ -81,6 +79,49 @@ jobs:
...
+            // Multi-arch images are OCI Image Indexes whose platform-specific
+            // child manifests appear as untagged versions in the Packages API.
+            // We must not delete those or the tagged image breaks.
+            async function getProtectedDigests(pkg, protectedTags) {
+              const digests = new Set();
+              if (protectedTags.length === 0) return digests;
+
+              const tokenRes = await fetch(
+                `https://ghcr.io/token?service=ghcr.io&scope=repository:kagenti/kagenti/${pkg}:pull`,
+                { headers: { 'Authorization': `Basic ${Buffer.from(`x:${process.env.GITHUB_TOKEN}`).toString('base64')}` } }
```
huang195, line 96: **Must-fix.** `process.env.GITHUB_TOKEN` is `undefined` inside `actions/github-script`. The action exposes the token only via the `github`/`octokit` Octokit client — it never writes `GITHUB_TOKEN` to `process.env` (confirmed in [main.ts](https://github.com/actions/github-script/blob/main/src/main.ts), which calls `core.getInput('github-token')` and passes it to `getOctokit`, with nothing setting `process.env`).

So `Buffer.from("x:" + process.env.GITHUB_TOKEN)` becomes the literal string `"x:undefined"`, the GHCR `/token` endpoint rejects it, `tokenRes.ok` is false, `getProtectedDigests` returns `null`, and the fail-safe skips every untagged version on every package. The intent — "delete orphaned untagged but preserve referenced child manifests" — silently degrades to "skip all untagged cleanup."

Two-line fix:

```yaml
- name: Cleanup stale SHA-tagged container images
  uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea
  env:
    GITHUB_TOKEN: ${{ github.token }}
  with:
    script: |
      ...
```

This would also be visible in your existing test-plan checkbox: a `dry_run: true` run today would log `Skipping N untagged versions (could not resolve protected digests)` for every package. Worth executing that step against a feature branch before merge.

.github/workflows/cleanup-container-images.yaml
```diff
@@ -81,6 +79,49 @@ jobs:
...
+            // Resolve child digests referenced by protected manifest lists.
+            // Multi-arch images are OCI Image Indexes whose platform-specific
+            // child manifests appear as untagged versions in the Packages API.
+            // We must not delete those or the tagged image breaks.
+            async function getProtectedDigests(pkg, protectedTags) {
+              const digests = new Set();
+              if (protectedTags.length === 0) return digests;
+
+              const tokenRes = await fetch(
+                `https://ghcr.io/token?service=ghcr.io&scope=repository:kagenti/kagenti/${pkg}:pull`,
```
huang195: Suggestion: `kagenti/kagenti/${pkg}` is hardcoded here and on the manifests URL below. Use `${{ github.repository }}` (returns `owner/repo`) for portability — this workflow is otherwise generic and could be reused as-is in a sibling repo. Same for the `manifests/${tag}` URL.

.github/workflows/cleanup-container-images.yaml
```diff
@@ -81,6 +79,49 @@ jobs:
...
+              if (!tokenRes.ok) {
+                console.log(`Warning: could not get registry token for ${pkg}, skipping untagged cleanup`);
+                return null;
+              }
+              const { token } = await tokenRes.json();
+
+              for (const tag of protectedTags) {
+                try {
+                  const res = await fetch(
+                    `https://ghcr.io/v2/kagenti/kagenti/${pkg}/manifests/${tag}`,
```
huang195: Nit: defense in depth — `manifests/${encodeURIComponent(tag)}` in case a future tag contains `+`, `:` or another character valid in OCI tags but reserved in URLs. Not an issue for current tag conventions, but URL-encoding here is essentially free.

huang195 (changes requested):
## Summary

The phasing into `taggedToDelete` / `untaggedCandidates` / `protectedDigests` is correct, the fail-safe direction is right, and the OCI manifest list parsing handles both single-arch (406) and multi-arch indexes cleanly. The blocker is `process.env.GITHUB_TOKEN`: `actions/github-script` does NOT auto-expose it, so the registry token fetch fails on every run, `getProtectedDigests` returns `null`, and the fail-safe permanently skips ALL untagged cleanup — the platform-breakage bug is fixed, but the cleanup degrades into a no-op for the untagged backlog.

**Areas reviewed**: GitHub Actions workflow (JavaScript inside actions/github-script)
**Commits**: 1, signed-off
**CI status**: lint/static green; `Deploy & Test` failure is unrelated (this PR only touches the cleanup workflow); E2E HyperShift gated on `/run-e2e`.

Requesting changes on the GITHUB_TOKEN issue. The other two notes are non-blocking.

Assisted-By: Claude Code

.github/workflows/cleanup-container-images.yaml
```diff
@@ -81,6 +79,49 @@ jobs:
...
+                } catch (e) {
+                  console.log(`Warning: could not resolve manifest for ${tag}: ${e.message}`);
+                }
+              }
+              return digests;
+            }
+
             // Rotate package order by day-of-year so cap pressure is spread
             // fairly across all packages during backlog drain.
             const dayOfYear = Math.floor((Date.now() - new Date(new Date().getFullYear(), 0, 0).getTime()) / 86400000);
```
rubambiza, line 136: **Suggestion**: `getProtectedDigests` issues one fetch per protected tag with no pacing. If a package accumulates many protected tags (e.g., 20+ release candidates), this fires 20+ sequential requests against `ghcr.io/v2/` with no rate-limit check in between.

The workflow already has `checkRateLimit()` for GitHub API calls — consider adding a small delay or batch check inside the `for (const tag of protectedTags)` loop to avoid hitting GHCR rate limits on packages with many tags.

Low risk today (most packages have <10 protected tags), but worth a `// TODO` at minimum.

rubambiza, line 136: **Nit**: At L48 in the new code, `await res.json()` is called after `if (!res.ok) continue` — but if the registry returns a 200 with a non-JSON body (unlikely but possible during GHCR incidents), parsing throws and the outer `catch` at L54 handles it. The error message `could not resolve manifest for ${tag}: ${e.message}` would say something like `Unexpected token` — a bit opaque. Consider `await res.json().catch(() => null)` with a null-check, so the warning message can say `malformed manifest response` explicitly.

rubambiza (commented):
Agreeing with Hai's review — the `GITHUB_TOKEN` env issue is the blocker. Two additional non-blocking notes inline.

The phased approach (classify → resolve protected digests → filter → delete) is well-designed and the fail-safe is correct in principle. Once the token plumbing is fixed, the core logic looks sound.

**Areas reviewed**: GitHub Actions workflow (JavaScript in actions/github-script)
**Commits**: 1, signed-off
**CI status**: Static checks green; Deploy & Test failure unrelated (workflow-only change)

Assisted-By: Claude Code

huang195 (approved):
Mamma mia, perfetto! All five items from the prior round landed cleanly in 2cf5506:

- `GITHUB_TOKEN: ${{ github.token }}` mapped at the step level so `process.env.GITHUB_TOKEN` is actually defined when the registry token exchange runs
- `context.repo` replacing the hardcoded `'kagenti'` (and propagated through `org: owner` on both the list and delete calls)
- `encodeURIComponent(tag)` on the manifest URL
- Defensive JSON parse with `.catch(() => null)` plus an explicit malformed-response warning
- TODO comment acknowledging the >20-tag pacing follow-up

The phased structure (classify → resolve protected digests → filter → delete) reads cleanly, and the `protectedDigests === null` fail-safe correctly degrades to "skip this package's untagged cleanup" rather than nuking children. Bellissimo. Clearing my prior CHANGES_REQUESTED.

**Areas reviewed**: GitHub Actions workflow (JavaScript inside actions/github-script)
**CI status**: All static checks green; E2E HyperShift gated on `/run-e2e`

Assisted-By: Claude Code

PR #1652, opened 2026-05-22
fix(release): automate image tag pinning across both charts

pdettori:
## Summary

- Adds `scripts/pin-release-tags.sh` — a single command that pins all kagenti-built image tags across both `charts/kagenti/` and `charts/kagenti-deps/` values files
- Extends `scripts/check-release-pins.sh` to validate `kagenti-deps` and detect tag drift between the two charts  
- Fixes `build.yaml` to also pin `mlflow-oauth-secret` and `spiffe-idp-setup` at OCI chart-publish time
- Updates release documentation and the `/release` skill to reference the new automation

## Problem

The `spiffe-idp-setup` image tag in `charts/kagenti-deps/values.yaml` was pinned at `v0.6.0-rc.1` and never updated through rc.2–rc.5. Users testing rc.5 from a repo checkout (the standard install path) pulled the stale rc.1 image.

Root cause: The release pinning process was manual and only covered `charts/kagenti/values.yaml`. The `kagenti-deps` chart and the `mlflow-oauth-secret` image were missed by both:
1. The manual `chore(release): pin image tags` commits
2. The CI `build.yaml` chart-publish job

## How the pin script works

```bash
# Pin all images to a release version
bash scripts/pin-release-tags.sh v0.6.0-rc.6

# Preview without modifying
bash scripts/pin-release-tags.sh v0.6.0-rc.6 --dry-run

# Verify images exist before pinning
bash scripts/pin-release-tags.sh v0.6.0-rc.6 --verify-images
```

## Test plan

- [x] `pin-release-tags.sh --dry-run` shows correct before/after for all 7 images
- [x] `check-release-pins.sh` detects the existing drift (`spiffe-idp-setup` at rc.1 vs platform at alpha.1) 
- [x] Pre-commit hooks pass
- [ ] CI passes on this PR

Assisted-By: Claude Code

### DIFF
--- a/scripts/check-release-pins.sh
+++ b/scripts/check-release-pins.sh
@@ -404,7 +404,7 @@ if [[ -f "$DEPS_VALUES_FILE" ]] && command -v yq >/dev/null 2>&1; then
 
         if [[ -n "$platform_tag" ]] && [[ "$platform_tag" != "null" ]] \
            && [[ "$spiffe_tag" != "$platform_tag" ]]; then
-            add_warning "charts/kagenti-deps/values.yaml" \
+            add_error "charts/kagenti-deps/values.yaml" \
                 "spiffeIdp.image.tag ($spiffe_tag) differs from platform tag ($platform_tag) — run scripts/pin-release-tags.sh"
         else
             add_ok "charts/kagenti-deps/values.yaml" \
--- a/scripts/check-release-pins.sh
+++ b/scripts/check-release-pins.sh
@@ -20,6 +20,9 @@ VALUES_FILE="$CHARTS_DIR/values.yaml"
 CHART_FILE="$CHARTS_DIR/Chart.yaml"
 TEMPLATES_DIR="$CHARTS_DIR/templates"
 
+DEPS_CHARTS_DIR="$REPO_ROOT/charts/kagenti-deps"
+DEPS_VALUES_FILE="$DEPS_CHARTS_DIR/values.yaml"
+
 # ---------------------------------------------------------------------------
 # Options
 # ---------------------------------------------------------------------------
@@ -363,7 +366,55 @@ else
 fi
 
 # ---------------------------------------------------------------------------
-# Check 5 (optional): Verify pinned GHCR images exist in registry
+# Check 5: No "tag: latest" in kagenti-deps values.yaml
+# ---------------------------------------------------------------------------
+
+if [[ ! -f "$DEPS_VALUES_FILE" ]]; then
+    add_warning "charts/kagenti-deps/values.yaml" "file not found (skipping)"
+else
+    matches=$(grep -Fn 'tag: latest' "$DEPS_VALUES_FILE" || true)
+
+    if [[ -z "$matches" ]]; then
+        add_ok "charts/kagenti-deps/values.yaml" "no tag: latest found"
+    else
+        while IFS= read -r match; do
+            if [[ -z "$match" ]]; then
+                continue
+            fi
+
+            line_number="${match%%:*}"
+            add_error "charts/kagenti-deps/values.yaml" "$line_number" "tag: latest"
+        done <<< "$matches"
+    fi
+fi
+
+# ---------------------------------------------------------------------------
+# Check 6: kagenti-deps images built by this repo are pinned consistently
+#
+# The spiffe-idp-setup image lives in kagenti-deps but is built by this repo.
+# Its tag must not drift from the main chart's platform image tags.
+# ---------------------------------------------------------------------------
+
+if [[ -f "$DEPS_VALUES_FILE" ]] && command -v yq >/dev/null 2>&1; then
+    spiffe_tag=$(yq eval '.spiffeIdp.image.tag' "$DEPS_VALUES_FILE" 2>/dev/null || echo "")
+
+    if [[ -n "$spiffe_tag" ]] && [[ "$spiffe_tag" != "null" ]]; then
+        # Compare against the first platform image tag in the main chart
+        platform_tag=$(yq eval '.ui.frontend.tag' "$VALUES_FILE" 2>/dev/null || echo "")
+
+        if [[ -n "$platform_tag" ]] && [[ "$platform_tag" != "null" ]] \
+           && [[ "$spiffe_tag" != "$platform_tag" ]]; then
+            add_warning "charts/kagenti-deps/values.yaml" \
+                "spiffeIdp.image.tag ($spiffe_tag) differs from platform tag ($platform_tag) — run scripts/pin-release-tags.sh"
+        else
+            add_ok "charts/kagenti-deps/values.yaml" \
+                "spiffeIdp.image.tag ($spiffe_tag) consistent with platform"
+        fi
+    fi
+fi
+
+# ---------------------------------------------------------------------------
+# Check 7 (optional): Verify pinned GHCR images exist in registry
 #
 # Extract image: <ref> lines from values.yaml. Only GHCR is checked.
 # docker is required in this mode; fail with a clear message if missing.
--- a/scripts/pin-release-tags.sh
+++ b/scripts/pin-release-tags.sh
@@ -1,0 +1,247 @@
+#!/usr/bin/env bash
+#
+# pin-release-tags.sh — Pin all kagenti-built image tags to a release version.
+#
+# Updates image tags across both charts/kagenti/ and charts/kagenti-deps/ so
+# that a tagged release checkout always pulls the correct images when installed
+# from source (via the setup-kagenti.sh installers).
+#
+# Usage:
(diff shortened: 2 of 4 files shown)


scripts/check-release-pins.sh
```diff
@@ -363,7 +366,55 @@ else
...
+if [[ -f "$DEPS_VALUES_FILE" ]] && command -v yq >/dev/null 2>&1; then
+    spiffe_tag=$(yq eval '.spiffeIdp.image.tag' "$DEPS_VALUES_FILE" 2>/dev/null || echo "")
+
+    if [[ -n "$spiffe_tag" ]] && [[ "$spiffe_tag" != "null" ]]; then
+        # Compare against the first platform image tag in the main chart
+        platform_tag=$(yq eval '.ui.frontend.tag' "$VALUES_FILE" 2>/dev/null || echo "")
+
+        if [[ -n "$platform_tag" ]] && [[ "$platform_tag" != "null" ]] \
+           && [[ "$spiffe_tag" != "$platform_tag" ]]; then
+            add_warning "charts/kagenti-deps/values.yaml" \
```
huang195: The drift-between-charts check is `add_warning`, but this is the exact failure mode the PR exists to prevent (`spiffe-idp-setup` stuck at rc.1 through rc.5 while the platform moved on). A warning won't fail `check-release-pins.sh`, so a release attempt with drift still passes the gate. Consider promoting to `add_error` so CI fails when a release PR has cross-chart tag drift — the message already directs the user to the new pin script.

scripts/pin-release-tags.sh
```diff
@@ -0,0 +1,247 @@
...
+# ---------------------------------------------------------------------------
+
+echo "Pinning image tags to ${VERSION}..."
+echo ""
+
+for entry in "${PINNABLE_IMAGES[@]}"; do
+    IFS='|' read -r target_file yq_path image_name <<< "$entry"
+
+    # Get current value
+    current=$(yq eval "$yq_path" "$target_file" 2>/dev/null || echo "")
```
huang195: `current=$(yq eval "$yq_path" "$target_file" 2>/dev/null || echo "")` — yq returns the literal string `"null"` with exit 0 when the path doesn't exist (verified: `echo 'foo: bar' | yq eval '.nonexistent' -` → `null`, exit 0). So `current` becomes `"null"`, the skip-already-pinned check at L196 never matches, and the diff log prints e.g. `mlflow-oauth-secret (charts/kagenti/values.yaml): null → v0.6.0-rc.6`, which reads like a bug.

Two low-effort options:

```bash
# Option 1: explicit guard
if [[ -z "$current" || "$current" == "null" ]]; then
    echo "error: yq path $yq_path not found in $target_file" >&2
    exit 1
fi

# Option 2: yq fallback
current=$(yq eval "$yq_path // \"\"" "$target_file")
```

Protects against typos in `PINNABLE_IMAGES` and makes the logs honest.

scripts/pin-release-tags.sh
```diff
@@ -0,0 +1,247 @@
...
+    "$KAGENTI_VALUES|.apiOAuthSecret.tag|api-oauth-secret"
+    "$KAGENTI_VALUES|.mlflowOAuthSecret.tag|mlflow-oauth-secret"
+    "$KAGENTI_DEPS_VALUES|.spiffeIdp.image.tag|spiffe-idp-setup"
+)
+
+# ---------------------------------------------------------------------------
+# Colors
+# ---------------------------------------------------------------------------
+
+RED=$'\033[0;31m'
```
huang195: Nit: color codes are emitted unconditionally to stdout. CI logs and any piped output get raw `\033[...]` escapes. Standard pattern: detect TTY + honor `NO_COLOR`:

```bash
if [[ -t 1 && -z "${NO_COLOR:-}" ]]; then
    RED=$'\033[0;31m'; GREEN=$'\033[0;32m'; YELLOW=$'\033[1;33m'; NC=$'\033[0m'
else
    RED=''; GREEN=''; YELLOW=''; NC=''
fi
```

scripts/pin-release-tags.sh
```diff
@@ -0,0 +1,247 @@
...
+# produces the same result.
+#
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+
+KAGENTI_VALUES="$REPO_ROOT/charts/kagenti/values.yaml"
+KAGENTI_DEPS_VALUES="$REPO_ROOT/charts/kagenti-deps/values.yaml"
+KAGENTI_CHART="$REPO_ROOT/charts/kagenti/Chart.yaml"
```
huang195, line 26: Nit: `--chart-version` updates `charts/kagenti/Chart.yaml` only; `charts/kagenti-deps/Chart.yaml` is untouched. If both should be version-synced at release time (matches the spirit of the rest of the script — "keep both charts in step"), include it; if not, a one-line comment explaining the asymmetry would help future readers.

huang195 (approved):
## Summary

Solid fix for a real problem — manual chart pinning was missing both the `kagenti-deps` chart and `mlflow-oauth-secret`, which is what allowed `spiffe-idp-setup` to sit at rc.1 through rc.5. The script is idempotent, has a sane interface (`--dry-run`, `--verify-images`, `--chart-version`), is `set -euo pipefail` clean, validates version format with regex, and the `build.yaml` addition closes the gap on the OCI publish side. `check-release-pins.sh` extension catches the same class of drift on PRs.

**Areas reviewed**: Shell scripts, GitHub Actions workflow, Docs
**Commits**: 1, signed-off
**CI status**: lint/static green; `Deploy & Test` failure is unrelated (same flake on parallel PR #1651, no nexus to this PR's diff).

Approving — the inline notes are non-blocking. Two of them (the warning-vs-error severity and the `"null"` return from yq) feel worth addressing before this script gets used for an actual release.

Assisted-By: Claude Code

scripts/pin-release-tags.sh
```diff
@@ -0,0 +1,247 @@
...
+done
+
+if [[ -z "$VERSION" ]]; then
+    echo "error: VERSION argument is required" >&2
+    echo "" >&2
+    usage
+fi
+
+# Validate version format (must start with v and contain at least major.minor.patch)
+if [[ ! "$VERSION" =~ ^v[0-9]+\.[0-9]+\.[0-9]+ ]]; then
```
rubambiza: **Suggestion**: The regex anchors the start but not the end:

```bash
if [[ ! "$VERSION" =~ ^v[0-9]+\.[0-9]+\.[0-9]+ ]]; then
```

This passes strings like `v1.2.3"injected` or `v1.2.3 trailing`. Since `$VERSION` is interpolated into a yq expression at L205, a crafted suffix containing `"` could break the expression. Low practical risk (release engineers pass known tags), but tightening with an end anchor is cheap:

```bash
if [[ ! "$VERSION" =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-[a-zA-Z0-9.]+)?$ ]]; then
```

scripts/pin-release-tags.sh
```diff
@@ -0,0 +1,247 @@
...
+
+    if [[ "$current" == "$VERSION" ]]; then
+        printf '%s[SKIP]%s %s (%s) — already %s\n' "$GREEN" "$NC" "$image_name" "$rel_path" "$VERSION"
+        continue
+    fi
+
+    if [[ "$DRY_RUN" == "true" ]]; then
+        printf '%s[DRY]%s  %s (%s): %s → %s\n' "$YELLOW" "$NC" "$image_name" "$rel_path" "$current" "$VERSION"
+    else
+        yq eval "$yq_path = \"$VERSION\"" -i "$target_file"
```
rubambiza: **Suggestion**: `$VERSION` is interpolated directly into the yq expression string. If it ever contained a `"` (the regex doesn't block this today — see L95 comment), it would break the expression or cause unintended YAML mutations.

Using `--arg` avoids the interpolation entirely:

```bash
yq eval --arg v "$VERSION" "$yq_path = \$v" -i "$target_file"
```

This is more idiomatic for mikefarah/yq and sidesteps any quoting edge cases.

rubambiza (approved):
Solid automation that addresses a real release pain point. CI static checks all pass (Deploy & Test failure is the known flake also visible on PR #1651). The script is well-structured with `set -euo pipefail`, prerequisite checks, and clear output formatting.

Two non-blocking suggestions inline — both are about defensive hardening, not correctness bugs.

**Areas reviewed**: Shell scripts, GitHub Actions workflow, Docs
**Commits**: 1, signed-off
**CI status**: Static checks all green; Deploy & Test failure is unrelated flake (same on #1651)

Assisted-By: Claude Code

PR #1653, opened 2026-05-22
feat: align kagenti with kagenti-extensions go-spiffe SDK migration

huang195:
Companion to [kagenti-extensions#432](https://github.com/kagenti/kagenti-extensions/pull/432).

## Why

The new authbridge image (PR #432) replaces the bundled `spiffe-helper` binary with an in-process `go-spiffe` SDK integration. That changes the authbridge config schema in two ways:

- A top-level `spiffe:` block signals the framework to build the in-process Provider (X.509 source for mTLS + lazy JWT source for tokenexchange).
- A per-plugin `tokenexchange.identity.jwt_audience` field carries the JWT-SVID client-assertion audience that previously lived in `helper.conf`.

Without these chart updates, the new image fails to boot in SPIRE-enabled deployments — either skipping Provider construction (no mTLS, spiffe identity unwired) or failing Configure with `identity.type=spiffe requires identity.jwt_audience to be set`.

## What

Two code paths emit authbridge config in this repo. Both are updated:

| Source | What changed |
|---|---|
| `charts/kagenti/templates/authbridge-template-configmaps.yaml` and `charts/kagenti/templates/agent-namespaces.yaml` | Render `spiffe: {}` (when `spire.enabled=true`) and `tokenexchange.identity.jwt_audience` (when `clientAuthType=federated-jwt`). Drop the unused `JWT_AUDIENCE` env var from `authbridge-config` ConfigMap. |
| `kagenti/backend/app/routers/agents.py` (`_build_authbridge_runtime_yaml`) + tests | Same schema change for backend-API-created agents. |
| `kagenti/ui-v2/src/types/index.ts` (`IdentityConfig`) | Add optional `jwt_audience` field for type-safety in the UI form/validation surface. |

## Sequencing / backward compatibility

This PR is **safe to land before** kagenti-extensions#432:

- New `spiffe:` block + `jwt_audience` field are additive. The OLD authbridge image's YAML decoder silently drops unknown keys, so deployments still running the old image keep working with the updated chart.
- `JWT_AUDIENCE` env var being removed has no consumer (verified across kagenti, kagenti-extensions, and kagenti-operator — only the operator parsed it into a struct field that nothing reads). Old image unaffected.
- `spiffe-helper-config` ConfigMap and the legacy `helper.conf` rendering are intentionally LEFT in place. They're no-ops against the new image (no spiffe-helper binary to read them) and required by the old image during a phased rollout. Cleanup can land in a follow-up PR after all old-image deployments are gone.

## Test plan

- [ ] `cd kagenti/backend && uv run pytest tests/test_authbridge_runtime_config.py -v` passes
- [ ] `helm template charts/kagenti --set spire.enabled=true --set authBridge.clientAuthType=federated-jwt` renders `spiffe: {}` and `tokenexchange.identity.jwt_audience` in both `agent-namespaces.yaml` and `authbridge-template-configmaps.yaml` outputs
- [ ] `helm template charts/kagenti --set spire.enabled=true --set authBridge.clientAuthType=client-secret` does NOT render `jwt_audience` (only `type: client-secret`)
- [ ] `helm template charts/kagenti --set spire.enabled=false` does NOT render the top-level `spiffe:` block
- [ ] Local Kind: deploy + chat with weather-agent in federated-jwt mode succeeds (verified)

Assisted-By: Claude Code

### DIFF
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -2085,33 +2085,46 @@ def _build_authbridge_runtime_yaml(
     # AgentRuntime.Spec.AuthBridgeMode → namespace ConfigMap →
     # cluster default (proxy-sidecar). Hardcoding it here would pin
     # every backend-rendered ConfigMap to one shape regardless of CR.
-    config = {
-        "pipeline": {
-            "inbound": {
-                "plugins": [
-                    {
-                        "name": "jwt-validation",
-                        "config": {
-                            "issuer": issuer,
-                            "keycloak_url": keycloak_url,
-                            "keycloak_realm": realm,
-                        },
-                    }
-                ]
-            },
-            "outbound": {
-                "plugins": [
-                    {
-                        "name": "token-exchange",
-                        "config": {
-                            "keycloak_url": keycloak_url,
-                            "keycloak_realm": realm,
-                            "default_policy": "passthrough",
-                            "identity": {"type": identity_type},
-                        },
-                    }
-                ]
-            },
+    identity: dict[str, str] = {"type": identity_type}
+    if identity_type == "spiffe":
+        # JWT-SVID client-assertion audience — must match what Keycloak's
+        # SPIFFE IdP expects (the realm issuer URL). Required by
+        # authbridge's token-exchange plugin in the spiffe identity path.
+        # See kagenti-extensions#332.
+        identity["jwt_audience"] = issuer
+    config: dict[str, object] = {}
+    if spire_enabled:
+        # Empty spiffe block signals authbridge to construct the in-process
+        # SPIFFE provider (X.509 source for mTLS + lazy JWT source for
+        # tokenexchange). All fields default: socket points at the standard
+        # SPIRE agent socket, mirror writes /opt/svid*.pem on rotation.
+        # The JWT audience lives per-tokenexchange (under identity above).
+        config["spiffe"] = {}
+    config["pipeline"] = {
+        "inbound": {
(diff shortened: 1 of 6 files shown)


kagenti/backend/app/routers/agents.py
```diff
@@ -2085,33 +2085,46 @@ def _build_authbridge_runtime_yaml(
...
-                            "default_policy": "passthrough",
-                            "identity": {"type": identity_type},
-                        },
-                    }
-                ]
-            },
+    identity: dict[str, str] = {"type": identity_type}
+    if identity_type == "spiffe":
+        # JWT-SVID client-assertion audience — must match what Keycloak's
+        # SPIFFE IdP expects (the realm issuer URL). Required by
```
pdettori, line 2091: suggestion (non-blocking): The `issuer` param used as `jwt_audience` is semantically the Keycloak realm URL — consider renaming to `jwt_audience` at the function boundary for clarity, since it's not really an "issuer" in the OIDC sense when used as a SPIFFE JWT-SVID audience claim. Current code is correct either way.

pdettori (approved):
Clean, well-structured PR. Config schema changes are consistent across all three paths (Helm agent-namespaces, Helm authbridge-template-configmaps, Python backend). The validation helper in `_helpers.tpl` provides good fail-fast behavior for misconfiguration. Tests cover both positive (spiffe + jwt_audience present) and negative (client-secret without jwt_audience) cases. CI fully green.

**Areas reviewed**: Helm/K8s, Python, TypeScript, Tests

PR #1655, opened 2026-05-23
fix(ocp): skip remote tag detection when --kagenti-repo is provided

### LINKED ISSUE #1654: OCP setup script overrides pinned image tags with latest remote tag
## Problem

When running `setup-kagenti.sh --kagenti-repo .` on OpenShift, the UI shows `v0.7.0-alpha.1` instead of the expected `v0.6.0-rc.6` that is pinned in the chart values.

## Root Cause

The "Build UI helm flags" section (around line 954) unconditionally fetches the latest git tag from the remote repo:

```bash
LATEST_TAG=$(git ls-remote --tags --sort="v:refname" https://github.com/kagenti/kagenti.git | tail -n1 | ...)
KAGENTI_UI_FLAGS+=(--set "ui.frontend.tag=v${LATEST_TAG}")
KAGENTI_UI_FLAGS+=(--set "ui.backend.tag=v${LATEST_TAG}")
```

This overrides whatever is pinned in `charts/kagenti/values.yaml`, making the pinning commit (`995424b9`) ineffective on OCP installs.

Kind installs work correctly because `kind-full-test.sh` does not have this tag detection logic.

## Expected Behavior

When `--kagenti-repo` points to a local directory (or any explicit repo), the chart's pinned image tags should be authoritative. Remote tag detection should only apply when no repo is specified (i.e., installing from upstream main).

## Steps to Reproduce

1. Checkout a commit with pinned tags (e.g. `v0.6.0-rc.6`)
2. Run `./scripts/ocp/setup-kagenti.sh --kagenti-repo . --with-all`
3. Observe UI shows `v0.7.0-alpha.1` (or whatever the latest remote tag is)

pdettori:
## Summary

- Fixes the OCP setup script overriding chart-pinned image tags with the latest remote tag when `--kagenti-repo` is explicitly provided
- Remote tag detection now only runs when no repo is specified (using upstream main)
- When a local repo is passed, the chart's `values.yaml` image tags are used as-is

Fixes #1654

## Test plan

- [ ] Run `./scripts/ocp/setup-kagenti.sh --kagenti-repo . --with-all` on OCP — verify UI shows the version pinned in chart values
- [ ] Run `./scripts/ocp/setup-kagenti.sh` (no `--kagenti-repo`) — verify it still detects and uses the latest remote tag
- [ ] Kind install continues to work as before (unaffected code path)

Assisted-By: Claude Code

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -38,6 +38,11 @@ updates:
         update-types: [minor, patch]
       major:
         update-types: [major]
+    ignore:
+      # cryptography 47+ crashes with SIGILL on ARM64 Kind clusters.
+      # Pin is intentional — block all bumps above the current upper bound.
+      - dependency-name: "cryptography"
+        update-types: [version-update:semver-major, version-update:semver-minor]
 
   # npm (UI) — React frontend dependencies
   - package-ecosystem: npm
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -41,6 +41,7 @@ updates:
     ignore:
       # cryptography 47+ crashes with SIGILL on ARM64 Kind clusters.
       # Pin is intentional — block all bumps above the current upper bound.
+      # Effect: only patch bumps allowed until the SIGILL is resolved upstream.
       - dependency-name: "cryptography"
         update-types: [version-update:semver-major, version-update:semver-minor]
 
--- a/scripts/ocp/setup-kagenti.sh
+++ b/scripts/ocp/setup-kagenti.sh
@@ -221,6 +221,7 @@ _clone_kagenti() {
 }
 
 KAGENTI_SOURCE=""
+KAGENTI_REPO_ORIGINAL="$KAGENTI_REPO"
 if [ -z "$KAGENTI_REPO" ]; then
   # No --kagenti-repo given: always clone fresh from upstream main
   KAGENTI_SOURCE="$KAGENTI_GITHUB_URL"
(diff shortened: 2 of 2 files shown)


.github/dependabot.yml
```diff
@@ -32,6 +32,11 @@ updates:
     groups:
       minor-and-patch:
         update-types: [minor, patch]
+    ignore:
+      # cryptography 47+ crashes with SIGILL on ARM64 Kind clusters.
+      # Pin is intentional — block all bumps above the current upper bound.
```
huang195, line 43: Nit: the existing comment explains *why* the pin exists (47+ ARM64 SIGILL) but not what's *allowed*. With both `version-update:semver-major` and `version-update:semver-minor` ignored, only patch bumps get through — worth one extra line for the next maintainer:

```yaml
# cryptography 47+ crashes with SIGILL on ARM64 Kind clusters.
# Pin is intentional — block all bumps above the current upper bound.
# Effect: only patch bumps allowed for cryptography until the SIGILL is resolved upstream.
```

huang195 (approved):
## Summary

Tight, correct fix. The original code used `[ -z "$KAGENTI_REPO" ]` after `_clone_kagenti` had already clobbered `KAGENTI_REPO` with the cache dir, so the "no flag" branch ran for *every* invocation — including explicit `--kagenti-repo <local-path>`. Capturing `KAGENTI_REPO_ORIGINAL` before the clone block resolves it cleanly. The `else` arm with the explicit log message is a nice touch (operators see exactly which decision was made).

One PR-scope note (non-blocking): the title and body only describe the OCP setup fix, but the diff also includes an unrelated `cryptography` dependabot ignore. Either split into two PRs (cleaner revert story) or amend the description to cover both. The dependabot change itself is sensible — the cryptography 47+ ARM64 SIGILL is real and the pin is intentional.

**Areas reviewed**: Shell (setup-kagenti.sh), CI config (dependabot.yml)
**CI status**: pending checks queued; no failures.

Assisted-By: Claude Code

PR #1660, opened 2026-05-24
fix(backend): use agent card url for A2A JSON-RPC invoke endpoint

jordigilh:
## Problem

The backend constructs the agent URL from the Kubernetes service name and port via `resolve_agent_url()`, then POSTs directly to the root path (`/`). This breaks agents that mount their A2A JSON-RPC handler at a sub-path (e.g. `/a2a/invoke`), returning a 404.

Per the [A2A spec](https://a2a-protocol.org/latest/specification/#446-agentinterface), the agent card `url` field is defined as the endpoint that accepts JSON-RPC requests. The backend already fetches the agent card (for the `/agent-card` display endpoint) but ignores the `url` field when sending messages.

## Fix

Add `_resolve_invoke_url()` that fetches the agent card and uses its `url` field to determine the correct POST path. Falls back to the bare service URL when the card is unavailable, so existing agents serving at `/` are unaffected.

Path traversal sequences (`..`) in the card URL are rejected to prevent directory traversal (SI-10).

Two call sites updated:
- `send_message` (`/send` endpoint)
- `stream_message` (`/stream` endpoint)

## Impact

- **Agents at root `/`**: No change — the card `url` typically has no path or `/`, so the fallback kicks in.
- **Agents at sub-paths**: Now reachable. The kubernaut API Frontend (`/a2a/invoke`), LangChain Agent Server (`/a2a/{assistant_id}`), and others work correctly.

## Related

- jordigilh/kubernaut#1268 — AF-side compatibility alias (`POST /`)
- jordigilh/kubernaut#1269 — AF PR implementing the alias

## Ecosystem context

This is a well-known A2A ecosystem issue:
- [a2aproject/A2A#1599](https://github.com/a2aproject/A2A/issues/1599) — agent card advertises internal URL instead of exposed URL
- [docker/docker-agent#1679](https://github.com/docker/docker-agent/issues/1679) — agent card URL breaks clients
- [VoltAgent/voltagent#1199](https://github.com/VoltAgent/voltagent/pull/1199) — fixed cards to advertise `/a2a/{serverId}`
- [kagent-dev/kagent#536](https://github.com/kagent-dev/kagent/issues/536) — dynamic address in agent card URL (closed as stale)


jordigilh:
## CI fix pushed (2622bf2e)

Pushed a follow-up commit that addresses the 3 CI failures:

### Fixes

1. **CodeQL SSRF (py/ssrf/partial)** — Added RFC 1123 K8s name validation (`_K8S_NAME_RE`) for `name`/`namespace` before the agent card HTTP request. Invalid names skip the card fetch entirely, breaking the taint chain.

2. **CodeQL Log Injection (py/log-injection, CWE-117)** — Wrapped `namespace`/`name` in the `logger.debug()` call with the existing `sanitize_log()` helper to strip newlines/control characters.

3. **Ruff format** — Ran `ruff format` on the test file.

### Additional hardening

While fixing the CodeQL alerts I did an adversarial review of the path extraction logic and found that URL-encoded traversal sequences (`%2e%2e`) could bypass the `..` check. The fix:

- **`unquote()` before checking** — decode `%2e%2e` → `..` before the traversal filter
- **Positive path character allowlist** — only `[a-zA-Z0-9/-_.]` are permitted in the extracted path, rejecting query strings, fragments, and non-printable characters

Five new security test cases cover these scenarios.

### Broader findings (outside this PR's scope)

During the audit I noticed the same SSRF pattern exists in the pre-existing `get_agent_card()`, `send_message()`, and `stream_message()` endpoints on `main` — they all pass unvalidated path params into `resolve_agent_url()` → `httpx` requests. Happy to contribute a follow-up PR that adds input validation more broadly if maintainers are interested.

jordigilh:
@huang195 @pdettori @kellyaa PTAL — this PR makes the backend use the agent card `url` field for A2A invoke (so agents at sub-paths like `/a2a/invoke` work), with CodeQL SSRF/log-injection fixes and security hardening on the path extraction. CI workflows need approval to run on the latest commit.

kagenti/backend/app/routers/chat.py
```diff
@@ -19,14 +21,79 @@
...
+    query, when safe) to the internal service URL.  The host from the card
+    URL is always ignored — only the path and query are extracted — so an
+    agent card advertising an external hostname cannot redirect traffic.
+
+    Falls back to the bare service URL when the card is unavailable, has no
+    path, or fails validation.
+    """
+    name_match = _K8S_NAME_RE.fullmatch(name)
+    ns_match = _K8S_NAME_RE.fullmatch(namespace)
+    if not name_match or not ns_match:
```
pdettori, line 74: **Bug (causes CI failure):** `urlparse` separates the query string into `parsed.query`, so a card URL like `http://host/a2a?redirect=evil` yields `parsed.path = "/a2a"` which passes all four checks here. The test `test_rejects_path_with_query_string` correctly expects rejection but this code doesn't enforce it.

Add `not parsed.query and not parsed.fragment` to the condition:
```python
if (
    path
    and path != "/"
    and not parsed.query
    and not parsed.fragment
    and ".." not in path.split("/")
    and all(c in _SAFE_PATH_CHARS for c in path)
):
```

kagenti/backend/app/routers/chat.py
```diff
@@ -19,14 +21,79 @@
...
+
+
+async def _resolve_invoke_url(name: str, namespace: str, kube: KubernetesService) -> str:
+    """Resolve the A2A JSON-RPC invoke URL for an agent.
+
+    Per the A2A spec (§4.4.6) the ``AgentInterface.url`` field is a valid
+    absolute URL identifying the endpoint that accepts JSON-RPC requests.
+    Some agents mount this handler at a sub-path (e.g. ``/a2a/invoke``) or
+    include query parameters for routing (e.g. ``?assistant_id=123``).
+
```
pdettori, line 59: **Performance:** This function is called on every `/send` and `/stream` request, creating a new `httpx.AsyncClient` + TCP connection each time. For agents receiving frequent messages this adds non-trivial latency.

Consider a short-lived TTL cache keyed on `(name, namespace)` — even 30s would eliminate nearly all redundant agent-card fetches during a conversation. A simple dict + timestamp or `cachetools.TTLCache` would work (note `@lru_cache` doesn't support async).

pdettori (changes requested):
Good approach — using the agent card `url` field for A2A invoke is correct per spec, and the security hardening (SSRF, path traversal, log injection) is well-layered.

However there's a bug causing the `backend-tests` CI failure: `urlparse()` splits the query string into `parsed.query`, so the path-only validation doesn't catch URLs with `?param=value`. See inline comment for the fix.

Also flagged a performance concern (no caching on a per-message call path).

kagenti/backend/app/routers/chat.py
jordigilh: Thanks for the review! The latest commit (f7d0f7f2) actually takes a different approach here — rather than blanket-rejecting query strings, it preserves them when safe, per the A2A spec.

**[A2A spec §4.4.6](https://google.github.io/A2A/specification/#agentinterface)** defines `AgentInterface.url` as a full absolute URL (example: `"https://api.example.com/a2a/v1"`). Query parameters are a valid part of RFC 3986 URLs, and some agent frameworks use them for routing — e.g. a LangChain agent could mount at `/a2a?assistant_id=123`. Blanket-rejecting queries would break those agents.

The approach in the latest commit:
- Added `_SAFE_QUERY_CHARS` allowlist (alphanumeric, `-`, `_`, `.`, `=`, `&`, `+`) that **excludes** `:`, `/`, `#`, `%` to block URL injection and encoding attacks
- Path is validated independently — if the path passes but the query contains unsafe chars (e.g. `?redirect=http://evil.com` — the `:` and `/` fail), the query is dropped but the valid path is still used
- Fragments are implicitly rejected since `urlparse` separates them and we never include `parsed.fragment`

Two new tests cover this:
- `test_drops_unsafe_query_string` — malicious query stripped, path preserved
- `test_preserves_safe_query_string` — benign `?assistant_id=123` is kept

The old `test_rejects_path_with_query_string` was renamed/updated to reflect the new behavior.

kagenti/backend/app/routers/chat.py
jordigilh: Good call — adding a TTL cache. Will push a follow-up commit with a `cachetools.TTLCache` keyed on `(name, namespace)` with a 30s TTL.

jordigilh:
@pdettori can you approve again? I fixed the 2 issues and then rebase commit and the approval was lost in the process.

pdettori (approved):
LGTM

PR #1661, opened 2026-05-24
fix(backend): skip Service creation for Sandbox agents

pdettori:
## Summary

- Remove backend-managed ClusterIP Service creation for Sandbox workload types
- The Sandbox controller manages its own headless Service (`ClusterIP=None`); the backend creating a regular ClusterIP Service caused an immutable field conflict that blocked the Sandbox from reaching Ready
- Simplify `_create_or_replace_service` to early-return for both Job and Sandbox workloads

## Root Cause

When the backend created a Sandbox CR, it also created a regular `ClusterIP` Service with the same name. The Sandbox controller then tried to adopt/create its own headless Service (`ClusterIP: None`), but Kubernetes rejected the mutation because `spec.clusterIP` is immutable. This left the Sandbox permanently stuck in `ReconcilerError` state.

## Test plan

- [ ] Deploy agent with workload type `sandbox` on OpenShift — verify Sandbox reaches Ready
- [ ] Verify no ClusterIP Service is created by the backend for sandbox agents
- [ ] Verify the Sandbox controller creates its own headless Service
- [ ] Verify non-sandbox agents (Deployment, StatefulSet) still get their Service created normally
- [ ] Kind e2e tests pass (existing Sandbox test coverage)

Assisted-By: Claude Code

### DIFF
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -2687,23 +2687,13 @@ def _create_or_replace_service(
     service_manifest: dict,
     workload_type: str,
 ) -> None:
-    """Create the Service for an agent / tool, with the Sandbox-controller-race
-    handling that kagenti#1581 introduced.
+    """Create the Service for an agent / tool.
 
-    Returns silently for ``WORKLOAD_TYPE_JOB`` since Jobs don't need a Service
-    (the operator's `AgentCardReconciler` doesn't reconcile Jobs).
-
-    For ``WORKLOAD_TYPE_SANDBOX`` an existing Service may have been created by
-    the agent-sandbox controller before our admission window — kagenti#1581
-    handles that 409 by deleting the controller's short-lived Service and
-    replacing it with the backend-managed ClusterIP shape. Other workload
-    types (Deployment, StatefulSet) re-raise on 409.
-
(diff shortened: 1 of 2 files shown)


esnible (approved):
## Summary

Small, focused fix (173 diff lines, 2 files). Root cause analysis is correct — `spec.clusterIP` is immutable, so the backend's regular Service collided with the Sandbox controller's headless Service (`ClusterIP=None`). Removing backend Service creation for Sandbox workloads is the right call since the controller is the sole owner.

**Areas reviewed**: Python (backend router), tests
**Commits**: 1 commit, signed-off ✓
**CI status**: All required checks pass; E2E HyperShift pending maintainer `/run-e2e`

### Notes (non-blocking)

- The Shipwright finalize flow at `kagenti/backend/app/routers/agents.py:3787` still calls `_build_service_manifest` + `_create_or_replace_service` unconditionally. Behavior is correct (helper early-returns for Sandbox/Job), but the image-create flow now guards with an `if workloadType not in (...)` while the Shipwright flow doesn't — minor asymmetry that could invite future drift.
- `tests/test_sandbox_manifest.py:276` uses the literal `"deployment"` instead of the `WORKLOAD_TYPE_DEPLOYMENT` constant that's already exported.

Assisted-By: Claude Code

PR #1665, opened 2026-05-25
feat: gate OTel Helm resources for operator-managed bootstrap

Bobbins228:
> **Merge after:** https://github.com/kagenti/kagenti-operator/pull/376

## Summary

Adds an `otelBootstrap.operatorManaged` toggle to the `kagenti-deps` Helm chart and an `--otel-operator-managed` flag to the OCP setup script. When enabled, the Helm chart defers OTel collector ConfigMap assembly and ingress CA trust to the kagenti-operator's bootstrap runnable instead of rendering them itself.

## Changes

- **`charts/kagenti-deps/values.yaml`** — new `otelBootstrap.operatorManaged` flag (default `false`)
- **`charts/kagenti-deps/templates/otel-collector.yaml`** — gate the `otel-collector-config` ConfigMap behind `{{- if not .Values.otelBootstrap.operatorManaged }}`
- **`charts/kagenti-deps/templates/otel-ingress-ca-job.yaml`** — gate the entire Job behind the same toggle
- **`scripts/ocp/setup-kagenti.sh`** — new `--otel-operator-managed` flag that:
  - Skips injecting MLflow OTel values into the Helm args
  - Passes `--set otelBootstrap.operatorManaged=true` to `kagenti-deps`
  - Skips the deferred `helm upgrade` for OTel endpoint wiring
  - MLflow provisioning (DSC, CR, RBAC) remains unchanged

## Test Plan

- [x] `helm template` renders ConfigMap and Job when `otelBootstrap.operatorManaged` is unset
- [x] `helm template` omits ConfigMap and Job when `otelBootstrap.operatorManaged=true`
- [x] `setup-kagenti.sh --otel-operator-managed` logs skip messages and passes the flag to Helm
- [x] Full end-to-end: setup script + operator with `otelBootstrap.enable=true` on OCP cluster

Made with [Cursor](https://cursor.com)

### DIFF
--- a/charts/kagenti-deps/templates/otel-collector.yaml
+++ b/charts/kagenti-deps/templates/otel-collector.yaml
@@ -155,6 +155,7 @@ spec:
       port: 8335
       targetPort: 8335
   type: ClusterIP
+{{- if not .Values.otelBootstrap.operatorManaged }}
 ---
 apiVersion: v1
 kind: ConfigMap
@@ -166,6 +167,7 @@ metadata:
 data:
   base.yaml: |
     {{- include "kagenti.otel.collectorConfig" . | nindent 4 }}
+{{- end }}
 {{- end }}
 {{- if and .Values.components.otel.enabled (.Capabilities.APIVersions.Has "security.istio.io/v1") }}
 ---
--- a/charts/kagenti-deps/templates/otel-collector.yaml
+++ b/charts/kagenti-deps/templates/otel-collector.yaml
@@ -155,6 +155,8 @@ spec:
       port: 8335
       targetPort: 8335
   type: ClusterIP
+{{- /* ConfigMap is operator-managed when otelBootstrap.operatorManaged=true;
+       Service and Deployment above remain Helm-managed in all modes. */}}
 {{- if not .Values.otelBootstrap.operatorManaged }}
 ---
 apiVersion: v1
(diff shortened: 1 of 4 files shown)


charts/kagenti-deps/templates/otel-collector.yaml
```diff
@@ -155,6 +155,7 @@ spec:
       port: 8335
       targetPort: 8335
   type: ClusterIP
+{{- if not .Values.otelBootstrap.operatorManaged }}
```
rubambiza, line 160: **Nit**: The new `operatorManaged` guard gates only the ConfigMap (and the ingress CA job in the other template). The Service and Deployment above still render when `operatorManaged=true`. This seems intentional (operator manages *config* not *lifecycle*), but a one-line comment here like `{{- /* ConfigMap is operator-managed when otelBootstrap.operatorManaged=true */}}` would make the design choice explicit for future readers.

scripts/ocp/setup-kagenti.sh
```diff
@@ -667,8 +671,12 @@ _helm_kagenti_deps() {
   done
 
   # Build MLflow OTEL flags: enable the pipeline and point it at the DSC-managed endpoint.
+  # When --otel-operator-managed is set, the operator handles ConfigMap assembly
```
rubambiza, line 674: **Suggestion**: When `OTEL_OPERATOR_MANAGED=true` AND `MLFLOW_TRACES_ENDPOINT` is non-empty (user sets both), the operator-managed branch wins silently. Consider emitting a warning like `log_warn "--otel-operator-managed set; ignoring MLFLOW_TRACES_ENDPOINT"` to make the precedence obvious for operators troubleshooting why their endpoint isn't being used.

rubambiza (approved):
Well-scoped PR that introduces an opt-in toggle (`otelBootstrap.operatorManaged: false`) to defer OTel collector ConfigMap assembly and ingress CA job to the kagenti-operator. Helm template nesting is correct — the new guard only gates the ConfigMap and CA job while leaving Service/Deployment Helm-managed. Shell script logic correctly prioritizes `--otel-operator-managed` over `MLFLOW_TRACES_ENDPOINT` with clear log messages at each skip point.

**Areas reviewed**: Helm templates, Helm values, Shell script
**CI status**: All passing

Two non-blocking notes inline.

Assisted-By: Claude Code

PR #1668, opened 2026-05-26
fix(skills): add --with-skills flag, fix image loading, and update docs

### LINKED ISSUE #1667: fix(skills): setup script and docs issues for Skills feature
## Problem

Several issues were found when following the Skills feature demo end-to-end on WSL2:

1. `--with-skills` flag missing from `setup-kagenti.sh` — `KAGENTI_FEATURE_FLAG_SKILLS` env var was silently ignored
2. `featureFlags.skills` not passed to Helm, so the backend always started with the flag disabled
3. `docker build --load` fails (`--load` is a `buildx`-only flag)
4. `kind load docker-image` fails with "failed to detect containerd snapshotter" on WSL2
5. Registry `/etc/hosts` entry was only written on fresh cluster creation; re-running with `--skip-cluster` left a stale ClusterIP causing image pull failures
6. `sed -i` on `/etc/hosts` fails inside Kind nodes (bind-mount rename not allowed)
7. Docs used wrong kubectl label selector (`app=kagenti-backend` → `app.kubernetes.io/name=kagenti-backend`)
8. Demo doc lacked troubleshooting guidance for common setup failures

eranra:
Closes #1667.

## Summary

- Add `--with-skills` flag to `setup-kagenti.sh` and wire it to `featureFlags.skills` in Helm so the Skills feature flag actually reaches the backend (previously the env var export was silently ignored)
- Include `--with-skills` in `--with-all`
- Replace `docker build --load` with `docker buildx build --load` (`--load` is a buildx-only flag)
- Replace `kind load docker-image` with a `load_image_into_kind()` helper that uses `docker save | ctr import` — avoids "failed to detect containerd snapshotter" on WSL2
- Fix registry `/etc/hosts` upsert: run unconditionally (not only on fresh cluster), use `grep -v + cat >` instead of `sed -i` (bind-mount safe)
- Fix kubectl label selector in docs: `app=kagenti-backend` → `app.kubernetes.io/name=kagenti-backend`
- Add Troubleshooting sections to `docs/skills.md` and `docs/demos/demo-generic-agent-skill.md`

## Test plan

- [ ] `scripts/kind/setup-kagenti.sh --with-skills --skip-cluster` enables skills on existing cluster
- [ ] `--with-all` enables skills
- [ ] `--build-images` completes without `--load` or snapshotter errors on WSL2
- [ ] Re-running with `--with-builds --skip-cluster` correctly updates the registry hosts entry
- [ ] `kubectl logs -n kagenti-system -l app.kubernetes.io/name=kagenti-backend | grep "skills routes registered"` returns output when flag is on

Assisted-By: Claude Code

### DIFF
--- a/charts/kagenti/templates/ui.yaml
+++ b/charts/kagenti/templates/ui.yaml
@@ -393,11 +393,16 @@ rules:
     resources: ["pods"]
     verbs: ["get", "list"]
   # ConfigMaps: list/get for reads; create/update/patch for AuthBridge finalize
-  # (authbridge-config, envoy-config, spiffe-helper-config, authproxy-routes merge);
-  # delete required for skill deletion via the Skills API.
+  # (authbridge-config, envoy-config, spiffe-helper-config, authproxy-routes merge).
   - apiGroups: [""]
     resources: ["configmaps"]
-    verbs: ["get", "list", "create", "update", "patch", "delete"]
+    verbs: ["get", "list", "create", "update", "patch"]
+  {{- if .Values.featureFlags.skills }}
+  # Skills-specific: delete required for skill deletion via the Skills API.
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    verbs: ["delete"]
+  {{- end }}
   - apiGroups: [""]
     resources: ["secrets"]
     verbs: ["get", "list"]
--- a/charts/kagenti/templates/ui.yaml
+++ b/charts/kagenti/templates/ui.yaml
@@ -393,10 +393,11 @@ rules:
     resources: ["pods"]
     verbs: ["get", "list"]
   # ConfigMaps: list/get for reads; create/update/patch for AuthBridge finalize
-  # (authbridge-config, envoy-config, spiffe-helper-config, authproxy-routes merge)
+  # (authbridge-config, envoy-config, spiffe-helper-config, authproxy-routes merge);
+  # delete required for skill deletion via the Skills API.
   - apiGroups: [""]
     resources: ["configmaps"]
-    verbs: ["get", "list", "create", "update", "patch"]
+    verbs: ["get", "list", "create", "update", "patch", "delete"]
   - apiGroups: [""]
     resources: ["secrets"]
     verbs: ["get", "list"]
--- a/scripts/kind/setup-kagenti.sh
+++ b/scripts/kind/setup-kagenti.sh
@@ -48,6 +48,7 @@ WITH_BUILDS=false
 WITH_KIALI=false
 WITH_KUADRANT=false
 WITH_AGENT_SANDBOX=false
+WITH_SKILLS=false
 WITH_ALL=false
 SKIP_CLUSTER=false
 SKIP_MLFLOW=false
@@ -84,6 +85,16 @@ run_cmd() {
   if $DRY_RUN; then echo "  [dry-run] $*"; else "$@"; fi
 }
 
+# Load a single image into the Kind node via docker save piped to ctr import.
+# Avoids 'kind load docker-image' failures (e.g. "failed to detect containerd
+# snapshotter") on WSL2 and Rancher Desktop.
+load_image_into_kind() {
+  local img="$1"
+  $CONTAINER_ENGINE save "$img" | \
+    $CONTAINER_ENGINE exec -i "${CLUSTER_NAME}-control-plane" \
+      ctr --namespace=k8s.io images import -
+}
+
 # ── Argument parsing ────────────────────────────────────────────────────────
 while [[ $# -gt 0 ]]; do
   case "$1" in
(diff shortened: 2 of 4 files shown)


eranra:
Added a second commit (`b5e7e2ce`) with an additional fix:

**`fix(rbac): add delete verb to configmaps in kagenti-backend ClusterRole`**

Skill deletion via the UI was silently failing — the Delete button appeared to work but nothing happened. Root cause: `kagenti-backend` ClusterRole was missing `delete` on `configmaps`, so Kubernetes returned 403. The mutation had no `onError` handler, making the failure invisible to the user.

Fix: added `"delete"` to the configmaps verbs in `charts/kagenti/templates/ui.yaml`.

scripts/kind/setup-kagenti.sh
```diff
@@ -192,7 +204,8 @@ echo "    OTel:          $WITH_OTEL"
 echo "    MLflow:        $WITH_MLFLOW"
 echo "    Builds:        $WITH_BUILDS"
 echo "    Kiali:         $WITH_KIALI"
-echo "    Agent Sandbox: $WITH_AGENT_SANDBOX"
+echo "    Agent Sandbox: $WITH_AGENT_SANDBOX
+    Skills:        $WITH_SKILLS"
 echo "    Skip cluster:  $SKIP_CLUSTER"
```
pdettori, line 223: **must-fix**: The "Skills" status line is embedded inside the "Agent Sandbox" echo string rather than being a separate `echo` statement. This breaks the pattern every other line uses and produces output like:
```
    Agent Sandbox: true
    Skills:        true
```
…but as a single `echo`, which will cause misalignment if the value length changes and makes the script harder to read.

Should be:
```bash
echo "    Agent Sandbox: $WITH_AGENT_SANDBOX"
echo "    Skills:        $WITH_SKILLS"
```

charts/kagenti/templates/ui.yaml
```diff
@@ -393,10 +393,11 @@ rules:
     resources: ["pods"]
     verbs: ["get", "list"]
   # ConfigMaps: list/get for reads; create/update/patch for AuthBridge finalize
-  # (authbridge-config, envoy-config, spiffe-helper-config, authproxy-routes merge)
+  # (authbridge-config, envoy-config, spiffe-helper-config, authproxy-routes merge);
+  # delete required for skill deletion via the Skills API.
   - apiGroups: [""]
```
pdettori, line 397: **suggestion**: The `delete` verb is added unconditionally to the configmaps ClusterRole rule, but it's only needed for the Skills API delete endpoint. Consider gating it behind `{{- if .Values.featureFlags.skills }}` (similar to how sandbox gets its own expanded block below). This follows least-privilege — clusters without skills enabled shouldn't grant configmap delete to the backend service account.

scripts/kind/setup-kagenti.sh
```diff
@@ -580,9 +593,10 @@ if $WITH_BUILDS && ! $SKIP_CLUSTER; then
   if ! $DRY_RUN; then
     CLUSTER_IP=$(kubectl get svc "$REGISTRY_NAME" -n "$REGISTRY_NS" -o jsonpath='{.spec.clusterIP}' 2>/dev/null || true)
     if [ -n "$CLUSTER_IP" ]; then
-      # Add registry DNS to Kind node's /etc/hosts
+      # Upsert registry DNS in Kind node's /etc/hosts (replace stale entry if present).
```
pdettori, line 610: **nit**: The `/etc/hosts` upsert uses `grep -v '${REGISTRY_HOST}' /etc/hosts > /tmp/hosts.tmp && ...`. If every line in `/etc/hosts` matches the pattern (unlikely but possible in a minimal container), `grep -v` returns exit code 1 (no output lines), which would abort under `set -e`. A trailing `|| true` on the grep makes this bulletproof:
```bash
sh -c "{ grep -v '${REGISTRY_HOST}' /etc/hosts || true; } > /tmp/hosts.tmp && cat /tmp/hosts.tmp > /etc/hosts && echo '${CLUSTER_IP} ${REGISTRY_HOST}' >> /etc/hosts"
```

pdettori (changes requested):
Good PR overall — adds real usability improvements for Skills setup and fixes the WSL2/Rancher Desktop image-loading issues. One blocking formatting bug and two suggestions below.

**Areas reviewed**: Shell, Helm/K8s, Docs
**Commits**: 2 commits, all signed-off ✓
**CI**: passing

scripts/kind/setup-kagenti.sh
eranra: Fixed. Split into two separate `echo` statements in [scripts/kind/setup-kagenti.sh:221-222](https://github.com/kagenti/kagenti/blob/fix/skills-setup-and-docs/scripts/kind/setup-kagenti.sh#L221-L222).

charts/kagenti/templates/ui.yaml
eranra: Done. Kept base configmaps rule as `get/list/create/update/patch`, added a separate `delete`-only rule gated behind `{{- if .Values.featureFlags.skills }}`.

scripts/kind/setup-kagenti.sh
eranra: Fixed. Added `|| true` using the brace-group form `{ grep -v ...; || true; }` as suggested.

pdettori (approved):
All review comments addressed. LGTM.

- Echo split: ✅
- Delete verb gated behind `featureFlags.skills`: ✅
- `grep -v || true` for set -e safety: ✅

CI green. Ship it.

PR #1671, opened 2026-05-26
fix(backend): restore backend-managed Service for Sandbox agents

pdettori:
## Summary

- Reverts the PR #1661 skip logic that removed Service creation for Sandbox workloads
- Backend now creates a ClusterIP Service (port 8080→8000) for Sandbox agents, restoring the port translation needed by the platform
- Sets `spec.service: false` on the Sandbox CR to suppress the agent-sandbox controller's conflicting headless Service (v0.4.6+ behavior)
- Aligns `scripts/openshell/deploy-shared.sh` to agent-sandbox v0.4.6 (matching `kind/setup-kagenti.sh` and `ocp/setup-kagenti.sh`)
- Also fixes the GET agent path to correctly fetch Service info for Sandbox agents

## Context

The agent-sandbox controller creates a headless Service (`ClusterIP=None`) which cannot perform port translation—kube-proxy is not in the path for headless Services. Kagenti agents listen on port 8000 but the platform addresses them at port 8080, requiring a ClusterIP Service for the 8080→8000 translation.

With agent-sandbox v0.4.6 ([PR #775](https://github.com/kubernetes-sigs/agent-sandbox/pull/775), [PR #800](https://github.com/kubernetes-sigs/agent-sandbox/pull/800)), Service creation became opt-in via `spec.service: true`. Setting `spec.service: false` explicitly suppresses it, preventing conflicts with the backend-managed Service.

## Test plan

- [x] Unit tests pass (18/18 in `test_sandbox_manifest.py`)
- [ ] Kind: deploy contact-extractor agent, verify agent card fetchable at port 8080
- [ ] OpenShift: same verification with agent-sandbox v0.4.6

Assisted-By: Claude Code

### DIFF
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -866,7 +866,7 @@ async def get_agent(
 
     # Try to get the associated Service (not applicable for Jobs)
     service = None
-    if workload_type not in (WORKLOAD_TYPE_JOB, WORKLOAD_TYPE_SANDBOX):
+    if workload_type != WORKLOAD_TYPE_JOB:
         try:
             service = kube.get_service(namespace=namespace, name=name)
         except ApiException as e:
@@ -2807,11 +2807,12 @@ def _create_or_replace_service(
 ) -> None:
     """Create the Service for an agent / tool.
 
-    Returns silently for ``WORKLOAD_TYPE_JOB`` and ``WORKLOAD_TYPE_SANDBOX``
-    since Jobs don't need a Service and the Sandbox controller manages its own
-    headless Service (ClusterIP=None).
+    Returns silently for ``WORKLOAD_TYPE_JOB`` since Jobs don't need a Service.
+    Sandbox agents get a backend-managed ClusterIP Service for port translation
+    (8080→8000); the agent-sandbox controller's own Service is suppressed via
+    ``spec.service: false`` on the Sandbox CR (v0.4.6+).
     """
-    if workload_type in (WORKLOAD_TYPE_JOB, WORKLOAD_TYPE_SANDBOX):
+    if workload_type == WORKLOAD_TYPE_JOB:
         return
     # Strip CR/LF before logging — name and namespace come from the FastAPI
     # request body. Kubernetes will reject non-DNS-1123 names so this is
@@ -3124,6 +3125,7 @@ def _build_sandbox_manifest(
         },
         "spec": {
             "replicas": 1,
+            "service": False,
             "podTemplate": {
                 "metadata": {
                     "labels": {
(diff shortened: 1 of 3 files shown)


kagenti/backend/app/routers/agents.py
```diff
@@ -3124,6 +3125,7 @@ def _build_sandbox_manifest(
         },
         "spec": {
             "replicas": 1,
+            "service": False,
```
huang195, line 3128: Migration risk: `spec.service: false` requires agent-sandbox **v0.4.6+** — earlier CRDs reject (or silently drop) the field, and the controller still creates its headless Service → 409 on backend Service create. The setup scripts pin v0.4.6 now (perfetto), but operators with existing clusters running v0.3.x need to upgrade the CRD first. Worth a note in the upgrade docs or a controller-version check.

kagenti/backend/tests/test_sandbox_manifest.py
```diff
@@ -217,17 +218,17 @@ def test_job_workload_skips_service_creation(self):
...
+        Service is suppressed via spec.service: false on the Sandbox CR."""
         from app.routers.agents import _create_or_replace_service
 
         kube = MagicMock()
         manifest = self._service_manifest("sb")
         _create_or_replace_service(kube, "team1", "sb", manifest, WORKLOAD_TYPE_SANDBOX)
 
-        kube.create_service.assert_not_called()
-        kube.delete_service.assert_not_called()
+        kube.create_service.assert_called_once_with(namespace="team1", body=manifest)
```
huang195, line 231: Nit: for parity with `test_deployment_creates_service` at line 242, consider also asserting `kube.delete_service.assert_not_called()` here. Same pattern, same workload-type gate, easy symmetry.

kagenti/backend/tests/test_sandbox_manifest.py
```diff
@@ -240,18 +241,19 @@ def test_deployment_creates_service(self):
...
 
-    def test_sandbox_409_not_applicable(self):
-        """Sandbox workloads are skipped entirely — no race handling needed
-        since the backend never creates a Service for Sandbox agents."""
+    def test_sandbox_409_propagates(self):
+        """Sandbox workloads now create a Service — a 409 is a real conflict
+        and must propagate (same as Deployment workloads)."""
         from app.routers.agents import _create_or_replace_service
+        from kubernetes.client import ApiException
+        import pytest
```
huang195, line 249: Nit: `import pytest` inside the test function — pytest is implicitly available in test files. Top-of-file import (or none — `pytest.raises` works without an explicit import in pytest-collected modules) would be cleaner.

huang195 (approved):
Mamma mia, Paolo — bel lavoro! 🇮🇹 Clean, surgical fix that goes right to the heart of the problem. The diff tells the whole story: revert the skip, add `spec.service: false` to suppress the controller, and update the tests to match. Bellissimo when a fix is this focused.

All required CI checks pass. The one red mark — `OpenShell PoC (Kind)` — is a Dockerfile build issue unrelated to your changes (missing `/usr/local/bin/openshell-sandbox` artifact). Worth noting that the `deploy-shared.sh` v0.4.6 bump didn't actually get exercised end-to-end this run, so a manual smoke on Kind/OCP per your test plan would be the responsible thing.

One follow-up question (out of scope, just curious): now that Sandbox agents get a real Service at port 8080, the `_ensure_agentruntime` and HTTPRoute gates at `agents.py:3348` and `:3361` (and the equivalents in the shipwright finalize at `:3932` / `:3949`) still skip Sandbox. Is that intentional — do Sandbox agents not need sidecar injection or external HTTPRoute exposure? Either answer is fine, but if the answer is "they should have them eventually," a TODO comment would help future-Paolo.

Suggestions and nits inline. Approving — ship it. 🍝

**Areas reviewed**: Python (backend + tests), Shell (deploy-shared.sh)
**Verified externally**: agent-sandbox v0.4.6 SandboxSpec has `Service *bool`; kind/ocp setup scripts already on v0.4.6.

PR #1672, opened 2026-05-26
feat(installer): add --skip-mlflow and --skip-kuadrant to Kind installer

### LINKED ISSUE #1670: feat(installer): add --skip-mlflow and --skip-kuadrant options to Kind installer
## Summary

Add `--skip-mlflow` and `--skip-kuadrant` flags to the Kind installer script (`.github/scripts/local-setup/kind-full-test.sh`) to allow users to exclude resource-heavy components from the installation.

## Motivation

Kuadrant and MLflow consume significant CPU and memory resources. The current installer requires at least **6 vCPUs and 18 GB** of memory in the Podman Desktop VM to run all components. For developers with smaller VMs (e.g., 4 vCPUs / 12–16 GB memory), these components make the install impractical.

Adding skip flags enables a lighter-weight local development experience without requiring a full-resource VM.

## Proposed Changes

- Add `--skip-mlflow` flag to exclude MLflow deployment
- Add `--skip-kuadrant` flag to exclude Kuadrant deployment
- Both flags default to off (components installed by default)
- Document the flags in the script's usage/help output

## Acceptance Criteria

- [ ] `kind-full-test.sh --skip-mlflow` completes without deploying MLflow
- [ ] `kind-full-test.sh --skip-kuadrant` completes without deploying Kuadrant
- [ ] Flags can be combined: `--skip-mlflow --skip-kuadrant`
- [ ] Existing behavior (no flags) is unchanged
- [ ] Help/usage output documents the new flags

## Resource Context

| Configuration | vCPUs | Memory | Notes |
|--------------|-------|--------|-------|
| Full install | 6+ | 18 GB+ | All components |
| Skip MLflow + Kuadrant | 4 | 12–16 GB | Lighter footprint |

Assisted-By: Claude Code

pdettori:
## Summary

- Adds `--skip-mlflow` and `--skip-kuadrant` flags to both `setup-kagenti.sh` and `kind-full-test.sh`
- `--with-all` expansion is deferred until after argument parsing, making `--skip-*` flags order-independent
- Enables lightweight local dev on 4 vCPU / 12–16 GB VMs

## Changes

- `scripts/kind/setup-kagenti.sh`: new `--skip-mlflow` / `--skip-kuadrant` flags; deferred `--with-all` expansion so skip flags work in any position
- `.github/scripts/local-setup/kind-full-test.sh`: passes skip flags through to `setup-kagenti.sh`

## Usage

```bash
# Direct usage (setup-kagenti.sh) — flags are order-independent
scripts/kind/setup-kagenti.sh --with-all --skip-mlflow --skip-kuadrant
scripts/kind/setup-kagenti.sh --skip-mlflow --with-all   # same result

# Via kind-full-test.sh wrapper
./.github/scripts/local-setup/kind-full-test.sh --skip-mlflow --skip-kuadrant --skip-cluster-destroy

# Skip only MLflow
./.github/scripts/local-setup/kind-full-test.sh --skip-mlflow --skip-cluster-destroy

# Full install (unchanged default behavior)
./.github/scripts/local-setup/kind-full-test.sh --skip-cluster-destroy
```

## Test plan

- [x] `bash -n` syntax check passes
- [ ] `--skip-mlflow --skip-kuadrant` installs without MLflow/Kuadrant pods
- [ ] Default (no skip flags) still deploys all components via `--with-all`
- [ ] Flags combine with existing `--skip-cluster-create`, `--skip-cluster-destroy`
- [x] `--skip-*` flags work regardless of position relative to `--with-all`

Closes #1670

Assisted-By: Claude Code

### DIFF
--- a/scripts/kind/setup-kagenti.sh
+++ b/scripts/kind/setup-kagenti.sh
@@ -48,7 +48,10 @@ WITH_BUILDS=false
 WITH_KIALI=false
 WITH_KUADRANT=false
 WITH_AGENT_SANDBOX=false
+WITH_ALL=false
 SKIP_CLUSTER=false
+SKIP_MLFLOW=false
+SKIP_KUADRANT=false
 BUILD_IMAGES=false
 PRELOAD_IMAGES=false
 DRY_RUN=false
@@ -95,15 +98,10 @@ while [[ $# -gt 0 ]]; do
     --with-builds)      WITH_BUILDS=true; shift ;;
     --with-kiali)       WITH_KIALI=true; shift ;;
     --with-agent-sandbox) WITH_AGENT_SANDBOX=true; shift ;;
-    --with-all)
-      WITH_ISTIO=true; WITH_SPIRE=true; WITH_BACKEND=true; WITH_UI=true
-      WITH_MCP_GATEWAY=true; WITH_KUADRANT=true; WITH_OTEL=true
-      WITH_MLFLOW=true; WITH_BUILDS=true; WITH_KIALI=true
-      WITH_AGENT_SANDBOX=true
-      shift ;;
+    --with-all)         WITH_ALL=true; shift ;;
     --skip-cluster)     SKIP_CLUSTER=true; shift ;;
-    --skip-mlflow)      WITH_MLFLOW=false; shift ;;
-    --skip-kuadrant)    WITH_KUADRANT=false; shift ;;
+    --skip-mlflow)      SKIP_MLFLOW=true; shift ;;
+    --skip-kuadrant)    SKIP_KUADRANT=true; shift ;;
     --build-images)     BUILD_IMAGES=true; shift ;;
     --preload-images)   PRELOAD_IMAGES=true; shift ;;
     --secrets-file)     SECRETS_FILE_ARG="$2"; shift 2 ;;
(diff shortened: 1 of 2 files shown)


scripts/kind/setup-kagenti.sh
```diff
@@ -101,6 +102,8 @@ while [[ $# -gt 0 ]]; do
       WITH_AGENT_SANDBOX=true
       shift ;;
     --skip-cluster)     SKIP_CLUSTER=true; shift ;;
+    --skip-mlflow)      WITH_MLFLOW=false; shift ;;
```
huang195: Order dependence: `--skip-mlflow` and `--skip-kuadrant` only override `--with-all` if they come *after* it in argv. The wrapper in `kind-full-test.sh` always emits the right order, so it works there, but a user calling `setup-kagenti.sh --skip-mlflow --with-all` directly will silently get MLflow back. Either (a) note this in the help text at line 133 ("must follow --with-all"), or (b) defer the `--with-all` expansion to after argv parsing (set a `WITH_ALL=true` flag, then expand at the end). Option (b) is the more robust fix; (a) is a one-line stopgap.

huang195 (approved):
Bel lavoro, Paolo — clean, focused enablement for resource-constrained dev. The dynamic `SETUP_ARGS` array in `kind-full-test.sh` is a pulita way to handle conditional flags, and the help-text update on `setup-kagenti.sh` reads well.

All required CI checks pass. I ran `bash -n` on both scripts locally — clean. Verified that `WITH_MLFLOW` / `WITH_KUADRANT` already exist in upstream and the new skip flags correctly toggle them off.

One real callout inline (order dependence on `--skip-*` vs `--with-all` for direct callers of `setup-kagenti.sh`). All four boxes in your test plan are unchecked — would be nice to confirm at least the lightweight install path before merge.

Two follow-up thoughts (out of scope, just thinking out loud):
- The same pattern would generalize to `--skip-otel`, `--skip-builds`, `--skip-kiali` etc. if `--with-all` ever needs more granular subtraction. The current design scales linearly with each skip flag added — fine for now, but worth thinking about whether `--with-all --without otel,builds` (comma-separated) becomes attractive past 4–5 skips.
- Same change is also needed in `scripts/ocp/setup-kagenti.sh` if OCP users hit the same pain. Out of scope for this PR but worth a follow-up issue.

Approving — ship it. 🇮🇹

**Areas reviewed**: Shell (2 scripts)
**Verified externally**: variable existence, `bash -n` clean, CI all green.

PR #1674, opened 2026-05-26
fix(backend): deduplicate env vars to prevent Sandbox creation failure

### LINKED ISSUE #1673: Duplicate env vars prevent Sandbox from starting when importing agents
## Description

When importing an agent through the UI with environment variables (via "Import Environment Variables"), duplicate env var names can occur in the generated pod spec. This prevents the Sandbox controller from starting the agent pod.

## Root Cause

Both `_build_env_vars()` in `agents.py` and `_build_tool_env_vars()` in `tools.py` start with `DEFAULT_ENV_VARS` (which includes `PORT`, `HOST`, `OTEL_EXPORTER_OTLP_ENDPOINT`, `KEYCLOAK_URL`, `UV_CACHE_DIR`) and then **append** user-provided env vars without checking for name collisions.

If a user imports a `.env` file containing any variable already in `DEFAULT_ENV_VARS`, the resulting container spec has duplicate env var entries.

## Impact

- Agent builds successfully but **no Sandbox pod is started**
- Affects the Sandbox workload type (the agent-sandbox controller likely rejects or mishandles duplicates)
- Affects any user who imports env vars overlapping with platform defaults

## Reproduction

1. Import an agent via the UI (e.g., `git-issue-agent`)
2. Use "Import Environment Variables" with a `.env` file that contains a `PORT` variable (or any other var in `DEFAULT_ENV_VARS`)
3. Select Sandbox workload type
4. Observe: build succeeds, but no sandbox/pod is created

## Fix

Deduplicate environment variables at the end of `_build_env_vars()` and `_build_tool_env_vars()`, keeping the last occurrence so that user-provided values override platform defaults.

## Reporter

@davidhadas

pdettori:
## Summary

- Deduplicate environment variables in `_build_env_vars()` (agents) and `_build_tool_env_vars()` (tools), keeping the last occurrence so user-provided values override `DEFAULT_ENV_VARS`
- Adds unit tests verifying no duplicate env var names appear in Sandbox manifests
- Prevents Sandbox controller from failing when imported `.env` files contain variables already in platform defaults (e.g. `PORT`, `HOST`)

Fixes #1673

## Test plan

- [x] Existing sandbox manifest tests pass (20/20)
- [x] New tests verify: user env vars override defaults, no duplicates in final list
- [ ] Manual: import git-issue-agent with `.env.openai`, verify Sandbox pod starts

Assisted-By: Claude Code

### DIFF
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -2556,8 +2556,9 @@ def _build_env_vars(request: "CreateAgentRequest") -> List[dict]:
 
                 env_vars.append(env_entry)
 
-    # Deduplicate environment variables, keeping the last occurrence
-    # This allows user-provided envVars to override DEFAULT_ENV_VARS
+    # Deduplicate environment variables, keeping the last occurrence.
+    # Precedence (last wins): DEFAULT_ENV_VARS < AGENT_ENDPOINT/SKILL_FOLDERS < user envVars.
+    # User overrides of AGENT_ENDPOINT or SKILL_FOLDERS are intentional (advanced use).
     seen = {}
     for env in env_vars:
         seen[env["name"]] = env
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -2555,7 +2555,13 @@ def _build_env_vars(request: "CreateAgentRequest") -> List[dict]:
                     }
 
                 env_vars.append(env_entry)
-    return env_vars
+
+    # Deduplicate environment variables, keeping the last occurrence
+    # This allows user-provided envVars to override DEFAULT_ENV_VARS
+    seen = {}
+    for env in env_vars:
+        seen[env["name"]] = env
+    return list(seen.values())
 
 
 def _build_common_labels(
--- a/kagenti/backend/app/routers/tools.py
+++ b/kagenti/backend/app/routers/tools.py
@@ -388,7 +388,13 @@ def _build_tool_env_vars(
                     }
 
                 env_vars.append(env_entry)
-    return env_vars
+
+    # Deduplicate environment variables, keeping the last occurrence
+    # This allows user-provided envVars to override DEFAULT_ENV_VARS
+    seen = {}
+    for env in env_vars:
+        seen[env["name"]] = env
+    return list(seen.values())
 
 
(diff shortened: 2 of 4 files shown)


kagenti/backend/app/routers/tools.py
```diff
@@ -388,7 +388,13 @@ def _build_tool_env_vars(
                     }
 
                 env_vars.append(env_entry)
-    return env_vars
+
+    # Deduplicate environment variables, keeping the last occurrence
+    # This allows user-provided envVars to override DEFAULT_ENV_VARS
+    seen = {}
+    for env in env_vars:
+        seen[env["name"]] = env
```
mrsabath, line 396: Same dedup fix is applied here as in `agents.py`, but `test_sandbox_manifest.py` only exercises the `_build_env_vars` path. Consider adding a parallel test for `_build_tool_env_vars` so a future regression in either path is caught.

kagenti/backend/app/routers/agents.py
```diff
@@ -2555,7 +2555,13 @@ def _build_env_vars(request: "CreateAgentRequest") -> List[dict]:
...
 
                 env_vars.append(env_entry)
-    return env_vars
+
+    # Deduplicate environment variables, keeping the last occurrence
+    # This allows user-provided envVars to override DEFAULT_ENV_VARS
+    seen = {}
+    for env in env_vars:
+        seen[env["name"]] = env
+    return list(seen.values())
```
mrsabath, line 2567: Last-wins dedup also lets user-supplied envVars override platform-injected `AGENT_ENDPOINT` and `SKILL_FOLDERS` (both appended before user envVars). For `PORT`/`HOST` this is the desired behavior, but an accidental user override of `AGENT_ENDPOINT` would silently break agent discovery. Consider either documenting the precedence in the comment, or excluding platform-managed names from override.

mrsabath (approved):
Tight, well-scoped fix. The last-wins dedup correctly lets user-provided envVars override `DEFAULT_ENV_VARS` (PORT, HOST, etc.), which is the documented intent. Tests target the bug class directly and are assertive. Two non-blocking suggestions inline.

**Areas reviewed**: Python (backend routers + tests)
**Commits**: 1 commit, signed-off, links #1673
**CI status**: required checks passing (Deploy & Test still running; E2E HyperShift awaits maintainer trigger)

PR #1677, opened 2026-05-26
fix(backend): clean up existing Build/BuildRun on re-import

### LINKED ISSUE #1676: Failed builds not visible in UI; re-import blocked by leftover Build/BuildRun resources
## Problem

When a build-from-source for an agent or tool **fails**, the following issues occur:

1. **Failed builds are invisible** — there is no UI surface showing past builds or their status. The only way to see build progress is during the import flow itself (via `BuildProgressPage`). If the user navigates away or closes the tab, the failed build is effectively hidden.

2. **Re-import is blocked** — attempting to re-import the same agent fails because the leftover `Build` and `BuildRun` custom resources still exist in the namespace. The backend tries to create a new Build with the same name and gets a conflict (409).

### Current behavior

- Build/BuildRun resources persist indefinitely after failure (no cleanup).
- The only cleanup path is deleting the agent entirely (which removes associated BuildRuns via label selector).
- No "Builds" section exists in the UI navigation — builds are only accessible inline during the import flow.

### Expected behavior

- Users can see failed builds and understand what went wrong.
- Users can re-import or rebuild without manual intervention or kubectl cleanup.

---

### Option A: Add a "Builds" section to the UI

Add a new **"Builds"** page under the "Agentic Workloads" section in the sidebar navigation (alongside Agents, Tools, Skills).

**Scope:**
- New route: `/builds` listing all Build/BuildRun resources across namespaces
- Table columns: Name, Namespace, Agent/Tool, Status (Pending/Running/Succeeded/Failed), Started, Duration
- Row actions: View logs, Delete build, Retry build
- Filter/sort by status, namespace, agent name
- Clicking a build row shows the existing `BuildProgressView` component (reuse)

**Backend changes:**
- The `GET /shipwright/builds` endpoint already exists (`shipwright_builds.py`) — extend it to support filtering and expose BuildRun failure messages
- Add `DELETE /shipwright/builds/{namespace}/{name}` endpoint to delete Build + associated BuildRuns

**Pros:**
- Full visibility into build history (successes and failures)
- Enables workflows like "check what's building" without navigating to each agent
- Consistent with how other platforms (OpenShift, Tekton Dashboard) surface builds

**Cons:**
- Larger UI surface to maintain
- Adds navigation complexity
- Most users may only care about builds when something fails

---

### Option B: Add build cleanup to the agent import/build flow

Add a **"Clean up previous builds"** mechanism so re-importing doesn't conflict.

**Scope:**
- Backend: Before creating a new Build, check if one already exists with the same name. If so, delete the old Build and its BuildRuns, then proceed with the new one (idempotent import).
- UI (Import page): If a previous build exists, show a warning: _"A previous build exists for this agent (status: Failed). It will be replaced."_ with a confirm action.
- UI (Build page): Add a "Delete Build" button on the build progress page when status is Failed, allowing manual cleanup.

**Backend changes:**
- Modify build creation logic in `agents.py` to detect and clean up existing builds before creating a new one
- Add `DELETE /agents/{namespace}/{name}/build` endpoint for explicit cleanup

**Pros:**
- Minimal UI changes — works within existing flows
- Solves the immediate blocker (re-import conflict) directly
- Lower implementation effort

**Cons:**
- Doesn't provide build history visibility
- Users still can't see failed builds they navigated away from
- Cleanup is implicit — could surprise users if they had intentional builds

---

## Recommendation

**Option B (cleanup on re-import) for v0.6.0-rc.8, with Option A as a follow-up for v0.7.0.**

Rationale:
- The immediate pain point is the **re-import blocker** — Option B solves this with minimal changes.
- Option B can ship in days; Option A requires new pages, routes, navigation changes, and testing.
- Option A is valuable long-term (especially for operators managing many agents) but not critical for rc.8.
- The two options are additive — Option B's cleanup endpoint would be reused by Option A's delete action.

### Suggested implementation for Option B (v0.6.0-rc.8)

1. **Backend** (`agents.py`): In the build creation path, if a Build CR already exists:
   - Delete all associated BuildRuns (existing pattern from agent deletion)
   - Delete the Build CR
   - Create the new Build CR
2. **UI** (`ImportAgentPage.tsx`): No change needed if backend handles cleanup transparently. Optionally show a toast: "Previous build cleaned up."
3. **UI** (`BuildProgressPage.tsx`): Add a "Delete Build" button when status is `Failed`, calling the new cleanup endpoint.

---

## Related code

- Backend build creation: `kagenti/backend/app/routers/agents.py` (trigger_buildrun, finalize_shipwright_build)
- Backend build listing: `kagenti/backend/app/services/shipwright_builds.py`
- UI build progress: `kagenti/ui-v2/src/pages/BuildProgressPage.tsx`
- UI import form: `kagenti/ui-v2/src/pages/ImportAgentPage.tsx`
- Build cleanup on delete: `kagenti/backend/app/routers/agents.py:1088-1118`

pdettori:
## Summary

- Adds `cleanup_existing_build()` helper in `shipwright_builds` service that deletes existing Build + BuildRun resources before creating new ones
- Called in both agent and tool source-deploy paths to prevent 409 conflicts on re-import
- Best-effort cleanup: 404s are silently ignored, other errors are logged but don't block the new build

## Problem

When a build-from-source fails and the user tries to re-import the same agent/tool, the request fails with a 409 because the leftover Build CR still exists. The only workaround was manual `kubectl delete` or deleting the entire agent.

## Changes

| File | Change |
|------|--------|
| `app/services/shipwright_builds.py` | New `cleanup_existing_build()` — deletes BuildRuns by label, then the Build CR |
| `app/routers/agents.py` | Calls cleanup before Build creation in source-deploy path |
| `app/routers/tools.py` | Same cleanup for tools |
| `tests/test_build_cleanup.py` | 8 tests (6 unit + 2 integration) |

## Test plan

- [x] 8 new unit/integration tests pass
- [x] All 358 existing backend tests pass (no regressions)
- [x] Linter clean
- [x] Manually verified on Kind cluster — re-import succeeds after failed build

Closes #1676

Assisted-By: Claude Code

### DIFF
--- a/kagenti/backend/app/services/shipwright_builds.py
+++ b/kagenti/backend/app/services/shipwright_builds.py
@@ -17,9 +17,65 @@ from app.core.constants import (
     SHIPWRIGHT_CRD_GROUP,
     SHIPWRIGHT_CRD_VERSION,
     SHIPWRIGHT_BUILDS_PLURAL,
+    SHIPWRIGHT_BUILDRUNS_PLURAL,
 )
 from app.models.shipwright import ShipwrightBuildListItem
 from app.services.kubernetes import KubernetesService
+
+
+logger = logging.getLogger(__name__)
+
+
+def cleanup_existing_build(
+    kube: KubernetesService,
+    namespace: str,
+    build_name: str,
+) -> None:
+    """Delete an existing Shipwright Build and its BuildRuns (best-effort).
+
+    Called before creating a new Build to prevent 409 conflicts on re-import.
+    Silently handles 404s (resource already gone) and logs warnings for other errors.
+    """
+    # Delete BuildRuns first (they reference the Build)
+    try:
+        buildruns = kube.list_custom_resources(
+            group=SHIPWRIGHT_CRD_GROUP,
+            version=SHIPWRIGHT_CRD_VERSION,
+            namespace=namespace,
+            plural=SHIPWRIGHT_BUILDRUNS_PLURAL,
+            label_selector=f"kagenti.io/build-name={build_name}",
+        )
+        for buildrun in buildruns:
+            br_name = buildrun.get("metadata", {}).get("name")
+            if br_name:
+                try:
+                    kube.delete_custom_resource(
+                        group=SHIPWRIGHT_CRD_GROUP,
+                        version=SHIPWRIGHT_CRD_VERSION,
+                        namespace=namespace,
+                        plural=SHIPWRIGHT_BUILDRUNS_PLURAL,
+                        name=br_name,
+                    )
+                except ApiException as e:
+                    if e.status != 404:
+                        logger.warning("Failed to delete BuildRun '%s': %s", br_name, e.reason)
+    except ApiException as e:
+        if e.status != 404:
+            logger.warning("Failed to list BuildRuns for '%s': %s", build_name, e.reason)
+        return
+
+    # Delete the Build CR
+    try:
+        kube.delete_custom_resource(
+            group=SHIPWRIGHT_CRD_GROUP,
+            version=SHIPWRIGHT_CRD_VERSION,
+            namespace=namespace,
+            plural=SHIPWRIGHT_BUILDS_PLURAL,
+            name=build_name,
+        )
+    except ApiException as e:
+        if e.status != 404:
+            logger.warning("Failed to delete Shipwright Build '%s': %s", build_name, e.reason)
 
 
 def format_shipwright_build_timestamp(timestamp) -> Optional[str]:
--- a/kagenti/backend/app/services/shipwright_builds.py
+++ b/kagenti/backend/app/services/shipwright_builds.py
@@ -58,13 +58,12 @@ def cleanup_existing_build(
                     )
                 except ApiException as e:
                     if e.status != 404:
-                        logger.warning("Failed to delete BuildRun '%s': %s", br_name, e.reason)
+                        logger.warning("Failed to delete BuildRun: status=%d", e.status)
     except ApiException as e:
         if e.status != 404:
-            logger.warning("Failed to list BuildRuns for '%s': %s", build_name, e.reason)
-        return
+            logger.warning("Failed to list BuildRuns for cleanup: status=%d", e.status)
 
-    # Delete the Build CR
+    # Delete the Build CR (independent of BuildRun list success)
     try:
         kube.delete_custom_resource(
             group=SHIPWRIGHT_CRD_GROUP,
@@ -75,7 +74,7 @@ def cleanup_existing_build(
         )
     except ApiException as e:
         if e.status != 404:
-            logger.warning("Failed to delete Shipwright Build '%s': %s", build_name, e.reason)
+            logger.warning("Failed to delete Shipwright Build: status=%d", e.status)
 
 
 def format_shipwright_build_timestamp(timestamp) -> Optional[str]:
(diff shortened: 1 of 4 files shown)


kagenti/backend/app/services/shipwright_builds.py
```diff
@@ -17,11 +17,67 @@
...
+            logger.warning("Failed to list BuildRuns for '%s': %s", build_name, e.reason)
+        return
+
+    # Delete the Build CR
+    try:
+        kube.delete_custom_resource(
+            group=SHIPWRIGHT_CRD_GROUP,
+            version=SHIPWRIGHT_CRD_VERSION,
+            namespace=namespace,
+            plural=SHIPWRIGHT_BUILDS_PLURAL,
```
mrsabath, line 72: Early `return` after a non-404 list-BuildRuns failure means the Build CR delete is skipped on transient list errors. If the goal is best-effort cleanup, consider attempting the Build delete anyway — listing BuildRuns and deleting the Build are independent operations, and the bug this PR fixes is precisely a leftover Build CR causing 409.

kagenti/backend/tests/test_build_cleanup.py
```diff
@@ -0,0 +1,266 @@
...
+        kube.list_custom_resources.side_effect = ApiException(status=404)
+
+        # Should not raise
+        cleanup_existing_build(kube, namespace="team1", build_name="new-agent")
+
+        # Should not attempt to delete anything
+        kube.delete_custom_resource.assert_not_called()
+
+    def test_skips_buildrun_without_name(self):
+        """BuildRuns missing metadata.name are skipped."""
```
mrsabath, line 96: nit: Name says `propagates_non_404_build_delete_error` but the assertion is "does not raise" (matches the actual best-effort behavior). Consider renaming to `test_swallows_non_404_build_delete_error` so the test name reflects the contract.

kagenti/backend/tests/test_build_cleanup.py
```diff
@@ -0,0 +1,266 @@
...
+        from fastapi.testclient import TestClient
+        from app.routers import agents
+        from app.services.kubernetes import get_kubernetes_service
+
+        app = FastAPI()
+        app.include_router(agents.router, prefix="/api/v1")
+
+        kube = MagicMock()
+        kube.list_custom_resources.return_value = [
+            {"metadata": {"name": "test-agent-run-old"}},
```
mrsabath, line 161: `assert len(delete_calls) >= 1` is too loose — the contract is "delete each BuildRun, then delete the Build". With this assertion, a regression that deletes the BuildRun but skips the Build CR would still pass. Tighten to `>= 2` (1 old BuildRun + 1 old Build) or, better, assert the exact delete sequence.

mrsabath (approved):
Solid fix for the re-import 409 problem. Cleanup ordering (BuildRuns before Build) is correct since Shipwright doesn't set ownerReferences on BuildRuns. Best-effort semantics are appropriate for a pre-create cleanup. Three non-blocking suggestions inline.

One thing to manually verify (not in scope for this PR): that the Build CR has no finalizer that would delay deletion past the subsequent create. K8s deletion is async, so if a finalizer holds the Build, the new create could still 409.

**Areas reviewed**: Python (backend service + routers + tests)
**Commits**: 1 commit, signed-off, links #1676
**CI status**: required checks passing (Deploy & Test pending; E2E HyperShift awaits maintainer trigger)

PR #1684, opened 2026-05-27
fix: hide registry secret field for OpenShift Internal Registry

### LINKED ISSUE #1683: UI: Registry secret field should be hidden for OpenShift Internal Registry
## Description

When selecting **OpenShift Internal Registry** as the container registry for builds, the UI shows a "Registry Secret Name" field pre-filled with `openshift-registry-secret`. This is incorrect — the OpenShift internal registry does not require a separate registry secret because pods access it via their service account.

## Current Behavior

1. Go to Import Agent/Tool → Build from Source
2. Select "OpenShift Internal Registry" 
3. The "Registry Secret Name" field appears with a default value of `openshift-registry-secret`
4. An "Authentication Required" alert is shown

## Expected Behavior

When "OpenShift Internal Registry" is selected:
- The registry secret field should be hidden (same as "Local Registry")
- The "Authentication Required" alert should not be shown
- The registry namespace field should still appear (maps to the OpenShift project)

## Affected Components

- `kagenti/ui-v2/src/pages/ImportAgentPage.tsx`
- `kagenti/ui-v2/src/pages/ImportToolPage.tsx`

pdettori:
## Summary

- Hide the "Registry Secret Name" field and "Authentication Required" alert when OpenShift Internal Registry is selected for builds
- The OpenShift internal registry uses service account authentication, not a separate registry secret
- Applies to both Import Agent and Import Tool pages

Fixes #1683

## Test plan

- [x] Select "OpenShift Internal Registry" → verify registry secret field is hidden
- [x] Select "Quay.io" or "Docker Hub" → verify registry secret field still appears with default value
- [x] Select "Local Registry" → verify registry secret field is hidden (existing behavior preserved)
- [ ] Verify registry namespace field still appears for OpenShift (needed for project name)

Assisted-By: Claude Code

### DIFF
--- a/kagenti/ui-v2/src/pages/ImportAgentPage.tsx
+++ b/kagenti/ui-v2/src/pages/ImportAgentPage.tsx
@@ -130,11 +130,12 @@ export const ImportAgentPage: React.FC = () => {
   const [registrySecret, setRegistrySecret] = useState('');
 
   // Update registry secret default when registry type changes
+  // OpenShift internal registry doesn't require credentials (uses service account)
   React.useEffect(() => {
-    if (registryType !== 'local') {
-      setRegistrySecret(`${registryType}-registry-secret`);
-    } else {
+    if (registryType === 'local' || registryType === 'openshift') {
       setRegistrySecret('');
+    } else {
+      setRegistrySecret(`${registryType}-registry-secret`);
     }
   }, [registryType]);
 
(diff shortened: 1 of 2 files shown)


evaline-ju (approved):
Correct fix: splits the `registryType !== 'local'` JSX block so the Namespace field remains visible for OpenShift (needed for project name), while the Registry Secret and Authentication Required alert are hidden. State reset in `useEffect` is symmetric across both pages. The backend safely handles the empty string that `ImportAgentPage` still sends at submission (`registryType !== 'local' ? registrySecret : undefined` → `''` for OpenShift) because `shipwright.py:254` uses a truthiness check on `pushSecretName`.

**nit**: `ImportToolPage` already uses `registrySecret || undefined` at submission (line 465), which sends `undefined`. `ImportAgentPage` sends `''` instead. Functionally safe, but worth aligning in a follow-up.

**Areas reviewed**: Frontend (TSX), YAML, Commit conventions, Backend submission logic
**Commits**: 2, all signed-off: yes
**CI status**: DCO/CodeQL pass; E2E pending maintainer trigger

evaline-ju (commented):
will require rebase for conflict

PR #1687, opened 2026-05-27
fix(ocp): set components.mlflow.enabled in kagenti chart install

pdettori:
## Summary

- The OCP installer provisions MLflow via RHOAI (Step 3d) but never passes `components.mlflow.enabled=true` to the main kagenti Helm chart (Step 4)
- The `ui-routes-job` helm hook renders its route discovery list gated on `components.mlflow.enabled`, so the MLflow route is excluded from the `kagenti-ui-config` ConfigMap
- This causes the UI observability page to hide the MLflow card (conditional render: `{mlflowUrl && (...)}`)
- Fix: add `--set components.mlflow.enabled=...` based on `SKIP_MLFLOW`, matching the existing pattern for `kagenti-operator-chart.mlflow.enable`

## Test plan

- [ ] Deploy on OCP with `./scripts/ocp/setup-kagenti.sh --kagenti-repo . --with-all`
- [ ] Verify `kubectl get configmap kagenti-ui-config -n kagenti-system -o yaml` contains `MLFLOW_DASHBOARD_URL` with a valid route URL
- [ ] Verify MLflow card appears on the observability page
- [ ] Deploy with `--skip-mlflow` and verify MLflow card does NOT appear

Assisted-By: Claude Code

### DIFF
--- a/charts/kagenti-deps/values.yaml
+++ b/charts/kagenti-deps/values.yaml
@@ -236,6 +236,7 @@ otel:
             span:
               - 'IsMatch(name, "^a2a\\..*")'
               - 'attributes["http.method"] != nil'
+              - 'attributes["openinference.span.kind"] != nil'
               - 'attributes["mcp.method.name"] == "initialize"'
               - 'attributes["mcp.method.name"] == "notifications/initialized"'
               - 'attributes["mcp.method.name"] == "notifications/cancelled"'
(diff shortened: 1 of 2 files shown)


huang195 (approved):
Bellissimo, three small fixes in one tight bundle. All verified:

- **Commit 1 (Dependabot)**: ignore syntax is correct — `[major, minor]` lets patch updates through while blocking the v47+ SIGILL territory.
- **Commit 2 (OCP MLflow)**: confirmed `components.mlflow.enabled` exists in `charts/kagenti/values.yaml` (default `false`) and is gated by `ui-routes-job.yaml` for the `MLFLOW_DASHBOARD_URL` route entry. The new `--set` line mirrors the existing `kagenti-operator-chart.mlflow.enable` pattern on the next line. Perfetto.
- **Commit 3 (OpenInference filter)**: kept symmetric across `filter/mlflow` and `filter/phoenix` — same rule in the same position in both processors. Nice catch that LangGraph/LangChain spans were silently dropping.

Tiny nit (non-blocking): the PR body only describes commit 2 — commits 1 and 3 are both independently merge-worthy fixes that aren't mentioned anywhere. A one-liner in the summary about the Dependabot ignore + OpenInference filter would help future git-archeologists. Not worth a re-push.

**Areas reviewed**: Dependabot config, Helm values (OTel filter), OCP install script
**Commits**: 3, all signed-off
**CI status**: Static checks green; E2E HyperShift gated on `/run-e2e`

Assisted-By: Claude Code

PR #1693, opened 2026-05-27
feat(skills): external skill registry references (#1691)

### LINKED ISSUE #1691: feat: external skill registries — fetch skills from remote registries at agent startup
## Summary

Add support for **external skill registries** in Kagenti. Currently, skills are stored as Kubernetes ConfigMaps containing the full skill file content. This feature introduces a lightweight **external skill reference**: a ConfigMap that holds only a pointer (URL + metadata) to a skill in a remote registry. At agent pod startup, an init container fetches the skill archive from the registry and writes it to a shared `emptyDir` volume. The agent mounts the result identically to a local skill — no agent framework changes required.

The initial target registry is [skillberry-store](https://github.ibm.com/skillberry/skillberry-store/), which serves skill archives as `tar.gz` bundles via REST API.

**Feature flag:** `kagenti_feature_flag_external_skills` (default: `False`)

**Design spec:** `docs/superpowers/specs/2026-05-27-external-skill-registries-design.md`

---

## Motivation

- Teams maintaining skills in a shared registry should not have to upload file content into every Kagenti namespace separately.
- Skills in a registry can be versioned and updated centrally; agents always fetch the declared version at startup.
- Enables integration with skill marketplaces (e.g., skillberry-store) without changes to the agent framework.

---

### New: External Skill Reference ConfigMap

```yaml
labels:
  kagenti.io/type: skill
  kagenti.io/source: external        # new
  kagenti.io/registry-type: skillberry
annotations:
  kagenti.io/registry-url: "https://skillberry.example.com"
  kagenti.io/registry-skill-name: "code-review"
  kagenti.io/registry-skill-version: "1.2.0"
data: {}   # intentionally empty
```

### New: `kagenti-skill-fetcher-scripts` ConfigMap

Kagenti-managed ConfigMap (created lazily in each team namespace) containing a shell script per registry type (`skillberry.sh`, `generic.sh`). Scripts handle download, retry (3×), and extraction. Adding a new registry type requires only a new script key — no backend code changes.

### New: Init Container Injection

For each external skill in an agent, the backend injects:
- An `emptyDir` volume for the fetched content
- An init container (`alpine:3`) that runs the registry-type script, writes to the `emptyDir`
- The main container mounts the `emptyDir` at `/app/skills/<name>` (identical to local skill mounts)

Local skills continue to use ConfigMap volume mounts unchanged.

### Backend (`skills.py`)
- New endpoint: `POST /skills/external`
- `Skill` model gains optional `source` and `externalInfo` fields
- `SkillDetail` for external skills returns empty `files` + registry metadata

### Backend (`agents.py`)
- `_get_linked_skill_mounts()` — split into local and external paths
- New `_ensure_fetcher_scripts_cm(kube, namespace)` — idempotent upsert of scripts ConfigMap
- New `_get_external_skill_init_containers()` — builds init container + volume specs

### Frontend
- Skill catalog: `External` badge on external skills
- Skill detail: "Registry Info" card replaces file tree for external skills
- Import page: new "From Registry" tab (feature-flagged)

---

## Acceptance Criteria

- [ ] `POST /skills/external` creates a reference ConfigMap with correct labels/annotations
- [ ] `GET /skills` lists external skills alongside local skills, with `source: "external"` in response
- [ ] Agent created with an external skill has an init container in the pod spec
- [ ] Init container fetches from the registry URL and extracts to the correct mount path
- [ ] `SKILL_FOLDERS` env var on the agent container includes the external skill path
- [ ] Pod fails to start (init container error) if registry is unreachable
- [ ] `kagenti-skill-fetcher-scripts` ConfigMap is created in the agent namespace before pod creation
- [ ] All behaviour is gated behind `kagenti_feature_flag_external_skills=False` by default
- [ ] UI shows "External" badge in skill catalog
- [ ] UI shows registry metadata (not file tree) on external skill detail page
- [ ] "From Registry" tab appears in import page when feature flag is enabled
- [ ] Linting and unit tests pass

---

## Out of Scope

- Caching fetched skill content across pod restarts
- Registry authentication (designed to accept credentials via env vars in a future PR)
- Helm-chart-level delivery of the fetcher scripts ConfigMap

---

Assisted-By: Claude Code

eranra:
## Summary

Adds support for **external skill registries** in Kagenti. Skills can now be registered as lightweight ConfigMap references (`kagenti.io/source=external`) pointing to a remote registry (e.g. skillberry-store) rather than embedding full file content. At agent pod startup, an `alpine:3` init container fetches the skill archive and mounts it transparently — the agent sees identical mount paths for local and external skills.

- New endpoint: `POST /skills/external` creates a registry-reference ConfigMap
- `GET /skills` and `GET /skills/{ns}/{name}` include `source`/`externalInfo` fields
- Agent manifest builders inject init containers + emptyDir volumes for external skills
- A Kagenti-managed `kagenti-skill-fetcher-scripts` ConfigMap holds per-registry shell scripts (skillberry, generic); new registry types require only a new script key
- Gated behind `kagenti_feature_flag_external_skills` (default: `False`)

## Frontend

- Skill catalog: "External" badge on registry-reference skills
- Skill detail: Registry info card (type, URL, skill name, version) replaces file tree for external skills
- Import page: "From Registry" tab (shown when feature flag is enabled)

## Design

Design spec: `docs/superpowers/specs/2026-05-27-external-skill-registries-design.md`
Implementation plan: `docs/superpowers/plans/2026-05-27-external-skill-registries.md`

## Test plan

- [ ] Enable `KAGENTI_FEATURE_FLAG_EXTERNAL_SKILLS=true` and `KAGENTI_FEATURE_FLAG_SKILLS=true`
- [ ] `POST /skills/external` with skillberry payload — verify ConfigMap created with correct labels/annotations
- [ ] `GET /skills?namespace=team1` — verify external skill appears with `source: "external"` and `externalInfo`
- [ ] Create agent referencing the external skill — verify pod spec has `fetch-skill-0` init container with correct env vars
- [ ] Verify `SKILL_FOLDERS` env var includes the external skill mount path
- [ ] Disable flag — verify `POST /skills/external` returns 404
- [ ] UI: catalog shows "External" badge; detail shows registry card; import shows "From Registry" tab

Closes kagenti/kagenti#1691

Assisted-By: Claude Code


eranra:
@pavelanni FYI ^^^ + please add additional members from RH that need to be looped-in

kagenti/backend/app/routers/agents.py
```diff
@@ -2522,12 +2533,213 @@ def _get_linked_skill_mounts(
...
+#!/bin/sh
+set -e
+
+apk add -q --no-cache curl 2>/dev/null || true
+echo "Fetching skill from ${REGISTRY_URL}"
+
+RETRIES=3
+DELAY=2
+for i in $(seq 1 $RETRIES); do
+    if curl -fsSL -o /tmp/skill.tar.gz "${URL}"; then
```
pdettori: **must-fix (correctness bug):** `${URL}` is undefined in `_GENERIC_SH`. Should be `${REGISTRY_URL}`.

The skillberry script correctly constructs a local `URL` variable from `${REGISTRY_URL}/skills/${SKILL_NAME}/export-anthropic`, but the generic script never defines `URL` — it only sets `REGISTRY_URL` via the env var. This means the generic fetcher will always curl an empty string and fail.

Note: the test file (`test_agents_external_skills.py`) has the correct version with `${REGISTRY_URL}`, so the bug was masked by the test/source divergence.

kagenti/backend/app/routers/agents.py
```diff
@@ -2522,12 +2533,213 @@ def _get_linked_skill_mounts(
...
+
+        init_containers.append(
+            {
+                "name": f"fetch-skill-{index}",
+                "image": SKILL_FETCHER_IMAGE,
+                "command": [
+                    "/bin/sh",
+                    "-c",
+                    (
+                        f"SCRIPT=/fetcher-scripts/{registry_type}.sh; "
```
pdettori: **suggestion:** `registry_type` is interpolated into a shell command via Python f-string. While K8s label validation restricts the charset (no shell metacharacters possible), defense-in-depth says prefer the env var reference:

```python
'SCRIPT=/fetcher-scripts/${REGISTRY_TYPE}.sh; '
```

The `REGISTRY_TYPE` env var is already being set in the `env` list below, so this is a one-line change that avoids Python string interpolation of user-controlled data into shell code.

kagenti/backend/app/routers/skills.py
```diff
@@ -426,6 +491,74 @@ async def create_skill(
         raise HTTPException(status_code=exc.status or 500, detail=str(exc))
 
 
+@router.post(
+    "/external",
+    response_model=CreateSkillResponse,
+    dependencies=[Depends(require_roles(ROLE_OPERATOR))],
+)
+async def create_external_skill(
```
pdettori, line 516: **suggestion:** No validation on `registryUrl` beyond `registryType` being non-empty. Consider at minimum:
1. Validate scheme is `http://` or `https://` (rejects `file://`, `ftp://`, etc.)
2. Validate `registrySkillName` against `^[a-zA-Z0-9][a-zA-Z0-9._-]*$` to prevent path traversal in mount paths
3. Add `max_length` constraints to string fields in `CreateExternalSkillRequest`

The init container will curl whatever URL is stored here, so this is the right place to enforce URL safety.

kagenti/backend/app/core/constants.py
```diff
@@ -172,6 +172,16 @@
...
 
+# External skill registry constants
+SKILL_SOURCE_LABEL = "kagenti.io/source"
+SKILL_SOURCE_EXTERNAL = "external"
+SKILL_REGISTRY_TYPE_LABEL = "kagenti.io/registry-type"
+SKILL_REGISTRY_URL_ANNOTATION = "kagenti.io/registry-url"
+SKILL_REGISTRY_SKILL_NAME_ANNOTATION = "kagenti.io/registry-skill-name"
+SKILL_REGISTRY_SKILL_VERSION_ANNOTATION = "kagenti.io/registry-skill-version"
+SKILL_FETCHER_SCRIPTS_CM = "kagenti-skill-fetcher-scripts"
+SKILL_FETCHER_IMAGE = "alpine:3"
```
pdettori: **suggestion:** `alpine:3` is a mutable tag that resolves to different images over time. For reproducibility and supply-chain safety, consider pinning to a specific minor version (e.g., `alpine:3.21.3`) or a SHA digest.

kagenti/backend/app/routers/agents.py
```diff
@@ -2522,12 +2533,213 @@ def _get_linked_skill_mounts(
...
+                    "configMap": {"name": SKILL_FETCHER_SCRIPTS_CM},
+                }
+            )
+            fetcher_vol_added = True
+
+        volumes.append({"name": emptydir_vol_name, "emptyDir": {}})
+
+        init_containers.append(
+            {
+                "name": f"fetch-skill-{index}",
```
pdettori, line 2690: **suggestion:** The init container has no `resources` block. Without limits, a malicious registry could serve a multi-GB archive causing OOM. Consider adding:

```python
"resources": {
    "requests": {"memory": "32Mi", "cpu": "50m"},
    "limits": {"memory": "128Mi", "cpu": "200m"},
},
```

Also consider adding `--max-filesize 50M` to the curl commands in both fetcher scripts.

pdettori (changes requested):
Well-structured feature with good separation (feature flag, design docs, incremental commits). The critical issue is a correctness bug where `_GENERIC_SH` uses an undefined `${URL}` variable — the generic registry fetcher will fail on every invocation. The test mock has the correct variable name, which masked the bug.

**Areas reviewed:** Python (backend), Helm/K8s, Shell scripts, Frontend (TypeScript), Docs
**Commits:** 32 commits, all signed-off ✓
**CI status:** All passing ✓

Assisted-By: Claude Code

kagenti/backend/app/routers/agents.py
eranra: Fixed in commit `287ff67f`. The curl command in `_GENERIC_SH` now uses `${REGISTRY_URL}` directly — the skillberry script defines a local `URL` variable because it appends a path suffix, but the generic script should curl `${REGISTRY_URL}` as-is.

kagenti/backend/app/routers/agents.py
eranra: Fixed in commit `f399c2ee`. Replaced `f"SCRIPT=/fetcher-scripts/{registry_type}.sh; "` with `'SCRIPT=/fetcher-scripts/${REGISTRY_TYPE}.sh; '`. The `REGISTRY_TYPE` env var is already injected into the container, so the shell expands it directly — no Python string interpolation of user-supplied data into the shell command.

kagenti/backend/app/routers/skills.py
eranra: Fixed in commit `4600c226`. Added to `CreateExternalSkillRequest`:\n- `@field_validator("registryUrl")` — rejects anything other than `http://` or `https://` schemes\n- `@field_validator("registrySkillName")` — enforces `^[a-zA-Z0-9][a-zA-Z0-9._-]*$`\n- `Field(..., max_length=...)` constraints on all string fields (253 for names, 63 for namespace/type/version, 2048 for URL)

kagenti/backend/app/core/constants.py
eranra: Fixed in commit `2a04b5ba`. Changed `SKILL_FETCHER_IMAGE` from `"alpine:3"` to `"alpine:3.21.3"` in `constants.py`. Used `3.21.3` as suggested; can be bumped to a newer patch in a follow-on if needed.

kagenti/backend/app/routers/agents.py
eranra: Fixed in commit `7f0d5ba6`. Added a `resources` block (32Mi/50m request, 128Mi/200m limit) to the init container spec, and `--max-filesize 52428800` (50 MiB) to the `curl` invocation in both `skillberry.sh` and `generic.sh`.

kagenti/backend/app/routers/agents.py
```diff
@@ -2522,12 +2533,217 @@ def _get_linked_skill_mounts(
...
+        return labels.get(SKILL_SOURCE_LABEL) == SKILL_SOURCE_EXTERNAL
+    except ApiException:
+        return False
+
+
+_SKILLBERRY_SH = """\
+#!/bin/sh
+set -e
+
+apk add -q --no-cache curl unzip 2>/dev/null || true
```
pdettori, line 2552: **suggestion**: Shell scripts use `set -e` but not `set -eu`. Adding `-u` would cause the script to fail immediately if an expected environment variable (e.g., `REGISTRY_URL`, `SKILL_NAME`) is accidentally omitted from the init container spec, rather than silently interpolating an empty string into the curl URL. Since these are `/bin/sh` scripts, `pipefail` isn't available, but `set -eu` is portable POSIX.

kagenti/ui-v2/src/utils/validation.ts
```diff
@@ -51,3 +51,32 @@ export const isValidImageTag = (tag: string): boolean => {
...
+ * Substitutes port 8002 (UI) for the stored API port and appends the skill path.
+ * Returns '' if registryUrl is not a valid URL.
+ */
+export const getSkillberryUiUrl = (registryUrl: string, skillName: string): string => {
+  try {
+    const url = new URL(registryUrl);
+    url.port = '8002';
+    return `${url.origin}/skills/${skillName}`;
+  } catch {
+    return '';
```
pdettori, line 80: **suggestion**: `skillName` is interpolated directly into the URL path without encoding. If a skill name were to contain `/` or other URL-special characters (unlikely given the backend regex validation, but defense-in-depth), the derived URL would be malformed. Consider:
```ts
return `${url.origin}/skills/${encodeURIComponent(skillName)}`;
```

pdettori (approved):
Well-structured feature addition gated behind `kagenti_feature_flag_external_skills` (off by default). Security concerns are properly addressed: image pinned to `alpine:3.21.3`, input validation on URL scheme and skill name format, resource limits on init containers, 50MB download cap, retry logic with proper error handling, and RBAC enforcement (`ROLE_OPERATOR`). CI is fully green. All 38 commits have DCO sign-off.

Two non-blocking suggestions below.

PR #1700, opened 2026-05-27
fix(otel): remove span filter processors that cause orphaned traces

### LINKED ISSUE #1606: bug: OTel Collector span filters cause orphaned spans in trace views
## Description

The OpenTelemetry Collector pipelines for both Phoenix and MLflow backends include `filter/` processors that drop spans matching certain patterns (A2A protocol spans, HTTP-attributed spans, and several MCP method spans). Dropping these intermediate spans breaks the parent-child span hierarchy — child spans that reference a dropped parent appear detached from their actual ancestor, connecting only to the trace root.

This makes traces misleading and hard to interpret: agent interaction flows appear as disconnected fragments rather than coherent call trees.

## Dropped span patterns

```yaml
filter/phoenix:  # (and identical filter/mlflow)
  traces:
    span:
      - 'IsMatch(name, "^a2a\\..*")'
      - 'attributes["http.method"] != nil'
      - 'attributes["mcp.method.name"] == "initialize"'
      - 'attributes["mcp.method.name"] == "notifications/initialized"'
      - 'attributes["mcp.method.name"] == "notifications/cancelled"'
      - 'attributes["mcp.method.name"] == "tools/list"'
```

## Expected Behavior

Traces display complete parent-child span hierarchies. Agent interactions (A2A, MCP) are visible as coherent nested call trees in Phoenix and MLflow.

## Actual Behavior

Child spans whose parent was filtered appear as orphans — they attach directly to the trace root, losing all intermediate context. This is especially problematic for A2A spans (`^a2a\..*`) which often serve as parents for downstream LLM/tool spans.

## Root Cause

The OTel Collector `filter` processor drops spans without re-parenting their children. When a mid-tree span is dropped, its children retain a `parent_span_id` that no longer exists in the exported trace, causing backends to display them as detached.

## Workaround

Remove the `filter/phoenix` and `filter/mlflow` processors entirely from `charts/kagenti-deps/values.yaml`, and update the pipeline definitions to skip the filter stage:

```yaml
# Before
processors: [memory_limiter, filter/phoenix, transform/genai_to_openinference, batch]
# After
processors: [memory_limiter, transform/genai_to_openinference, batch]
```

```yaml
# Before
processors: [memory_limiter, filter/mlflow, batch]
# After
processors: [memory_limiter, batch]
```

The observability backends (Phoenix, MLflow) can handle the full span volume, and complete traces are more valuable than reduced storage costs.

## Environment

- Component: `charts/kagenti-deps/values.yaml`
- Affected pipelines: `traces/phoenix`, `traces/mlflow`
- OTel Collector configuration section: `otel.phoenixConfig.processors`, `otel.mlflowConfig.processors`

yoavkatz:
## Summary

- Remove `filter/phoenix` and `filter/mlflow` OTel Collector processors that drop intermediate spans (A2A protocol, HTTP-attributed, MCP lifecycle spans)
- These filters break parent-child span hierarchy — child spans referencing a dropped parent appear detached from their actual ancestor, connecting only to the trace root
- Traces now preserve complete span trees for accurate debugging and observability

Closes #1606

## Test plan

- [ ] Deploy to Kind cluster and verify OTel collector starts without errors
- [ ] Send agent requests and confirm trace views in Phoenix show complete span hierarchies
- [ ] Verify MLflow trace ingestion shows full parent-child relationships
- [ ] Confirm no orphaned/detached spans appear in trace visualizations

Assisted-By: Claude Code

### DIFF
--- a/charts/kagenti-deps/values.yaml
+++ b/charts/kagenti-deps/values.yaml
@@ -231,16 +231,6 @@ otel:
           tls:
             insecure: true
       processors:
-        filter/phoenix:
-          traces:
-            span:
-              - 'IsMatch(name, "^a2a\\..*")'
-              - 'attributes["http.method"] != nil'
-              - 'attributes["openinference.span.kind"] != nil'
-              - 'attributes["mcp.method.name"] == "initialize"'
-              - 'attributes["mcp.method.name"] == "notifications/initialized"'
-              - 'attributes["mcp.method.name"] == "notifications/cancelled"'
-              - 'attributes["mcp.method.name"] == "tools/list"'
         transform/genai_to_openinference:
           trace_statements:
             - context: span
@@ -276,7 +266,7 @@ otel:
         pipelines:
           traces/phoenix:
             receivers: [otlp]
-            processors: [memory_limiter, filter/phoenix, transform/genai_to_openinference, batch]
+            processors: [memory_limiter, transform/genai_to_openinference, batch]
             exporters: [otlp/phoenix]
     mlflowConfig:
       exporters:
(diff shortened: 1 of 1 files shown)


pdettori (approved):
Clean, well-scoped fix. Removing the span filter processors restores correct parent-child trace hierarchy — the observability correctness gain outweighs any noise increase.

Verified:
- YAML structure valid in both phoenix and mlflow config sections after removal
- Pipeline processor lists correctly updated (no dangling references)
- Commit signed-off, conventional format
- All CI checks passing

PR #1701, opened 2026-05-27
fix(mlflow): unblock AI Gateway endpoints and start job runner

### LINKED ISSUE #1605: bug: MLflow AI Gateway and evaluation jobs non-functional — OIDC auth blocks endpoints, job runner not started
## Description

MLflow's AI Gateway feature (`/endpoints/`, `/gateway/`, `/api/2.0/endpoints/`) and evaluation jobs are completely non-functional in the current deployment. This blocks using MLflow's AI capabilities (LLM evaluation, model comparison, prompt engineering) on the platform.

Three related problems prevent AI Gateway and job execution from working:

1. **Gateway endpoints blocked by OIDC auth** — The `/endpoints/`, `/gateway/`, and `/api/2.0/endpoints/` paths are protected by both the FastAPI permission validator and the Flask `before_request` hook from mlflow-oidc-auth. These endpoints are called internally by MLflow itself (not by end users with sessions), so they fail with 401.

2. **Huey job runner never started** — Our deployment bypasses the `mlflow server` CLI (to apply OIDC patches), but the CLI is responsible for launching the Huey consumer subprocess that executes async jobs. Without it, evaluation jobs are queued but never run. The job runner also requires `_MLFLOW_HUEY_STORAGE_PATH` and `MLFLOW_GATEWAY_URI` environment variables that were never set.

3. **Health probes use wrong path and are auth-blocked** — Readiness/liveness probes target `/version`, which is protected by OIDC auth. Kubelet probes carry no session cookie, so they get 401s, potentially causing pod restarts. The `/health` endpoint is the correct probe target but also needs to be exempted from auth.

4. **Memory limit too low** — Running MLflow server + job runner subprocess + OIDC auth within 2Gi causes OOM kills under evaluation workloads.

## Expected Behavior

- MLflow AI Gateway endpoints are accessible for internal LLM routing
- Evaluation jobs submitted through the UI or API execute successfully
- Health probes pass without authentication
- Pod remains stable under evaluation workloads

## Actual Behavior

- Gateway endpoints return 401 Unauthorized
- Jobs are queued but never execute (no consumer process)
- Probes intermittently fail causing pod restarts
- Pod OOM-killed during evaluation runs

## Workaround

A workaround is implemented in `charts/kagenti-deps/templates/mlflow.yaml`:

### 1. Set environment variables for job runner

```bash
export _MLFLOW_HUEY_STORAGE_PATH=$(mktemp -d /dev/shm/mlflow-huey-XXXXXX)
export MLFLOW_GATEWAY_URI="http://localhost:5000"
```

### 2. Exempt health probes and gateway endpoints from FastAPI OIDC auth

```python
original_is_unprotected = AuthMiddleware._is_unprotected_route
_UNPROTECTED_PREFIXES = ("/v1/traces", "/health", "/version", "/ping", "/endpoints/", "/gateway/", "/api/2.0/endpoints/")
def patched_is_unprotected(self, path: str) -> bool:
    if path.startswith(_UNPROTECTED_PREFIXES):
        return True
    return original_is_unprotected(self, path)
AuthMiddleware._is_unprotected_route = patched_is_unprotected
```

### 3. Skip FastAPI permission validation for gateway endpoints

```python
original_find_validator = fpm._find_fastapi_validator
def patched_find_validator(path: str):
    if path.startswith(("/v1/traces", "/endpoints/", "/gateway/", "/api/2.0/endpoints/")):
        return None
    return original_find_validator(path)
fpm._find_fastapi_validator = patched_find_validator
```

### 4. Exempt gateway endpoints from Flask-level auth (before_request hook)

```python
import mlflow_oidc_auth.hooks.before_request as _br
_original_flask_is_unprotected = _br._is_unprotected_route
def _patched_flask_is_unprotected(path: str) -> bool:
    if path.startswith(("/endpoints/", "/gateway/", "/api/2.0/endpoints/")):
        return True
    return _original_flask_is_unprotected(path)
_br._is_unprotected_route = _patched_flask_is_unprotected
```

### 5. Launch Huey job runner as background subprocess

```python
import subprocess, sys, time as _time
_server_up_time = str(int(_time.time() * 1000))
_job_runner_proc = subprocess.Popen(
    [sys.executable, "-m", "mlflow.server.jobs._job_runner"],
    env={
        **os.environ,
        "MLFLOW_SERVER_PID": str(os.getpid()),
        "_MLFLOW_SERVER_UP_TIME": _server_up_time,
        "MLFLOW_TRACKING_URI": os.environ["MLFLOW_BACKEND_STORE_URI"],
        "MLFLOW_DEPLOYMENTS_TARGET": "http://localhost:5000",
    },
)
```

### 6. Switch probe paths and bump memory

```yaml
readinessProbe:
  httpGet:
    path: /health  # was /version
    port: 5000
livenessProbe:
  httpGet:
    path: /health  # was /version
    port: 5000
resources:
  limits:
    memory: "4Gi"  # was 2Gi
```

These patches are fragile and tightly coupled to mlflow-oidc-auth internals. A proper fix should be contributed upstream to mlflow-oidc-auth (configurable route exclusions) or handled via a dedicated sidecar/init approach for the job runner.

## Environment

- Component: `charts/kagenti-deps/templates/mlflow.yaml`
- MLflow version: deployed via kagenti-deps chart
- Auth plugin: mlflow-oidc-auth v7+

yoavkatz:
## Summary

- Exempt `/endpoints/`, `/gateway/`, `/api/2.0/endpoints/` from OIDC auth at both FastAPI middleware and Flask `before_request` hook layers — these AI Gateway endpoints need unauthenticated access for LLM routing (Istio mTLS provides transport security)
- Start the Huey job runner subprocess and configure required env vars (`_MLFLOW_HUEY_STORAGE_PATH`, `MLFLOW_GATEWAY_URI`) that are normally set by the `mlflow server` CLI which we bypass
- Bump memory limit from 2Gi to 4Gi to accommodate job runner + gateway overhead

Closes #1605

## Test plan

- [ ] Deploy to Kind cluster with MLflow enabled
- [ ] Verify `/endpoints/` API responds without 401/403 (unauthenticated)
- [ ] Verify `/gateway/` routes LLM calls correctly
- [ ] Confirm Huey job runner process is running (`ps aux | grep job_runner`)
- [ ] Submit an evaluation job and verify it completes
- [ ] Confirm health/version probes pass without auth errors in pod logs

Assisted-By: Claude Code

### DIFF
--- a/charts/kagenti-deps/templates/mlflow.yaml
+++ b/charts/kagenti-deps/templates/mlflow.yaml
@@ -175,6 +175,10 @@ spec:
           export _MLFLOW_SERVER_FILE_STORE="postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@postgres:5432/mlflow" && \
           export MLFLOW_HOST="0.0.0.0" && \
           export MLFLOW_PORT="5000" && \
+          # Job execution requires Huey storage (normally set by `mlflow server` CLI)
+          export _MLFLOW_HUEY_STORAGE_PATH=$(mktemp -d /dev/shm/mlflow-huey-XXXXXX) && \
+          # Gateway URI for job subprocesses to route LLM calls through the server
+          export MLFLOW_GATEWAY_URI="http://localhost:5000" && \
           # Initialize MLflow database (create tables + run alembic upgrades)
           # Uses _initialize_tables which does create_all() first (safe for fresh DB)
           # then _upgrade_db() for any pending migrations.
@@ -199,17 +203,17 @@ spec:
           assert hasattr(fpm, "_find_fastapi_validator"), \
               "mlflow-oidc-auth API changed: _find_fastapi_validator not found"
 
-          # Patch 1: Exclude /v1/traces from OIDC auth
-          # The OTel collector authenticates via OAuth2 client credentials (Bearer token)
-          # but may not have a user profile in the MLflow users table.
-          # Skip auth entirely — Istio mTLS protects this path.
+          # Patch 1: Exclude health/version probes and /v1/traces from OIDC auth
+          # - /health, /version: kubelet probes have no session/cookie
+          # - /v1/traces: OTel collector uses Bearer token, no MLflow user profile
           original_is_unprotected = AuthMiddleware._is_unprotected_route
+          _UNPROTECTED_PREFIXES = ("/v1/traces", "/health", "/version", "/ping", "/endpoints/", "/gateway/", "/api/2.0/endpoints/")
           def patched_is_unprotected(self, path: str) -> bool:
-              if path.startswith("/v1/traces"):
+              if path.startswith(_UNPROTECTED_PREFIXES):
                   return True
               return original_is_unprotected(self, path)
           AuthMiddleware._is_unprotected_route = patched_is_unprotected
-          print("Excluded /v1/traces from OIDC auth (mTLS via Istio)")
+          print("Excluded /health, /version, /ping, /v1/traces from OIDC auth")
 
           # Patch 2: Skip permission validation for /v1/traces
           # mlflow-oidc-auth v7+ adds a FastAPI permission middleware that checks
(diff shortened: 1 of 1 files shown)


yoavkatz:
This PR was need to allow to run MLFlow built in analysis.  Since a new trace analysis component (#2134 ) is introduced in Kagenti, this is no longer needed.

PR #1704, opened 2026-05-27
Feat: Add --with-examples to kind setup-kagenti.sh to simplify tests and demos

esnible:
See https://github.com/kagenti/kagenti/issues/1524#issuecomment-4421457959 for rationale.

## Summary

This saves time for Kagenti core developers who frequently re-install on Kind.
This will also make it easier to write docs and tutorials, as most of them begin with lengthly UI steps to get a working agent.

- First run `./scripts/kind/setup-kagenti.sh --with-istio --with-spire --with-ui --with-backend --with-mlflow --with-examples`
- Then start Ollama with `ollama start`
- Then do `kubectl -n team1 set env deployment/weather-service LLM_API_BASE=http://host.docker.internal:11434/v1` so that the Weather service uses Ollama
- Then try the weather tool and weather agent through the Kagenti UI

### DIFF
--- a/scripts/kind/setup-kagenti.sh
+++ b/scripts/kind/setup-kagenti.sh
@@ -55,6 +55,7 @@ SKIP_MLFLOW=false
 SKIP_KUADRANT=false
 BUILD_IMAGES=false
 PRELOAD_IMAGES=false
+INSTALL_EXAMPLES=false
 DRY_RUN=false
 SECRETS_FILE_ARG=""
 CONTAINER_ENGINE="${CONTAINER_ENGINE:-docker}"
@@ -121,6 +122,7 @@ while [[ $# -gt 0 ]]; do
     --domain)           DOMAIN="$2"; shift 2 ;;
     --kagenti-values)   KAGENTI_VALUES_FILES+=("--values" "$2"); shift 2 ;;
     --kagenti-deps-values) KAGENTI_DEPS_VALUES_FILES+=("--values" "$2"); shift 2 ;;
+    --with-examples)    INSTALL_EXAMPLES=true; shift ;;
     --dry-run)          DRY_RUN=true; shift ;;
     -h|--help)
       echo "Usage: $0 [OPTIONS]"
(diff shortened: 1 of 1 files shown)


pdettori:
I like the idea @esnible !

scripts/kind/setup-kagenti.sh
```diff
@@ -158,6 +160,7 @@ while [[ $# -gt 0 ]]; do
       echo "                      Helm override file to apply to Kagenti chart"
       echo "  --kagenti-deps-values FILE"
       echo "                      Helm override file to apply to Kagenti-deps chart"
+      echo "  --with-examples     Install weather agent and weather tool examples"
```
pdettori, line 163: **nit / discoverability**: `--with-examples` is placed in "Other options" between two `--values` config-file flags. Users scanning for what components to install look in the "Component flags" block at the top of the help. Moving it there (or adding a short cross-reference) makes it easier to find.

pdettori, line 163: **suggestion**: The help text doesn't mention prerequisites. The PR description documents them (run with `--with-istio --with-spire --with-ui --with-backend`), but users calling `--help` won't see that. Consider: `Install weather agent and tool examples (requires --with-backend --with-istio for full functionality)`

scripts/kind/setup-kagenti.sh
```diff
@@ -1236,6 +1242,26 @@ else
 fi
 echo ""
 
+# ============================================================================
+# Step 9b: Install Examples
+# ============================================================================
+log_info "Step 9b: Agent and tool examples (weather)"
+
+if $INSTALL_EXAMPLES; then
```
pdettori, line 1250: **nit / UX**: Other optional flags with implicit deps auto-enable their prerequisites (e.g. `--with-ui` auto-enables `--with-backend`). A soft guard here — even just a `log_warn` when `--with-examples` is set without `--with-backend` — would prevent a confusing half-deployed demo state.

pdettori (approved):
Clean, focused addition. The `--with-examples` flag is well-structured: it defaults to false, respects `--dry-run`, is intentionally excluded from `--with-all` with an explanatory comment, and shows a useful post-install LLM endpoint hint.

The **Deploy & Test** CI failure is a pre-existing regression affecting multiple open PRs (#1705, #1711) — not caused by this change. All other checks pass.

**CI**: Deploy & Test failing but pre-existing; all other checks pass ✅

PR #1705, opened 2026-05-27
fix(ocp): fix Kiali and MLflow dashboard links on OpenShift

pdettori:
## Summary
- Fix MLflow route lookup to use configurable namespace (`redhat-ods-applications` by default) instead of hardcoded `kagenti-system`, matching where RHOAI creates the Route
- Increase Kiali route poll timeout from 5s to 30s to allow Kiali operator time to reconcile the Route after CR creation
- Remove `localtest.me` fallback for Kiali URL in the backend — the UI already hides the card when URL is empty

## Test plan
- [ ] `helm template charts/kagenti/ --set openshift=true --set components.mlflow.enabled=true` shows `redhat-ods-applications,mlflow,...` in ROUTES_DATA
- [ ] Unit tests pass: `uv run pytest kagenti/backend/tests/test_dashboard_config.py -v`
- [ ] Deploy on OCP with RHOAI MLflow and verify MLflow link appears in Observability page
- [ ] Verify Kiali link resolves after operator reconciliation (within 30s)
- [ ] Kind (non-OpenShift) path unaffected — route job is only rendered when `openshift=true`

Assisted-By: Claude Code

### DIFF
--- a/charts/kagenti/values.yaml
+++ b/charts/kagenti/values.yaml
@@ -34,8 +34,10 @@ components:
   mlflow:
     enabled: false
     # Namespace where the MLflow Route lives on OpenShift.
-    # RHOAI-managed MLflow creates its Route in redhat-ods-applications.
-    routeNamespace: "redhat-ods-applications"
+    # Defaults to the release namespace (kagenti-system).
+    # RHOAI-managed MLflow creates its Route in redhat-ods-applications;
+    # set routeNamespace: "redhat-ods-applications" when using RHOAI.
+    routeNamespace: ""
 
 # Secrets — override via a .secrets.yaml file or --set secrets.<key>=<value>.
 # githubUser/githubToken are optional: only needed to deploy agents/tools from
--- a/charts/kagenti/values.yaml
+++ b/charts/kagenti/values.yaml
@@ -33,6 +33,9 @@ components:
     enabled: false
   mlflow:
     enabled: false
+    # Namespace where the MLflow Route lives on OpenShift.
+    # RHOAI-managed MLflow creates its Route in redhat-ods-applications.
+    routeNamespace: "redhat-ods-applications"
 
 # Secrets — override via a .secrets.yaml file or --set secrets.<key>=<value>.
 # githubUser/githubToken are optional: only needed to deploy agents/tools from
--- a/kagenti/backend/app/routers/config.py
+++ b/kagenti/backend/app/routers/config.py
@@ -105,7 +105,7 @@ async def get_dashboard_config() -> DashboardConfigResponse:
 
     return DashboardConfigResponse(
         traces=settings.traces_dashboard_url,
-        network=settings.network_dashboard_url or f"http://kiali.{domain}:8080",
+        network=settings.network_dashboard_url,
         mlflow=settings.mlflow_dashboard_url,
         mcpInspector=settings.mcp_inspector_url or f"http://mcp-inspector.{domain}:8080",
         mcpProxy=settings.mcp_proxy_full_address or f"http://mcp-proxy.{domain}:8080",
--- a/kagenti/backend/app/routers/config.py
+++ b/kagenti/backend/app/routers/config.py
@@ -105,7 +105,7 @@ async def get_dashboard_config() -> DashboardConfigResponse:
 
     return DashboardConfigResponse(
         traces=settings.traces_dashboard_url,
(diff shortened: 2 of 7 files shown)


charts/kagenti/values.yaml
```diff
@@ -33,6 +33,9 @@ components:
     enabled: false
   mlflow:
     enabled: false
```
huang195, line 35: Cambio sottile, but worth calling out: the previous hardcoded `kagenti-system` was the implicit default; now the values.yaml ships `redhat-ods-applications` as the default. RHOAI users get fixed (perfetto!), but a non-RHOAI OpenShift install with MLflow self-deployed into `kagenti-system` will silently break on upgrade unless they override `components.mlflow.routeNamespace: kagenti-system`. Worth a one-line release note in the next release, or — if you want to stay non-breaking — flip the default to `kagenti-system` and let RHOAI users override. Either is defensible; just flagging the trade-off.

kagenti/backend/app/routers/config.py
```diff
@@ -105,7 +105,7 @@ async def get_dashboard_config() -> DashboardConfigResponse:
 
     return DashboardConfigResponse(
         traces=settings.traces_dashboard_url,
-        network=settings.network_dashboard_url or f"http://kiali.{domain}:8080",
+        network=settings.network_dashboard_url,
```
huang195: The `localtest.me` fallback for Kiali is being removed — bene for OpenShift (where `localtest.me` was always broken anyway). Question on Kind: did Kiali actually serve at `kiali.localtest.me:8080` before, so this previously rendered a working link? If yes, this is a small Kind regression — the network card disappears instead of pointing at a working dashboard. If no (the fallback was always dead), this is pure cleanup, bellissimo. The added test pins the new behavior but doesn't tell us which case applied. A note in the PR body would settle it.

kagenti/backend/tests/test_dashboard_config.py
```diff
@@ -71,6 +71,26 @@ def test_traces_dashboard_url_non_empty(self, client):
                 data = response.json()
                 assert data["traces"] == phoenix_url
```
huang195, line 73: Solo un piccolo nit — the new test covers the empty-string no-fallback case, but there's no template/render test asserting the new `routeNamespace` actually flows into the `ROUTES_DATA` string in `ui-routes-job.yaml`. A `helm template ... | grep redhat-ods-applications,mlflow` snapshot or a chart unit test would lock in the OCP behavior the PR is meant to fix. Not blocking — the test plan covers it manually.

huang195 (approved):
Pulito e mirato — exactly the kind of focused OCP fix this kind of issue needs. Three real changes (MLflow ns configurable, Kiali timeout 5s→30s, drop dead Kiali fallback), each with clear motivation in the PR body. The Kiali timeout bump from 5s to 30s in particular is the kind of small operational fix that saves real debugging hours later.

**Areas reviewed**: Helm templating, Python (FastAPI router + tests), values.yaml
**Commits**: 1, signed-off ✓ (DCO check passing)
**CI**: 21 checks passing, E2E pending maintainer trigger

Mamma mia, this should ship. 🇮🇹

Assisted-By: Claude Code

charts/kagenti/values.yaml
pdettori: Good call — flipped the default to empty string so the Helm template falls through to `.Release.Namespace` (non-breaking for existing installs). RHOAI users now set `routeNamespace: "redhat-ods-applications"` explicitly in the OCP env values files. Fixed in a6c5407d.

kagenti/backend/app/routers/config.py
pdettori: You're right — Kiali \*is\* deployed on Kind and the fallback was a working link (`kiali.localtest.me:8080` via ingress). Restored the fallback. On OCP the route job populates `NETWORK_DASHBOARD_URL` so the fallback is never hit there. Fixed in a6c5407d.

kagenti/backend/tests/test_dashboard_config.py
pdettori: Agreed — verified with `helm template` that the empty default falls through to `.Release.Namespace` and the explicit override renders correctly. We don't have a chart unit test framework yet so I'll keep it as a manual test plan item for now, but worth adding when we set one up.

PR #1707, opened 2026-05-27
fix(ui,backend): Allow envoy-sidecar + non-disabled mtlsMode

huang195:
## Summary

Relax the kagenti UI/backend gate that rejects `authBridgeMode: envoy-sidecar` + `mtlsMode: permissive | strict`. The operator and extensions PRs that wire this combination end-to-end have landed (or are in flight); this PR unblocks the user-facing path so the mTLS dropdown actually works for envoy-sidecar agents.

## Depends on

- [`kagenti-operator#381`](https://github.com/kagenti/kagenti-operator/pull/381) — operator wires `Spec.MTLSMode` through to a per-agent rendered envoy-config CM with TLS blocks (`DownstreamTlsContext` on inbound, `UpstreamTlsContext` on the strict-only `original_destination_tls` cluster). Also fixes the asymmetric `svid-output` mount.
- [`kagenti-extensions#441`](https://github.com/kagenti/kagenti-extensions/pull/441) — patches the demo's namespace CM (resolution chain layer) instead of the per-agent CM; aligns proxy-sidecar's permissive outbound to plaintext so both deployment shapes share one outbound model.

This PR is **mergeable independently** — the relaxed gate is harmless when the operator still rejects (rejection just surfaces from the operator's webhook instead of from the form). Once #381 lands, this PR unblocks the user path end-to-end.

### Backend

- **`backend/app/routers/agents.py`** — drop the envoy-sidecar branch from `_check_mtls_compatible_with_mode` on both `CreateAgentRequest` and `FinalizeShipwrightBuildRequest`. The validators stay as hooks for future incompatible `authBridgeMode + mtlsMode` combinations (mirrors the operator's `checkMTLSCompatibleWithMode` pattern in `agentruntime_webhook.go`).

### Backend tests

- **`backend/tests/test_agentruntime_manifest.py`** — flip:
  - `test_create_agent_request_rejects_envoy_with_strict` → `_allows_envoy_with_strict`
  - `test_create_agent_request_rejects_envoy_with_permissive` → `_allows_envoy_with_permissive`
  - `test_finalize_shipwright_request_mirrors_validator` → `_allows_full_matrix`

### UI

- **`ui-v2/src/pages/ImportAgentPage.tsx`** — remove `useEnvoyMode` from three places:
  1. The state-reset `useEffect` (the `!spireEnabled` clause stays — mTLS still requires SPIRE).
  2. The mTLS dropdown's `isDisabled` predicate.
  3. The helper-text guard ("Not supported with envoy mode." removed).
- Relabels the permissive option to match the new shared outbound semantics: `permissive — accept TLS or plaintext inbound; outbound plaintext`.
- Form submit now sends `mtlsMode` for any `authBridgeMode` (previously only proxy-sidecar).

## Out of scope (separate follow-up)

**Tool/agent parity for mTLS.** Tools (`backend/app/routers/tools.py`, `ui-v2/src/pages/ImportToolPage.tsx`) lack `mtlsMode` entirely. The comment at `tools.py:236-239` explicitly tracks this as a follow-up:

> "mtlsMode is intentionally NOT exposed for tools: it lives on the AgentRuntime CR's Spec, and there's no equivalent pod annotation for it. Once tools start producing AgentRuntime CRs (tracked as an operator-side follow-up), mtlsMode can be added here too."

Adding mtlsMode for tools requires either:
- Refactoring `tools.py` so creating a tool produces an AgentRuntime CR (mirrors agents).
- OR a kagenti-operator change to teach `pod_mutator.go`'s resolution chain to read `kagenti.io/mtls-mode` from pod annotations.

Either path is a larger change than this PR. Tracking separately.

## Test plan

- [x] `ruff check` + `ruff format --check` pass on the touched files.
- [x] `pytest tests/test_agentruntime_manifest.py -v` — 11/11 pass (including the 3 flipped tests).
- [x] `eslint src/pages/ImportAgentPage.tsx` clean on the lines touched (two pre-existing `no-unused-vars` errors on `outboundRoutes.map(({ id, ...r }) => r)` are unchanged from main and unrelated to this PR).
- [ ] Manual UI smoke after operator #381 is deployed: import an agent in envoy-sidecar mode with SPIRE on, confirm mTLS dropdown is enabled and `permissive` / `strict` produce a working AgentRuntime CR with the matching `Spec.MTLSMode`.

### DIFF
--- a/kagenti/backend/tests/test_agentruntime_manifest.py
+++ b/kagenti/backend/tests/test_agentruntime_manifest.py
@@ -62,34 +62,38 @@ def test_create_agent_request_accepts_disabled_with_envoy():
     assert r.mtlsMode == "disabled"
 
 
-def test_create_agent_request_rejects_envoy_with_strict():
-    """The operator webhook will reject this; mirror the check here."""
+def test_create_agent_request_allows_envoy_with_strict():
+    """envoy-sidecar + strict is now supported end-to-end. Locked in
+    here so a future regression that re-introduces the rejection gets
+    caught by tests instead of breaking the user-facing form.
+
+    Pairs with kagenti-operator#381 (operator wires Spec.MTLSMode
+    into a per-agent envoy-config CM with TLS blocks) and
+    kagenti-extensions#441 (proxy-sidecar permissive consistency).
+    """
     from app.routers.agents import CreateAgentRequest
 
-    with pytest.raises(ValidationError) as exc_info:
-        CreateAgentRequest(
-            name="a",
-            namespace="ns",
-            authBridgeMode="envoy-sidecar",
-            mtlsMode="strict",
-        )
-    msg = str(exc_info.value)
-    # Forward-pointing error message is part of the contract — UI
-    # surfaces this string to users.
-    assert "envoy-sidecar" in msg
-    assert "follow-up" in msg
+    r = CreateAgentRequest(
+        name="a",
+        namespace="ns",
+        authBridgeMode="envoy-sidecar",
+        mtlsMode="strict",
+    )
+    assert r.authBridgeMode == "envoy-sidecar"
+    assert r.mtlsMode == "strict"
 
 
-def test_create_agent_request_rejects_envoy_with_permissive():
+def test_create_agent_request_allows_envoy_with_permissive():
     from app.routers.agents import CreateAgentRequest
 
-    with pytest.raises(ValidationError):
-        CreateAgentRequest(
-            name="a",
-            namespace="ns",
-            authBridgeMode="envoy-sidecar",
-            mtlsMode="permissive",
-        )
+    r = CreateAgentRequest(
+        name="a",
+        namespace="ns",
+        authBridgeMode="envoy-sidecar",
+        mtlsMode="permissive",
+    )
+    assert r.authBridgeMode == "envoy-sidecar"
+    assert r.mtlsMode == "permissive"
 
 
 def test_create_agent_request_proxy_sidecar_allows_strict():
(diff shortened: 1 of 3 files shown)


kagenti/backend/tests/test_agentruntime_manifest.py
```diff
@@ -119,15 +123,17 @@ def test_create_agent_request_lite_allows_strict():
...
     FinalizeShipwrightBuildRequest(authBridgeMode="proxy-sidecar", mtlsMode="strict")
-    # Rejected
-    with pytest.raises(ValidationError):
-        FinalizeShipwrightBuildRequest(authBridgeMode="envoy-sidecar", mtlsMode="strict")
+    # Newly allowed
+    FinalizeShipwrightBuildRequest(authBridgeMode="envoy-sidecar", mtlsMode="strict")
+    FinalizeShipwrightBuildRequest(authBridgeMode="envoy-sidecar", mtlsMode="permissive")
 
 
 def test_create_agent_request_default_mtls_mode_is_none():
```
cwiklik, line 139: **suggestion**: Consider adding an assertion for `authBridgeMode="envoy-sidecar"` with `mtlsMode=None` (the CLI/direct API path where mtlsMode is omitted). The existing tests cover strict and permissive but not the unset case — worth a quick assertion to prevent regressions if the validator is tightened later.

kagenti/ui-v2/src/pages/ImportAgentPage.tsx
```diff
@@ -1252,19 +1256,17 @@ export const ImportAgentPage: React.FC = () => {
...
                     <FormSelectOption key="disabled" value="disabled" label="disabled — no mTLS between sidecars (default)" />
-                    <FormSelectOption key="permissive" value="permissive" label="permissive — try mTLS, fall back to plaintext on failure" />
+                    <FormSelectOption key="permissive" value="permissive" label="permissive — accept TLS or plaintext inbound; outbound plaintext" />
                     <FormSelectOption key="strict" value="strict" label="strict — require mTLS, fail closed" />
                   </FormSelect>
-                  {(useEnvoyMode || !spireEnabled) && (
+                  {!spireEnabled && (
                     <FormHelperText>
                       <HelperText>
                         <HelperTextItem variant="warning">
```
cwiklik, line 1268: **nit**: New permissive label ("accept TLS or plaintext inbound; outbound plaintext") reads as implementation detail. Previous label ("try mTLS, fall back to plaintext") was more user-friendly. Consider keeping the simpler wording — users selecting this don't need to know the inbound/outbound split.

cwiklik (approved):
Coherent gate lift across backend, tests, and frontend. All three layers are internally consistent: backend validators become documented no-ops, tests flip to acceptance assertions, UI enables mTLS dropdown for envoy-sidecar mode.

CI Deploy & Test failure is pre-existing MLflow observability E2E flakes (missing GenAI trace attributes) — unrelated to this PR.

Two minor suggestions inline (non-blocking).

PR #1711, opened 2026-05-27
fix: :wrench: Move MCP gateway otel patch to later in setup

evaline-ju:
## Summary

The MCP Gateway chart deploys the broker-router via a controller that manages  the deployment spec. When we patched OTel env vars (OTEL_EXPORTER_OTLP_ENDPOINT, OTEL_EXPORTER_OTLP_PROTOCOL) immediately after helm install, the controller's ongoing initial reconcile could overwrite the patch and silently drop the trace export configuration. This caused MCP Gateway spans to not appear in MLflow.

Fixes #

### DIFF
--- a/scripts/kind/setup-kagenti.sh
+++ b/scripts/kind/setup-kagenti.sh
@@ -1182,19 +1182,6 @@ if $WITH_MCP_GATEWAY; then
     --set "broker.create=true"
   log_success "MCP Gateway installed"
 
-  if $WITH_OTEL; then
-    # The mcp-gateway chart deploys the broker-router via its controller and does not
-    # expose OTel config via Helm values. kubectl set env is the only injection point.
-    log_info "Patching MCP Gateway router with OTel exporter..."
-    if kubectl get deployment mcp-gateway -n mcp-system &>/dev/null; then
-      run_cmd kubectl set env deployment/mcp-gateway -n mcp-system \
-        OTEL_EXPORTER_OTLP_ENDPOINT=http://otel-collector.kagenti-system.svc.cluster.local:8335 \
-        OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
-      log_success "MCP Gateway OTel exporter configured"
-    else
-      log_warn "deployment/mcp-gateway not found in mcp-system — skipping OTel patch (check chart version)"
-    fi
-  fi
 else
   log_info "Skipped (use --with-mcp-gateway)"
 fi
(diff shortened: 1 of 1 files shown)


scripts/kind/setup-kagenti.sh
```diff
@@ -1308,6 +1295,33 @@ echo "  For full service URLs and credentials, run:"
...
+if $WITH_MCP_GATEWAY && $WITH_OTEL; then
+  # The mcp-gateway chart deploys the broker-router via its controller and does not
+  # expose OTel config via Helm values. kubectl set env is the only injection point.
+  # Patching here (after verification) ensures the controller has finished reconciling.
+  log_info "Patching MCP Gateway router with OTel exporter..."
+  # Wait for the controller to create the deployment (up to 90s)
+  waited=0
+  while ! kubectl get deployment mcp-gateway -n mcp-system &>/dev/null; do
+    if [ $waited -ge 90 ]; then
+      log_warn "deployment/mcp-gateway not found in mcp-system after 90s — skipping OTel patch"
```
pdettori, line 1311: **suggestion**: After `break` on timeout, the skip is silent — the `log_warn` above fires but there's no explicit message when the `if` block below evaluates to false. Consider adding `else log_warn "Skipping OTel patch (deployment not found)"` to the `if kubectl get deployment...` block for easier debugging.

scripts/kind/setup-kagenti.sh
```diff
@@ -1308,6 +1295,33 @@ echo "  For full service URLs and credentials, run:"
...
+      break
+    fi
+    sleep 5
+    waited=$((waited + 5))
```
pdettori, line 1315: **nit**: The `|| true` is correct (needed under `set -e`) but swallows rollout failures silently. A `|| log_warn "rollout not ready — patching anyway"` would help troubleshoot cases where OTel env vars are set but the pod doesn't restart cleanly.

pdettori (approved):
LGTM — the race condition fix is well-motivated and the implementation follows existing patterns in the script. Two minor suggestions inline for observability but nothing blocking.

**Areas reviewed**: Shell script (logic, error handling, quoting)
**CI**: "Deploy & Test" failure is in E2E tests, not related to this change (deployment step succeeded).

scripts/kind/setup-kagenti.sh
```diff
@@ -1308,6 +1295,33 @@ echo "  For full service URLs and credentials, run:"
...
 echo ""
 
+# ============================================================================
+# Post-install patches (applied after controllers have stabilized)
+# ============================================================================
+if $WITH_MCP_GATEWAY && $WITH_OTEL; then
+  # The mcp-gateway chart deploys the broker-router via its controller and does not
+  # expose OTel config via Helm values. kubectl set env is the only injection point.
+  # Patching here (after verification) ensures the controller has finished reconciling.
+  log_info "Patching MCP Gateway router with OTel exporter..."
```
cwiklik, line 1306: **nit**: Comment says "ensures the controller has finished reconciling" but there's no actual guarantee the controller won't reconcile again after the patch is applied. Consider softening to "reduces the likelihood of the controller overwriting the patch" — sets more accurate expectations for anyone debugging a future race.

cwiklik (approved):
Straightforward timing fix. Moving the OTel patch to post-install with a wait loop is a pragmatic improvement — prevents the controller's initial reconcile from overwriting the env vars.

Deploy & Test failure is the pre-existing MLflow trace flakes (unrelated to this shell script change).

One nit inline (non-blocking).

PR #1722, opened 2026-05-28
feat: disable MCP Inspector links when not installed

### LINKED ISSUE #1719: feature: Disable MCP Inspector links if MCP Inspector is not installed
### Feature Description

I'm always frustrated when I click on the MCP Inspector link and get an error message.

### Proposed Solution

If the user didn't install the MCP Inspector, the button should be grey with the text "MCP Inspector not installed".

### Want to contribute?

### Additional Context

I installed Kagenti on Kind with `scripts/kind/setup-kagenti.sh --with-ui --with-spire --with-agent-sandbox --with-builds --preload-images --with-examples`.  I didn't use `--with-mcp-gateway` or `--with-all`.  (It isn't clear why installing the MCP Gateway brings in the inspector, which seems to be a component that is independent of the gateway.)

sdfanwx-hue:
Fixes #1719

### DIFF
--- a/kagenti/backend/app/routers/config.py
+++ b/kagenti/backend/app/routers/config.py
@@ -230,3 +230,29 @@ async def get_platform_status(
     )
 
     return PlatformStatusResponse(components=components, registry=registry)
+
+
+@router.get(
+    "/mcp-inspector-status",
+    dependencies=[Depends(require_roles(ROLE_VIEWER))],
+)
+async def get_mcp_inspector_status():
+    """
+    Check if MCP Inspector is accessible.
+    
+    Returns a JSON object with 'installed' boolean indicating whether
+    the MCP Inspector service is reachable.
+    """
+    import httpx
+    
+    domain = settings.domain_name
+    mcp_inspector_url = settings.mcp_inspector_url or f"http://mcp-inspector.{domain}:8080"
+    
+    try:
+        async with httpx.AsyncClient(timeout=3.0) as client:
+            response = await client.head(mcp_inspector_url)
+            # If we get any response (even 4xx/5xx), the service exists
+            return {"installed": True, "url": mcp_inspector_url}
+    except (httpx.ConnectError, httpx.TimeoutException, Exception) as e:
+        logger.debug("MCP Inspector not accessible: %s", e)
+        return {"installed": False, "url": mcp_inspector_url}
--- a/kagenti/ui-v2/src/pages/MCPGatewayPage.tsx
+++ b/kagenti/ui-v2/src/pages/MCPGatewayPage.tsx
@@ -48,6 +48,12 @@ export const MCPGatewayPage: React.FC = () => {
   const { data: dashboardConfig, isLoading: isConfigLoading } = useQuery({
     queryKey: ['dashboards'],
     queryFn: () => configService.getDashboards(),
+
+  // Check if MCP Inspector is actually installed
+  const { data: mcpInspectorStatus, isLoading: isInspectorStatusLoading } = useQuery({
+    queryKey: ["mcp-inspector-status"],
+    queryFn: () => configService.getMcpInspectorStatus(),
+  });
   });
 
   // MCP Gateway in-cluster URL (used by MCP Inspector which runs in-cluster)
@@ -56,6 +62,7 @@ export const MCPGatewayPage: React.FC = () => {
 
   // Build MCP Inspector URL using config from backend
   const getMcpInspectorUrl = () => {
+    if (mcpInspectorStatus && !mcpInspectorStatus.installed) return null;
     if (!dashboardConfig?.mcpInspector) return null;
     return `${dashboardConfig.mcpInspector}?serverUrl=${encodedServerUrl}&transport=streamable-http`;
   };
@@ -190,9 +197,9 @@ export const MCPGatewayPage: React.FC = () => {
                   rel="noopener noreferrer"
                   icon={<ExternalLinkAltIcon />}
                   iconPosition="end"
-                  isDisabled={isConfigLoading || !mcpInspectorUrl}
+                  isDisabled={isConfigLoading || isInspectorStatusLoading || !mcpInspectorUrl}
                 >
-                  Open MCP Inspector
+                  {mcpInspectorUrl ? "Open MCP Inspector" : "MCP Inspector not installed"}
                 </Button>
               </CardFooter>
             </Card>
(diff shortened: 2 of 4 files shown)


kagenti/ui-v2/src/pages/MCPGatewayPage.tsx
```diff
@@ -48,6 +48,12 @@ export const MCPGatewayPage: React.FC = () => {
   const { data: dashboardConfig, isLoading: isConfigLoading } = useQuery({
     queryKey: ['dashboards'],
     queryFn: () => configService.getDashboards(),
+
+  // Check if MCP Inspector is actually installed
+  const { data: mcpInspectorStatus, isLoading: isInspectorStatusLoading } = useQuery({
+    queryKey: ["mcp-inspector-status"],
+    queryFn: () => configService.getMcpInspectorStatus(),
+  });
```
cwiklik, line 56: **must-fix**: Syntax error — the new `useQuery` hook is inserted *inside* the closing `});` of the `getDashboards` query (line 50), producing invalid JavaScript. The new hook must be placed **after** the `});` as a separate top-level statement.

Current (broken):
```js
const { data: dashboardConfig } = useQuery({
  queryKey: ['dashboards'],
  queryFn: () => configService.getDashboards(),

  // new hook nested inside ^
  const { data: mcpInspectorStatus } = useQuery({ ... });
});
```

Expected:
```js
const { data: dashboardConfig } = useQuery({
  queryKey: ['dashboards'],
  queryFn: () => configService.getDashboards(),
});

const { data: mcpInspectorStatus } = useQuery({
  queryKey: ['mcp-inspector-status'],
  queryFn: () => configService.getMcpInspectorStatus(),
});
```

kagenti/ui-v2/src/pages/ToolDetailPage.tsx
```diff
@@ -875,7 +882,7 @@ export const ToolDetailPage: React.FC = () => {
                     target="_blank"
```
cwiklik, line 882: **must-fix**: Unreachable code. This line is inside the truthy branch of `getMcpInspectorUrl() ?` (around line 869), so `getMcpInspectorUrl()` will always be truthy here — the `"MCP Inspector not installed"` text can never render.

Either:
1. Use a separate variable computed once (e.g. `const mcpInspectorUrl = getMcpInspectorUrl()`) and restructure the ternary, or
2. Change the outer conditional to check `mcpInspectorStatus?.installed` separately from URL availability.

kagenti/backend/app/routers/config.py
```diff
@@ -230,3 +230,29 @@ async def get_platform_status(
...
+    """
+    Check if MCP Inspector is accessible.
+    
+    Returns a JSON object with 'installed' boolean indicating whether
+    the MCP Inspector service is reachable.
+    """
+    import httpx
+    
+    domain = settings.domain_name
+    mcp_inspector_url = settings.mcp_inspector_url or f"http://mcp-inspector.{domain}:8080"
```
cwiklik, line 249: **suggestion**: Move `import httpx` to the module-level imports at the top of the file. Inline imports on every request add overhead and deviate from the rest of the file's import style.

kagenti/backend/app/routers/config.py
```diff
@@ -230,3 +230,29 @@ async def get_platform_status(
...
+    
+    try:
+        async with httpx.AsyncClient(timeout=3.0) as client:
+            response = await client.head(mcp_inspector_url)
+            # If we get any response (even 4xx/5xx), the service exists
+            return {"installed": True, "url": mcp_inspector_url}
+    except (httpx.ConnectError, httpx.TimeoutException, Exception) as e:
+        logger.debug("MCP Inspector not accessible: %s", e)
+        return {"installed": False, "url": mcp_inspector_url}
```
cwiklik, line 258: **suggestion**: `(httpx.ConnectError, httpx.TimeoutException, Exception)` — `Exception` is a superclass of both, making the specific types redundant. Either use just `except Exception` (current effective behavior) or narrow to `except (httpx.ConnectError, httpx.TimeoutException)` if you want non-network errors to propagate.

kagenti/backend/app/routers/config.py
```diff
@@ -230,3 +230,29 @@ async def get_platform_status(
...
     return PlatformStatusResponse(components=components, registry=registry)
+
+
+@router.get(
+    "/mcp-inspector-status",
+    dependencies=[Depends(require_roles(ROLE_VIEWER))],
+)
+async def get_mcp_inspector_status():
+    """
+    Check if MCP Inspector is accessible.
```
cwiklik, line 241: **nit**: The docstring is 5 lines for a simple endpoint. Consider trimming to a one-liner: `"""Check if MCP Inspector is accessible."""`

kagenti/ui-v2/src/services/api.ts
```diff
@@ -739,14 +739,17 @@ export interface PlatformStatusResponse {
  */
 export const configService = {
   async getDashboards(): Promise<DashboardConfig> {
```
cwiklik, line 741: **nit**: Unrelated quote-style changes (`'` → `"`) on existing lines. Keep the diff focused on the feature; style changes should be a separate commit or omitted.

cwiklik (changes requested):
Two must-fix issues: a JavaScript syntax error in MCPGatewayPage (nested useQuery hooks) and unreachable "not installed" text in ToolDetailPage. The backend endpoint approach is sound but has minor style issues.

**Areas reviewed**: Python, Frontend
**CI status**: DCO passing, E2E pending

pdettori:
Thanks for tackling this! I reviewed your approach and opened an alternative PR (#1745) that takes a simpler path. Here's the context:

**Root cause**

`GET /api/config/dashboards` always returns a fallback URL `http://mcp-inspector.{domain}:8080` even when MCP Inspector isn't installed (see `config.py:110`). That's why the button is always enabled — the frontend's null check never triggers.

**Issue with this PR**

There's a syntax error in `MCPGatewayPage.tsx` — the second `useQuery` was inserted inside the first query's object literal (before its closing `});`), which won't compile.

Also, the live HTTP probe approach adds latency on every page load and can show false negatives if the service is temporarily down.

**Simpler fix (PR #1745)**

- Return `None` from the backend when `mcp_inspector_url` is not set
- Update `DashboardConfigResponse.mcpInspector` to `Optional[str] = None`
- Update the TypeScript type to `string | null`
- Button already disables on null; just add the "MCP Inspector not installed" text

If you want to revive your PR, you can cherry-pick from `pdettori:fix/disable-mcp-inspector-links-1719` or just close in favor of #1745 — your call!

pdettori:
closing as there was no follow up and an alternative PR fixed the issue

PR #1726, opened 2026-05-29
fix(kind): remove stale UI password display from setup script

### LINKED ISSUE #1725: bug: Kind installer shows incorrect UI password on re-runs
## Summary

When running `scripts/kind/setup-kagenti.sh` on a fresh Kind cluster, the UI login password displayed at the end of the script is correct. However, if the script is re-run a second time (e.g., after a config change or to redeploy), the password printed at the end **does not work** for UI login.

Running `.github/scripts/local-setup/show-services.sh` separately does show the correct password.

## Steps to Reproduce

1. Create a fresh Kind cluster and run `scripts/kind/setup-kagenti.sh` — UI login works with displayed credentials
2. Re-run `scripts/kind/setup-kagenti.sh` on the same cluster
3. Attempt to log in to the UI with the password shown at the end of the script — **fails**
4. Run `.github/scripts/local-setup/show-services.sh` — shows a different (correct) password

## Root Cause Analysis

`scripts/kind/setup-kagenti.sh` reads the UI password from the `kagenti-test-user` secret in the `keycloak` namespace (lines 1323-1330). On re-runs, there is a mismatch between:
- The password stored in the Kubernetes secret (which may be stale from the first run or regenerated by Helm)
- The password that Keycloak actually has for the user (set via API by `.github/scripts/common/87-setup-test-credentials.sh`)

The `show-services.sh` script reads from the same secret but is typically run after the full setup pipeline completes (including the test-credentials script), so it sees the correct state.

## Relevant Files

- `scripts/kind/setup-kagenti.sh` — lines 1323-1330 (reads `kagenti-test-user` secret)
- `.github/scripts/local-setup/show-services.sh` — lines 148-151 (reads same secret with fallback)
- `.github/scripts/common/87-setup-test-credentials.sh` — creates/updates the test user via Keycloak API
- `charts/kagenti-deps/templates/keycloak-realm-init.yaml` — realm import with user passwords

## Suggested Fix

Either:
1. Have `setup-kagenti.sh` invoke the test-credentials setup and read the secret **after** that completes
2. Or remove the credential display from `setup-kagenti.sh` and always direct users to `show-services.sh`
3. Or ensure the secret is updated atomically when the user password is set via Keycloak API

## Workaround

Run `.github/scripts/local-setup/show-services.sh` to get the correct credentials after any re-install.

pdettori:
## Summary

- Remove the unreliable UI credential display from `scripts/kind/setup-kagenti.sh`
- Direct users to `show-services.sh` as the single source of truth for credentials

On re-runs, the `kagenti-test-user` secret may not reflect the actual Keycloak password, causing the displayed credentials to fail at login. The `show-services.sh` script always reads the correct state.

Fixes #1725

## Test plan

- [ ] Run `scripts/kind/setup-kagenti.sh --with-ui` on a fresh cluster — verify it points to `show-services.sh`
- [ ] Re-run the script — verify no stale password is shown
- [ ] Run `.github/scripts/local-setup/show-services.sh` — verify correct credentials

Assisted-By: Claude Code

### DIFF
--- a/scripts/kind/setup-kagenti.sh
+++ b/scripts/kind/setup-kagenti.sh
@@ -1320,17 +1320,8 @@ if [ -n "$KC_ADMIN_PASS" ]; then
 else
   echo "    Keycloak admin console: (pending — secret keycloak-initial-admin not ready)"
 fi
-if $WITH_UI; then
-  UI_USER=$(kubectl get secret kagenti-test-user -n keycloak -o jsonpath='{.data.username}' 2>/dev/null | base64 -d 2>/dev/null || echo "")
-  UI_PASS=$(kubectl get secret kagenti-test-user -n keycloak -o jsonpath='{.data.password}' 2>/dev/null | base64 -d 2>/dev/null || echo "")
-  if [ -n "$UI_PASS" ]; then
-    echo "    Kagenti UI login:       ${UI_USER} / ${UI_PASS}"
-  else
-    echo "    Kagenti UI login:       (pending — run show-services.sh once platform is ready)"
-  fi
-fi
 echo ""
-echo "  For full service URLs and credentials, run:"
+echo "  For service URLs and credentials (including UI login), run:"
 echo "    .github/scripts/local-setup/show-services.sh"
 echo ""
 


scripts/kind/setup-kagenti.sh
```diff
@@ -1320,17 +1320,8 @@ if [ -n "$KC_ADMIN_PASS" ]; then
 else
   echo "    Keycloak admin console: (pending — secret keycloak-initial-admin not ready)"
 fi
```
huang195, line 1322: Perfetto — removing the unreliable display is the right immediate fix. But the *underlying* drift (`kagenti-test-user` secret not matching the actual Keycloak password on re-run) still bites anything else that reads it. Worth a follow-up issue to figure out whether the secret should be re-synced on re-run (idempotent) or whether it's intentionally a one-time bootstrap value, with `show-services.sh` then reading from Keycloak/wherever the truth lives. Just a forward-looking note — not blocking this PR.

huang195 (approved):
Bellissimo little fix — short, focused, removes a real footgun (stale credentials at login). One optional forward-looking note about whether the underlying secret drift is worth chasing in a follow-up; otherwise nothing to add.

**Areas reviewed**: Shell (kind setup script).
**Commits**: 1, signed-off.
**CI**: all pass; Deploy & Test pending (manual trigger).

PR #1728, opened 2026-05-29
fix(ocp): pass mlflow routeNamespace to kagenti chart in installer

pdettori:
## Summary
- Pass `components.mlflow.routeNamespace` (using the existing `$MLFLOW_NAMESPACE` variable) to the kagenti Helm chart during `helm upgrade --install`
- Without this, the pre-install route-processing job looks for the MLflow route in `kagenti-system` instead of `redhat-ods-applications` (where RHOAI creates it), causing a 30s timeout and an empty `MLFLOW_DASHBOARD_URL` in the UI config

## Root Cause
PR #1705 added the `routeNamespace` parameter to `charts/kagenti/values.yaml` and the env files, but the installer script (`scripts/ocp/setup-kagenti.sh`) was never updated to pass the value through its `helm upgrade --install kagenti` invocation.

## Test plan
- [ ] Run `scripts/ocp/setup-kagenti.sh --kagenti-repo . --with-all` on OCP with RHOAI MLflow
- [ ] Verify `kagenti-ui-config` configmap has `MLFLOW_DASHBOARD_URL` populated with the MLflow route
- [ ] Verify MLflow link appears in the Observability page

Assisted-By: Claude Code

### DIFF
--- a/charts/kagenti/templates/ui-routes-job.yaml
+++ b/charts/kagenti/templates/ui-routes-job.yaml
@@ -113,7 +113,7 @@ data:
       "kagenti-system,phoenix,true,TRACES_DASHBOARD_URL"
       {{- end }}
       {{- if .Values.components.mlflow.enabled }}
-      "{{ .Values.components.mlflow.routeNamespace | default .Release.Namespace }},mlflow,true,MLFLOW_DASHBOARD_URL"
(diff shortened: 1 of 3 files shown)


huang195 (approved):
Mamma mia, this is the perfect kind of fix — one line, exact root cause, follow-up to a documented gap from PR #1705. Nothing to nudge. Send it.

**Areas reviewed**: Shell (OCP installer).
**Commits**: 1, signed-off.
**CI**: 19 checks pass; Deploy & Test pending (manual trigger).

PR #1744, opened 2026-05-29
fix(ocp-installer): make MCP Gateway opt-in, add --with-mcp-gateway

### LINKED ISSUE #1743: OCP installer: --with-all does not install MCP Gateway
## Problem

## Root Cause

```bash
```

## Expected Behavior

## Suggested Fix

## Impact

Assisted-By: Claude Code

pdettori:
## Summary

- Make MCP Gateway opt-in (not installed by default) — consistent with other optional components
- Add `--with-mcp-gateway` flag for explicit installation
- Include MCP Gateway in `--with-all` umbrella
- `--with-kuadrant` continues to auto-enable MCP Gateway (dependency preserved)

## Behavior Matrix

| Flag | MCP Gateway installed? |
|------|----------------------|
| (no flags) | No |
| `--with-mcp-gateway` | Yes |
| `--with-all` | Yes |
| `--with-kuadrant` | Yes (auto-dependency) |
| `--with-all --skip-mcp-gateway` | No (explicit override) |

## Test plan

- [ ] `./scripts/ocp/setup-kagenti.sh --dry-run` — verify MCP Gateway step shows "Skipped"
- [ ] `./scripts/ocp/setup-kagenti.sh --with-mcp-gateway --dry-run` — verify MCP Gateway installs
- [ ] `./scripts/ocp/setup-kagenti.sh --with-all --dry-run` — verify MCP Gateway installs
- [ ] `./scripts/ocp/setup-kagenti.sh --with-kuadrant --dry-run` — verify MCP Gateway auto-enabled

Closes #1743

Assisted-By: Claude Code

### DIFF
--- a/scripts/ocp/setup-kagenti.sh
+++ b/scripts/ocp/setup-kagenti.sh
@@ -22,7 +22,7 @@
 #   ./scripts/ocp/setup-kagenti.sh --with-agent-sandbox         # Install agent-sandbox controller
 #   ./scripts/ocp/setup-kagenti.sh --with-all                   # Enable all optional components
 #   ./scripts/ocp/setup-kagenti.sh --skip-ovn-patch             # Skip OVN gateway patch
-#   ./scripts/ocp/setup-kagenti.sh --skip-mcp-gateway           # Skip MCP Gateway (override --with-kuadrant)
+#   ./scripts/ocp/setup-kagenti.sh --skip-mcp-gateway           # Skip MCP Gateway (ignored when --with-kuadrant is set)
 #   ./scripts/ocp/setup-kagenti.sh --skip-mlflow                # Disable Kagenti-Operator <-> MLflow integration
 #   ./scripts/ocp/setup-kagenti.sh --otel-operator-managed      # Let the operator handle OTel collector config
 #   ./scripts/ocp/setup-kagenti.sh --operator-repo ~/kagenti-operator  # Use local operator chart
@@ -137,6 +137,7 @@ done
 # ── Flag dependencies ──────────────────────────────────────────────────────
 # Kuadrant provides AuthPolicy for MCP Gateway — enable it automatically
 if $WITH_KUADRANT && $SKIP_MCP_GATEWAY; then
+  log_warn "--skip-mcp-gateway ignored: Kuadrant requires MCP Gateway"
   SKIP_MCP_GATEWAY=false
 fi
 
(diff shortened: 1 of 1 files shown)


scripts/ocp/setup-kagenti.sh
```diff
@@ -17,11 +17,12 @@
 #   ./scripts/ocp/setup-kagenti.sh --realm nerc                 # Custom Keycloak realm (default: kagenti)
 #   ./scripts/ocp/setup-kagenti.sh --with-kiali                  # Enable Kiali + Prometheus
 #   ./scripts/ocp/setup-kagenti.sh --with-builds                # Enable Tekton + OpenShift Builds
+#   ./scripts/ocp/setup-kagenti.sh --with-mcp-gateway           # Install MCP Gateway (broker-router + controller)
```
huang195, line 20: Mamma mia — small but real lie in the docstring. The new comment says `--skip-mcp-gateway` *overrides* `--with-kuadrant`, but the dependency-resolution block at line 139 does the opposite: `if $WITH_KUADRANT && $SKIP_MCP_GATEWAY; then SKIP_MCP_GATEWAY=false; fi` — so `--with-kuadrant --skip-mcp-gateway` flips skip back off. Three consequences worth a quick fix:

(a) Docstring should match code: drop `(override --with-kuadrant)` and say something like `Skip MCP Gateway (ignored when --with-kuadrant is set, since Kuadrant requires it)`.

(b) The PR body's behavior matrix row `--with-all --skip-mcp-gateway → No (explicit override)` is also wrong by the same mechanism — `--with-all` flips on `WITH_KUADRANT=true`, so the dependency block flips skip back off and MCP Gateway *does* install.

(c) Cleanest alternative: print a warning (or error) when both `--with-kuadrant` and `--skip-mcp-gateway` are passed, instead of silently winning one over the other.

(a) is the smallest fix and matches the existing comment directly above ("Kuadrant provides AuthPolicy for MCP Gateway — enable it automatically"), which is the truthful version.

huang195 (approved):
Bellissimo opt-in flip — straightforward, matches the project's flag conventions, CI fully green. Just one doc/behavior mismatch around `--skip-mcp-gateway` "overriding" `--with-kuadrant`: the new docstring promises something the dependency-resolution block doesn't deliver, and the PR body's behavior matrix has the same off-by-one for `--with-all --skip-mcp-gateway`. Easy to align by dropping the override claim or by making the script warn on the conflicting combo.

**Areas reviewed**: Shell (OCP installer flag parsing + dependency block).
**Commits**: 1, signed-off.
**CI**: 21 checks pass (incl. Deploy & Test).

scripts/ocp/setup-kagenti.sh
pdettori: Good catch — fixed all three points:

(a) Updated the docstring in the usage header and `--help` output to say `Skip MCP Gateway (ignored when --with-kuadrant is set, since Kuadrant requires it)` — matching the actual behavior.

(b) Fixed the behavior matrix in the PR body: `--with-all --skip-mcp-gateway` now correctly shows "Yes" since `--with-all` enables Kuadrant, which forces MCP Gateway on (with a warning).

(c) Already implemented — the `log_warn` at line 140 prints a message when the conflict is detected.

PR #1745, opened 2026-05-29
fix: disable MCP Inspector button when not installed

### LINKED ISSUE #1719: feature: Disable MCP Inspector links if MCP Inspector is not installed

pdettori:
## Summary

- **Root cause**: `GET /api/config/dashboards` always returned a default `mcpInspector` URL (`http://mcp-inspector.{domain}:8080`), even when MCP Inspector wasn't installed. This made the button always appear enabled.
- **Fix**: Return `None` from the backend when `mcp_inspector_url` is not configured. The existing `!mcpInspectorUrl` null-check in the frontend then correctly disables the button.
- **Button text**: Updated `MCPGatewayPage.tsx` to show `"MCP Inspector not installed"` when the URL is null.
- **Type safety**: Updated `DashboardConfig.mcpInspector` to `string | null` in `api.ts` to match the nullable backend response.

**No new endpoint or live-probing needed** — the dashboard config is the right signal for install state.

Closes #1719

> Note: PR #1722 addresses the same issue but has a syntax error in `MCPGatewayPage.tsx` (second `useQuery` inserted inside first query's object literal) and uses a fragile live HTTP probe approach. This PR takes the minimal, correct fix.

## Test plan

- [ ] Deploy with `--with-mcp-gateway` — button shows "Open MCP Inspector" and is enabled
- [ ] Deploy without `--with-mcp-gateway` — button shows "MCP Inspector not installed" and is disabled (grey)
- [ ] Verify `GET /api/config/dashboards` returns `mcpInspector: null` when not configured

Assisted-By: Claude Code


kagenti/backend/app/routers/config.py
```diff
@@ -107,7 +107,7 @@ async def get_dashboard_config() -> DashboardConfigResponse:
         traces=settings.traces_dashboard_url,
         network=settings.network_dashboard_url or f"http://kiali.{domain}:8080",
         mlflow=settings.mlflow_dashboard_url,
-        mcpInspector=settings.mcp_inspector_url or f"http://mcp-inspector.{domain}:8080",
+        mcpInspector=settings.mcp_inspector_url or None,
         mcpProxy=settings.mcp_proxy_full_address or f"http://mcp-proxy.{domain}:8080",
```
huang195: Bellissimo on the Inspector fix — but note the line right below has the *exact same shape* for `mcpProxy`: `settings.mcp_proxy_full_address or f"http://mcp-proxy.{domain}:8080"`. If MCP Gateway is opt-in (per #1744), then `mcp_proxy_full_address` will also be unset on installs without the gateway, and `mcpProxy` returns a fallback URL pointing at nothing. Today that's not a user-visible bug because only the Inspector button consumes nullability — the proxy URL isn't rendered as a clickable button on `MCPGatewayPage.tsx`. But if any future card surfaces `mcpProxy` the same way, the same trap is waiting. Worth either (a) symmetric fix in this PR (also flip `mcpProxy` to `Optional[str] = None`), or (b) a follow-up issue capturing it so it doesn't get forgotten.

kagenti/backend/app/routers/config.py
```diff
@@ -107,7 +107,7 @@ async def get_dashboard_config() -> DashboardConfigResponse:
         traces=settings.traces_dashboard_url,
         network=settings.network_dashboard_url or f"http://kiali.{domain}:8080",
         mlflow=settings.mlflow_dashboard_url,
-        mcpInspector=settings.mcp_inspector_url or f"http://mcp-inspector.{domain}:8080",
+        mcpInspector=settings.mcp_inspector_url or None,
```
huang195, line 110: Two small notes on this line, take or leave:

**Nit on `or None`**: technically redundant — when `mcp_inspector_url` is `None`, the `or None` is a no-op; when it's `""` (empty string), the `or` flips to `None`. Net effect: empty string is treated identically to None, which is reasonable — but the comment alone (or just `settings.mcp_inspector_url`) would do the same job for the common case. The explicit form does signal intent.

**Suggestion on testing**: The test plan is manual-only. A small unit test on `GET /api/config/dashboards` asserting `mcpInspector` is `null` when `settings.mcp_inspector_url` is unset (and the URL string when it is set) would lock the behavior in. Cheap to add and prevents a future "fix the fallback again" round-trip if someone later changes the response shape.

huang195 (approved):
Perfetto fix — minimal, correct, and uses the existing frontend null-check rather than threading a new "installed?" boolean through the API. One symmetric concern: `mcpProxy` follows the same fallback pattern and is technically vulnerable to the same bug, just not currently surfaced as a button. Otherwise just polish: redundant `or None` and a missing automated test.

**Areas reviewed**: Python (backend models + router), TypeScript (page + types), uv.lock.
**Commits**: 1, signed-off.

kagenti/backend/app/routers/config.py
pdettori: Done — applied the symmetric fix: `mcpProxy` now returns `None` when `mcp_proxy_full_address` is unset, and `DashboardConfigResponse.mcpProxy` is `Optional[str] = None`. Updated `api.ts` to match (`string | null`).

kagenti/backend/app/routers/config.py
pdettori: Added `test_config_dashboards_api.py` with two cases: null when unconfigured, URL when set. Both for `mcpInspector` and `mcpProxy`. The `or None` is intentional — `mcp_inspector_url` defaults to `""` (empty string) in settings, so the explicit form handles both the empty-string and None cases cleanly.

PR #1747, opened 2026-05-29
fix(ui): show real MCP Gateway status instead of hardcoded Running

### LINKED ISSUE #1720: :bug: Kagenti reports Incorrect status of MCP Gateway
### Describe the bug

Kagenti http://kagenti-ui.localtest.me:8080/mcp-gateway reports an incorrect status for MCP Gateway.  It reports it as `Running`.

Image

### Steps To Reproduce

- Install Kagenti with `scripts/kind/setup-kagenti.sh --with-ui --with-spire --with-agent-sandbox --preload-images`
- Go to http://kagenti-ui.localtest.me:8080/mcp-gateway and log in.  Verify it says the Gateway is `Running`.
- Verify `kubectl get ns` doesn't even show a _gateway-system_ namespace.

### Expected Behavior

Kagenti should report the status as "Not installed" or something similar.

### Additional Context

_No response_

pdettori:
## Summary

- Add backend `GET /config/mcp-gateway-status` endpoint that checks the `mcp-gateway-istio` deployment in `gateway-system` namespace
- Replace hardcoded green "Running" label in the MCP Gateway page with a real API call
- Display "Running" (green), "Degraded" (orange), or "Not Installed" (grey) based on actual cluster state

Closes #1720

## Test plan

- [ ] Unit tests pass: `uv run pytest tests/test_mcp_gateway_status.py -v` (4 tests covering Ready/Degraded/Missing/API-error)
- [ ] Deploy to Kind without MCP Gateway → page shows "Not Installed"
- [ ] Deploy to Kind with MCP Gateway → page shows "Running"
- [ ] TypeScript compiles cleanly (`npx tsc --noEmit`)

Assisted-By: Claude Code

### DIFF
--- a/kagenti/backend/app/routers/config.py
+++ b/kagenti/backend/app/routers/config.py
@@ -230,3 +230,22 @@ async def get_platform_status(
     )
 
     return PlatformStatusResponse(components=components, registry=registry)
+
+
+class MCPGatewayStatusResponse(BaseModel):
+    """Status of the MCP Gateway component."""
+
+    status: Literal["Ready", "Degraded", "Missing"]
+
+
+@router.get(
+    "/mcp-gateway-status",
+    response_model=MCPGatewayStatusResponse,
+    dependencies=[Depends(require_roles(ROLE_VIEWER))],
+)
+async def get_mcp_gateway_status(
+    kube: KubernetesService = Depends(get_kubernetes_service),
+) -> MCPGatewayStatusResponse:
+    """Return the health status of the MCP Gateway deployment."""
+    status = _check_deployment_ready(kube, "gateway-system", "mcp-gateway-istio")
+    return MCPGatewayStatusResponse(status=status)
--- a/kagenti/backend/app/routers/config.py
+++ b/kagenti/backend/app/routers/config.py
@@ -247,5 +247,6 @@ async def get_mcp_gateway_status(
     kube: KubernetesService = Depends(get_kubernetes_service),
 ) -> MCPGatewayStatusResponse:
     """Return the health status of the MCP Gateway deployment."""
+    # TODO: parameterize namespace/deployment if install location ever varies
     status = _check_deployment_ready(kube, "gateway-system", "mcp-gateway-istio")
     return MCPGatewayStatusResponse(status=status)
--- a/kagenti/ui-v2/src/pages/MCPGatewayPage.tsx
+++ b/kagenti/ui-v2/src/pages/MCPGatewayPage.tsx
@@ -34,14 +34,9 @@ import { useQuery } from '@tanstack/react-query';
 import { configService } from '@/services/api';
 
 export const MCPGatewayPage: React.FC = () => {
-  // Placeholder query - will be replaced with actual API call
-  const { isLoading } = useQuery({
+  const { data: gatewayStatus, isLoading } = useQuery({
     queryKey: ['mcp-gateway-status'],
-    queryFn: async () => {
-      // Placeholder - return mock data for now
-      return { status: 'running', tools: 12, uptime: '5d 12h' };
-    },
-    enabled: false, // Disabled until API is ready
+    queryFn: () => configService.getMCPGatewayStatus(),
   });
 
   // Fetch dashboard config for MCP Inspector URL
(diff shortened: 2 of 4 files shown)


kagenti/ui-v2/src/pages/MCPGatewayPage.tsx
```diff
@@ -103,7 +98,15 @@ export const MCPGatewayPage: React.FC = () => {
...
+                        {gatewayStatus?.status === 'Ready' && (
+                          <Label color="green">Running</Label>
+                        )}
+                        {gatewayStatus?.status === 'Degraded' && (
+                          <Label color="orange">Degraded</Label>
+                        )}
+                        {(gatewayStatus?.status === 'Missing' || !gatewayStatus) && (
+                          <Label color="grey">Not Installed</Label>
+                        )}
                       </DescriptionListDescription>
```
huang195, line 111: Mamma mia — small UX wart: while the query is loading (`gatewayStatus === undefined`), the third branch fires and shows "Not Installed" briefly before the real status arrives. Same when the API errors out for transient reasons (network blip, backend restart): `gatewayStatus` stays undefined → "Not Installed". Two cheap fixes: (a) gate on `isLoading`/`isError` from the same `useQuery` hook (you already destructure `isLoading` at line 36) and render `<Label color="grey">Checking…</Label>` for loading and `<Label color="red">Status unknown</Label>` for error; (b) keep current code but key off `gatewayStatus?.status === 'Missing'` only (drop `|| !gatewayStatus`), so undefined renders nothing rather than misleading "Not Installed". Pick whichever — (a) is friendlier UX.

kagenti/backend/app/routers/config.py
```diff
@@ -230,3 +230,22 @@ async def get_platform_status(
...
+@router.get(
+    "/mcp-gateway-status",
+    response_model=MCPGatewayStatusResponse,
+    dependencies=[Depends(require_roles(ROLE_VIEWER))],
+)
+async def get_mcp_gateway_status(
+    kube: KubernetesService = Depends(get_kubernetes_service),
+) -> MCPGatewayStatusResponse:
+    """Return the health status of the MCP Gateway deployment."""
+    status = _check_deployment_ready(kube, "gateway-system", "mcp-gateway-istio")
```
huang195, line 251: Two notes on this line, take or leave:

**Nit on hardcoded names**: The namespace + deployment name are hardcoded (`gateway-system` / `mcp-gateway-istio`). For now this is fine — that's where the chart installs it — but if any deployment ever lands the gateway elsewhere (custom helm overrides, future restructure), the endpoint silently returns "Missing." Worth either pulling these into `settings.mcp_gateway_namespace` / `settings.mcp_gateway_deployment_name` (with sensible defaults), or at least leaving a `# TODO: parameterize if install location ever varies` so a future maintainer doesn't re-debug the same thing.

**Suggestion on cross-PR symmetry**: Bellissimo separation — gateway-status is a runtime/health property, dashboard-config is a static-URL property, two endpoints is the right shape. Worth noting though that PR #1745 used "URL is null → not installed" as the signal for MCP Inspector. Two different patterns for the same conceptual question ("is X installed?") is fine for now, but if a third opt-in component lands, picking one canonical approach (deployment-health check seems more robust) and migrating the Inspector to it would prevent drift. Maybe a follow-up issue rather than an in-this-PR change.

huang195 (approved):
Solid fix — backend endpoint with proper Ready/Degraded/Missing taxonomy, four-case unit test coverage, frontend wired through `configService`. Headline polish: the "loading or error" → "Not Installed" rendering should distinguish "we don't know yet" from "we checked and it's not there." Hardcoded namespace+deployment is fine for today but worth a TODO. Conceptual symmetry with PR #1745's nullable-URL approach is a separate concern for a future cleanup pass.

**Areas reviewed**: Python (backend router + tests), TypeScript (page + service + types).
**Commits**: 1, signed-off.

kagenti/ui-v2/src/pages/MCPGatewayPage.tsx
pdettori: Fixed in c32f35e4 — destructured `isError` from useQuery and render `<Label color="red">Status unknown</Label>` on error. Dropped the `|| !gatewayStatus` fallback so undefined renders nothing (skeleton handles the loading state already).

kagenti/backend/app/routers/config.py
pdettori: Added a TODO comment on the hardcoded names. Agree that settings-based parameterization makes sense if the install location ever varies.

On the cross-PR symmetry with #1745: noted — deployment-health check is more robust than nullable-URL as a canonical "is X installed?" signal. Worth a follow-up issue if a third opt-in component lands.

PR #1749, opened 2026-05-29
Feat: UI verbiage simplification and outbound table for AuthBridge

### LINKED ISSUE #1309: feature: Simplify Import Agent AuthBridge UI
### Feature Description

The Import Agent UI AuthBridge options are stale.

"Enable AuthBridge sidecar injection" explicitly names the K8s sidecar container names, and these have changed.  We also name the technology used "sidecar injection" and the user-facing verbiage should discuss the capabilities.  Use "Enable AuthBridge security constraints" (or something) that doesn't mention sidecars.

The "Enable SPIRE Identity" checkbox seems more independent of AuthBridge settings than it is.
The Advanced Injection Controls names the K8s sidecars (which may now be stale and it isn't clear to me if these are still needed and if they can be set independently.

Consider labeling the checkobx "Enable Kagenti Authbridge", rather than merely "Enable Authbridge", to avoid confusion with [this company](https://authbridge.com/).

### Proposed Solution

Simplify the screen based on the security config options we wish to expose, following the suggestions above.

### Want to contribute?

### Additional Context

There is additional AuthBridge config in "import tools", but it currently doesn't work.  If it won't work soon remove it; if it will work soon but doesn't work yet please indicate that with help text and disabled control.

esnible:
## Summary

Resolves #1309

In addition to simplifying verbiage, this PR tries to make it easier to understand the three components of outbound route rules, and make the UI behave more like similar UIs for table editing.

The usual -- check out this PR, then

```bash
make build-load-ui-frontend &&
  UI_TAG=$(docker images | grep kagenti-ui-v2 | tail -n 1 | awk 'END {print $2}') &&
  echo UI_TAG is $UI_TAG &&
  oc -n kagenti-system set image deployment/kagenti-ui frontend=ghcr.io/kagenti/kagenti-ui-v2:${UI_TAG} &&
```

### DIFF
--- a/kagenti/ui-v2/src/pages/ImportAgentPage.tsx
+++ b/kagenti/ui-v2/src/pages/ImportAgentPage.tsx
@@ -203,13 +203,19 @@ export const ImportAgentPage: React.FC = () => {
   });
   const [outboundRoutes, setOutboundRoutes] = useState<RouteRow[]>([makeEmptyRoute()]);
   const isEmptyRoute = (r: RouteRow) => !r.host && !r.target_audience;
+  const isCommittedRoute = (r: RouteRow) => !!r.host && !!r.target_audience;
   const removeRoute = (i: number) => setOutboundRoutes((prev) => prev.filter((_, idx) => idx !== i));
   const updateRoute = (i: number, field: 'host' | 'target_audience' | 'token_scopes', value: string) =>
     setOutboundRoutes((prev) => {
       const updated = [...prev];
       updated[i] = { ...updated[i], [field]: value };
-      // If the user edited the trailing empty row, append a fresh empty row.
-      if (i === prev.length - 1 && (value || updated[i].host || updated[i].target_audience)) {
+      // Once the trailing row has both Host Pattern and Target OIDC Audience
+      // filled in, promote it by appending a fresh empty row below.
+      if (
+        i === prev.length - 1 &&
+        updated[i].host &&
+        updated[i].target_audience
+      ) {
         updated.push(makeEmptyRoute());
       }
       return updated;
@@ -1179,13 +1185,13 @@ export const ImportAgentPage: React.FC = () => {
               <ExpandableSection
                 toggleText={(() => {
                   const n = outboundRoutes.filter((r) => !isEmptyRoute(r)).length;
-                  return `Outbound Authorization header control (${n} route${n !== 1 ? 's' : ''})`;
+                  return `Outbound OIDC token exchange rules (${n} route${n !== 1 ? 's' : ''})`;
                 })()}
                 isExpanded={showOutboundRouting}
                 onToggle={(_event, expanded) => setShowOutboundRouting(expanded)}
               >
                 <Text component="p" style={{ marginBottom: '8px' }}>
-                  RFC 8693 / OAuth 2.0 Token Exchange - Restrict outbound to certain hosts and OIDC audiences and scopes.
+                  RFC 8693 / OAuth 2.0 Token Exchange - Restrict outbound to certain hosts, OIDC audiences, and scopes.
                 </Text>
                 <Table aria-label="Outbound routes" variant="compact">
                   <Thead>
@@ -1198,7 +1204,7 @@ export const ImportAgentPage: React.FC = () => {
                   </Thead>
                   <Tbody>
                     {outboundRoutes.map((route, index) => {
-                      const isTrailingEmpty = isEmptyRoute(route) && index === outboundRoutes.length - 1;
+                      const showRemove = isCommittedRoute(route);
                       return (
                         <Tr key={route.id}>
                           <Td>
@@ -1226,7 +1232,7 @@ export const ImportAgentPage: React.FC = () => {
                             />
                           </Td>
                           <Td>
-                            {!isTrailingEmpty && (
+                            {showRemove && (
                               <Button variant="link" onClick={() => removeRoute(index)}>
                                 Remove
                               </Button>
--- a/kagenti/ui-v2/src/pages/ImportAgentPage.tsx
+++ b/kagenti/ui-v2/src/pages/ImportAgentPage.tsx
@@ -35,6 +35,7 @@ import {
   Checkbox,
 } from '@patternfly/react-core';
 import { TrashIcon, PlusCircleIcon, UploadIcon } from '@patternfly/react-icons';
+import { Table, Thead, Tr, Th, Tbody, Td } from '@patternfly/react-table';
 import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
 
 import { agentService, ShipwrightBuildConfig, skillService } from '@/services/api';
@@ -1147,7 +1148,7 @@ export const ImportAgentPage: React.FC = () => {
               </FormGroup>
 
               {authBridgeEnabled && (
-                <div style={{ marginLeft: '24px' }}>
+                <FormGroup fieldId="useEnvoyMode" style={{ marginLeft: '24px' }}>
                   <Checkbox
                     id="useEnvoyMode"
                     label="Sandbox networking with Envoy proxy"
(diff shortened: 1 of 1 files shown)


esnible:
authbridge-ui-2

kagenti/ui-v2/src/pages/ImportAgentPage.tsx
```diff
@@ -525,7 +542,9 @@ export const ImportAgentPage: React.FC = () => {
         // and envoy-sidecar support the full disabled/permissive/strict
         // matrix end-to-end (kagenti-operator#381 + extensions#441).
         mtlsMode: authBridgeEnabled ? mtlsMode : undefined,
-        outboundRoutes: authBridgeEnabled && outboundRoutes.length > 0 ? outboundRoutes.map(({ id, ...r }) => r) : undefined,
+        outboundRoutes: authBridgeEnabled && outboundRoutes.some((r) => !isEmptyRoute(r))
```
huang195: **must-fix**: This filter uses `!isEmptyRoute(r)` (and the duplicate at line 587 in the update path), but `isEmptyRoute` requires **both** `host` AND `target_audience` to be empty. So a half-filled row (e.g., user typed `github.com` in Host but never filled Target) passes the filter and gets submitted as `{host: "github.com", target_audience: "", token_scopes: "openid"}`.

Commit `5b67bfb`'s title says "Don't accept a outbound route exchange until it has been filled in" — to honor that, this filter should use `isCommittedRoute(r)` (both fields non-empty), which is the helper you already defined and use to gate the Remove button.

The expandable-section title bar at line 1185 also uses `!isEmptyRoute`, so consider whether the count should reflect committed-only or all-non-empty (loose count for "in progress" + strict filter for submit is also defensible — just be deliberate). Apply the same fix at the duplicate site (line 587) so create and update flows agree.

kagenti/ui-v2/src/pages/ImportAgentPage.tsx
```diff
@@ -1125,26 +1146,26 @@ export const ImportAgentPage: React.FC = () => {
...
                       setUseEnvoyMode(false);
                     }
                   }}
-                  description="When enabled, the operator injects a combined AuthBridge sidecar for inbound JWT validation and outbound token exchange. Defaults to proxy-sidecar mode (HTTP_PROXY)."
+                  description="When enabled, the agent will reject inbound messages without valid JWT and can perform outbound token exchange."
               />
               </FormGroup>
 
               {authBridgeEnabled && (
                 <FormGroup fieldId="useEnvoyMode" style={{ marginLeft: '24px' }}>
```
huang195, line 1167: **suggestion**: "Use envoy-sidecar mode" → "Use Envoy + iptables interception" trades clarity-for-non-experts for technical precision. The mode is named `envoy-sidecar` consistently in docs, operator code, and troubleshooting (`kagenti-operator#381`, `extensions#441`) — losing that label here creates a gap when an operator searches for the term they saw in their pod. Consider keeping the mode name and using the description (line 1166) to explain what it means: e.g., `label="Use envoy-sidecar mode (Envoy + iptables interception)"` keeps the searchable string + the technical hint.

kagenti/ui-v2/src/pages/ImportAgentPage.tsx
```diff
@@ -1213,7 +1251,7 @@ export const ImportAgentPage: React.FC = () => {
               <ExpandableSection
                 toggleText="AuthBridge Advanced Configuration"
```
huang195, line 1257: **nit**: "Disable outbound security for some ports" reads a bit alarming — the change actually disables AuthBridge proxy *interception* on those ports; other security primitives (mTLS at L4, NetworkPolicy, etc.) still apply. Something like "Bypass AuthBridge proxy on these outbound ports" matches what the underlying script does without implying broader safety regression. Same applies to "Disable AuthBridge inbound security on ports" at line 1273.

kagenti/ui-v2/src/pages/ImportAgentPage.tsx
```diff
@@ -190,19 +191,35 @@ export const ImportAgentPage: React.FC = () => {
...
+  const makeEmptyRoute = (): RouteRow => ({
+    id: newRouteRowId(),
+    host: '',
+    target_audience: '',
+    token_scopes: 'openid',
+  });
+  const [outboundRoutes, setOutboundRoutes] = useState<RouteRow[]>([makeEmptyRoute()]);
+  const isEmptyRoute = (r: RouteRow) => !r.host && !r.target_audience;
+  const isCommittedRoute = (r: RouteRow) => !!r.host && !!r.target_audience;
+  const removeRoute = (i: number) => setOutboundRoutes((prev) => prev.filter((_, idx) => idx !== i));
```
huang195, line 210: **suggestion**: Bellissimo on the auto-trailing-empty-row UX — way friendlier than the old "click + add row" flow. One subtle gotcha: the promotion check `updated[i].host && updated[i].target_audience` only fires from `updateRoute`. If a user pastes both fields in a single keystroke event into the same row, only the field that actually fired `onChange` triggers the check; the other field's value is read fresh.

Today that's fine because each TextInput has its own onChange — but if anything ever batches multi-field updates, the trailing row may not auto-promote. Worth a one-line comment explaining the invariant: "promotion runs on every per-field onChange; both fields are read fresh inside the setter so order doesn't matter." Just a paper trail for the next maintainer.

huang195 (changes requested):
Tasty UX upgrade: PatternFly Table + auto-trailing-empty-row removes a click and the verbiage rewrites are mostly clearer for operators doing this for the first time. One real bug worth flagging: the submit filter (lines 545 and 587) uses `!isEmptyRoute(r)` where the commit history says it should use `isCommittedRoute(r)`, so half-filled rows will leak through to the backend. Two suggestions on label wording (envoy-sidecar mode discoverability + "disable security" tone) and a couple of nits. Once the filter is fixed, this is APPROVE.

**Areas reviewed**: TypeScript / React (single-file ImportAgentPage).
**Commits**: 6, all signed-off.

kagenti/ui-v2/src/pages/ImportAgentPage.tsx
esnible: I chose the wording because I believe our users will do a web search, rather than consult our documentation.  My hope is that our documentation will follow the wording we choose in the UI.  I don't want to introduce Kagenti-specific terms, such as "envoy-sidecar mode".  Maybe "Envoy + iptables" is too technical for our users.  I had originally used the term "security sandbox" here, and Opus 4.7 suggested the phrasing that it is now complaining about!  Do you have any suggestions you like?

kagenti/ui-v2/src/pages/ImportAgentPage.tsx
```diff
@@ -35,6 +35,7 @@ import {
   Checkbox,
 } from '@patternfly/react-core';
 import { TrashIcon, PlusCircleIcon, UploadIcon } from '@patternfly/react-icons';
+import { Table, Thead, Tr, Th, Tbody, Td } from '@patternfly/react-table';
 import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
 
 import { agentService, ShipwrightBuildConfig, skillService } from '@/services/api';
```
huang195, line 41: `isCommittedRoute` silently drops rows where the user typed something into one field but never completed the other. The old behavior preserved every user-added row. Probably fine given the new auto-trailing-row UX, but consider: if a user types a host pattern, gets distracted, and submits, their partial work disappears with no warning. Not a blocker — flagging as an intentional UX tradeoff to confirm.

huang195 (approved):
Clean refactor of the outbound routes UI to use a PatternFly `Table` with the always-on trailing empty row pattern. Promotion logic is correct (reads post-update state, order-independent), `showRemove` correctly hides the trash on the trailing row, and serialization filters out incomplete rows on submit. Verbiage simplifications are good — "Secure with Kagenti AuthBridge" lands much better than the previous technical phrasing.

**Areas reviewed**: Frontend (React/TypeScript)
**Commits**: 10, all signed-off
**CI status**: passing (HyperShift E2E pending maintainer trigger)

PR #1753, opened 2026-05-30
fix(release): pin Chart.yaml version by default in release scripts

pdettori:
## Summary

- `pin-release-tags.sh` now updates `Chart.yaml` version and appVersion to match the release VERSION argument **by default** — no need to pass `--chart-version` separately
- `check-release-pins.sh` gains a new Check 3b that validates Chart.yaml version/appVersion consistency with pinned image tags
- Release skill updated to reflect the new default behavior
- Fixes pre-existing bug in `check-release-pins.sh` spiffe check (missing line-number argument to `add_error`)

## Context

RC tags v0.6.0-rc.7 through rc.11 all shipped with `Chart.yaml version: 0.6.0-rc.6` and `appVersion: 0.6.0-rc.5` because the release workflow didn't pass `--chart-version` to `pin-release-tags.sh`. This change makes chart version pinning automatic so it can't drift again.

## Test plan

- [x] `pin-release-tags.sh v0.6.0-rc.12 --dry-run` — shows Chart.yaml version update
- [x] `pin-release-tags.sh v0.6.0-rc.12 --dry-run --skip-chart-version` — correctly skips Chart.yaml
- [x] `check-release-pins.sh` — new consistency check passes on main (version matches tags)
- [ ] Pre-commit hooks pass

Assisted-By: Claude Code

### DIFF
--- a/scripts/check-release-pins.sh
+++ b/scripts/check-release-pins.sh
@@ -362,27 +362,32 @@ fi
 # appVersion should be 0.6.0-rc.11 (without the 'v' prefix).
 # ---------------------------------------------------------------------------
 
-if [[ -f "$CHART_FILE" ]] && [[ -f "$VALUES_FILE" ]] && command -v yq >/dev/null 2>&1; then
-    chart_version=$(yq eval '.version' "$CHART_FILE" 2>/dev/null || echo "")
-    chart_app_version=$(yq eval '.appVersion' "$CHART_FILE" 2>/dev/null || echo "")
-    platform_tag=$(yq eval '.ui.frontend.tag' "$VALUES_FILE" 2>/dev/null || echo "")
+if [[ -f "$CHART_FILE" ]] && [[ -f "$VALUES_FILE" ]]; then
+    if ! command -v yq >/dev/null 2>&1; then
+        add_warning "charts/kagenti/Chart.yaml" \
+            "yq not found — skipping Chart.yaml version/appVersion consistency check"
+    else
+        chart_version=$(yq eval '.version' "$CHART_FILE" 2>/dev/null || echo "")
+        chart_app_version=$(yq eval '.appVersion' "$CHART_FILE" 2>/dev/null || echo "")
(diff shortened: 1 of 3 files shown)


scripts/check-release-pins.sh
```diff
@@ -354,6 +354,39 @@ else
...
 
+# ---------------------------------------------------------------------------
+# Check 3b: Chart.yaml version and appVersion match image tags
+#
+# The chart version and appVersion should be consistent with the pinned image
+# tags. If image tags are pinned to e.g. v0.6.0-rc.11, the chart version and
+# appVersion should be 0.6.0-rc.11 (without the 'v' prefix).
+# ---------------------------------------------------------------------------
+
+if [[ -f "$CHART_FILE" ]] && [[ -f "$VALUES_FILE" ]] && command -v yq >/dev/null 2>&1; then
```
huang195: Check 3b silently skips when `yq` isn't installed — consistent with the other yq-dependent checks, but silent gaps in release validation are worth flagging. Consider emitting a `warn "yq not found — skipping Chart.yaml version check"` when the `command -v yq` guard fails.

huang195 (approved):
Mamma mia, this is a clean fix for a real problem — rc.6/rc.5 vs rc.11 drift is exactly the kind of silent breakage that causes pain at GA. Making `--chart-version` default to `VERSION` is the right call: the existing stripping logic (`${CHART_VERSION#v}`) handles the 'v' prefix correctly, and Check 3b in `check-release-pins.sh` closes the loop by catching any future drift before it ships. The spiffe `add_error` line-number fix is a nice cleanup too — bellissimo!

**Areas reviewed**: Shell scripts, docs/skills
**CI status**: All checks passing ✅

scripts/check-release-pins.sh
pdettori: Fixed — both Check 3b and Check 6 now emit a warning when `yq` is not found instead of silently skipping. See d24410ca.

PR #1755, opened 2026-05-30
fix(helm): restart otel-collector on config change

pdettori:
## Summary

Add `checksum/config` annotation to the `otel-collector` Deployment pod template so that `helm upgrade` triggers a rolling restart whenever the rendered OTEL collector configuration changes.

Key changes:
- Add `checksum/config: {{ include "kagenti.otel.collectorConfig" . | sha256sum }}` annotation to the pod template metadata in `charts/kagenti-deps/templates/otel-collector.yaml`

**Root cause**: On a fresh OCP install with RHOAI MLflow, `setup-kagenti.sh` performs multiple Helm upgrades. The first upgrade deploys the OTEL collector with a config that only has the `debug` exporter (because `MLFLOW_TRACES_ENDPOINT` is not yet known). Subsequent upgrades update the ConfigMap to add the `otlphttp/mlflow` exporter, but without this annotation the pod never restarts — so it keeps running the initial config and silently drops all traces to MLflow.

The annotation ensures the pod spec changes whenever the config changes, triggering a Kubernetes rolling update automatically.

Assisted-By: Claude Code

### DIFF
--- a/charts/kagenti-deps/templates/otel-collector.yaml
+++ b/charts/kagenti-deps/templates/otel-collector.yaml
@@ -46,6 +46,8 @@ spec:
       app: otel-collector
   template:
     metadata:
+      annotations:
+        checksum/config: {{ include "kagenti.otel.collectorConfig" . | sha256sum }}
       labels:
         app: otel-collector
     spec:


huang195 (approved):
Bellissimo pattern — `checksum/config` on the pod template is exactly the right Helm idiom for this. The `kagenti.otel.collectorConfig` helper merges all config variants (phoenix, mlflow, auth, RHOAI) so any config change will invalidate the hash and trigger a rolling restart automatically. No more silent trace drops after multi-step upgrades.

**Areas reviewed**: Helm/K8s
**CI status**: All checks passing ✅ (E2E HyperShift pending maintainer trigger — expected)

PR #1759, opened 2026-06-01
fix(ocp-installer): guard MLflow kubectl-run from set -e exit

### LINKED ISSUE #1743: OCP installer: --with-all does not install MCP Gateway

pdettori:
## Summary

- Guard `kubectl run mlflow-exp-create` command substitution with `|| true` to prevent `set -e` from killing the script when the curl pod exits non-zero
- Guard the `python3` JSON parse on the same path for belt-and-suspenders
- PR #1754 fixed the token creation line but missed these two lines — same root cause, same fix pattern

## Root Cause

With `set -euo pipefail`, this line:
```bash
_EXP_RESP=$(kubectl run mlflow-exp-create --rm -i ... 2>/dev/null)
```
exits the script when the pod returns non-zero (MLflow not ready, network timeout, image pull failure). The `2>/dev/null` suppresses stderr but NOT the exit code in a variable assignment under `set -e`. Steps 5, 5b, and 6 (Kuadrant, MCP Gateway, verification) are silently skipped.

## Test plan

- [ ] Run `scripts/ocp/setup-kagenti.sh --with-all` on OpenShift
- [ ] Verify Steps 5b and 6 execute (MCP Gateway gets installed)
- [ ] Verify MLflow experiment creation still succeeds on healthy clusters

Closes #1743 (follow-up to #1754)

Assisted-By: Claude Code

### DIFF
--- a/scripts/ocp/setup-kagenti.sh
+++ b/scripts/ocp/setup-kagenti.sh
@@ -1473,9 +1473,9 @@ else
         -H "x-mlflow-workspace: '"$_EXP_WS"'" \
         -H "Content-Type: application/json" \
         -d "{\"name\":\"'"$_EXP_NAME"'\"}" \
-        "https://mlflow.'"${MLFLOW_NAMESPACE}"'.svc.cluster.local:8443/api/2.0/mlflow/experiments/create"' 2>/dev/null)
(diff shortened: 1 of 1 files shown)


rubambiza (approved):
LGTM.

PR #1765, opened 2026-06-01
fix(ocp-installer): upgrade MCP Gateway to v0.6.0 and fix hostname routing

pdettori:
## Summary

- Upgrades MCP Gateway chart from v0.5.1 to v0.6.0 on OpenShift installer, fixing a bug where the controller pod was included in the `mcp-gateway` Service endpoints (causing ext_proc gRPC to connect to the wrong pod ~50% of the time)
- Fixes Gateway listener hostname mismatch on OpenShift — the listener now uses the detected apps domain to match the auto-generated Route hostname
- Adds `mcpGateway.openshiftDomain` value so the kagenti Helm chart can construct the correct Gateway listener hostname on OCP

## Root Cause

Chart v0.5.1 labels the controller pod with `app.kubernetes.io/name: mcp-gateway` (same as the broker-router). The Service selector `{name: mcp-gateway}` matches both pods, so the Istio gateway's ext_proc gRPC randomly connects to the controller (which doesn't serve on port 50051) → connection reset → HTTP 500. Chart v0.6.0 fixes this by labeling the controller `mcp-gateway-controller`.

## Test plan

- [x] Verified on OCP cluster: `curl -X POST https://mcp-gateway-gateway-system.apps.kagenti1.kubestellar.org/mcp` returns valid MCP initialize response
- [x] Verified internal cluster path works (Inspector proxy → Istio Gateway → MCP Gateway)
- [ ] Run OCP installer from scratch on a clean cluster
- [ ] Verify Kind install still works (no regression — Kind already uses v0.6.0)

Assisted-By: Claude Code

### DIFF
--- a/charts/kagenti/templates/mcp-gateway.yaml
+++ b/charts/kagenti/templates/mcp-gateway.yaml
@@ -20,6 +20,9 @@ spec:
   listeners:
     - name: mcp
       {{- if .Values.openshift }}
+      {{- if not .Values.mcpGateway.openshiftDomain }}
+      {{- fail "mcpGateway.openshiftDomain is required on OpenShift (set by scripts/ocp/setup-kagenti.sh)" }}
+      {{- end }}
       hostname: "mcp-gateway-{{ .Values.mcpGateway.namespaces.gatewaySystem }}.{{ .Values.mcpGateway.openshiftDomain }}"
       {{- else }}
       hostname: "{{ .Values.mcpGateway.hostname }}"
--- a/charts/kagenti/templates/mcp-gateway.yaml
+++ b/charts/kagenti/templates/mcp-gateway.yaml
@@ -19,7 +19,11 @@ spec:
   gatewayClassName: istio
   listeners:
     - name: mcp
+      {{- if .Values.openshift }}
+      hostname: "mcp-gateway-{{ .Values.mcpGateway.namespaces.gatewaySystem }}.{{ .Values.mcpGateway.openshiftDomain }}"
+      {{- else }}
       hostname: "{{ .Values.mcpGateway.hostname }}"
+      {{- end }}
       port: 8080
       protocol: HTTP
       allowedRoutes:
--- a/deployments/envs/ocp_values.yaml
+++ b/deployments/envs/ocp_values.yaml
@@ -110,6 +110,9 @@ charts:
         useServiceAccountCA: true
       keycloak:
         url: http://keycloak-service.keycloak:8080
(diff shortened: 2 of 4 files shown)


charts/kagenti/templates/mcp-gateway.yaml
```diff
@@ -19,7 +19,11 @@ spec:
   gatewayClassName: istio
   listeners:
     - name: mcp
+      {{- if .Values.openshift }}
+      hostname: "mcp-gateway-{{ .Values.mcpGateway.namespaces.gatewaySystem }}.{{ .Values.mcpGateway.openshiftDomain }}"
```
huang195, line 26: Perfetto on the conditional! One small thing to consider: if someone deploys the chart on OCP without the installer (forgetting to pass `mcpGateway.openshiftDomain`), the listener hostname renders as `mcp-gateway-gateway-system.` with no domain suffix. Could add a template guard:

```yaml
{{- if not .Values.mcpGateway.openshiftDomain }}
{{ fail "mcpGateway.openshiftDomain is required on OpenShift" }}
{{- end }}
```

Not blocking since the installer always provides it and the pattern of installer-managed values is already established — just thinking about the direct-Helm-install case.

deployments/envs/ocp_values.yaml
```diff
@@ -110,6 +110,9 @@ charts:
         useServiceAccountCA: true
       keycloak:
         url: http://keycloak-service.keycloak:8080
+      # mcpGateway.openshiftDomain is set dynamically by the installer
+      # (scripts/ocp/setup-kagenti.sh) using the detected apps domain.
+      # Example: apps.kagenti1.kubestellar.org
```
huang195: Nice that you documented this as installer-managed — very helpful. Could add a pointer to the exact `--set` flag or script (`scripts/ocp/setup-kagenti.sh`) for someone reading this file without running the installer. Tiny nit, no action needed.

huang195 (approved):
Bellissimo fix! This addresses a real production issue — a 50% gRPC failure rate is pretty brutal, and the root cause (shared label between broker-router and controller) was genuinely subtle. The fix is minimal and correct: Kind clusters are safely guarded by `openshift: false` in all dev_values files, and `DOMAIN` is already validated with a prompt fallback in the script, so no empty-domain risk. All CI checks pass.

**Areas reviewed**: Helm/K8s, Shell, YAML
**Commits**: 1 commit, signed-off ✓, `Assisted-By` attribution correct ✓
**CI status**: All passing (CodeQL, Deploy & Test, E2E HyperShift pending as expected)

PR #1767, opened 2026-06-01
Docs: Kagenti needs 6 cpus, but docs say 4

esnible:
## Summary

This fixes an important mistake in our install instructions -- mac users should have at least 6 CPUs in their podman setup to use kind with Kagenti.

Also other minor doc fixes found while reading docs.


cwiklik (approved):
Verified all changes against `upstream/main`:

1. **CPU fix** (`--cpus 4` → `--cpus 6`): correct, platform needs 6+ CPUs with Istio+SPIRE
2. **Tornjak API curl removal**: confirmed `tornjak-api` HTTPRoute is not created by the installer — only `tornjak-ui` is set up, so these curl examples were broken
3. **Ansible installer → `setup-kagenti.sh`**: confirmed `deployments/ansible/run-install.sh` no longer exists on `main`
4. **`--with-examples` flag**: confirmed the flag exists in `setup-kagenti.sh`
5. **Stale `run-install.sh` comments** removed from `dev_values_federated-jwt.yaml`

**Areas reviewed**: Docs, YAML
**Commits**: 3, all signed-off ✓
**CI**: all passing

PR #1771, opened 2026-06-01
fix(kagenti-deps): derive KC_HOSTNAME from keycloak.publicUrl when set

### LINKED ISSUE #1770: kagenti-deps: KC_HOSTNAME ignores keycloak.publicUrl, breaking OIDC discovery for non-OpenShift HTTPS deployments
## Summary

Operators who deploy `kagenti-deps` on a non-OpenShift cluster (Kind, vanilla Kubernetes, etc.) and front Keycloak with an HTTPS ingress find that **browser login times out with `ERR_CONNECTION_TIMED_OUT`**, even with `keycloak.publicUrl` set to the right HTTPS URL.

Cause: the Keycloak StatefulSet is started with `KC_HOSTNAME` hardcoded to `http://keycloak.<domain>:8080/`. The chart never reads `keycloak.publicUrl` for this env var, even though it uses `publicUrl` correctly everywhere else. Keycloak embeds `KC_HOSTNAME` into its OIDC discovery document, so every OIDC client gets pointed at an unreachable URL.

The fix is a one-line change to honor `keycloak.publicUrl` when set.

## Background — what's actually broken

When a user opens the Kagenti UI logged-out, this happens:

1. Kagenti UI: "you're not logged in, redirecting you to Keycloak."
2. Browser asks Keycloak: "what URL should I send the login form to?"
   Keycloak answers based on its own `KC_HOSTNAME` env var.
3. Browser POSTs the login form to whatever URL Keycloak gave it.
4. Keycloak verifies, issues a token, redirects back to Kagenti.

Step 2 is where it goes wrong. The `kagenti-deps` chart starts Keycloak with `KC_HOSTNAME=http://keycloak.<domain>:8080/`. So the answer Keycloak gives in step 2 is "POST your login form to `http://keycloak.<domain>:8080/...`" — wrong protocol (HTTP), wrong port (8080, which isn't open externally on most clusters fronted by an HTTPS ingress).

Step 3 then hangs and times out. The user sees `ERR_CONNECTION_TIMED_OUT`.

The chart already has the right value sitting in `keycloak.publicUrl` (e.g. `https://keycloak.my-cluster.example.com`). Six other places in the chart use it correctly — `keycloak-realm-init.yaml`, `authbridge-template-configmaps.yaml`, `ui.yaml`, etc. Only the `KC_HOSTNAME` env var ignores it and uses a hardcoded fallback unconditionally.

## Reproduction

Cluster: non-OpenShift (e.g. Kind), `kagenti-deps` v0.1.0, ingress terminating TLS at `https://keycloak.<domain>`.

```sh
# 1. Confirm publicUrl is set
$ helm get values kagenti-deps -n kagenti-system | grep publicUrl
publicUrl: https://keycloak.163-75-85-180.sslip.io

# 2. Inspect what the StatefulSet actually got
$ kubectl get sts -n keycloak keycloak \
    -o jsonpath='{range .spec.template.spec.containers[0].env[?(@.name=="KC_HOSTNAME")]}{.value}{end}'
http://keycloak.163-75-85-180.sslip.io:8080/

# 3. OIDC discovery is wrong
$ curl -sk https://keycloak.163-75-85-180.sslip.io/realms/kagenti/.well-known/openid-configuration \
    | jq .authorization_endpoint
"http://keycloak.163-75-85-180.sslip.io:8080/realms/kagenti/protocol/openid-connect/auth"
```

Expected: `KC_HOSTNAME=https://keycloak.163-75-85-180.sslip.io` and OIDC discovery advertising `https://...` endpoints on port 443.
Actual: any browser following the OIDC redirect hits `http://...:8080/` (closed externally) and times out.

## Root cause

`charts/kagenti-deps/templates/keycloak-k8s.yaml`, lines 241–242, inside the `kagenti.keycloak.defaultEnv` named template:

```yaml
- name: KC_HOSTNAME
  value: {{ printf "http://keycloak.%s:8080/" .Values.domain | quote }}
```

This unconditionally hardcodes `http://`, port `8080`, and the cluster-internal-style hostname `keycloak.<domain>`. Other consumers of `publicUrl` in the same chart use it correctly (e.g. `keycloak-realm-init.yaml:217`, `authbridge-template-configmaps.yaml:26`, `ui.yaml:148`).

## Proposed fix

```yaml
- name: KC_HOSTNAME
  value: {{ .Values.keycloak.publicUrl | default (printf "http://keycloak.%s:8080/" .Values.domain) | quote }}
```

Backward-compatible (fallback unchanged) and matches the rest of the chart.

## Manual workaround

For operators on affected versions:

```sh
kubectl set env sts/keycloak -n keycloak \
  KC_HOSTNAME=https://keycloak.<your-public-domain>
```

After Keycloak restarts, OIDC discovery returns clean `https://...` URLs and login works.

## OpenShift scope

Scoped to the **non-OpenShift k8s StatefulSet path** (`keycloak-k8s.yaml`, gated `if (not .Values.openshift)`). The OpenShift path uses the RHBK operator (`keycloak-operand.yaml`) which sets `hostname.strict: false` with no explicit `hostname.hostname` — it relies on the OpenShift Route's hostname plus `KC_PROXY_HEADERS=forwarded` for issuer derivation. Different mechanism, not affected.

moonlight16:
Closes #1770.

## Summary

The `kagenti-deps` chart's Keycloak StatefulSet (non-OpenShift path) hardcoded `KC_HOSTNAME` to `http://keycloak.<domain>:8080/` regardless of `keycloak.publicUrl`. Every other consumer of `publicUrl` in the chart already honors it; only `KC_HOSTNAME` itself did not. As a result, Keycloak's OIDC discovery advertised `http://...:8080/...` URLs that browsers couldn't reach externally, breaking login on any cluster where Keycloak is fronted by an HTTPS ingress on a different host/port.

This change has the chart use `keycloak.publicUrl` when set, with the previous `http://keycloak.<domain>:8080/` as the fallback for users who don't set it.

## Before / after rendered manifest

`helm template kagenti-deps --set domain=example.com --set openshift=false --set keycloak.publicUrl=https://keycloak.example.com`:

Before:
```yaml
- name: KC_HOSTNAME
  value: "http://keycloak.example.com:8080/"
```

After:
```yaml
- name: KC_HOSTNAME
  value: "https://keycloak.example.com"
```

When `keycloak.publicUrl` is unset, output is identical to today (`http://keycloak.<domain>:8080/`).

## Test plan

- [x] `helm template` with `keycloak.publicUrl` set; `KC_HOSTNAME` is the HTTPS public URL.
- [x] `helm template` with `keycloak.publicUrl` unset; `KC_HOSTNAME` matches previous behavior.
- [x] `helm template --set openshift=true`; `keycloak-k8s.yaml` does not render (operand path unaffected).
- [ ] Live cluster (verified by issue reporter on a Kind-class cluster fronted by an HTTPS ALB): `kubectl get sts -n keycloak keycloak ... KC_HOSTNAME ...` returns the HTTPS URL; OIDC discovery's `authorization_endpoint` is `https://...`; end-to-end login flow through the kagenti UI completes without `ERR_CONNECTION_TIMED_OUT`.

## Scope

Non-OpenShift k8s StatefulSet path (`keycloak-k8s.yaml`, gated `if (not .Values.openshift)`). The OpenShift RHBK-operator path is unaffected — it uses `hostname.strict: false` with no explicit `hostname.hostname` and relies on the Route's hostname plus `KC_PROXY_HEADERS=forwarded` for issuer derivation. Different mechanism, no fix needed there.

### DIFF
--- a/charts/kagenti-deps/tests/keycloak-k8s_test.yaml
+++ b/charts/kagenti-deps/tests/keycloak-k8s_test.yaml
@@ -31,7 +31,7 @@ tests:
           path: spec.template.spec.containers[0].env
           content:
             name: KC_HOSTNAME
-            value: "http://keycloak.localtest.me:8080/"
+            value: "http://keycloak.localtest.me:8080"
         documentIndex: 2
       - contains:
           path: spec.template.spec.containers[0].env
@@ -112,7 +112,7 @@ tests:
           path: spec.template.spec.containers[0].env
           content:
             name: KC_HOSTNAME
-            value: "http://keycloak.localtest.me:8080/"
+            value: "http://keycloak.localtest.me:8080"
         documentIndex: 2
 
   - it: appends new env var from extraEnvVars
--- a/charts/kagenti-deps/templates/keycloak-k8s.yaml
+++ b/charts/kagenti-deps/templates/keycloak-k8s.yaml
@@ -239,7 +239,7 @@ extraEnvVars with the same name override these defaults.
 - name: KC_HOSTNAME_STRICT
   value: "false"
 - name: KC_HOSTNAME
-  value: {{ printf "http://keycloak.%s:8080/" .Values.domain | quote }}
+  value: {{ .Values.keycloak.publicUrl | default (printf "http://keycloak.%s:8080/" .Values.domain) | quote }}
 - name: KC_HOSTNAME_BACKCHANNEL_DYNAMIC
   value: "true"
 - name: KC_HEALTH_ENABLED


charts/kagenti-deps/tests/keycloak-k8s_test.yaml
```diff
@@ -112,7 +112,7 @@ tests:
           path: spec.template.spec.containers[0].env
           content:
             name: KC_HOSTNAME
-            value: "http://keycloak.localtest.me:8080/"
+            value: "http://keycloak.localtest.me:8080"
```
pdettori, line 115: **suggestion**: After this PR the default `KC_HOSTNAME` becomes `http://keycloak.localtest.me:8080` (no trailing slash, sourced from `publicUrl` in values.yaml). This `notContains` still checks for the old value **with** trailing slash, so it trivially passes regardless of whether `extraEnvVars` actually removes the default.

Consider updating to:
```yaml
            value: "http://keycloak.localtest.me:8080"
```
…so the assertion meaningfully validates the override mechanism.

pdettori (approved):
Clean, minimal fix aligning `KC_HOSTNAME` with how the rest of the chart already consumes `keycloak.publicUrl`. Single commit, properly signed off, CI all green.

One minor suggestion on test coverage below — not blocking.

PR #1774, opened 2026-06-01
fix(kagenti): trigger pod rollouts on config changes via checksum/config

### LINKED ISSUE #1773: kagenti: kagenti-backend and kagenti-ui Deployments don't roll out on config changes (helm upgrade leaves stale pods serving old auth config)
## Summary

**Symptom**: After running `helm upgrade kagenti` with values that should change Kagenti's authentication or dashboard configuration (Keycloak URL, realm, domain, etc.), users **can no longer log into the Kagenti UI**. The browser is redirected to the OLD Keycloak URL — the one that was active *before* the upgrade. Login either fails (404 / "realm not found") or hangs and times out (`ERR_CONNECTION_TIMED_OUT`), depending on whether the old hostname still resolves.

**Cause**: The `kagenti-backend` and `kagenti-ui` Deployments source most of their runtime config from the in-chart `kagenti-ui-config` ConfigMap and the runtime-generated `kagenti-ui-oauth-secret` Secret. Neither Deployment's pod template embeds a `checksum/config` annotation. As a result, `helm upgrade` correctly rewrites those resources, but the running pods don't restart — they keep their old configuration cached in-process and keep serving the old URLs from `/api/v1/auth/config` until an operator manually runs `kubectl rollout restart`.

**Fix**: The standard Helm "automatically roll deployments" pattern: hash the relevant inputs into a pod-template annotation so any meaningful change forces a fresh ReplicaSet.

## Background — why `helm upgrade` looks like it worked but didn't

Kubernetes only rolls out a Deployment when its `spec.template` differs between revisions. If the only thing that changed is a referenced ConfigMap or Secret (which are themselves separate resources, referenced by name from the Deployment), the pod template is byte-identical and Kubernetes does nothing — the existing pods keep running with their old environment.

The Kagenti chart hits this whenever a value flows into the `kagenti-ui-config` ConfigMap or affects what `ui-oauth-secret-job` writes into the `kagenti-ui-oauth-secret` Secret. Triggers in practice:

- Changing `domain` (flows into `DOMAIN_NAME` and the Keycloak/Kagenti hostnames the OAuth job writes back).
- Changing `keycloak.publicUrl` (flows into `AUTH_ENDPOINT`, `REDIRECT_URI`, `KEYCLOAK_CONSOLE_URL`).
- Changing `keycloak.realm` (flows into all OIDC URLs).
- Rotating the `kagenti-ui-config` dashboard URLs (`MLFLOW_DASHBOARD_URL`, etc.).
- Anything else that gets re-rendered into the ConfigMap or causes the OAuth job to rewrite the Secret.

In all these cases:

1. Helm correctly updates the ConfigMap and Secret.
2. Helm reports `Upgrade complete`. Operators reasonably assume the upgrade took effect.
3. The pods keep running with stale env vars baked into the process at startup time.
4. The first user-visible symptom is usually a broken login flow: `/api/v1/auth/config` returns the old Keycloak URL, the browser is redirected to a hostname that no longer works (or never worked), login fails or hangs.
5. `kubectl rollout restart deploy/kagenti-backend deploy/kagenti-ui -n kagenti-system` fixes it. The `kagenti-ui` frontend specifically also benefits, because it caches the backend's `/api/v1/auth/config` response on first load.

The standard Kubernetes/Helm fix is to add a `checksum/config` annotation to the Deployment's pod template, computed as the sha256 of the inputs. ConfigMap content changes → checksum changes → pod template differs → rollout. Documented in [Helm's chart tips](https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments). The Kagenti chart doesn't currently use this pattern (`grep -rn "checksum/config\|sha256sum" charts/` returns no matches).

## Reproduction

The simplest concrete trigger is a `keycloak.publicUrl` change — but the same shape reproduces with any of the values listed above.

```sh
# Initial install at one Keycloak public URL.
helm install kagenti . -n kagenti-system -f values.yaml \
  --set keycloak.publicUrl=https://keycloak.cluster-a.example.com

# Wait for everything to come up, log in via the UI -> works.

# Change the public URL (e.g. you moved Keycloak behind a different ingress).
helm upgrade kagenti . -n kagenti-system --reuse-values \
  --set keycloak.publicUrl=https://keycloak.cluster-b.example.com

# ConfigMap and Secret have new values:
kubectl get cm -n kagenti-system kagenti-ui-config -o jsonpath='{.data.KEYCLOAK_CONSOLE_URL}'
# https://keycloak.cluster-b.example.com/admin/...

# But pods are unchanged:
kubectl get pods -n kagenti-system -l app.kubernetes.io/name=kagenti-backend
# NAME                          READY   AGE
# kagenti-backend-abc-xyz       1/1     24h     # <-- old pod, old env

# /api/v1/auth/config still returns the OLD Keycloak URL, browsers redirect
# there, login fails:
curl -sk https://kagenti-ui.<domain>/api/v1/auth/config
# {"keycloak_url": "https://keycloak.cluster-a.example.com", ...}
```

## Root cause

`charts/kagenti/templates/ui.yaml`:

- `kagenti-backend` Deployment, lines 42–263. The pod template at lines 57–263 has no `metadata.annotations` block. Its env vars come from `kagenti-ui-config` and `kagenti-ui-oauth-secret`, neither of which is reflected in the pod template — so Kubernetes never sees a diff.
- `kagenti-ui` Deployment, lines 287–344. Same — no annotations block. The frontend has no values-driven env vars, but caches the backend's `/api/v1/auth/config` response and benefits from being rolled when configuration changes.

## Proposed fix

Add to both Deployments' `spec.template.metadata`:

```yaml
metadata:
  annotations:
    checksum/config: {{ include (print $.Template.BasePath "/ui.yaml") . | sha256sum }}
    checksum/oauth-config: {{ printf "%s|%s|%s|%s" .Values.keycloak.url .Values.keycloak.publicUrl .Values.keycloak.realm .Values.ui.namespace | sha256sum }}
  labels:
    ...
```

The `ui.yaml`-self-hash is a slight over-hash (the file also contains the Deployments themselves), but it remains stable across no-op renders. A tighter alternative is to extract the ConfigMap into its own template file and hash only that — happy to refactor if maintainers prefer.

`checksum/oauth-config` hashes the values that determine what `ui-oauth-secret-job.yaml` writes into `kagenti-ui-oauth-secret`. We can't hash a Secret that doesn't exist yet at install time, so we hash the inputs.

## Manual workaround

```sh
kubectl rollout restart deploy/kagenti-backend deploy/kagenti-ui -n kagenti-system
```

Operators must remember to run this after every `helm upgrade` that touches values flowing through the ConfigMap or OAuth Secret.

## Scope note

`keycloak.publicUrl` and related values are also consumed by the AuthBridge sidecar's ConfigMaps (`charts/kagenti/templates/authbridge-template-configmaps.yaml`) and per-namespace agent ConfigMaps. Those workloads probably need similar treatment but are out of scope for this issue — addressing them would be a follow-up.

moonlight16:
Closes #1773.

## Summary

Adds `checksum/config` annotations to the `kagenti-backend` and `kagenti-ui` Deployments so that `helm upgrade` triggers a rollout whenever the in-chart `kagenti-ui-config` ConfigMap or the inputs that determine the runtime-generated `kagenti-ui-oauth-secret` Secret change.

Today, those Deployments' pod templates have no annotations block. Since their env vars come from a referenced ConfigMap and Secret (not embedded directly in the pod template), Kubernetes sees a byte-identical pod template across `helm upgrade` revisions and doesn't roll out new pods. The result is that values like `keycloak.publicUrl`, `keycloak.realm`, and dashboard URLs land in the cluster but the running pods keep their old configuration cached in-process. `/api/v1/auth/config` then serves stale URLs and login breaks.

This is the standard Helm "automatically roll deployments" pattern (see [Helm chart tips](https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments)).

## Before / after rendered manifest

`helm template kagenti --set ui.domainName=example.com --set openshift=false` (relevant fragment for the `kagenti-backend` Deployment):

Before:
```yaml
spec:
  template:
    metadata:
      labels:
        app.kubernetes.io/name: kagenti-backend
        ...
    spec:
      containers: ...
```

After:
```yaml
spec:
  template:
    metadata:
      annotations:
        checksum/config: 307e5279e5dcccc276a43e63f2c3dd21fad8b9eb62149e0f97cf954eac66df31
        checksum/oauth-config: 5f1a650fad76b3719d9bd6c332eb702d294209563eaa494d32a7ccde606740da
      labels:
        app.kubernetes.io/name: kagenti-backend
        ...
    spec:
      containers: ...
```

Re-rendering with `--set ui.domainName=other.example.com` produces different `checksum/config` values, so the pod template now differs across upgrades that touch the domain (or any other ConfigMap/Secret input).

## Test plan

- [x] `helm template` twice with identical values → identical `checksum/*` values (idempotent).
- [x] `helm template` with two different `ui.domainName` values → different `checksum/config` (sensitive to relevant changes).
- [ ] Live cluster:
  - Install kagenti at domain A. Log in. Confirm `/api/v1/auth/config` returns A's URLs.
  - `helm upgrade --reuse-values -f domain-B.yaml` (domain change only).
  - Without manual intervention, `kubectl get pods` shows new `kagenti-backend` and `kagenti-ui` pods (recently created), and `/api/v1/auth/config` returns B's URLs.
  - Login through the UI completes successfully against the new Keycloak hostname.
- [ ] Confirm no spurious rollouts when running `helm upgrade --reuse-values` with no value changes.

## Notes

- `checksum/config` hashes the values that flow into the `kagenti-ui-config` ConfigMap (`ui.domainName`, `ui.api.hostname`, `ui.namespace`, `ui.configmap`, `keycloak.realm`, and the phoenix/mlflow component toggles). A self-`include` of `ui.yaml` would have been an even tighter "hash the rendered ConfigMap" approach, but Helm's template engine treats that as recursion when the file also contains the Deployment doing the include — hashing the inputs avoids that and is what the official chart-tips example does for ConfigMaps in the same template file.
- `checksum/oauth-config` hashes the values that determine what `ui-oauth-secret-job.yaml` writes into `kagenti-ui-oauth-secret`. We can't directly hash a Secret that doesn't exist yet at install time, so we hash the inputs.
- Other workloads (AuthBridge sidecars, agent namespaces) consume similar config from ConfigMaps and would benefit from the same pattern. Out of scope here; this PR matches the user-observable workaround (`kubectl rollout restart deploy/kagenti-backend deploy/kagenti-ui`).

### DIFF
--- a/charts/kagenti/templates/ui.yaml
+++ b/charts/kagenti/templates/ui.yaml
@@ -56,6 +56,11 @@ spec:
       app.kubernetes.io/name: kagenti-backend
   template:
     metadata:
+      annotations:
+        # Force pod restart whenever the in-chart ConfigMap or OAuth-secret inputs change.
+        # See https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments
+        checksum/config: {{ printf "%s|%s|%s|%s|%s|%t|%t" .Values.ui.domainName .Values.ui.api.hostname .Values.ui.namespace .Values.ui.configmap .Values.keycloak.realm .Values.components.phoenix.enabled .Values.components.mlflow.enabled | sha256sum }}
+        checksum/oauth-config: {{ printf "%s|%s|%s|%s" .Values.keycloak.url .Values.keycloak.publicUrl .Values.keycloak.realm .Values.ui.namespace | sha256sum }}
       labels:
         app.kubernetes.io/name: kagenti-backend
         app.kubernetes.io/component: backend
@@ -300,6 +305,11 @@ spec:
       app.kubernetes.io/name: kagenti-ui
   template:
     metadata:
+      annotations:
+        # Force pod restart whenever the in-chart ConfigMap or OAuth-secret inputs change.
+        # See https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments
+        checksum/config: {{ printf "%s|%s|%s|%s|%s|%t|%t" .Values.ui.domainName .Values.ui.api.hostname .Values.ui.namespace .Values.ui.configmap .Values.keycloak.realm .Values.components.phoenix.enabled .Values.components.mlflow.enabled | sha256sum }}
+        checksum/oauth-config: {{ printf "%s|%s|%s|%s" .Values.keycloak.url .Values.keycloak.publicUrl .Values.keycloak.realm .Values.ui.namespace | sha256sum }}
       labels:
         app.kubernetes.io/name: kagenti-ui
         app.kubernetes.io/component: frontend


pdettori (approved):
Correct application of the standard Helm "automatically roll deployments" pattern. Verified:

- `checksum/config` covers all values that flow into the `kagenti-ui-config` ConfigMap (domainName, api.hostname, namespace, configmap, realm, phoenix/mlflow toggles)
- `checksum/oauth-config` covers the inputs to `ui-oauth-secret-job.yaml` (keycloak.url, publicUrl, realm, ui.namespace)
- Both Deployments get identical annotations (they consume the same ConfigMap/Secret)
- Idempotent: same values → same checksums → no spurious rollouts
- All CI passing (Helm lint, chart tests, Deploy & Test)

Areas reviewed: Helm template (charts/kagenti/templates/ui.yaml), cross-referenced with values.yaml and ui-oauth-secret-job.yaml inputs.
Commit: 1 commit, DCO signed, conventional format.

PR #1779, opened 2026-06-01
feat(kagenti-deps): Add Tekton and Shipwright support for vanilla Kubernetes

### LINKED ISSUE #1778: Support Tekton and Shipwright on vanilla Kubernetes clusters
## Problem

The kagenti-deps Helm chart currently only installs Tekton Pipelines and Shipwright Build on OpenShift clusters via OLM operators. Vanilla Kubernetes clusters are missing these critical components, which prevents Kagenti's build functionality from working.

## Current Behavior

- On OpenShift (openshift=true): Tekton and Shipwright are installed via operators
- On vanilla k8s (openshift=false): No Tekton or Shipwright installation occurs
- Result: Build features fail on non-OpenShift clusters

## Proposed Solution

Add conditional templates to kagenti-deps that install upstream Tekton and Shipwright when openshift=false:

1. **tekton-pipelines.yaml** - Apply Tekton Pipelines latest release manifest
   - URL: https://storage.googleapis.com/tekton-releases/pipeline/latest/release.yaml
   - Only when: components.tekton.enabled=true AND openshift=false

2. **shipwright-build.yaml** - Install Shipwright Build Controller + setup
   - Apply Shipwright manifests
   - Configure build strategies
   - Only when: components.shipwright.enabled=true AND openshift=false

3. **Installation ordering** - Use Helm hooks to ensure Tekton installs before Shipwright (Shipwright depends on Tekton)

## Implementation Notes

- Use pattern: `{{- if and .Values.components.tekton.enabled (not .Values.openshift) }}`
- Existing OpenShift operator templates remain unchanged
- Helm pre-install/pre-upgrade hooks for proper ordering
- Test with `helm template` to verify rendering

## Benefits

- Enables Kagenti build features on any Kubernetes cluster
- Maintains existing OpenShift operator-based approach
- No breaking changes to existing deployments

moonlight16:
# Add vanilla-Kubernetes support for Tekton + Shipwright via shared bash library

## Summary

Adds support for installing Kagenti's optional Tekton + Shipwright build stack on **vanilla Kubernetes clusters** (K3s, kubeadm, EKS, GKE, AKS, IBM Cloud, etc.) — not just Kind or OpenShift. This unblocks Kagenti deployments on non-Kind, non-OCP environments without requiring users to reimplement the install logic out-of-band.

## Background

The first revision of this PR added Helm chart templates (`templates/tekton-pipelines.yaml`, `templates/shipwright-build.yaml`) that installed Tekton/Shipwright via in-cluster install Jobs when `openshift=false`. Reviewer @pdettori flagged six concerns ([review thread](https://github.com/kagenti/kagenti/pull/1779)):

1. **must-fix:** Conflict with the existing Kind installer (which already pins versions and disables the chart components).
2. **must-fix:** Unpinned `latest` versions are a reproducibility hazard.
3. **must-fix:** Bound to `cluster-admin` — overly broad.
4. **must-fix:** Missing cert-manager TLS setup for Shipwright webhook.
5. **suggestion:** `hook-succeeded` delete policy hurts debugging.
6. **suggestion:** No uninstall path.

In the discussion, @pdettori suggested **"extending the kind installer to work with vanilla (pre-existing) Kubernetes as well"** rather than adding a parallel install path inside the Helm chart. This commit takes that approach.

## Approach (Path A from review discussion)

Extract the Tekton/Shipwright install logic from `scripts/kind/setup-kagenti.sh` into a shared bash library, then add a thin vanilla-Kubernetes entry point that reuses it. The Helm chart templates added in the previous revision are removed.

### New layout

```
scripts/
├── lib/
│   └── install-deps.sh          # shared installer functions
├── kind/
│   ├── setup-kagenti.sh         # creates Kind cluster, sources lib
│   └── cleanup-kagenti.sh       # unchanged
└── k8s/
    ├── setup-kagenti.sh         # vanilla-k8s entry point (new)
    └── cleanup-kagenti.sh       # vanilla-k8s cleanup (new)
```

### Shared library functions

`scripts/lib/install-deps.sh` provides:

- `install_tekton [version]` — applies the upstream Tekton release manifest at a pinned version (default `v0.66.0`).
- `install_shipwright [version]` — applies the upstream Shipwright release manifest at a pinned version (default `v0.14.0`), then runs `configure_shipwright_webhook_tls`, applies sample build strategies, and installs `buildah-insecure-push`.
- `configure_shipwright_webhook_tls` — wires up cert-manager (self-signed ClusterIssuer → CA Certificate → Issuer → webhook Certificate with DNS SANs → CRD CA-injection annotations → webhook restart).
- `install_buildah_insecure_push_strategy` — applies the buildah-insecure-push ClusterBuildStrategy.

The lib is **stateless and dependency-injected**: it reads `DRY_RUN` from the caller's scope and uses the caller's `log_*`, `run_cmd`, and `_wait_deployment_ready` helpers. No globals defined in the lib itself.

### Kind entry point

`scripts/kind/setup-kagenti.sh` now:
- Sources `../lib/install-deps.sh` once at startup.
- Replaces the inline Tekton install (Step 3b) with `install_tekton "$TEKTON_VERSION"`.
- Replaces the inline Shipwright install + cert-manager TLS + build strategies (Step 6b, ~163 lines) with `install_shipwright "$SHIPWRIGHT_VERSION"`.
- Honors `KAGENTI_SETUP_FLAVOR` env var to relabel the banner when invoked via the k8s entry point.

Behavior is unchanged. Verified with `bash scripts/kind/setup-kagenti.sh --dry-run --skip-cluster --with-builds`.

### Vanilla-k8s entry point

`scripts/k8s/setup-kagenti.sh`:
- Verifies `kubectl` can reach a cluster (clear error if not).
- Sets `KAGENTI_SETUP_FLAVOR=k8s` so the banner reads "(Kubernetes)".
- `exec`s the Kind entry point with `--skip-cluster` forced.
- Forwards all other flags (`--with-*`, `--skip-*`, `--domain`, etc.) unchanged.

`scripts/k8s/cleanup-kagenti.sh`:
- Symmetric wrapper around `scripts/kind/cleanup-kagenti.sh`.
- Rejects `--destroy-cluster` with a clear error (Kind-only flag).

### Removed

- `charts/kagenti-deps/templates/tekton-pipelines.yaml`
- `charts/kagenti-deps/templates/shipwright-build.yaml`

## How this addresses each review comment

| # | Comment | Resolution |
|---|---------|-----------|
| 1 | Conflict with Kind installer | **Eliminated.** Single install path. Both Kind and k8s entry points call the same shared lib. The Kind installer's existing `--set components.tekton.enabled=false --set components.shipwright.enabled=false` is unchanged. No risk of double-install at different versions. |
| 2 | Unpinned `latest` | **Fixed.** Lib functions accept a version argument. Defaults match the Kind installer's existing pinned constants (`TEKTON_VERSION=v0.66.0`, `SHIPWRIGHT_VERSION=v0.14.0`). Callers can override per call. |
| 3 | `cluster-admin` overly broad | **Eliminated by removing in-cluster Jobs.** Bash scripts run with the operator's local kubectl credentials. No ServiceAccounts, ClusterRoleBindings, or third-party kubectl image with cluster-admin to compromise. |
| 4 | Missing cert-manager TLS | **Fixed.** `configure_shipwright_webhook_tls` is now part of `install_shipwright`. The vanilla-k8s entry point inherits it for free: self-signed ClusterIssuer → CA Certificate → Issuer → webhook Certificate with DNS SANs → CRD CA-injection annotations → webhook restart. Identical to what the Kind installer already does. |
| 5 | `hook-succeeded` debugging | **N/A.** No Helm hooks remain. Failures halt the script via `set -e` and the user sees them immediately on stderr. |
| 6 | No uninstall path | **Fixed.** `scripts/k8s/cleanup-kagenti.sh` mirrors `scripts/kind/cleanup-kagenti.sh`, providing symmetric uninstall: deletes CRs, uninstalls Helm releases, deletes PVCs, clears stale Istio CA artifacts, removes namespaces. |

## Testing

- ✅ `bash scripts/kind/setup-kagenti.sh --dry-run --skip-cluster --with-builds` — dry-run output unchanged from baseline; Tekton + Shipwright steps now flow through the lib.
- ✅ `bash scripts/k8s/setup-kagenti.sh --dry-run --with-builds` — banner says "(Kubernetes)", same install steps.
- ✅ Live install on a fresh Kind cluster via the **k8s entry point** with `--with-builds`. End-to-end result:
  - All Kagenti pods Running across `kagenti-system`, `keycloak`, `cert-manager`, `istio-system`, `gateway-system`, `cr-system`, `tekton-pipelines`, `shipwright-build`.
  - `tekton-pipelines-controller`, `tekton-pipelines-webhook`, `tekton-events-controller` Running.
  - `shipwright-build-controller`, `shipwright-build-webhook` Running with cert-manager-issued TLS.
  - All 9 ClusterBuildStrategies registered, including `buildah-insecure-push`.
  - "Skipping Kind registry-DNS step on vanilla Kubernetes" message confirmed; the Kind-only `docker exec <control-plane>` containerd config block is correctly bypassed.
- 📝 `bash -n` syntax check passes for all scripts and the lib.

## Optional follow-up: full lib extraction

This PR keeps the scope minimal — only Tekton, Shipwright, and the build-strategy install moved into the lib. The other cluster-agnostic install steps (cert-manager, Istio, SPIRE, Gateway API CRDs, Helm chart installs, MCP Gateway, Kuadrant) still live inline in `scripts/kind/setup-kagenti.sh` and are reachable from the k8s entry point via `--skip-cluster` passthrough.

If desired, a follow-up PR could extract those steps into the lib as well, leaving both entry points as thin orchestrators. Happy to do that here or in a follow-up — let me know which you'd prefer.

## Out of scope for this PR

- No changes to `scripts/ocp/setup-kagenti.sh`.
- No changes to any Helm chart values, templates, or behavior on OpenShift.
- No changes to component versions — the lib uses the same pinned versions the Kind installer already uses.

Closes #1778

Signed-off-by: Jeremy Cohn <jacohn1@gmail.com>

### DIFF
--- a/scripts/k8s/cleanup-kagenti.sh
+++ b/scripts/k8s/cleanup-kagenti.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+# ============================================================================
+# KAGENTI CLEANUP FOR VANILLA KUBERNETES
+# ============================================================================
+# Reverses the install performed by scripts/k8s/setup-kagenti.sh: uninstalls
+# Helm releases, deletes namespaces, and clears stale Istio CA artifacts.
+# Does NOT delete the underlying cluster — that's the operator's call.
+#
+# Usage:
+#   scripts/k8s/cleanup-kagenti.sh
+#
+# Prerequisites: kubectl context pointing at the target cluster.
+# ============================================================================
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/../.." && pwd)"
+KIND_CLEANUP="$REPO_ROOT/scripts/kind/cleanup-kagenti.sh"
+
+if [[ ! -x "$KIND_CLEANUP" ]]; then
+  echo "Error: $KIND_CLEANUP not found or not executable." >&2
+  exit 1
+fi
+
+if ! command -v kubectl >/dev/null 2>&1; then
+  echo "Error: kubectl not found in PATH." >&2
+  exit 1
+fi
+if ! kubectl cluster-info >/dev/null 2>&1; then
+  echo "Error: kubectl cannot reach a cluster." >&2
+  exit 1
+fi
+
+# Reject the Kind-only flag explicitly so users don't get confused by silent
+# passthrough.
+for arg in "$@"; do
+  if [[ "$arg" == "--destroy-cluster" ]]; then
+    echo "Error: --destroy-cluster is Kind-only. Delete the cluster manually if needed." >&2
+    exit 1
+  fi
+done
+
+exec "$KIND_CLEANUP" "$@"
--- a/scripts/k8s/setup-kagenti.sh
+++ b/scripts/k8s/setup-kagenti.sh
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+# ============================================================================
+# KAGENTI PLATFORM SETUP FOR VANILLA KUBERNETES
+# ============================================================================
+# Installs the Kagenti stack onto a pre-existing Kubernetes cluster (K3s,
+# kubeadm, EKS, GKE, AKS, or any other distribution that exposes a working
+# kubectl context).
+#
+# Same component set as scripts/kind/setup-kagenti.sh — the only difference
+# is that cluster bring-up is the operator's responsibility, not this
+# script's. The shared installer logic lives in scripts/lib/install-deps.sh.
+#
+# Usage:
+#   scripts/k8s/setup-kagenti.sh                          # Core only
+#   scripts/k8s/setup-kagenti.sh --with-all               # Everything
+#   scripts/k8s/setup-kagenti.sh --with-builds            # Tekton + Shipwright
+#   scripts/k8s/setup-kagenti.sh --domain example.com     # Custom hostname domain
+#
+# Prerequisites:
+#   - kubectl context pointing at the target cluster
+#   - helm v3
+#   - cluster has enough capacity for the requested components
+#
+# Run scripts/k8s/setup-kagenti.sh --help for the full flag set.
+# ============================================================================
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/../.." && pwd)"
+KIND_SETUP="$REPO_ROOT/scripts/kind/setup-kagenti.sh"
+
+if [[ ! -x "$KIND_SETUP" ]]; then
+  echo "Error: $KIND_SETUP not found or not executable." >&2
+  echo "       The Kubernetes entry point delegates installation to the shared" >&2
+  echo "       implementation in scripts/kind/setup-kagenti.sh + scripts/lib/install-deps.sh." >&2
+  exit 1
+fi
+
+# Verify a usable kubectl context before delegating. Fail fast with a clear
+# message rather than letting a missing kubeconfig bubble out of helm or
+# kubectl partway through the install.
+if ! command -v kubectl >/dev/null 2>&1; then
+  echo "Error: kubectl not found in PATH." >&2
+  exit 1
+fi
+if ! kubectl cluster-info >/dev/null 2>&1; then
+  echo "Error: kubectl cannot reach a cluster." >&2
+  echo "       Set KUBECONFIG or run 'kubectl config use-context <name>' to point at" >&2
+  echo "       the target cluster, then re-run this script." >&2
+  exit 1
+fi
+
+# Reject Kind-only flags with a clear error rather than letting them fail
+# deep in the script when they hit a `docker exec <kind-container>` call.
+for arg in "$@"; do
+  case "$arg" in
+    --build-images|--preload-images)
+      echo "Error: $arg is Kind-only (loads images into the Kind node)." >&2
+      echo "       On vanilla Kubernetes, push images to your cluster's" >&2
+      echo "       registry and reference them by tag instead." >&2
+      exit 1
+      ;;
+  esac
+done
+
+# Forward the user's flags to the shared setup, but force --skip-cluster so
+# the Kind-specific create/preload steps are skipped. KAGENTI_SETUP_FLAVOR
+# changes the banner from "(Kind)" to "(Kubernetes)" so users running this
+# entry point see an accurate label, and gates the Kind-only registry-DNS
+# configuration block.
+KAGENTI_SETUP_FLAVOR=k8s exec "$KIND_SETUP" --skip-cluster "$@"
--- a/scripts/kind/setup-kagenti.sh
+++ b/scripts/kind/setup-kagenti.sh
@@ -185,6 +185,11 @@ except Exception:
   fi
 }
 
+# ── Shared dependency installers (Tekton, Shipwright, build strategies) ─────
+# Sourced after log_*/run_cmd are defined; the lib reads DRY_RUN from this
+# scope and uses the helpers above.
+. "$REPO_ROOT/scripts/lib/install-deps.sh"
+
 # Load a single image into the Kind node via docker save piped to ctr import.
 # Avoids 'kind load docker-image' failures (e.g. "failed to detect containerd
 # snapshotter") on WSL2 and Rancher Desktop.
@@ -321,7 +326,10 @@ START_SECONDS=$SECONDS
 
 echo ""
 echo "============================================"
-echo "  Kagenti Platform Setup (Kind)"
+case "${KAGENTI_SETUP_FLAVOR:-kind}" in
+  k8s) echo "  Kagenti Platform Setup (Kubernetes)" ;;
+  *)   echo "  Kagenti Platform Setup (Kind)" ;;
+esac
 echo "============================================"
 echo ""
 echo "  Cluster:       $CLUSTER_NAME"
@@ -589,10 +597,7 @@ fi
 # ============================================================================
 if $WITH_BUILDS; then
   log_info "Step 3b: Tekton"
-  log_info "Installing Tekton ${TEKTON_VERSION}..."
-  run_cmd kubectl apply --server-side \
-    -f "https://storage.googleapis.com/tekton-releases/pipeline/previous/${TEKTON_VERSION}/release.yaml"
-  log_success "Tekton applied"
+  install_tekton "$TEKTON_VERSION"
   echo ""
 fi
 
@@ -741,7 +746,14 @@ log_success "kagenti-deps installed"
 echo ""
 
 # ── Configure Kind node to reach in-cluster container registry ──────────────
-if $WITH_BUILDS; then
+# Kind-only: writes /etc/hosts and /etc/containerd/certs.d on the Kind
+# control-plane container so kubelet can pull images from the in-cluster
+# registry by its cluster-DNS hostname. Vanilla Kubernetes operators need
+# to configure their nodes via the distribution's normal mechanism (e.g.
+# /etc/rancher/k3s/registries.yaml on K3s, /etc/containerd/config.toml +
+# restart on kubeadm); this script intentionally doesn't try to SSH into
+# nodes or guess the distro. Skip the block under the k8s flavor.
+if $WITH_BUILDS && [[ "${KAGENTI_SETUP_FLAVOR:-kind}" != "k8s" ]]; then
   REGISTRY_NAME="registry"
   REGISTRY_NS="cr-system"
   REGISTRY_HOST="${REGISTRY_NAME}.${REGISTRY_NS}.svc.cluster.local"
@@ -774,6 +786,14 @@ TOML
     fi
   fi
   echo ""
+elif $WITH_BUILDS; then
+  log_info "Skipping Kind registry-DNS step on vanilla Kubernetes."
+  log_info "  To pull images from the in-cluster registry by its cluster-DNS"
+  log_info "  hostname (registry.cr-system.svc.cluster.local:5000), configure"
+  log_info "  containerd on each node via your distribution's mechanism:"
+  log_info "    K3s:     /etc/rancher/k3s/registries.yaml + restart k3s/k3s-agent"
+  log_info "    kubeadm: /etc/containerd/config.toml + systemctl restart containerd"
+  echo ""
 fi
 
 # ============================================================================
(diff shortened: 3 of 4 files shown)


charts/kagenti-deps/templates/tekton-pipelines.yaml
```diff
@@ -0,0 +1,85 @@
+{{- if and .Values.components.tekton.enabled (not .Values.openshift) }}
```
pdettori: **must-fix: Conflict with Kind installer**

The existing Kind installer (`scripts/kind/setup-kagenti.sh`) already installs Tekton and Shipwright with:
- Pinned versions (`TEKTON_VERSION=v0.66.0`, `SHIPWRIGHT_VERSION=v0.14.0`)
- cert-manager TLS integration for Shipwright webhook
- Custom `buildah-insecure-push` ClusterBuildStrategy
- Registry DNS configuration

And it **explicitly disables** these chart components:
```bash
--set "components.tekton.enabled=false"
--set "components.shipwright.enabled=false"
```

This means the PR templates will never activate in the Kind workflow. If they were enabled, they'd double-install at a potentially different version (`latest` vs `v0.66.0`). Need to clarify the intended relationship with the existing installer.

charts/kagenti-deps/templates/tekton-pipelines.yaml
```diff
@@ -0,0 +1,85 @@
...
+      restartPolicy: OnFailure
+      containers:
+      - name: installer
+        image: {{ .Values.common.kubectlImage }}
+        command:
+        - /bin/sh
+        - -c
+        - |
+          set -e
+          echo "Installing Tekton Pipelines from upstream release..."
```
pdettori: **must-fix: Unpinned `latest` is a production hazard**

Using `tekton-releases/pipeline/latest/release.yaml` means:
- Builds are not reproducible
- A Tekton breaking change can silently break the platform
- Version skew between environments (Kind installer pins v0.66.0)
- Every `helm upgrade` could jump multiple major versions

Should pin via a values parameter:
```yaml
# values.yaml
components:
  tekton:
    version: "v0.66.0"
```

Then use:
```
https://storage.googleapis.com/tekton-releases/pipeline/previous/{{ .Values.components.tekton.version }}/release.yaml
```

Same issue applies to the Shipwright template.

charts/kagenti-deps/templates/tekton-pipelines.yaml
```diff
@@ -0,0 +1,85 @@
...
+  name: tekton-installer
+  labels:
+    {{- include "kagenti.labels" . | nindent 4 }}
+  annotations:
+    helm.sh/hook: pre-install,pre-upgrade
+    helm.sh/hook-weight: "-5"
+    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
```
pdettori: **must-fix: `cluster-admin` is overly broad**

Binding to `cluster-admin` grants unrestricted access during Job execution. While Tekton installation does require creating CRDs, namespaces, and cluster-scoped resources, a purpose-scoped ClusterRole (e.g., create/patch/get on CRDs, Namespaces, Deployments, Services, ClusterRoles, ClusterRoleBindings, ServiceAccounts) would significantly reduce blast radius.

If the Job image were compromised or the manifest URL MITM'd, `cluster-admin` enables full cluster takeover. The `common.kubectlImage` is already a third-party image pulling a manifest from an external URL — this is a supply-chain risk surface that warrants least-privilege.

Same issue in `shipwright-build.yaml`.

charts/kagenti-deps/templates/shipwright-build.yaml
```diff
@@ -0,0 +1,91 @@
...
+      containers:
+      - name: installer
+        image: {{ .Values.common.kubectlImage }}
+        command:
+        - /bin/sh
+        - -c
+        - |
+          set -e
+          echo "Installing Shipwright Build Controller..."
+          kubectl apply --filename https://github.com/shipwright-io/build/releases/latest/download/release.yaml --server-side=true
```
pdettori: **must-fix: Missing cert-manager TLS setup for Shipwright webhook**

The Kind installer (`scripts/kind/setup-kagenti.sh:651-727`) performs extensive cert-manager integration that this template omits:
1. Creates a `shipwright-selfsigned-issuer` ClusterIssuer
2. Creates a CA Certificate (`shipwright-ca`)
3. Creates a webhook Certificate with proper DNS SANs
4. Annotates Shipwright CRDs for CA injection
5. Restarts the webhook deployment to pick up TLS

Without this, the Shipwright admission webhook will not have valid TLS certificates and will reject or fail to validate `Build`/`BuildRun` resources. This is a blocking functional gap.

charts/kagenti-deps/templates/shipwright-build.yaml
```diff
@@ -0,0 +1,91 @@
...
+kind: Job
+metadata:
+  name: shipwright-build-installer
+  namespace: shipwright-build
+  labels:
+    {{- include "kagenti.labels" . | nindent 4 }}
+  annotations:
+    helm.sh/hook: pre-install,pre-upgrade
+    helm.sh/hook-weight: "10"
+    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
```
pdettori: **suggestion: `hook-succeeded` delete policy limits debuggability**

With `before-hook-creation,hook-succeeded`, a failed Job is retained until the next install/upgrade attempt (when `before-hook-creation` cleans it up). This is OK for debugging the *current* failure, but if the failure is transient (network timeout), the next `helm upgrade` will silently delete the failed Job before you can inspect it.

Consider keeping the current policy but adding a note in docs that `kubectl get jobs -n shipwright-build` should be checked on failure before retrying.

charts/kagenti-deps/templates/tekton-pipelines.yaml
```diff
@@ -0,0 +1,85 @@
...
+  by applying the upstream release manifest. On OpenShift, Tekton is installed
+  via the OpenShift Pipelines Operator (see tekton-operator.yaml).
+
+  The installation uses a Job with a pre-install/pre-upgrade hook to ensure
+  Tekton is available before Shipwright attempts to install.
+*/}}
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: tekton-pipelines
```
pdettori: **suggestion: No uninstall/cleanup path**

Namespaces have `helm.sh/resource-policy: keep`, Jobs/SA/CRB use `hook-delete-policy`. On `helm uninstall kagenti-deps`, the installed Tekton (CRDs, controllers, webhooks, RBAC) persists permanently with no automated cleanup.

The Kind installer handles cleanup in `scripts/kind/cleanup-kagenti.sh`. For the chart path, either:
1. Add a pre-delete hook Job that reverses the installation, or
2. Document the manual cleanup steps (`kubectl delete -f <release-url>`), or
3. Accept that uninstall is manual (but state this explicitly in values/README)

`kubectl apply` without `--prune` also means resources removed in future Tekton releases become orphans on upgrade.

pdettori (changes requested):
## Review Summary

This PR adds Tekton and Shipwright installation for vanilla Kubernetes via Helm hook Jobs. CI passes and the conditional rendering logic is correct. However, there are significant concerns around **conflict with the existing Kind installer**, **security (cluster-admin)**, and **fragility** (unpinned versions, no uninstall path, missing cert-manager TLS for Shipwright).

### Key Issues

1. **Kind installer conflict**: `scripts/kind/setup-kagenti.sh` already installs Tekton (v0.66.0) and Shipwright (v0.14.0) with pinned versions, cert-manager TLS, custom build strategies, AND explicitly sets `components.tekton.enabled=false` / `components.shipwright.enabled=false` when deploying the chart. This means the PR templates will never activate on Kind as currently configured. If enabled, they'd double-install at potentially different versions.

2. **Unpinned `latest` versions**: Reproducibility and stability hazard. Every `helm upgrade` re-applies the latest upstream manifest, risking breaking changes.

3. **Missing Shipwright webhook TLS**: The Kind installer performs extensive cert-manager integration for Shipwright's admission webhook (ClusterIssuer, CA cert, webhook cert, CRD annotations). Without this, Shipwright webhook will lack valid TLS and fail to validate Build/BuildRun resources.

4. **cluster-admin overkill**: Both Jobs bind to `cluster-admin`. A scoped ClusterRole with only the needed permissions would significantly reduce blast radius.

5. **No uninstall/cleanup path**: `helm uninstall` will leave Tekton and Shipwright CRDs, controllers, and webhooks permanently installed with no automated cleanup.

### Design Question

Before addressing the details, the team needs to decide: should these chart templates **replace** the Kind installer's bash-based Tekton/Shipwright logic, or should they coexist for a different use case (e.g., a pure Helm-based deployment without the bash installer)? That decision drives everything else.

Assisted-By: Claude Code

pdettori:
@moonlight16 should we consider instead extending the kind installer to work with vanilla (pre-existing) Kubernetes as well ?

moonlight16:
@pdettori Thanks for the thorough review and the suggestion to extend the Kind installer rather than add a parallel Helm-templated install path. I've pushed a refactor commit (`e4049c76`) that takes that approach.

**TL;DR of the change**

The original Helm chart templates (`templates/tekton-pipelines.yaml`, `templates/shipwright-build.yaml`) are removed. Tekton/Shipwright/cert-manager TLS/build-strategy install logic is extracted from `scripts/kind/setup-kagenti.sh` into a new shared library `scripts/lib/install-deps.sh`, and a new `scripts/k8s/setup-kagenti.sh` entry point delegates to the Kind script with `--skip-cluster` (forwarding all other flags) plus a `KAGENTI_SETUP_FLAVOR=k8s` env var that gates Kind-only steps.

**Mapping to your inline review comments**

| Comment | Resolution |
|---------|-----------|
| must-fix: Conflict with Kind installer | Single install path now. Both entry points call the same shared lib. |
| must-fix: Unpinned `latest` | Lib uses pinned versions matching the Kind installer's existing constants (`v0.66.0`, `v0.14.0`). |
| must-fix: cluster-admin overly broad | No more in-cluster install Jobs. Bash uses the operator's local kubectl credentials. |
| must-fix: Missing cert-manager TLS for Shipwright webhook | `configure_shipwright_webhook_tls` is now part of `install_shipwright`. Vanilla-k8s inherits it for free. |
| suggestion: hook-succeeded debugging | N/A — no Helm hooks. |
| suggestion: No uninstall path | New `scripts/k8s/cleanup-kagenti.sh` mirrors `scripts/kind/cleanup-kagenti.sh`. |

**Live test on Kind**

Verified end-to-end on a fresh Kind cluster via the new k8s entry point (`bash scripts/k8s/setup-kagenti.sh --with-builds`): all pods Running across all platform namespaces, both Tekton and Shipwright deployments healthy, `buildah-insecure-push` ClusterBuildStrategy registered. The Kind-only `docker exec <control-plane>` containerd-config block is correctly skipped under the k8s flavor.

**One open question for you**

I kept the scope minimal in this PR — only Tekton/Shipwright/build-strategies moved into the lib. The other cluster-agnostic steps (cert-manager, Istio, SPIRE, Gateway API CRDs, Helm chart installs, Kuadrant, MCP Gateway) still live inline in `scripts/kind/setup-kagenti.sh` and are reachable from the k8s entry point via `--skip-cluster` passthrough.

Would you prefer:
- **(a)** Keep the focused scope here, land it, and do a follow-up PR that fully extracts the cluster-agnostic steps into the lib, or
- **(b)** Expand this PR to do the full extraction now?

Either is fine on my end — I went with (a) so the diff stays focused on the issues you flagged. Happy to fold in (b) here if you'd rather.

Assisted-By: Claude Code

pdettori (approved):
## Review Summary

All six original review comments have been thoroughly addressed by the refactor commit. The approach (shared bash library + thin vanilla-k8s entry point) fits cleanly with the existing Kind installer without introducing any functional risk.

### What was verified

- **No interface change**: `scripts/kind/setup-kagenti.sh` retains all existing flags. CI's call path (`kind-full-test.sh` → `setup-kagenti.sh --with-all --skip-cluster --build-images`) is unchanged.
- **Correct gating**: Kind-specific steps (`docker exec` containerd config, `kind load docker-image`) are properly skipped via `KAGENTI_SETUP_FLAVOR` and existing `--skip-cluster` / `--build-images` guards.
- **Library is stateless**: No globals defined; reads helpers (`log_info`, `run_cmd`, `_wait_deployment_ready`, `DRY_RUN`) from the sourcing scope. Source guard prevents double-execution.
- **Versions pinned**: Defaults match existing Kind installer constants (`v0.66.0` / `v0.14.0`).
- **cert-manager TLS**: Full Shipwright webhook TLS chain (ClusterIssuer → CA → Issuer → Certificate → CRD annotations → restart) is preserved in the shared lib.
- **Cleanup symmetric**: k8s cleanup delegates to Kind cleanup with `--destroy-cluster` properly rejected.

### Suggestion: squash commits

Commit 1 adds Helm chart templates that commit 2 removes. The net effect on `charts/` is zero change. I'd recommend squashing into a single commit (or rebasing to drop commit 1) so the history tells a clean story without the intermediate approach that was abandoned.

Assisted-By: Claude Code

pdettori:
@moonlight16 could you pls. rebase/resolve conflicts

PR #1780, opened 2026-06-01
fix(installer): handle stale MCPGatewayExtension stuck in deletion

pdettori:
## Summary

- Fixes MCP Gateway installer failure when a stale `MCPGatewayExtension` from a prior version is stuck in deletion with a finalizer the old controller can no longer process
- Before installing, detects resources with `deletionTimestamp` under both `mcp.kuadrant.io` and `mcp.kagenti.com` API groups and strips their finalizers
- Switches OCP installer from `helm install` to `helm upgrade --install` for idempotent re-runs (matches Kind installer)

## Context

When upgrading from MCP Gateway v0.5.x (API group `mcp.kagenti.com/v1alpha1`) to v0.6.0 (`mcp.kuadrant.io/v1alpha1`), a previously-deleted MCPGatewayExtension gets stuck because:
1. The old controller is gone and can't process the finalizer
2. The new controller sees `deletionTimestamp` and runs cleanup logic instead of creating the broker-router
3. The installer waits 5 minutes for a deployment that will never appear

## Test plan

- [x] Verified fix on OpenShift cluster with stuck MCPGatewayExtension (finalizer removal + helm upgrade recreates broker-router in <10s)
- [ ] Re-run OCP installer on cluster with existing healthy MCP Gateway (idempotent upgrade path)
- [ ] Run Kind installer with `--with-mcp-gateway` on fresh cluster

Assisted-By: Claude Code

### DIFF
--- a/scripts/kind/setup-kagenti.sh
+++ b/scripts/kind/setup-kagenti.sh
@@ -1221,6 +1221,28 @@ if $WITH_MCP_GATEWAY; then
   kubectl create namespace mcp-system --dry-run=client -o yaml | kubectl apply -f - 2>/dev/null || true
   kubectl create namespace gateway-system --dry-run=client -o yaml | kubectl apply -f - 2>/dev/null || true
 
+  # Clean up any MCPGatewayExtension stuck in deletion (e.g. leftover from a prior
+  # version that used a different API group). A stuck finalizer prevents the controller
+  # from creating the broker-router deployment on reinstall.
+  for _crd_group in mcp.kuadrant.io mcp.kagenti.com; do
+    _stuck=$(kubectl get mcpgatewayextensions.${_crd_group} -n mcp-system -o json 2>/dev/null \
+      | python3 -c "
+import sys, json
+data = json.load(sys.stdin)
+for item in data.get('items', []):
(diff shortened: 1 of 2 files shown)


rubambiza (approved):
Clean fix for a real upgrade pain point. The pre-install cleanup correctly handles both API groups (`mcp.kuadrant.io`, `mcp.kagenti.com`), only targets resources stuck in deletion (has `deletionTimestamp`), and fails gracefully when the CRDs don't exist (`2>/dev/null || true`). The `helm upgrade --install` switch on OCP matches the Kind installer pattern and enables idempotent re-runs.

Verified:
- Python3 JSON parsing follows existing script patterns (11 existing usages in each script, no jq)
- Kind uses `kubectl` directly, OCP uses `$KUBECTL` — both consistent with their scripts
- OCP cleanup is guarded by `! $DRY_RUN` (Kind has no dry-run mode)
- `sleep 2` after finalizer removal gives API server time to process deletions before helm proceeds

**Areas reviewed**: Shell scripts (Kind + OCP installers)
**Commits**: 1, signed-off
**CI status**: All passing (including Deploy & Test)

Assisted-By: Claude Code

PR #1781, opened 2026-06-01
fix(ocp-installer): use helm upgrade --install for mcp-gateway

pdettori:
## Summary

- Replace the `helm install` + "already installed — skipping" branches in Step 5b with a single `helm upgrade --install` call
- This makes the MCP Gateway step idempotent: every run of `setup-kagenti.sh` enforces the correct `gateway.publicHost` value regardless of prior installation state

## Root Cause

The mcp-gateway Helm chart templates the broker-router `--mcp-gateway-public-host` flag from `.Values.gateway.publicHost` (nested). The correct install argument is `--set gateway.publicHost=<host>`.

When mcp-gateway was re-installed manually with `--set publicHost=<host>` (top-level, wrong key), the template silently fell back to the default `mcp.127-0-0-1.sslip.io`, making MCP Inspector unable to connect. Re-running `setup-kagenti.sh` did not fix it because the "already installed" branch simply skipped the step.

## Test Plan

- [x] Verified fix manually on the OCP cluster: `helm upgrade` with `--set gateway.publicHost=mcp-gateway-gateway-system.apps.kagenti1.kubestellar.org` restored connectivity
- [x] `https://mcp-gateway-gateway-system.apps.kagenti1.kubestellar.org/mcp` returns HTTP 200
- [x] MCPGatewayExtension status: "successfully verified and configured"
- [ ] CI green

Assisted-By: Claude Code

### DIFF
--- a/scripts/ocp/setup-kagenti.sh
+++ b/scripts/ocp/setup-kagenti.sh
@@ -1558,11 +1558,11 @@ for item in data.get('items', []):
   fi
 
   MCP_GW_PUBLIC_HOST="mcp-gateway-gateway-system.${DOMAIN}"
-  log_info "Installing MCP Gateway v${MCP_GATEWAY_VERSION} (publicHost=${MCP_GW_PUBLIC_HOST})..."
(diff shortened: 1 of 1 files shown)


rubambiza (approved):
LGTM

PR #1783, opened 2026-06-02
fix(mlflow): redirect to root after OIDC login instead of user profile

### LINKED ISSUE #1752: bug: MLFlow observability link redirects to /oidc/ui/user instead of root on Kind
## Description

When kagenti is deployed on Kind with MLFlow OIDC auth enabled, clicking the MLFlow link from the Observability page redirects the user to `http://mlflow.localtest.me:8080/oidc/ui/user` (the OIDC user profile page) instead of `http://mlflow.localtest.me:8080` (the MLFlow root/experiments view).

## Steps to Reproduce

1. Deploy kagenti on Kind with `mlflow.auth.enabled=true`
2. Navigate to the Observability page in the kagenti UI
3. Click the MLFlow dashboard link
4. **Actual:** Browser navigates to `http://mlflow.localtest.me:8080/oidc/ui/user`
5. **Expected:** Browser navigates to `http://mlflow.localtest.me:8080` (the experiments list)

## Analysis

The `MLFLOW_DASHBOARD_URL` is correctly set to `http://mlflow.localtest.me:8080` in the ConfigMap (generated from `charts/kagenti/templates/ui.yaml` line 22). The UI correctly uses this URL to open MLFlow.

The problem is in the `mlflow-oidc-auth` plugin behavior: when an unauthenticated user hits the MLFlow root (`/`), the OIDC auth middleware redirects to the Keycloak login flow, and after successful authentication the post-login redirect lands on `/oidc/ui/user` (the user profile page) instead of the originally requested path (`/`).

This is likely a misconfiguration of the OIDC redirect URI or the plugin's post-login redirect behavior. The `mlflow-oidc-auth` plugin uses `/callback` as its redirect URI (configured in `kagenti/auth/mlflow-oauth-secret/mlflow_oauth_secret.py` line 363), but the post-login landing page defaults to the user profile rather than the original request.

## Possible Fixes

1. Configure `mlflow-oidc-auth` to preserve the original request URL through the OIDC flow (state parameter or session-based redirect)
2. Set a post-login redirect URL environment variable if supported by the plugin
3. Change the dashboard link to point directly to `/experiments` which may bypass the profile landing

## Related

- #1605 — MLFlow OIDC auth blocks AI Gateway and evaluation endpoints (same plugin, different symptom)

## Environment

- Deployment: Kind (localtest.me)
- MLFlow image: `ghcr.io/mlflow/mlflow:v3.9.0`
- Auth plugin: `mlflow-oidc-auth` (pip-installed at runtime)
- Keycloak OIDC enabled

pdettori:
## Summary

- Set `DEFAULT_LANDING_PAGE_IS_PERMISSIONS=false` in the MLflow OAuth secret generator and Helm template
- This configures `mlflow-oidc-auth` to redirect to `/` (experiments view) after Keycloak login, instead of `/oidc/ui/user` (permissions/profile page)

## Root Cause

The `mlflow-oidc-auth` plugin defaults `DEFAULT_LANDING_PAGE_IS_PERMISSIONS` to `true` ([source](https://github.com/mlflow-oidc/mlflow-oidc-auth/blob/main/mlflow_oidc_auth/config.py#L131)). When true, the OIDC callback handler redirects to the user profile page (`/oidc/ui/user`) instead of the originally requested path or root.

## Fix

Two-layer fix for robustness:
1. **Secret generator** (`mlflow_oauth_secret.py`): adds the key to the generated K8s secret so new deployments get it automatically
2. **Helm template** (`mlflow.yaml`): sets it as a direct env var, ensuring it takes effect even if the secret was created before this change

## Test plan

- [ ] CI passes (no behavioral change for Kind since MLflow auth is disabled)
- [ ] Verify on OpenShift cluster with MLflow OIDC auth enabled: clicking MLFlow link from Observability page should land on experiments view (`/`), not `/oidc/ui/user`

Closes #1752

Assisted-By: Claude Code

### DIFF
--- a/charts/kagenti-deps/templates/mlflow.yaml
+++ b/charts/kagenti-deps/templates/mlflow.yaml
@@ -397,6 +397,8 @@ spec:
           value: "postgresql://$(POSTGRES_USER):$(POSTGRES_PASSWORD)@postgres:5432/mlflow"
         - name: OIDC_ALEMBIC_VERSION_TABLE
           value: "mlflow_oidc_alembic_version"
+        - name: DEFAULT_LANDING_PAGE_IS_PERMISSIONS
+          value: "false"
         {{- end }}
         ports:
         - containerPort: 5000
--- a/kagenti/auth/mlflow-oauth-secret/mlflow_oauth_secret.py
+++ b/kagenti/auth/mlflow-oauth-secret/mlflow_oauth_secret.py
@@ -390,6 +390,7 @@ def main() -> None:
             "OIDC_REDIRECT_URI": redirect_uri,
             "OIDC_SCOPE": "openid email profile",
             "OIDC_GROUPS_CLAIM": "groups",
+            "DEFAULT_LANDING_PAGE_IS_PERMISSIONS": "false",
         }
 
         create_or_update_k8s_resource(


charts/kagenti-deps/templates/mlflow.yaml
```diff
@@ -397,6 +397,8 @@ spec:
           value: "postgresql://$(POSTGRES_USER):$(POSTGRES_PASSWORD)@postgres:5432/mlflow"
         - name: OIDC_ALEMBIC_VERSION_TABLE
           value: "mlflow_oidc_alembic_version"
+        - name: DEFAULT_LANDING_PAGE_IS_PERMISSIONS
```
huang195, line 400: nit (optional, no change needed) — nice belt-and-suspenders. Just for the record: this container reads the OAuth secret via `envFrom: secretRef` (L378), and in k8s a direct `env` entry always wins over `envFrom` for the same key. So this direct value will *shadow* whatever the secret carries — which is exactly what you want here (both are `false`, and it also covers pre-existing secrets created before this change that lack the key). Only flagging so nobody's surprised later if they try to flip it via the secret alone.

huang195 (approved):
### Summary

Bellissimo, Paolo — small, surgical, and correct. 👨‍🍳 I verified the whole chain against the upstream `mlflow-oidc-auth` source:

- `DEFAULT_LANDING_PAGE_IS_PERMISSIONS` is the right env var, default `True` (`config.py` L131) — so the redirect-to-permissions really is the current behavior you're fixing.
- `config_manager.get_bool` parses `"false"` → `False` correctly (`str(value).lower() in ("true","1","t")`), so the string value actually takes effect — not a silent no-op (the subtle trap with these string-bool configs).
- The env var lands inside the `{{- if .Values.mlflow.auth.enabled }}` guard, so Kind (auth disabled) is untouched — matches your test plan. Perfetto.
- The two-layer fix is consistent: secret via `envFrom` + direct `env`, both `false`; the direct env wins and also covers pre-existing secrets. Solid reasoning in the PR description.

One optional nit below, nothing blocking. 🟢

**Areas reviewed**: Helm template, Python secret generator, upstream mlflow-oidc-auth source, commit/PR conventions, CI
**CI status**: green (Deploy & Test / E2E HyperShift pending — awaiting maintainer `/run-e2e`, normal)

PR #1786, opened 2026-06-02
fix(mlflow): exempt /version, /health, /ping from OIDC auth and pin plugin

### LINKED ISSUE #1784: fix(e2e): MLflow /version endpoint returns 401 in RC Release Validation (Kind E2E)
## Summary

The `test_mlflow_version_endpoint` test fails in the **RC Release Validation** workflow (Kind E2E job) with a 401 response, while it passes in the regular `E2E K8s 1.35.0 (Kind)` workflow.

**Failing run:** https://github.com/kagenti/kagenti/actions/runs/26776833102
**Passing run (main, same day):** https://github.com/kagenti/kagenti/actions/runs/26786187730

## Error

```
AssertionError: MLflow /version endpoint failed: 401 - {"detail":"Authentication required"}
assert 401 in (200, 302, 307)
```

Test: `kagenti/tests/e2e/common/test_mlflow_auth.py::TestMLflowAuth::test_mlflow_version_endpoint`

### 1. `/version` is not in the unprotected routes list

The `mlflow-oidc-auth` plugin's `_is_unprotected_route()` only exempts `/v1/traces` (via the runtime patch in `charts/kagenti-deps/templates/mlflow.yaml` lines 206-211). The `/version` endpoint is not exempted, so it requires auth when OIDC is enabled.

### 2. 401 vs 302 depends on request type

The `mlflow-oidc-auth` plugin (v7.x) distinguishes between browser requests (returns 302 redirect to OIDC login) and API requests (returns 401 JSON). The test uses `httpx.AsyncClient` without `Accept: text/html`, so the plugin treats it as an API client and returns 401 instead of redirecting.

### 3. `mlflow-oidc-auth` is unpinned

```dockerfile
pip install --no-cache-dir --quiet "mlflow-oidc-auth"
```

This installs whatever version is latest (currently 7.3.1). The v7.2.0 release added "re-authenticate sessions against IdP token expiry" which changed redirect/auth-failure semantics. Without a version pin, behavior can change between runs.

### 4. K8s probes use `/version` but aren't affected

The readiness/liveness probes at lines 413-426 of `mlflow.yaml` use `/version` and work because they hit the pod directly (kubelet → pod IP, not through the auth middleware path). But external requests go through the full middleware stack.

## Proposed Fix

**Option A (recommended): Add `/version` to unprotected routes**

In `charts/kagenti-deps/templates/mlflow.yaml`, extend the runtime patch:

```python
def patched_is_unprotected(self, path: str) -> bool:
    if path.startswith("/v1/traces") or path in ("/version", "/health", "/ping"):
        return True
    return original_is_unprotected(self, path)
```

This is the correct behavior — health/version endpoints should never require auth.

**Option B (also recommended): Pin `mlflow-oidc-auth` version**

```dockerfile
pip install --no-cache-dir --quiet "mlflow-oidc-auth==7.3.1"
```

Prevents surprise behavior changes from upstream releases.

## Context

- Both workflows use `scripts/kind/setup-kagenti.sh --with-all --build-images`
- `--with-all` sets `mlflow.auth.enabled=true` via Helm `--set` (overrides `dev_values.yaml`)
- The mlflow-oauth-secret job completes successfully (OIDC client + K8s secret created)
- Test results in failing run: 1 failed, 71 passed, 30 skipped

## Files to Modify

- `charts/kagenti-deps/templates/mlflow.yaml` (unprotected routes patch, ~line 206)
- Optionally: `kagenti/tests/e2e/common/test_mlflow_auth.py` (accept 401 as valid when auth is enabled)

Assisted-By: Claude Code

pdettori:
## Summary

- Add `/version`, `/health`, and `/ping` to the mlflow-oidc-auth unprotected routes patch — these health/version endpoints should never require authentication
- Pin `mlflow-oidc-auth` to `>=7.3.1,<8` to prevent unexpected behavior changes from unpinned upstream releases (v7.2.0 changed auth-failure semantics)

Closes #1784

## Root Cause

The `mlflow-oidc-auth` plugin's `_is_unprotected_route()` only exempted `/v1/traces`. The `/version` endpoint requires auth when OIDC is enabled, returning 401 for API clients. This caused E2E test failures in the RC Release Validation workflow.

## Test Plan

- [ ] CI E2E tests pass (MLflow auth tests)
- [ ] RC Release Validation workflow passes with this change
- [ ] MLflow liveness/readiness probes unaffected (already use `/health`)

Assisted-By: Claude Code

### DIFF
--- a/charts/kagenti-deps/templates/mlflow.yaml
+++ b/charts/kagenti-deps/templates/mlflow.yaml
@@ -192,24 +192,41 @@ spec:
           # IMPORTANT: Patch BEFORE importing app (app is created at import time)
           from mlflow_oidc_auth.middleware.auth_middleware import AuthMiddleware
           import mlflow_oidc_auth.middleware.fastapi_permission_middleware as fpm
+          import mlflow_oidc_auth.hooks.before_request as _before_request
 
           # Assert patched APIs exist — fail fast on mlflow-oidc-auth upgrades
           assert hasattr(AuthMiddleware, "_is_unprotected_route"), \
               "mlflow-oidc-auth API changed: AuthMiddleware._is_unprotected_route not found"
           assert hasattr(fpm, "_find_fastapi_validator"), \
               "mlflow-oidc-auth API changed: _find_fastapi_validator not found"
+          assert hasattr(_before_request, "_is_unprotected_route"), \
+              "mlflow-oidc-auth API changed: before_request._is_unprotected_route not found"
 
-          # Patch 1: Exclude /v1/traces from OIDC auth
+          # Patch 1: Exclude /v1/traces, /version, /ping from OIDC auth (FastAPI gate)
           # The OTel collector authenticates via OAuth2 client credentials (Bearer token)
           # but may not have a user profile in the MLflow users table.
-          # Skip auth entirely — Istio mTLS protects this path.
+          # /version and /ping are health-check endpoints that should never require auth.
+          # Skip auth entirely — Istio mTLS protects /v1/traces.
           original_is_unprotected = AuthMiddleware._is_unprotected_route
           def patched_is_unprotected(self, path: str) -> bool:
-              if path.startswith("/v1/traces") or path in ("/version", "/health", "/ping"):
+              if path.startswith("/v1/traces") or path in ("/version", "/ping"):
                   return True
               return original_is_unprotected(self, path)
           AuthMiddleware._is_unprotected_route = patched_is_unprotected
-          print("Excluded /v1/traces, /version, /health, /ping from OIDC auth")
+          print("Excluded /v1/traces, /version, /ping from OIDC auth (FastAPI gate)")
+
+          # Patch 1b: Exclude /version, /ping from Flask before_request auth gate
+          # /version and /ping are Flask routes that pass through both FastAPI middleware
+          # AND the Flask before_request hook. Without this patch, Gate A (above) skips
+          # auth (so no username is set), then Gate B returns 401 because username is None.
+          # /health is already natively exempt at this layer.
+          _original_flask_is_unprotected = _before_request._is_unprotected_route
+          def _patched_flask_is_unprotected(path: str) -> bool:
+              if path in ("/version", "/ping"):
+                  return True
+              return _original_flask_is_unprotected(path)
+          _before_request._is_unprotected_route = _patched_flask_is_unprotected
+          print("Excluded /version, /ping from Flask before_request auth gate")
 
           # Patch 2: Skip permission validation for /v1/traces
           # mlflow-oidc-auth v7+ adds a FastAPI permission middleware that checks
--- a/charts/kagenti-deps/templates/mlflow.yaml
+++ b/charts/kagenti-deps/templates/mlflow.yaml
@@ -131,7 +131,7 @@ spec:
           export HOME=/tmp && \
           pip install --no-cache-dir --quiet "psycopg[binary]" && \
           {{- if .Values.mlflow.auth.enabled }}
-          pip install --no-cache-dir --quiet "mlflow-oidc-auth" && \
+          pip install --no-cache-dir --quiet "mlflow-oidc-auth>=7.3.1,<8" && \
           {{- end }}
           {{- if and .Values.openshift .Values.mlflow.auth.enabled }}
           # Combine system CAs with the ingress CA so Python can verify both
@@ -205,11 +205,11 @@ spec:
           # Skip auth entirely — Istio mTLS protects this path.
           original_is_unprotected = AuthMiddleware._is_unprotected_route
           def patched_is_unprotected(self, path: str) -> bool:
-              if path.startswith("/v1/traces"):
+              if path.startswith("/v1/traces") or path in ("/version", "/health", "/ping"):
                   return True
               return original_is_unprotected(self, path)
           AuthMiddleware._is_unprotected_route = patched_is_unprotected
-          print("Excluded /v1/traces from OIDC auth (mTLS via Istio)")
+          print("Excluded /v1/traces, /version, /health, /ping from OIDC auth")
 
           # Patch 2: Skip permission validation for /v1/traces
           # mlflow-oidc-auth v7+ adds a FastAPI permission middleware that checks


charts/kagenti-deps/templates/mlflow.yaml
```diff
@@ -205,11 +205,11 @@ spec:
           # Skip auth entirely — Istio mTLS protects this path.
           original_is_unprotected = AuthMiddleware._is_unprotected_route
           def patched_is_unprotected(self, path: str) -> bool:
-              if path.startswith("/v1/traces"):
+              if path.startswith("/v1/traces") or path in ("/version", "/health", "/ping"):
```
huang195: 🛑 **must-fix** — this opens Gate A, but `/version` and `/ping` will still 401. I traced the full path in `mlflow-oidc-auth` 7.3.1:

There are **two** independent auth gates:
1. `AuthMiddleware._is_unprotected_route` (this patch) — FastAPI/ASGI layer.
2. `mlflow_oidc_auth.hooks.before_request._is_unprotected_route` — the **Flask** before_request hook (not patched here).

The only FastAPI-native routes (which bypass Flask) are `/v1/traces`, `/gateway/*`, `/ajax-api/3.0/jobs/*`, `/ajax-api/3.0/mlflow/assistant/*` (see `app.py:_include_mlflow_fastapi_routers`). Everything else — including **`/version` and `/ping`** — falls through to the Flask WSGI mount (`oidc_app.mount("/", AuthAwareWSGIMiddleware(app))`).

So for `/version`: this patch makes Gate A skip auth (so **no username is set**), then the request hits Flask's `before_request_hook`, whose own `_is_unprotected_route` only exempts `/static, /favicon.ico, /health, /metrics, /docs, /redoc, /openapi.json` — **not `/version`/`/ping`** — so `username is None` → `make_auth_required_response()` → **401 again**. The patch just relocates the 401 from Gate A to Gate B.

This is the same shape as why `/v1/traces` needs *both* Patch 1 and Patch 2: skipping auth without a username makes the downstream layer 401. `/version`/`/ping` need the Flask equivalent — suggest adding a patch (next to Patch 2) that monkeypatches `mlflow_oidc_auth.hooks.before_request._is_unprotected_route` to also return True for `/version` and `/ping`.

(`/health` is already exempt at both gates, so that one's fine — and redundant in this list.)

Heads-up: Kind `Deploy & Test` runs with MLflow auth disabled (#1783), so it won't catch this — only the OCP E2E will.

huang195 (changes requested):
### Summary

Ciao Paolo — the version pin is perfetto (verified 7.3.1 is the latest on PyPI and `>=7.3.1,<8` correctly excludes the troublesome 7.2.0). But I traced the auth path through `mlflow-oidc-auth` 7.3.1 and there's a problem: this fix doesn't actually unblock `/version` (the endpoint from #1784). Mamma mia. 😅

`/version` and `/ping` are served by the **Flask** WSGI mount, not FastAPI, so they pass through *two* independent auth gates. This patch only opens the first (`AuthMiddleware._is_unprotected_route`); the second — the Flask `hooks/before_request._is_unprotected_route` hook — still 401s them because its hardcoded exempt list (`/static, /favicon.ico, /health, /metrics, /docs, /redoc, /openapi.json`) doesn't include `/version`/`/ping`. Net effect: the 401 just moves from one gate to the other. `/health` is the only one that works here — and it already worked (exempt at both gates). Details + suggested fix inline.

**Important:** the green Kind `Deploy & Test` doesn't validate this — Kind runs with MLflow auth disabled (#1783). Only the pending OCP `E2E HyperShift` exercises the OIDC path.

**Areas reviewed**: Helm template (Python startup patch), mlflow-oidc-auth 7.3.1 source (auth_middleware, before_request hook, app routing, fastapi permission middleware), PyPI version pin, commit/PR conventions, CI
**CI status**: green on Kind (auth disabled — does not cover this path); OCP E2E pending

pdettori:
## Addressed review feedback

**commit d9b83787**: Patch both auth gates for `/version` and `/ping`

- Added **Patch 1b**: monkeypatches `mlflow_oidc_auth.hooks.before_request._is_unprotected_route` to exempt `/version` and `/ping` at the Flask layer
- Removed `/health` from Patch 1 (redundant — already natively exempt at both gates)
- Added assertion for `_before_request._is_unprotected_route` existence (fail-fast on API changes)

The two gates are now both patched:
| Endpoint | Gate A (FastAPI) | Gate B (Flask) |
|----------|-----------------|----------------|
| `/v1/traces` | Patch 1 ✓ | N/A (FastAPI-native) |
| `/version` | Patch 1 ✓ | Patch 1b ✓ |
| `/ping` | Patch 1 ✓ | Patch 1b ✓ |
| `/health` | Natively exempt | Natively exempt |

---

**RE: E2E HyperShift failure** — unrelated. Deployment failed before tests ran:
```
Error from server (NotFound): customresourcedefinitions.apiextensions.k8s.io "mcpserverregistrations.mcp.kagenti.com" not found
```
This is a missing MCP Gateway CRD issue, not related to this MLflow auth fix.

huang195 (approved):
### Summary

Perfetto, Paolo — must-fix resolved in `d9b8378`. 👨‍🍳 You patched both auth gates:

- **Gate A** (Patch 1, FastAPI `AuthMiddleware._is_unprotected_route`): `/v1/traces`, `/version`, `/ping`.
- **Gate B** (Patch 1b, new, Flask `before_request._is_unprotected_route`): `/version`, `/ping`.
- `/health` correctly rides on the native exemption (present in both gates' stock lists) instead of being patched twice.

Verified the mechanics: the Flask hook calls `_is_unprotected_route` as a module global (resolved at call-time), so monkeypatching the module attribute takes effect; and you added the matching fail-fast `assert hasattr(...)` for the new symbol. Bravo. Version pin (`>=7.3.1,<8`) unchanged and still correct.

The only thing that'd give full end-to-end confidence is the OCP E2E (Kind runs with MLflow auth disabled, so it doesn't exercise this path) — but the logic is sound now.

**Areas reviewed**: Helm template (Python startup patch), mlflow-oidc-auth 7.3.1 source (both auth gates + app routing), monkeypatch mechanics, version pin, conventions, CI
**Commits**: 2, signed-off: yes
**CI status**: green (OCP E2E pending — maintainer trigger)

PR #1787, opened 2026-06-02
fix(e2e): skip MCP CRD validation when gateway not installed

### LINKED ISSUE #1785: fix(e2e): Update CRD validation for MCP Gateway v0.6.0 (HyperShift E2E)
## Summary

The HyperShift E2E job in the **RC Release Validation** workflow fails because it checks for the old `mcpserverregistrations.mcp.kagenti.com` CRD, which no longer exists after migrating to MCP Gateway chart v0.6.0.

**Failing run:** https://github.com/kagenti/kagenti/actions/runs/26776833102

## Error

```
ERROR: CRD mcpserverregistrations.mcp.kagenti.com not established
```

The script retries 12 times (~60s) checking for either `mcp.kagenti.com` or `mcp.kuadrant.io` domains, finds neither, defaults to `mcp.kagenti.com`, then fails when `wait_for_crd` can't find it.

## Root Cause

MCP Gateway v0.6.0 uses **different CRD names/groups** than the ones listed in `.github/scripts/kagenti-operator/41-wait-crds.sh`. The script checks for:

```bash
MCP_RESOURCES=(
    "mcpserverregistrations"
    "mcpvirtualservers"
    "mcpgatewayextensions"
)
```

With domain detection between `mcp.kagenti.com` and `mcp.kuadrant.io`. After the v0.6.0 migration, these CRD names are stale.

## Proposed Fix

Update `.github/scripts/kagenti-operator/41-wait-crds.sh` to:

1. Replace the `MCP_RESOURCES` array with the correct CRD names from MCP Gateway v0.6.0
2. Update the domain detection logic if the group changed
3. Remove the fallback to `mcp.kagenti.com` if that domain is no longer used

### Steps to determine the correct CRDs:

```bash
# On a cluster with MCP Gateway v0.6.0 deployed:
kubectl get crds | grep -i mcp
```

Or check the MCP Gateway v0.6.0 chart templates for CRD definitions.

## Context

- The script at `.github/scripts/kagenti-operator/41-wait-crds.sh` already has a comment acknowledging the domain rename (line 14-15)
- Kind CI was noted as using the "old version" and HyperShift the "new" — but now both should use v0.6.0
- Other CRDs (`agentcards.agent.kagenti.dev`, `agentruntimes.agent.kagenti.dev`) are present and healthy
- The cluster otherwise deploys successfully — controller-manager, backend, UI all Running

## Files to Modify

- `.github/scripts/kagenti-operator/41-wait-crds.sh` — update CRD names and domain detection

## Related

- PR #1780 (`fix/mcp-gateway-stale-finalizer`) — recent MCP Gateway fixes on main

Assisted-By: Claude Code

pdettori:
## Summary

- Skip MCP Gateway CRD validation in `41-wait-crds.sh` when MCP Gateway is not installed
- Fix domain detection order: try `mcp.kuadrant.io` first (actual v0.6.0 group), then `mcp.kagenti.com`
- Update fallback default to `mcp.kuadrant.io`

## Root Cause

The HyperShift E2E job in the RC Release Validation workflow fails because `41-wait-crds.sh` unconditionally waits for MCP Gateway CRDs. On HyperShift/OCP, the MCP Gateway chart is optional (`SKIP_MCP_GATEWAY=true` by default in `setup-kagenti.sh`), so CRDs are never registered — causing a 60s timeout followed by a hard failure.

## Fix

Check for the `mcp-gateway` Helm release in `mcp-system` (or existing MCP CRDs) before entering the wait loop. If MCP Gateway is not installed, log a skip message and continue.

Closes #1785

## Test plan

- [ ] RC Release Validation workflow passes (HyperShift E2E no longer fails on missing MCP CRDs)
- [ ] Kind E2E with `--with-all` still validates MCP CRDs correctly

Assisted-By: Claude Code

### DIFF
--- a/.github/scripts/kagenti-operator/41-wait-crds.sh
+++ b/.github/scripts/kagenti-operator/41-wait-crds.sh
@@ -7,48 +7,56 @@ source "$SCRIPT_DIR/../lib/k8s-utils.sh"
 
 log_step "41" "Waiting for Kagenti Operator CRDs"
 
-# NOTE: agents.agent.kagenti.dev is NOT required for E2E tests since we now
-# deploy agents using standard Kubernetes Deployments + Services directly.
-# Only keeping CRDs that are still actively used in CI.
-#
-# MCP gateway v0.6.0 renamed the CRD group from mcp.kuadrant.io to
-# mcp.kagenti.com. Kind CI still uses the old version; HyperShift uses the new.
-# Detect which domain is present, then wait for those CRDs.
-MCP_RESOURCES=(
-    "mcpserverregistrations"
-    "mcpvirtualservers"
-    "mcpgatewayextensions"
-)
-
-# Detect which MCP CRD domain is installed.
-# Retry up to 60s since operators may still be registering CRDs.
-MCP_DOMAIN=""
-for i in $(seq 1 12); do
-    if kubectl get crd "mcpserverregistrations.mcp.kagenti.com" &>/dev/null; then
-        MCP_DOMAIN="mcp.kagenti.com"
-        break
-    elif kubectl get crd "mcpserverregistrations.mcp.kuadrant.io" &>/dev/null; then
-        MCP_DOMAIN="mcp.kuadrant.io"
-        break
-    fi
-    log_info "MCP CRDs not yet available, retrying ($i/12)..."
-    sleep 5
-done
-if [ -z "$MCP_DOMAIN" ]; then
-    MCP_DOMAIN="mcp.kagenti.com"
-    log_info "Defaulting to $MCP_DOMAIN (neither domain detected after 60s)"
+# Check if MCP Gateway is installed before waiting for its CRDs.
+# The mcp-gateway chart is optional (requires --with-mcp-gateway or --with-all
+# on Kind, and is skipped by default on HyperShift/OCP).
+MCP_GATEWAY_INSTALLED=false
+if helm list -n mcp-system -q 2>/dev/null | grep -q "^mcp-gateway$"; then
+    MCP_GATEWAY_INSTALLED=true
+elif kubectl get namespace mcp-system &>/dev/null && \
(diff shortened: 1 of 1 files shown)


huang195 (approved):
### Summary

Perfetto, Paolo — clean, well-targeted CI fix. 🍝 I verified the key facts end-to-end:

- **The domain reorder is correct.** I pulled the actual chart CRDs (`helm show crds oci://ghcr.io/kuadrant/charts/mcp-gateway --version 0.6.0`): all three (`mcpserverregistrations`, `mcpvirtualservers`, `mcpgatewayextensions`) live in **`mcp.kuadrant.io`**. So trying kuadrant.io first and defaulting to it is right — the old comment claiming 0.6.0 renamed to `mcp.kagenti.com` was simply wrong. Buono that you corrected it.
- **The guard matches reality.** Release `mcp-gateway` in `mcp-system` is exactly how both Kind (`helm upgrade --install mcp-gateway -n mcp-system`) and OCP install it, and `SKIP_MCP_GATEWAY=true` is the OCP default — so the skip path triggers precisely when intended.
- **Shell-safe.** `set -euo pipefail` is on, but the helm/kubectl probes sit in `if`/`elif` conditions, so a no-match doesn't abort. 👍

One out-of-scope heads-up (not blocking, can't inline since it's outside this diff): `scripts/kind/cleanup-kagenti.sh:49` still deletes `mcpserverregistrations.mcp.kagenti.com`. Now that we've confirmed 0.6.0 uses `mcp.kuadrant.io`, that cleanup line silently no-ops (`--ignore-not-found … || true` hides it). Worth a quick follow-up so cleanup actually removes the resources.

**Areas reviewed**: Shell script (E2E CI), setup/cleanup scripts, mcp-gateway 0.6.0 chart CRDs, commit/PR conventions, CI
**CI status**: green (E2E HyperShift awaiting maintainer `/run-e2e`)

Issue #1768, opened 2026-06-01
feature: Consolidate AuthBridge config to a single source of truth


davidhadas:
### Feature Description

AuthBridge config has 3 independent sources that all need to agree.

  Here's the flow:

  1. Helm templates (initial deploy for pre-declared namespaces like team1/team2):
    - authbridge-template-configmaps.yaml — creates authbridge-runtime-config ConfigMap in kagenti-system as a template
    - agent-namespaces.yaml — creates the actual ConfigMaps in team1/team2 at install time
  2. Backend Python (agents.py:_build_authbridge_runtime_yaml()) — generates ConfigMaps dynamically when agents are created via the UI/API. This function hardcodes token-exchange at line 2073.
  3. Operator reconciliation — reads the template ConfigMap and applies it to namespaces, which can overwrite runtime patches.

 The real fix is consolidating the outbound plugin config into Helm values so there's a single source of truth.

### Proposed Solution

  Recommended approach:

  1. Add a Helm value like authbridge.outbound.plugin (defaulting to token-exchange for now) that both template files consume via {{ .Values.authbridge.outbound.plugin }}.
  2. Make the backend read from config — instead of hardcoding the plugin in _build_authbridge_runtime_yaml(), read it from settings (which can be populated from env vars set by the Helm chart).
  3. Switch via values — then toggling to token-broker becomes a single --set authbridge.outbound.plugin=token-broker plus the plugin-specific config block.

  This way:
  - Helm templates use {{ .Values }} — no awk patching needed
  - Backend reads the same setting from an env var injected by the chart
  - Operator reconciliation stays consistent because it uses the Helm-managed template ConfigMap

### Want to contribute?

- [x] I would like to work on this issue.

davidhadas:
@huang195 @pdettori @mrsabath

pdettori:
@davidhadas good point - I'm all for consolidation in single source of truth

davidhadas:
Fix merged

PR #1788, opened 2026-06-02
Single source of truth for AuthBridge Config + Envoy Config Fix

davidhadas:
## Summary

This PR comes to fix two issues, both need to be fixed preparing to support Token Broker plugin that was recently added to AuthBridge 

1. There are two multiple places in which AuthBridge config is defined while this PR changes this such that we have a single source of Truth about the expected AuthBridge config
2. The envoy config used lack message_timeout definition needed for token broker plugin

1  - A single source of Truth about the expected AuthBridge config

  The change consolidates the 3 independent definitions into one place — values.yaml authBridge.pipeline. The flow is now:

  values.yaml (authBridge.pipeline)
      → _helpers.tpl named template (tpl() evaluates Helm expressions)
          → authbridge-template-configmaps.yaml (kagenti-system ConfigMap)
          → agent-namespaces.yaml (team1/team2 ConfigMaps)
          → Mounted into backend pod → _get_authbridge_runtime_yaml() reads it
              → Copies to dynamically-created agent namespaces

  Files changed:
  - charts/kagenti/values.yaml — added authBridge.pipeline block
  - charts/kagenti/templates/_helpers.tpl — added kagenti.authbridge-runtime-config-yaml named template
  - charts/kagenti/templates/authbridge-template-configmaps.yaml — replaced inline pipeline with include
  - charts/kagenti/templates/agent-namespaces.yaml — replaced inline pipeline with include
  - charts/kagenti/templates/ui.yaml — added ConfigMap volume mount + checksum annotation
  - kagenti/backend/app/core/config.py — added authbridge_runtime_config_path setting
  - kagenti/backend/app/routers/agents.py — added _get_authbridge_runtime_yaml() that reads from mounted file, renamed old function as fallback

  SRE workflow to switch to token-broker: Edit authBridge.pipeline in values.yaml, run helm upgrade. Done — all consumers get the new config.

2 - missing message_timeout for envoy config

message_timeout: 310s (integrated) — Added to all 6 Envoy ext_proc filter blocks across:
  - charts/kagenti/templates/authbridge-template-configmaps.yaml (2 blocks)
  - charts/kagenti/templates/agent-namespaces.yaml (2 blocks)
  - kagenti/backend/app/core/constants.py DEFAULT_ENVOY_YAML (2 blocks)

Fixes # https://github.com/kagenti/kagenti/issues/1768

### DIFF
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -2072,54 +2072,118 @@ async def get_shipwright_build_info(
         raise HTTPException(status_code=e.status, detail=str(e.reason))
 
 
+def _get_authbridge_runtime_yaml() -> str:
+    """Read the authbridge runtime config from the Helm-managed ConfigMap.
+
+    The pipeline configuration is defined in values.yaml (authBridge.pipeline),
+    rendered by the Helm chart into the authbridge-runtime-config ConfigMap,
+    and mounted into the backend pod. This is the single source of truth.
+
+    Falls back to legacy in-memory construction when the file does not
+    exist (e.g. local development outside the cluster).
+    """
+    config_path = settings.authbridge_runtime_config_path
+    try:
+        with open(config_path, "r") as f:
+            content = f.read()
+        if content.strip():
+            return content
+    except FileNotFoundError:
+        logger.warning(
+            "AuthBridge runtime config not found at %s; "
+            "falling back to legacy in-memory construction",
+            config_path,
+        )
+    except OSError as e:
+        logger.warning(
+            "Failed to read AuthBridge runtime config from %s: %s; "
+            "falling back to legacy construction",
+            config_path,
+            e,
+        )
+
+    return _build_authbridge_runtime_yaml_fallback()
+
+
 def _build_authbridge_runtime_yaml(
     keycloak_url: str,
     realm: str,
     issuer: str,
-    spire_enabled: bool,
+    spire_enabled: bool = False,
 ) -> str:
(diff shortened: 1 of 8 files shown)


kagenti/backend/app/routers/agents.py
```diff
@@ -2072,54 +2072,118 @@ async def get_shipwright_build_info(
...
+
+    Used when the mounted ConfigMap is unavailable (local dev, tests).
+    The canonical source of truth is values.yaml authBridge.pipeline,
+    rendered by the Helm chart and mounted at the path specified by
+    settings.authbridge_runtime_config_path.
     """
+    keycloak_url = settings.keycloak_url or DEFAULT_KEYCLOAK_INTERNAL_URL
+    realm = settings.effective_keycloak_realm or DEFAULT_KEYCLOAK_REALM
+    issuer = f"{settings.effective_keycloak_url}/realms/{realm}"
+    spire_enabled = False
```
pdettori, line 2179: suggestion: The fallback hardcodes `spire_enabled = False`, meaning file-less deployments will never include the `spiffe: {}` block. If the intent is backward compatibility for local dev, this is fine — but a brief comment noting this divergence from the Helm-rendered path would help future readers understand that the mounted ConfigMap is authoritative for SPIRE-enabled clusters.

pdettori (approved):
Clean refactoring — single source of truth for AuthBridge config with proper fallback. All 6 ext_proc blocks get consistent `message_timeout: 310s`. CI passes. LGTM.

PR #1790, opened 2026-06-02
fix(ocp-installer): skip Kuadrant CR when CRD absent in v1.4+

pdettori:
## Summary
- Kuadrant operator v1.4.x removed the `Kuadrant` bootstrapping CRD, causing `setup-kagenti.sh --with-all` to fail at Step 5 with `no matches for kind "Kuadrant" in version "kuadrant.io/v1beta1"`
- Guards the CR creation behind a CRD existence check; dynamically resolves the served API version for older operators that still ship the CRD
- Verified: full cleanup + re-install passes cleanly on OCP cluster with Kuadrant 1.4.2

## Test plan
- [x] `scripts/ocp/cleanup-kagenti.sh --yes` followed by `scripts/ocp/setup-kagenti.sh --kagenti-repo . --with-all` completes with exit 0
- [x] No errors in Step 5 output — logs "Kuadrant CRD not present (v1.4+) — skipping CR creation"
- [ ] Regression: test with an older Kuadrant version that still has the CRD (if available)

Assisted-By: Claude Code

### DIFF
--- a/scripts/ocp/setup-kagenti.sh
+++ b/scripts/ocp/setup-kagenti.sh
@@ -1509,14 +1509,19 @@ if $WITH_KUADRANT; then
   if ! $DRY_RUN; then
     _wait_deployment_ready kuadrant-operator-controller-manager "$KUADRANT_NS" "Kuadrant operator"
 
-    log_info "Creating Kuadrant CR..."
-    $KUBECTL apply -f - <<EOF
-apiVersion: kuadrant.io/v1beta1
+    if $KUBECTL get crd kuadrants.kuadrant.io &>/dev/null; then
+      log_info "Creating Kuadrant CR..."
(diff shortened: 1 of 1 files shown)


esnible (approved):

mrsabath (approved):
Approving — small, surgical fix. Guards the `Kuadrant` CR creation behind a CRD existence check and dynamically resolves the served API version, which cleanly handles both pre-1.4 (CRD present) and 1.4+ (CRD removed) operator versions. CI green including the 23-min Deploy & Test. Style consistent with the rest of `setup-kagenti.sh` (`&>/dev/null`, `log_info`, etc.).

Non-blocking nit: `awk '{print $NF}'` over the served-versions list picks whichever version is listed last in the CRD YAML — fine in practice since Kuadrant has historically had only one served version at a time, but if a future release ships multiple served versions during a migration window, you'd want to prefer `storage: true` (always exactly one) or the highest version. Worth keeping in mind if you ever see this script pick a stale version.

PR #1795, opened 2026-06-02
fix: update preload-images path and trim to docker.io only

pdettori:
## Summary
- Move `PRELOAD_FILE` Makefile variable from `deployments/ansible/kind/preload-images.txt` to `scripts/kind/preload-images.txt` to match the current repo layout
- Remove ghcr.io and non-rate-limited images from the preload list — only docker.io images need pre-pulling to avoid rate-limiting during Kind cluster setup
- Add `kindest/kindnetd` and `kindest/local-path-provisioner` images that are pulled during Kind node startup

## Test plan
- [ ] Run `make preload-file` and verify it writes to `scripts/kind/preload-images.txt`
- [ ] Deploy a fresh Kind cluster and confirm all preloaded images are available

Assisted-By: Claude Code

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,7 +142,7 @@ help-ui:
 
 	
 # Define the path for the output file
-PRELOAD_FILE := deployments/ansible/kind/preload-images.txt
+PRELOAD_FILE := scripts/kind/preload-images.txt
 
 # The primary task to list and filter images
 .PHONY: preflight-check preload-file
--- a/scripts/kind/preload-images.txt
+++ b/scripts/kind/preload-images.txt
@@ -1,21 +1,10 @@
-# Docker Hub images pre-pulled on the host and loaded into Kind to avoid
-# rate-limiting during pod startup. Only docker.io images belong here —
-# ghcr.io and quay.io images are not rate-limited and pull fine on demand.
-#
-# Keep versions in sync with:
-#   - scripts/kind/setup-kagenti.sh  (ISTIO_VERSION)
-#   - charts/kagenti-deps/templates/ (phoenix, otel-collector)
-arizephoenix/phoenix:version-8.32.1
 docker.io/istio/install-cni:1.28.0-distroless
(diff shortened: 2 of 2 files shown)


mrsabath (approved):
Approving — the Makefile `PRELOAD_FILE` path was definitely stale (`deployments/ansible/kind/preload-images.txt` doesn't exist on main; `scripts/kind/preload-images.txt` does), and trimming ghcr.io images is consistent with the rate-limiting rationale (only docker.io is rate-limited). Phoenix is `enabled: false` in `dev_values.yaml`, so removing `arizephoenix/phoenix` is safe for the default `--with-all` run.

Non-blocking notes for follow-up:

1. **`kindest/kindnetd` and `kindest/local-path-provisioner` tags are very Kind-CLI-version-specific** (`v20251212-v0.29.0-alpha-105-g20ccfc88`). `.github/scripts/kind/create-cluster.sh` doesn't pin a `--image` for the cluster, so these tags only match when the runner's `kind` CLI version produces nodes with exactly these images. When the CI runner's kind upgrades, these preloads will silently mismatch and the cluster will pull from docker.io anyway. Worth either pinning the kind node image in `create-cluster.sh` or adding a comment noting this coupling.

2. **Unexplained nginx version downgrades** (`nginx-prometheus-exporter` 1.5.1→1.4.2 and `nginx-unprivileged` 1.29.5→1.29.1). I couldn't find these images referenced in any chart, template, or values file in the repo, so I can't verify which version the actual deployment pulls. If the new versions reflect what's actually being pulled at runtime, great; if not, preloading these tags is a no-op. A quick note in the PR body about where these versions come from would help future maintainers.

3. **The deleted header comment encoded the file's invariants** (only docker.io, keep versions in sync with chart templates, where to look for canonical versions). Consider adding a one-line replacement header so the next editor knows the policy.

Not worth blocking — Deploy & Test will catch a pure regression empirically. These are hygiene items for a follow-up.

PR #1804, opened 2026-06-03
fix: retry kubectl wait for CRD Established condition

pdettori:
## Summary

- Fixes intermittent `error: .status.conditions accessor error: <nil> is of the type <nil>, expected []interface{}` during Kind setup
- Root cause: `kubectl wait --for=condition=Established` fails if `.status.conditions` hasn't been populated yet on freshly-applied CRDs — a known kubectl race condition
- Introduces `_wait_crds_established()` helper that retries the wait up to 5 times with 2s backoff
- Applied to both Gateway API and agent-sandbox CRD waits

## Test plan

- [ ] Run `scripts/kind/setup-kagenti.sh` on a fresh Kind cluster — Gateway API CRDs should install without nil accessor errors
- [ ] Run with `--with-agent-sandbox` — sandbox CRD wait should also be resilient

Assisted-By: Claude Code

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -40,11 +40,11 @@ updates:
         update-types: [major]
     ignore:
       # cryptography 47+ crashes with SIGILL on ARM64 Kind clusters.
-      # Pin is intentional — block all bumps above the current upper bound.
-      # Effect: only patch bumps allowed until the SIGILL is resolved upstream.
+      # Pin is intentional — block major bumps above the current upper bound.
+      # Effect: only minor+patch bumps allowed until the SIGILL is resolved upstream.
       # TODO: remove once upstream fix lands (https://github.com/pyca/cryptography/issues/14764)
       - dependency-name: "cryptography"
-        update-types: [version-update:semver-major, version-update:semver-minor]
+        update-types: [version-update:semver-major]
 
   # npm (UI) — React frontend dependencies
   - package-ecosystem: npm
--- a/scripts/kind/setup-kagenti.sh
+++ b/scripts/kind/setup-kagenti.sh
@@ -281,6 +281,21 @@ _wait_deployment_ready() {
     log_warn "$label rollout not ready within timeout"
 }
 
+_wait_crds_established() {
+  # kubectl wait --for=condition=Established can fail with a nil accessor error
+  # if .status.conditions hasn't been populated yet on a freshly-applied CRD.
+  # Retry the wait to handle this race condition.
+  local timeout="${1:-60s}"; shift
+  local attempts=0 max_attempts=5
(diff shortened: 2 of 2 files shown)


scripts/kind/setup-kagenti.sh
```diff
@@ -230,6 +230,21 @@ _wait_deployment_ready() {
     log_warn "$label rollout not ready within timeout"
 }
 
+_wait_crds_established() {
+  # kubectl wait --for=condition=Established can fail with a nil accessor error
+  # if .status.conditions hasn't been populated yet on a freshly-applied CRD.
+  # Retry the wait to handle this race condition.
+  local timeout="${1:-60s}"; shift
+  local attempts=0 max_attempts=5
```
cwiklik, line 289: suggestion: `2>/dev/null` suppresses ALL stderr, not just the nil accessor race. If the failure is persistent (wrong CRD name, RBAC issue), the user gets 5 silent retries then a generic "not established" message with no root cause visible. Consider letting the last attempt print stderr — e.g. only redirect on retries 1–4 and let attempt 5 run without suppression.

.github/dependabot.yml
```diff
@@ -32,6 +32,11 @@ updates:
     groups:
       minor-and-patch:
         update-types: [minor, patch]
+    ignore:
+      # cryptography 47+ crashes with SIGILL on ARM64 Kind clusters.
+      # Pin is intentional — block all bumps above the current upper bound.
+      - dependency-name: "cryptography"
+        update-types: [version-update:semver-major, version-update:semver-minor]
```
cwiklik: nit: Ignoring `semver-minor` is broader than needed — the pin is `cryptography<47`, so minor bumps within 46.x are safe and wouldn't conflict. Only `semver-major` would hit the pin. This just reduces Dependabot noise, which may be intentional.

cwiklik (approved):
Clean fix for a known kubectl race condition where `kubectl wait --for=condition=Established` fails on freshly-applied CRDs with nil `.status.conditions`. The `_wait_crds_established()` helper with retry/backoff is the right pattern. The Dependabot cryptography ignore rule matches the existing `<47` pin in `pyproject.toml`.

**Areas reviewed**: Shell scripts, YAML (Dependabot)
**Commits**: 2, both signed-off ✅
**CI**: DCO pass, E2E pending (maintainer trigger) ✅

scripts/kind/setup-kagenti.sh
pdettori: Fixed — retries 1-4 still suppress stderr, but the final attempt now runs without `2>/dev/null` so persistent failures (wrong CRD name, RBAC, etc.) show the actual error.

.github/dependabot.yml
pdettori: Narrowed to `semver-major` only — 46.x minor bumps are safe with the `<47` pin.

PR #1791, opened 2026-06-02
feat: remove operator-managed resources from Helm chart

ChristianZaccaria:
Requires https://github.com/kagenti/kagenti-operator/pull/397 merged first.

## Summary
Removes resources from the kagenti Helm chart that the operator now manages directly. Eliminates the legacy keycloak-admin-secret bridge and authbridge-config ConfigMap templates, and cleans up the setup script.

## Changes
- Remove `authbridge-config` ConfigMap from `agent-namespaces.yaml` and `authbridge-template-configmaps.yaml` (operator reconciles this)
- Delete `keycloak-admin-secret-kagenti-system.yaml` (operator reads keycloak-initial-admin directly)
- Remove stale `adminUsername`/`adminPassword`/`adminExistingSecret` from values.yaml
- Remove `KC_ADMIN_FLAGS` block from setup-kagenti.sh (no longer needed)
- Update spireTrustDomain comment in values.yaml (auto-discovered by operator)

PR #1806, opened 2026-06-03
fix: wait for operator credentials secret before pod readiness check

### LINKED ISSUE #1805: bug: Deploy & Test flaky failure after PR #1791 removed keycloak-admin-secret from Helm chart
## Summary

Since PR #1791 (`feat: remove operator-managed resources from Helm chart`) was merged, the **Deploy & Test** CI check fails intermittently with a missing Kubernetes secret error. PR #1791 itself was merged with a failing Deploy & Test.

## Error

```
MountVolume.SetUp failed for volume "kagenti-keycloak-client-credentials-<hash>" :
  secret "kagenti-keycloak-client-credentials-<hash>" not found
```

The `weather-service` pod gets stuck in `ContainerCreating` because it tries to mount the client-credentials secret before the operator's Keycloak client registration controller creates it.

## Root Cause

PR #1791 removed `charts/kagenti/templates/keycloak-admin-secret-kagenti-system.yaml` from the Helm chart, expecting the operator to handle Keycloak client registration directly by reading the `keycloak-initial-admin` secret. However, there's now a race condition:

1. Helm deploys the weather-service Deployment (which references `kagenti-keycloak-client-credentials-*` as a volume mount)
2. The operator needs to start, discover the agent, register it with Keycloak, and create the credentials secret
3. If the pod starts before step 2 completes, it fails with `MountVolume.SetUp failed`

## Evidence

| Run | Branch | Result |
|-----|--------|--------|
| PR #1791 own CI | `create-keycloak-roles` | **Deploy & Test: FAILURE** (merged anyway) |
| PR #1803 run 1 (pre-rebase, before #1791 in base) | `fix/secret-values-path-1801` | pass |
| PR #1803 run 2 (post-rebase, includes #1791) | `fix/secret-values-path-1801` | fail |
| PR #1803 run 3 (re-run) | `fix/secret-values-path-1801` | fail |
| PR #1803 run 4 (re-run) | `fix/secret-values-path-1801` | fail |

Three consecutive failures on the same branch after rebasing onto main that includes #1791.

## Possible Fixes

1. **Add init-container or readiness gate** — make the weather-service pod wait for the credentials secret to exist before mounting
2. **Restore Helm-managed placeholder secret** — create an empty/placeholder secret in Helm that the operator later populates
3. **Increase deployment timeout** — give the operator more time to register the client (band-aid)
4. **Order dependency in Helm** — use a Job with `helm.sh/hook-weight` to ensure registration completes before agent deployments start

## Related

- Epic #1426: Eliminate Keycloak Admin Credentials and Client Secrets from Agent Pods
- PR #1421: Eliminate admin credentials from client registration using SPIFFE-based DCR
- PR #1791: The commit that introduced this regression

pdettori:
## Summary

Fixes the intermittent Deploy & Test CI failure introduced by PR #1791.

- **Root cause**: PR #1791 removed the Helm-managed `keycloak-admin-secret` from `kagenti-system`. The operator's `ClientRegistrationReconciler` now creates per-workload credential secrets asynchronously. The kagenti-webhook injects a volume mount for `kagenti-keycloak-client-credentials-<hash>` at pod admission time, but the pod stays in `ContainerCreating` until the operator finishes reconciliation and creates the Secret. No explicit wait existed for this.

- **Fix (41-wait-crds.sh)**: Wait for the `kagenti-operator` deployment to be Ready before proceeding to agent deployment scripts. Previously only CRD establishment was checked.

- **Fix (74-deploy-weather-agent.sh)**: Add explicit wait (up to 120s) for the `kagenti-keycloak-client-credentials-*` secret to appear in `team1` namespace, plus a `wait_for_deployment` readiness gate before proceeding. Previously the Kind path had no readiness check at all.

## Test plan

- CI Deploy & Test job should no longer fail with `MountVolume.SetUp failed for volume "kagenti-keycloak-client-credentials-*"`
- Run 3 consecutive CI cycles to confirm stability

Fixes #1805

Assisted-By: Claude Code

### DIFF
--- a/.github/scripts/kagenti-operator/74-deploy-weather-agent.sh
+++ b/.github/scripts/kagenti-operator/74-deploy-weather-agent.sh
@@ -148,33 +148,13 @@ log_success "Weather-service deployed via Deployment + Service (operator-indepen
 # Keycloak and creates the Secret.  Wait here to avoid a flaky timeout later.
 # ============================================================================
 log_info "Waiting for operator to create client credentials secret..."
-CREDS_TIMEOUT=120
-CREDS_FOUND=false
-for i in $(seq 1 $((CREDS_TIMEOUT / 5))); do
-    # grep -c exits 1 when count=0; wrap in subshell to avoid pipefail interaction
-    SECRET_COUNT=$(kubectl get secrets -n team1 -o name 2>/dev/null \
-        | { grep -c "kagenti-keycloak-client-credentials" || true; })
-    if [ "${SECRET_COUNT:-0}" -ge 1 ]; then
-        CREDS_FOUND=true
-        log_success "Client credentials secret created by operator"
-        break
-    fi
-    echo "[$i/$((CREDS_TIMEOUT / 5))] Credentials secret not yet created, waiting..."
-    sleep 5
-done
-
-if [ "$CREDS_FOUND" != "true" ]; then
-    log_warn "Credentials secret not found after ${CREDS_TIMEOUT}s — pod may be stuck in ContainerCreating"
-    log_info "All pods in kagenti-system:"
+if ! kubectl -n team1 wait --for=create secret/kagenti-keycloak-client-credentials --timeout=120s 2>/dev/null; then
+    log_warn "Credentials secret not found after 120s — pod may be stuck in ContainerCreating"
     kubectl get pods -n kagenti-system 2>&1 || true
-    log_info "All deployments in kagenti-system:"
-    kubectl get deployments -n kagenti-system 2>&1 || true
-    log_info "Operator logs (by any kagenti label):"
-    kubectl logs -n kagenti-system -l app.kubernetes.io/instance=kagenti --tail=20 2>&1 || true
-    log_info "Secrets in team1:"
     kubectl get secrets -n team1 2>&1 || true
-    log_info "keycloak-admin-secret in kagenti-system:"
-    kubectl get secret keycloak-admin-secret -n kagenti-system 2>&1 || echo "  NOT FOUND"
+    kubectl logs -n kagenti-system -l app.kubernetes.io/instance=kagenti --tail=20 2>&1 || true
+else
+    log_success "Client credentials secret created by operator"
 fi
 
 # WORKAROUND: Fix Service targetPort mismatch
--- a/.github/scripts/kagenti-operator/74-deploy-weather-agent.sh
+++ b/.github/scripts/kagenti-operator/74-deploy-weather-agent.sh
@@ -151,25 +151,30 @@ log_info "Waiting for operator to create client credentials secret..."
 CREDS_TIMEOUT=120
 CREDS_FOUND=false
 for i in $(seq 1 $((CREDS_TIMEOUT / 5))); do
+    # grep -c exits 1 when count=0; wrap in subshell to avoid pipefail interaction
     SECRET_COUNT=$(kubectl get secrets -n team1 -o name 2>/dev/null \
-        | grep -c "kagenti-keycloak-client-credentials" || echo "0")
-    if [ "$SECRET_COUNT" -ge 1 ]; then
+        | { grep -c "kagenti-keycloak-client-credentials" || true; })
+    if [ "${SECRET_COUNT:-0}" -ge 1 ]; then
         CREDS_FOUND=true
         log_success "Client credentials secret created by operator"
         break
     fi
-    echo "[$i/$((CREDS_TIMEOUT / 5))] Credentials secret not yet created, waiting... (${SECRET_COUNT} found)"
+    echo "[$i/$((CREDS_TIMEOUT / 5))] Credentials secret not yet created, waiting..."
     sleep 5
 done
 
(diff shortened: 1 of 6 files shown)


.github/scripts/kagenti-operator/74-deploy-weather-agent.sh
```diff
@@ -140,6 +140,43 @@ kubectl apply -f "$REPO_ROOT/kagenti/examples/agents/weather_service_service.yam
...
 
+# ============================================================================
+# Step 3: Wait for operator to create the client credentials secret
+# The webhook injects a volume mount for kagenti-keycloak-client-credentials-*
+# into the pod at admission time.  The pod stays in ContainerCreating until
+# the operator's ClientRegistrationReconciler registers the workload in
+# Keycloak and creates the Secret.  Wait here to avoid a flaky timeout later.
+# ============================================================================
+log_info "Waiting for operator to create client credentials secret..."
+CREDS_TIMEOUT=120
```
esnible: You should be able to do `kubectl -n team1 wait --for=create secret kagenti-keycloak-client-credentials --timeout=120s` and replace all this code

esnible (commented):
## Summary

Fix correctly addresses the CI regression from #1791 — verified against `upstream/main` that `authbridge-config` was indeed missing in both the per-namespace template and the release-namespace template, and that `keycloak-admin-secret` was no longer Helm-managed.

**Areas reviewed**: Helm templates, CI shell scripts, values.yaml
**Commits**: 6, all signed-off, all use `Assisted-By:` (per repo policy)
**CI status**: all green (incl. Deploy & Test and E2E HyperShift)

**Verdict**: COMMENT — no must-fix issues. Two suggestions inline; both are about consistency rather than correctness of the immediate fix.

charts/kagenti/templates/agent-namespaces.yaml
```diff
@@ -108,6 +108,26 @@ data:
...
+  namespace: {{ . }}
+data:
+  KEYCLOAK_URL: "http://keycloak-service.{{ $root.Values.keycloak.namespace }}.svc:8080"
+  KEYCLOAK_REALM: {{ $root.Values.keycloak.realm | quote }}
+  KEYCLOAK_NAMESPACE: "{{ $root.Values.keycloak.namespace }}"
+  ISSUER: "{{ $root.Values.keycloak.publicUrl }}/realms/{{ $root.Values.keycloak.realm }}"
+  SPIRE_ENABLED: {{ if $root.Values.spire.enabled }}"true"{{ else }}"false"{{ end }}
+  CLIENT_AUTH_TYPE: "{{ $root.Values.authBridge.clientAuthType }}"
+  SPIFFE_IDP_ALIAS: "{{ $root.Values.authBridge.spiffeIdpAlias }}"
+  EXPECTED_AUDIENCE: "{{ $root.Values.keycloak.publicUrl }}/realms/{{ $root.Values.keycloak.realm }}"
```
esnible, line 128: **suggestion** — The per-namespace `authbridge-config` defined here is missing `CREDENTIAL_WAIT_TIMEOUT`, which the parallel definition added to `authbridge-template-configmaps.yaml` (release-namespace, line 216 in the diff) does include.

Result: namespaces created at chart-install time will have a different key set than namespaces populated later by the operator's propagation logic (the template-configmaps copy). If the AuthBridge sidecar reads this key in any namespace, the asymmetry will manifest as inconsistent behavior depending on whether the namespace was created by Helm or by the operator.

Either add `CREDENTIAL_WAIT_TIMEOUT: {{ .Values.authBridge.credentialWaitTimeout | quote }}` here too, or drop it from the template-configmaps version — the two should be in sync since they're documented to produce identical content (see the comment at the top of `authbridge-template-configmaps.yaml`).

charts/kagenti/values.yaml
```diff
@@ -274,6 +274,7 @@ kagenti-operator-chart:
       - "--webhook-cert-path=/tmp/k8s-webhook-server/serving-certs"
       - "--config-path=/etc/kagenti/config.yaml"
       - "--feature-gates-path=/etc/kagenti/feature-gates/feature-gates.yaml"
+      - "--spire-trust-domain=localtest.me"
```
esnible: **suggestion** — Hardcoding `--spire-trust-domain=localtest.me` here contradicts the comment four lines below: *"SPIRE trust domain is auto-discovered by the operator from ZTWIM CR or spire-bundle ConfigMap."*

On non-Kind environments (OCP/HyperShift) `localtest.me` is not the correct trust domain, and there's no override in `deployments/envs/ocp_values.yaml` for `kagenti-operator-chart.controllerManager.container.args`. The commit message (`8f0ddb3`) notes that SPIRE-derived client IDs *require* this flag — which is exactly the case where Kind-only hardcoding will silently regress OCP.

Consider templating from a values key (e.g. `.Values.spire.trustDomain` or a new `kagenti-operator-chart.spireTrustDomain`) so each env can override. As-is, this PR fixes Kind CI but may break HyperShift the next time the operator's `ClientRegistrationReconciler` is exercised end-to-end on a non-`localtest.me` trust domain.

esnible (approved):

PR #1810, opened 2026-06-03
feat(chart): add optional SPARC reflection service component

OsherElhadad:
Adds the sparc-service Deployment, Service, and ConfigMap as an optional Helm component (components.sparcService.enabled, off by default) that backs the AuthBridge sparc plugin. The reflection LLM provider is configurable through values: watsonx by default, or ollama, openai, azure, or any litellm model.

Fixes #


huang195:
@OsherElhadad this might not be the right repo for these artifacts. For now, maybe a more appropriate place is in kagenti-extensions, under sparc directory. It could be mentioned in the instructions that these need to be deployed before sparc plugin can be enabled. What do you think?

PR #1829, opened 2026-06-05
fix(openshell): make kagenti-backend opt-in with --with-backend

pdettori:
## Summary

- Flip `STEP_BACKEND` default from `true` to `false` in `deploy-shared.sh` — most OpenShell deployments (especially on OpenShift where the full Kagenti platform is deployed separately) don't need a standalone `kagenti-backend` + PostgreSQL
- Introduce `--with-backend` flag for explicit opt-in
- Keep `--backend` and `--skip-backend` as backwards-compat aliases (both resolve to off)
- Fix broken `registry.redhat.io/rhel9/postgresql-16:1-123` tag → `:latest` (manifest unknown)

## Test plan

- [ ] `scripts/openshell/deploy-shared.sh --dry-run` shows `Backend + sessions: false`
- [ ] `scripts/openshell/deploy-shared.sh --with-backend --dry-run` shows `Backend + sessions: true`
- [ ] `--backend` and `--skip-backend` still parse without error
- [ ] CI (`openshell-full-test.sh`) passes — already handles missing backend dynamically

Assisted-By: Claude Code

### DIFF
--- a/deployments/sandbox/postgres-sessions.yaml
+++ b/deployments/sandbox/postgres-sessions.yaml
@@ -47,7 +47,7 @@ spec:
           type: RuntimeDefault
       containers:
       - name: postgres
-        image: registry.redhat.io/rhel9/postgresql-16:1-123
+        image: registry.redhat.io/rhel9/postgresql-16:latest
         securityContext:
           runAsNonRoot: true
           allowPrivilegeEscalation: false
--- a/deployments/sandbox/postgres-sessions.yaml
+++ b/deployments/sandbox/postgres-sessions.yaml
@@ -47,7 +47,7 @@ spec:
           type: RuntimeDefault
       containers:
       - name: postgres
-        image: registry.redhat.io/rhel9/postgresql-16:latest
+        image: registry.redhat.io/rhel9/postgresql-16:1
         securityContext:
           runAsNonRoot: true
           allowPrivilegeEscalation: false
--- a/scripts/openshell/deploy-shared.sh
+++ b/scripts/openshell/deploy-shared.sh
@@ -94,7 +94,7 @@ while [[ $# -gt 0 ]]; do
     --litellm)          STEP_LITELLM=true; shift ;;
     --pre-pull)         STEP_PREPULL=true; shift ;;
     --with-backend)     STEP_BACKEND=true; shift ;;
-    --backend|--skip-backend) STEP_BACKEND=false; shift ;;
+    --backend|--skip-backend) STEP_BACKEND=false; shift ;;  # compat aliases (no-op, kept for existing scripts)
     --kind-cluster)     KIND_CLUSTER="$2"; shift 2 ;;
     --dry-run)          DRY_RUN=true; shift ;;
     *)
@@ -115,7 +115,7 @@ echo "  cert-manager CA:    $STEP_TLS"
 echo "  Keycloak realm:     $STEP_KEYCLOAK"
 echo "  LiteLLM proxy:      $STEP_LITELLM"
 echo "  Image pre-pull:      $STEP_PREPULL"
-echo "  Backend + sessions: $STEP_BACKEND"
+echo "  Backend + sessions: $STEP_BACKEND$( $STEP_BACKEND || echo ' (use --with-backend to enable)' )"
 echo "  Keycloak namespace: $KEYCLOAK_NS"
 echo "  Dry run:            $DRY_RUN"
 echo ""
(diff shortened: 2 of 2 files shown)


deployments/sandbox/postgres-sessions.yaml
```diff
@@ -47,7 +47,7 @@ spec:
           type: RuntimeDefault
       containers:
       - name: postgres
-        image: registry.redhat.io/rhel9/postgresql-16:1-123
+        image: registry.redhat.io/rhel9/postgresql-16:latest
```
mrsabath: **suggestion**: `:latest` is fine functionally, but it pins the sandbox to whatever Red Hat happens to publish that day — reproducibility regresses vs. an explicit tag. Two more stable options: `1` (rolling major-tag), or a Red Hat–published digest from [catalog.redhat.com/.../rhel9/postgresql-16](https://catalog.redhat.com/software/containers/rhel9/postgresql-16). Either avoids the silent-bump-on-pull problem you'd otherwise re-hit. Not blocking — just a nudge if you want long-term repeatability.

scripts/openshell/deploy-shared.sh
```diff
@@ -45,7 +45,7 @@ STEP_TLS=true
 STEP_KEYCLOAK=true
 STEP_LITELLM=false
 STEP_PREPULL=false
-STEP_BACKEND=true
+STEP_BACKEND=false
```
mrsabath, line 48: **nit (off-diff but related)**: the startup banner at line 119 (`echo "  Backend + sessions: $STEP_BACKEND"`) doesn't hint at `--with-backend` when off. Could read `Backend + sessions: false (use --with-backend to enable)` to save a docs lookup. Minor.

scripts/openshell/deploy-shared.sh
```diff
@@ -94,8 +93,8 @@ while [[ $# -gt 0 ]]; do
     --keycloak-ns)      KEYCLOAK_NS="$2"; shift 2 ;;
     --litellm)          STEP_LITELLM=true; shift ;;
     --pre-pull)         STEP_PREPULL=true; shift ;;
-    --backend)          STEP_BACKEND=true; shift ;;
-    --skip-backend)     STEP_BACKEND=false; shift ;;
+    --with-backend)     STEP_BACKEND=true; shift ;;
+    --backend|--skip-backend) STEP_BACKEND=false; shift ;;
```
mrsabath: Nice touch keeping `--backend` and `--skip-backend` as no-op aliases — anything in user shell history or wrapper scripts continues to parse cleanly. One observation: the `Usage` block (line 77 in the new version) only documents `--with-backend`, so the aliases are silent. That's probably fine (less surface area in `--help`), but if you want to be explicit you could add `# Aliases: --backend, --skip-backend (both no-op, kept for compat)` as a comment near the case branch.

mrsabath (approved):
Bel lavoro, Paolo — small, surgical, and consistent with `scripts/kind/setup-kagenti.sh`'s `--with-*` convention. Backwards-compat aliases keep existing call sites safe, and `openshell-full-test.sh:362` already gates backend tests on `kubectl get deploy kagenti-backend`, so CI is unaffected.

**Areas reviewed:** Shell, K8s YAML
**Commits:** 1, signed-off, conventional-commits format ✓
**CI status:** All passing checks green; deploy/E2E/HyperShift pending maintainer `/run-e2e` (expected)

Left a few **non-blocking** suggestions inline + two off-diff observations below. None of them should hold this up.

### Off-diff observations

**1. `scripts/openshell/cleanup-shared.sh:47` — asymmetric default with deploy after this PR.** Cleanup still has `STEP_BACKEND=true` (opt-out via `--skip-backend`). After this PR, deploy is opt-in but cleanup is opt-out. Worth a follow-up — flipping cleanup's default has the opposite footgun (orphaned PostgreSQL PVCs after a `cleanup-shared.sh` that the user expected to be no-op for backend), so it deserves its own thought.

**2. `docs/sandbox-guide.md:67-69` — doc still shows bare `deploy-shared.sh`.** That command no longer brings up the backend. If the sandbox flow expects sessions/DB, the doc should say `--with-backend`. Probably a follow-up doc PR.

Approving as-is. Grazie!

pdettori:
## CI Failure Analysis — OpenShell PoC (Kind)

**Not caused by this PR.** Pre-existing test design issue.

### Failing tests
- `test_sandbox_exec_basic`
- `test_sandbox_exec_shell_interactive`

### Root cause

`_find_exec_pod()` picks the first running pod in the namespace and execs into its first container. When `openshell-server-0` is selected, that container is `gateway` — a distroless image with no `echo` or `sh`:

```
OCI runtime exec failed: exec: "echo": executable file not found in $PATH
```

### Why this isn't related to the PR

This PR only modifies `scripts/openshell/deploy-shared.sh` and `deployments/sandbox/postgres-sessions.yaml` — no image or container spec changes. The gateway image (`ghcr.io/kagenti/openshell/gateway:mvp-v2-3a1bbea`) is unchanged from prior commits.

### Follow-up needed

The test should explicitly target sandbox/agent pods rather than grabbing the first available pod. Will address in a separate fix.

---
*Assisted-By: Claude Code*

PR #1833, opened 2026-06-05
fix(openshell): remove cert-manager Certificate to fix TLS race condition

pdettori:
## Summary

- Remove `charts/openshell/templates/certificate.yaml` — the cert-manager Certificate resources
  were overwriting TLS secrets created by the certgen Helm hook with certs signed by a different CA
- Remove unused `tls.*` values from `values.yaml`
- Simplify `deploy-tenant.sh` Step 3 — replace the complex cert-manager retry loop with a simple
  secret-existence check (certgen hook now handles all TLS)

## Root Cause

The chart had **two conflicting TLS generators** targeting the same secrets (`openshell-server-tls`,
`openshell-client-tls`):

1. `certgen.yaml` — Helm pre-install hook that generates a self-signed CA + leaf certs
2. `certificate.yaml` — cert-manager Certificates signed by `openshell-ca-issuer` (different CA)

Race condition:
1. certgen hook runs → creates secrets with CA-A
2. StatefulSet starts → gateway loads certs signed by CA-A  
3. cert-manager reconciles → **overwrites** secrets with certs signed by CA-B
4. Sandbox pod starts → mounts CA-B → rejects gateway's CA-A cert → `UnknownCA`

## Fix

Remove `certificate.yaml` so certgen is the sole TLS source. Helm hook ordering
guarantees certgen completes before the StatefulSet deploys, eliminating the race.

## Test plan

- [ ] Deploy with `scripts/openshell/deploy-tenant.sh team1` on Kind
- [ ] Verify `openshell sandbox create` succeeds (no `UnknownCA` errors)
- [ ] Verify gateway pod serves cert matching the CA in the secret:
  ```bash
  kubectl get secret -n team1 openshell-server-tls -o jsonpath='{.data.ca\.crt}' | base64 -d > /tmp/ca.pem
  kubectl port-forward -n team1 svc/openshell-server 18080:8080 &
  echo | openssl s_client -connect localhost:18080 -CAfile /tmp/ca.pem 2>/dev/null | grep "Verify return code"
  # Expected: Verify return code: 0 (ok)
  ```

Assisted-By: Claude Code


scripts/openshell/deploy-tenant.sh
```diff
@@ -248,77 +248,17 @@ else
 fi
 echo ""
 
-# ── Step 3: Wait for certificates ──────────────────────────────────────────
-log_info "Step 3: Waiting for cert-manager certificates"
+# ── Step 3: Verify TLS secrets ─────────────────────────────────────────────
+log_info "Step 3: Verifying TLS secrets (created by certgen hook)"
```
mrsabath, line 252: **suggestion** (rotation): the deleted `Certificate` had `duration: 8760h` (1 year) with `renewBefore: 720h` (30 days), giving cert-manager auto-renew. With certgen, the secrets are created on `pre-install` / `pre-upgrade`, so rotation now requires a `helm upgrade` to re-run the hook. For an MVP that's a fine trade — but worth either a one-line note here (or in the gateway/install docs) so operators know the rotation model changed, or filing a tracking issue for built-in rotation when this graduates beyond MVP. Not a blocker for this PR.

mrsabath (approved):
Bel taglio, Paolo — netto, motivato, e i fatti tornano. I verified the key claims:

- ✅ certgen hook (`charts/openshell/templates/certgen.yaml`) writes both `openshell-server-tls` and `openshell-client-tls` with matching SANs (server, `.{tenant}`, `.svc`, `.svc.cluster.local`, `localhost`, `ingress.host`).
- ✅ Hook ordering: certgen Job uses `helm.sh/hook: pre-install,pre-upgrade` with `hook-weight: -20` (RBAC at `-30`), so it runs before the StatefulSet. Race eliminated.
- ✅ `gatewayJwt.enabled: true` is the default in `values.yaml:46`, so certgen always runs.
- ✅ `.Values.tls.*` is referenced *only* in the file being deleted — no leakage.
- ✅ −112 lines net; deploy/E2E pending the maintainer `/run-e2e` trigger.

**Areas reviewed:** Helm chart, Shell
**Commits:** 3 commits, all signed-off ✓ — net diff is only the 3 TLS files (the first two commits' content is already on main via `f8ba6ab5`/merge from upstream). Squash-on-merge would leave a single clean commit.
**CI status:** all green; deploy/E2E gated on maintainer trigger (expected).

Left a couple of **non-blocking** suggestions inline + two off-diff observations below. Approving as-is.

### Off-diff observations

**1. Stale `openshell-ca-issuer` setup/teardown.** `scripts/openshell/deploy-shared.sh:335-348` still installs a `ClusterIssuer openshell-ca-issuer`, and `cleanup-shared.sh:238` removes it. After this PR, **nothing in the chart consumes the issuer anymore** — it's installed and torn down for no purpose. Candidate for a follow-up cleanup PR.

**2. Research doc references.** `docs/research/openshell-mvp.md` (lines 791, 841, 867, 884) still references the deleted `tls.*` values + cert-manager flow as design. It's a research note rather than user docs, so deferring is fine — just flagging.

PR #1834, opened 2026-06-06
chore(openshell): bump gateway/supervisor to v0.0.56, document egress policies

pdettori:
## Summary

- Bump OpenShell gateway and supervisor images from `mvp-v2-3a1bbea` to `v0.0.56`
- Document network egress policies in `docs/sandbox-guide.md` (create with `--policy`, dynamic updates, git clone example)

## Motivation

v0.0.56 includes [kagenti/OpenShell#10](https://github.com/kagenti/OpenShell/pull/10) which fixes the sandbox egress 403 bug — the CONNECT proxy now falls through to OPA on identity resolution failure, and policies without `binaries` restrictions allow any binary.

## Changes

| File | Change |
|------|--------|
| `charts/openshell/values.yaml` | gateway + supervisor → `v0.0.56` |
| `charts/kagenti/values.yaml` | gateway → `v0.0.56` (compute-driver stays at `v0.1.0-rc.4`) |
| `docs/sandbox-guide.md` | New "Network Egress Policies" section |

## Test plan

- [x] Deployed to Kind cluster, created sandbox with `--policy github-egress.yaml -- sleep infinity`
- [x] Verified `curl https://github.com` returns 200 inside sandbox
- [x] Verified `git clone https://github.com/kagenti/kagenti.git` works
- [x] Verified `curl https://example.com` (not in policy) still returns 403
- [x] Verified dynamic `policy update --add-endpoint` / `--remove-endpoint` works on live sandbox

Assisted-By: Claude Code

### DIFF
--- a/charts/kagenti/values.yaml
+++ b/charts/kagenti/values.yaml
@@ -390,7 +390,7 @@ openshell:
     pullPolicy: IfNotPresent
   computeDriver:
     image: ghcr.io/kagenti/openshell-driver-openshift/compute-driver
-    tag: v0.0.56
+    tag: v0.1.0-rc.4
     pullPolicy: IfNotPresent
   credentialsDriver:
     image: ghcr.io/kagenti/openshell-credentials-keycloak/credentials-driver
--- a/charts/kagenti/values.yaml
+++ b/charts/kagenti/values.yaml
@@ -386,11 +386,11 @@ openshell:
   enabled: false
   gateway:
     image: ghcr.io/kagenti/openshell/gateway
-    tag: mvp-00a418f
+    tag: v0.0.56
     pullPolicy: IfNotPresent
   computeDriver:
     image: ghcr.io/kagenti/openshell-driver-openshift/compute-driver
-    tag: mvp-8c86191
+    tag: v0.0.56
     pullPolicy: IfNotPresent
   credentialsDriver:
     image: ghcr.io/kagenti/openshell-credentials-keycloak/credentials-driver
--- a/charts/openshell/values.yaml
+++ b/charts/openshell/values.yaml
@@ -5,7 +5,7 @@ tenant: ""
 images:
   gateway:
     repository: ghcr.io/kagenti/openshell/gateway
-    tag: mvp-v2-3a1bbea
+    tag: v0.0.56
     pullPolicy: IfNotPresent
   computeDriver:
     repository: ghcr.io/kagenti/openshell-driver-openshift/compute-driver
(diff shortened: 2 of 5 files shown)


pdettori (commented):
**Self-review notes**

Clean image-tag bump from dev-hash tags to semver releases, accompanied by good docs and a test fix. All commits are DCO-signed and follow conventional commit format.

**CI**: The "OpenShell PoC (Kind)" job has 2 failures in `test_T1_7_sandbox_connectivity` — `kubectl exec` into the gateway container fails because `echo`/`sh` are not in `$PATH` on `v0.0.56`. This suggests the new image is scratch/distroless. The tests assume a shell exists in the gateway container. Since these test `kubectl exec` (not egress policy), this is likely a pre-existing test gap now surfaced by the distroless image — worth confirming or skipping those tests for distroless gateways.

The "Deploy & Test" failure is at "Deploy weather-service Agent" — unrelated to OpenShell.

**Nit**: PR description says "compute-driver stays at v0.1.0-rc.4" but on `main` it's `mvp-8c86191`, so this is actually a version bump to `v0.1.0-rc.4`. Minor description inaccuracy.

huang195 (approved):
## Perfetto, Paolo! 🇮🇹

Clean, well-scoped bump. I verified the load-bearing bits instead of just eyeballing the tags:

- **Tag consistency** ✓ — `gateway` and `supervisor` → `v0.0.56` in both `charts/kagenti/values.yaml` and `charts/openshell/values.yaml`, and this aligns `charts/kagenti`'s `computeDriver` to the `v0.1.0-rc.4` that `charts/openshell` was already pinning. No skew between the two charts — bellissimo.
- **No tag left behind** — the second `gateway:`/`computeDriver:` block in `charts/openshell/values.yaml` is the `resources:` section (CPU/memory), not images, so nothing was missed.
- **Test T1_7 distroless skip** ✓ — confirmed the gateway runs as the `openshell-server` StatefulSet (container named `gateway`), so the `("openshell-server",)` prefix correctly skips the distroless v0.0.56 pods that lack a shell. Right fix for the right reason.
- **Test T4_3 ↔ docs parity** ✓ — `/dev/null` + `/dev/urandom` added to `read_write` matches the new docs example exactly; the two won't drift.
- The egress-policy walkthrough in `sandbox-guide.md` is clear and genuinely useful — the dynamic `policy update`/`set`/`get` section especially.

One small, non-blocking note (out of scope — just flagging while we're in the neighborhood): `credentialsDriver` still floats on `main-d7d8306` while everything else moved to versioned tags. Whenever it's convenient, pinning it to a release tag would make the OpenShell image set fully reproducible. Not for this PR.

LGTM — grazie! Just let CI go green before merging (Deploy & Test / OpenShell PoC were still pending at review time).

**Areas reviewed:** Helm values (both charts), docs, Python e2e tests. **Commits:** 6, all signed-off. **CI:** pending (not failing).

PR #1841, opened 2026-06-08
chore: remove MLflow UI config patching from setup script (now handled by operator)

ChristianZaccaria:
## Summary
Removes the MLflow dashboard URL patching from `setup-kagenti.sh`. This is now handled by the kagenti-operator at startup via `UIConfigBootstrapRunnable` (see [kagenti-operator#411](https://github.com/kagenti/kagenti-operator/pull/411)).

## Changes
- `scripts/ocp/setup-kagenti.sh` — removed the wait-and-patch block for `kagenti-ui-config` ConfigMap

## Test Plan
- [x] Operator patches `MLFLOW_DASHBOARD_URL` on startup (verified on OCP with RHOAI)
- [ ] Script still deploys MLflow OAuth proxy and Route correctly

### DIFF
--- a/scripts/ocp/setup-kagenti.sh
+++ b/scripts/ocp/setup-kagenti.sh
@@ -1210,30 +1210,6 @@ spec:
     termination: reencrypt
 OAUTH_EOF
   fi
-
-  # Wait for the oauth-proxy to become ready
-  local tries=0
-  while ! $KUBECTL get pods -n "$MLFLOW_NAMESPACE" \
-    -l app=mlflow-oauth-proxy -o jsonpath='{.items[0].status.phase}' 2>/dev/null | grep -q Running; do
-    tries=$((tries + 1))
-    if [ $tries -ge 24 ]; then
-      log_warn "MLflow OAuth proxy not ready after 2m — dashboard link may not work"
-      break
-    fi
-    sleep 5
-  done
-
-  local mlflow_host=""
-  mlflow_host=$($KUBECTL get route mlflow -n "$MLFLOW_NAMESPACE" \
-    -o jsonpath='{.spec.host}' 2>/dev/null || echo "")
-  if [ -n "$mlflow_host" ]; then
-    local mlflow_url="https://${mlflow_host}/mlflow/"
-    $KUBECTL patch configmap kagenti-ui-config -n kagenti-system \
-      --type merge -p "{\"data\":{\"MLFLOW_DASHBOARD_URL\":\"${mlflow_url}\"}}" 2>/dev/null || true
-    log_success "MLflow dashboard URL: $mlflow_url"
-  else
-    log_warn "Could not resolve MLflow Route host — UI link will be empty"
-  fi
 }
 log_info "Step 3d: MLflow provisioning (deferred)"
 _deferred_mlflow


scripts/ocp/setup-kagenti.sh
```diff
@@ -1210,30 +1210,6 @@ spec:
...
-    -o jsonpath='{.spec.host}' 2>/dev/null || echo "")
-  if [ -n "$mlflow_host" ]; then
-    local mlflow_url="https://${mlflow_host}/mlflow/"
-    $KUBECTL patch configmap kagenti-ui-config -n kagenti-system \
-      --type merge -p "{\"data\":{\"MLFLOW_DASHBOARD_URL\":\"${mlflow_url}\"}}" 2>/dev/null || true
-    log_success "MLflow dashboard URL: $mlflow_url"
-  else
-    log_warn "Could not resolve MLflow Route host — UI link will be empty"
-  fi
 }
```
cwiklik, line 1213: **suggestion:** The removed block actively **waited up to 2m** for the oauth-proxy/Route to be ready before patching. Its replacement — the operator's `UIConfigBootstrapRunnable` (#411) — runs **once at startup and silently no-ops if the MLflow CR isn't `Available` yet** (I flagged exactly this on #411, L52). On a cold install where MLflow lags operator startup, `MLFLOW_DASHBOARD_URL` may stay empty until the operator restarts — a regression versus the wait-and-retry this script did. Recommend confirming #411 handles the not-ready-yet case (watch/requeue) before removing the script's wait.

cwiklik (approved):
Correct, clean removal — the deleted block is self-contained and `_deferred_mlflow` still closes properly (Shell Script Lint passes). The diff itself is fine; the concerns are about the cross-repo migration to operator #411, not this change:

1. **Readiness timing (inline):** the removed block actively waited up to 2m for oauth-proxy/Route readiness before patching; the operator's `UIConfigBootstrapRunnable` (#411) runs once at startup and no-ops if MLflow isn't `Available` yet (flagged on #411, L52). On a cold install where MLflow lags operator startup, `MLFLOW_DASHBOARD_URL` may stay empty until the operator restarts.
2. **URL value parity — please verify:** the script wrote `https://<mlflow-route-host>/mlflow/` (note the `/mlflow/` path), whereas #411 writes `<MLflow CR status.url>` trimmed + `/` (its unit test yields `https://mlflow-gateway.apps.example.com/` — no `/mlflow/` path). If the CR `status.url` isn't already the `/mlflow/`-suffixed dashboard URL, the UI link will differ. The test plan confirms the operator *patches* the key on OCP — worth confirming the *value* still opens the dashboard.
3. **Cross-repo ordering:** this removes the script fallback in `kagenti/kagenti`, but #411 is in `kagenti/kagenti-operator`. Ensure the operator version the installer deploys includes #411 before this lands — otherwise nothing patches `MLFLOW_DASHBOARD_URL` in the gap.

**Security:** none (pure deletion). **Commits:** signed-off. Nit: `chore:` lacks the emoji prefix the convention favors; the 2nd test-plan box (script still deploys OAuth proxy/Route) is unchecked. All CI green.

scripts/ocp/setup-kagenti.sh
ChristianZaccaria: Thanks for the suggestion. I've added a a 4min window to retry. If MLflow takes significantly longer, a pod restart or manual reconciliation could handle this.

pdettori (approved):
LGTM — the removed block only patched `MLFLOW_DASHBOARD_URL` into the UI ConfigMap, which is now handled by kagenti-operator's `UIConfigBootstrapRunnable` (PR kagenti-operator#411, merged). MLFlow install, OAuth proxy deployment, and Route creation are all untouched by this change.

Verified that `--with-all` deploy order is safe: `_deferred_mlflow()` (Step 3d) creates the MLFlow Route **before** `helm install kagenti` (Step 4) deploys the operator, so the operator finds the Route at startup.

Assisted-By: Claude Code

PR #1856, opened 2026-06-08
Refactor: remove Keycloak operand/realm from chart (moved to operator)

ChristianZaccaria:
Companion to kagenti/kagenti-operator#412

## Summary

Removes OpenShift Keycloak resources from the kagenti-deps Helm chart that are now managed by the kagenti-operator's `KeycloakBootstrapRunnable`. The Kind path (ConfigMap-based realm import) is preserved.

## Changes
- `charts/kagenti-deps/templates/keycloak-operand.yaml` — deleted (Keycloak CR now created by operator)
- `charts/kagenti-deps/templates/keycloak-postgres.yaml` — deleted (Postgres infra now created by operator)
- `charts/kagenti-deps/templates/keycloak-realm-init.yaml` — removed OpenShift section (Secret + KeycloakRealmImport), kept Kind ConfigMap
- `scripts/ocp/setup-kagenti.sh` — updated comment reflecting operator ownership

## Test Plan
- [ ] Deploy on OpenShift with operator — verify Keycloak stack comes up via operator
- [ ] Deploy on Kind — verify ConfigMap realm import still works
- [ ] Confirm no duplicate resources between chart and operator


scripts/ocp/setup-kagenti.sh
```diff
@@ -698,8 +698,8 @@ EOF
     log_info "MLflow OTEL values: otel.mlflow.enabled=true, endpoint=${MLFLOW_TRACES_ENDPOINT}"
   fi
 
-  # Keycloak public URL is needed by the realm-init audience mapper.
-  # Construct from DOMAIN (known since Step 2) so it's correct on first install.
+  # Keycloak public URL — realm-init audience mapper is now handled by
+  # kagenti-operator, but the value is still passed for chart compatibility.
```
pdettori, line 659: The comment says realm-init is now handled by the operator, but `_wait_kagenti_deps_ready()` (line ~596) still has:
```bash
_wait_deployment_ready postgres-kc "$KC_NAMESPACE" "Keycloak PostgreSQL" statefulset
_wait_deployment_ready keycloak "$KC_NAMESPACE" Keycloak statefulset
```
These resources are deleted by this PR (from `keycloak-postgres.yaml` and `keycloak-operand.yaml`). Without the operator PR merged, the install script will hang here indefinitely.

Similarly, `_apply_operand_crs()` (line ~641) extracts post-install hooks from `kagenti-deps` and applies them — the deleted `keycloak-operand.yaml` was a post-install hook that created the Keycloak CR.

**This PR must not merge before kagenti/kagenti-operator#412.**

charts/kagenti-deps/templates/keycloak-realm-init.yaml
```diff
@@ -2,251 +2,14 @@
...
   Creates the demo realm with roles and test users (admin, dev-user, ns-admin).
 
   Kind:      ConfigMap mounted into Keycloak pod via --import-realm
-  OpenShift: KeycloakRealmImport CR managed by RHBK operator
-
-  NOTE: The UI OAuth client is currently registered in the MASTER realm.
-  These demo realm users are for future use when the backend migrates to
-  the demo realm. For current UI login, run kagenti/auth/create-test-users.sh
-  to create users in the master realm.
+  OpenShift: Managed by kagenti-operator (KeycloakRealmImport CR + kagenti-test-users Secret)
```
pdettori, line 9: The deleted `kagenti-test-users` Secret (random passwords for admin/dev-user/ns-admin) was consumed by `show-services.sh` to display credentials on OpenShift. Verify that the operator's `KeycloakBootstrapRunnable` creates an equivalent Secret with the same name (`kagenti-test-users`) and keys, or update `show-services.sh` to handle its absence gracefully.

pdettori (changes requested):
## Do not merge until kagenti/kagenti-operator#412 is merged

This PR removes Keycloak infrastructure from the Helm chart (OpenShift path) in favor of operator management. However, the companion operator PR (kagenti/kagenti-operator#412 — `KeycloakBootstrapRunnable`) is **still open**.

Merging this PR first breaks the OpenShift install path:
- `_wait_kagenti_deps_ready()` in `setup-kagenti.sh` will timeout waiting for `postgres-kc` and `keycloak` StatefulSets that are no longer created
- `_apply_operand_crs()` extracts post-install hooks to create the Keycloak CR — the deleted `keycloak-operand.yaml` was one of those hooks

CI passes because "Deploy & Test" only exercises the Kind path (where `openshift: false`), so the deleted OpenShift resources are never tested.

**Resolution**: Either merge the operator PR first, or ship them together with the setup script updated to skip/remove the waits for `postgres-kc` and `keycloak` StatefulSets.

charts/kagenti-deps/templates/keycloak-realm-init.yaml
ChristianZaccaria: Verified — the operator's `KeycloakBootstrapRunnable.ensureTestUsersSecret()` creates `kagenti-test-users` in the `keycloak` namespace with identical keys (admin-password, dev-user-password, ns-admin-password) and random passwords, matching what the chart previously generated.

Note: `show-services.sh` actually reads `kagenti-test-user` (singular) — a different secret created by 87-setup-test-credentials.sh with E2E test credentials (username, password, realm, client_id, client_secret). That script is unaffected by this change.

ChristianZaccaria:
Rebased

scripts/ocp/setup-kagenti.sh
```diff
@@ -699,8 +699,8 @@ EOF
     log_info "MLflow OTEL values: otel.mlflow.enabled=true, endpoint=${MLFLOW_TRACES_ENDPOINT}"
   fi
 
-  # Keycloak public URL is needed by the realm-init audience mapper.
-  # Construct from DOMAIN (known since Step 2) so it's correct on first install.
+  # Keycloak public URL — realm-init audience mapper is now handled by
```
pdettori, line 658: **must-fix**: The comment acknowledges the operator handles Keycloak now, but `_wait_kagenti_deps_ready()` at line ~594 still waits for `postgres-kc` and `keycloak` StatefulSets — which are no longer created by `kagenti-deps`. The operator that creates them isn't installed until `helm install kagenti` (line ~1329), which runs much later.

This causes a 5-minute timeout → exit 1 on fresh OpenShift installs, confirmed by the [E2E HyperShift failure](https://github.com/kagenti/kagenti/actions/runs/27783820666).

Please move the keycloak wait logic to after the kagenti chart install, or remove it from `_wait_kagenti_deps_ready()` entirely.

pdettori (changes requested):
## E2E HyperShift CI confirms installer breakage

The [E2E HyperShift run](https://github.com/kagenti/kagenti/actions/runs/27783820666) triggered on the latest commit (`2b5de2d`) fails at **Deploy Kagenti** step:

```
2026-06-18T19:38:56Z → Waiting for Keycloak PostgreSQL to appear...
2026-06-18T19:43:59Z ⚠ Keycloak PostgreSQL statefulset not found after 5m
##[error] Process completed with exit code 1.
```

### Root cause

`_wait_kagenti_deps_ready()` (line ~594 in `setup-kagenti.sh`) waits for `postgres-kc` and `keycloak` StatefulSets immediately after `helm install kagenti-deps`. This PR removes the chart templates that created those resources — they are now created by the operator's `KeycloakBootstrapRunnable`. But the operator is deployed later via `helm install kagenti` (line ~1329), so the resources don't exist yet when the wait fires.

The wait times out after 5 minutes and exits non-zero, which fails the CI step.

### Suggested fix

Either:
1. **Move the keycloak/postgres waits** from `_wait_kagenti_deps_ready()` to a new function called after the `kagenti` chart install (where the operator is deployed), or
2. **Guard the waits** with a check for whether the operator is managing Keycloak (e.g. skip when `openshift: true` and operator is expected), or
3. **Add the operator as a dependency of `kagenti-deps`** so it's available before the waits run.

Option 1 is the least invasive — just relocate lines 594-595 to run after line ~1329.

---

**Author**: ChristianZaccaria (CONTRIBUTOR — returning external)
**CI**: Kind path passes ✅ | OpenShift path fails ❌ (confirmed above)
**Companion PR**: kagenti/kagenti-operator#412 merged ✅

scripts/ocp/setup-kagenti.sh
ChristianZaccaria: Done, moved the `postgres-kc` and `keycloak` StatefulSet waits out of `_wait_kagenti_deps_ready()` into a new `_wait_keycloak_ready()` that runs after helm install kagenti, where the operator is actually deployed.

pdettori:
needs to be rebased on main, and pass hypershift CI

ChristianZaccaria:
@pdettori I've rebased. I believe the tests will pass once we upgrade the kagenti-operator chart used to the latest `v0.3.0-alpha.6`.

pdettori:
sorry there are still conflicts

ChristianZaccaria:
@pdettori, I've just rebased. Thanks!

pdettori (approved):
LGTM - HS CI passed

PR #1227, opened 2026-04-13
fix: return 502 instead of 500 when MCP tool is unreachable

pdettori:
## Summary

- Fix `connect_to_tool` and `invoke_tool` endpoints to catch `httpx` exceptions (not just Python's `ConnectionError`) when MCP tools are unreachable
- Returns 502 Bad Gateway for connection failures and 504 Gateway Timeout for timeouts, matching the existing pattern in `agents.py`
- Add E2E test validating correct status code for unreachable tools

## Root Cause

The MCP client (`streamablehttp_client`) uses httpx internally. `httpx.ConnectError` is NOT a subclass of Python's `ConnectionError` — it falls through to the generic `except Exception` handler → 500.

## Test plan

- [ ] E2E test `test_tool_connect_errors.py` verifies 502 for unreachable tools
- [ ] Deploy to Kind cluster, create a tool with wrong port, verify UI shows 502 instead of 500
- [ ] Verify existing tool connections still work (happy path)

Closes #1144

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1876, opened 2026-06-09
fix(backend): Python 3.14 follow-up to #1227 — pre-check MCP reachability

### LINKED ISSUE #1869: :bug: Python 3.14: MCP tool unreachable still returns 500 (regression of #1144 / follow-up to #1227)
## Describe the bug

The fix in #1227 catches `httpx.NetworkError` / `httpx.TimeoutException` /
`httpx.HTTPError` in `connect_to_tool` and `invoke_tool`. That works on
Python 3.11/3.12 where `httpx.ConnectError` propagates directly out of
`streamablehttp_client.__aenter__()` / `session.initialize()`.

On **Python 3.14**, the failure mode changes: the MCP SDK uses
`anyio.create_task_group()` inside `streamablehttp_client`. When the upstream
connect fails (e.g. DNS for an unreachable in-cluster service), the exception
is consumed by an anyio cancel scope, and what reaches our `try/except` is:

```
asyncio.exceptions.CancelledError: Cancelled via cancel scope <id>
```

raised at `await response_stream_reader.receive()` inside `session.initialize()`.
`CancelledError` is a `BaseException` (not `Exception`), and even with
`except BaseException` + `Task.cancelling()` discrimination it doesn't reliably
reach our handler — FastAPI returns the generic 500.

In addition, `NoCacheMiddleware` (which uses `BaseHTTPMiddleware`) hits a
separate `BaseExceptionGroup` trap inside Starlette's `call_next`, surfacing
as `RuntimeError("No response returned.")` and producing the same 500.

## Steps To Reproduce

1. Deploy a Kagenti backend image built on Python 3.14 (current default since v0.6.0).
2. Acquire a backend JWT (`kagenti-api-oauth-secret`).
3. POST to a non-existent tool route:
   ```bash
   curl -X POST -H "Authorization: Bearer $TOKEN" \
     https://<ui-route>/api/v1/tools/team1/nonexistent-tool/connect
   ```
4. Observe HTTP **500** in the response.
5. Equivalent pytest reproduction:
   ```bash
   uv run pytest kagenti/tests/e2e/common/test_tool_connect_errors.py -v
   ```
   → `Failed: Backend still returns 500 for unreachable tool (issue #1144 not fixed)`.

## Expected Behavior

Per the contract that #1227 introduced for older Python versions, the endpoint
should return **502 Bad Gateway** when the upstream MCP tool is unreachable
(or **504** for timeout), with a `detail` message naming the unreachable URL.

## Kagenti Version

v0.6.0-rc.13

### Fix direction

Two parts:

1. **Pre-check reachability with plain `httpx`** before invoking the MCP
   client. `httpx.AsyncClient` (without anyio task groups) raises
   `httpx.ConnectError` / `NetworkError` directly, so they reach our
   `except` clauses cleanly and return 502.
2. **Convert `NoCacheMiddleware` to pure ASGI** so it doesn't trip
   Starlette's `BaseHTTPMiddleware.call_next` `BaseExceptionGroup` handling.

Both are localized changes; #1227's existing handlers should be kept as
defense-in-depth for the post-pre-check MCP path.

### Verified locally

Smoke-tested live on `kagenti-hypershift-custom-ocp1` (HyperShift / OCP 4.20.21
/ Python 3.14 backend) via a ConfigMap overlay of the patched `tools.py` and
`main.py`:

- `POST /api/v1/tools/team1/nonexistent-tool-1144/connect` returns **502** (was 500)
- `uv run pytest kagenti/tests/e2e/common/test_tool_connect_errors.py` passes 2/2 (was failing 2/2)

webchang:
## Summary

Follow-up to #1227 for Python 3.14 backend images. #1227's
`httpx.NetworkError`/`TimeoutException` handlers don't catch the
failure on Python 3.14 because anyio's task group inside
`streamablehttp_client` transforms the underlying `httpx.ConnectError`
into `asyncio.CancelledError`, which escapes the `except` clauses.

- Adds a plain-httpx reachability probe in `connect_to_tool` and
	`invoke_tool` before invoking the MCP `streamablehttp_client`.
	`httpx.AsyncClient` doesn't use anyio task groups, so its
	`ConnectError` / `NetworkError` propagate cleanly to our handlers
	and return 502.
- Converts `NoCacheMiddleware` from `BaseHTTPMiddleware` to pure ASGI
	to dodge Starlette's `call_next` `BaseExceptionGroup` trap (separate
	but adjacent symptom on Python 3.14).
- Updates the existing test from #1227 (no behavior change — it now
	actually passes on Py3.14 backends).

#1227's handlers remain in place as defense-in-depth for the
post-probe MCP code path.

## Cluster tested on

`kagenti-hypershift-custom-ocp1` (HyperShift hosted cluster on AWS,
OCP 4.20.21, Python 3.14 backend, 3-worker NodePool).

## Kagenti version under test

v0.6.0-rc.13

## Test plan / results
Verified on 2026-06-08:

- [x] On the Python 3.14 backend pod, `POST /api/v1/tools/<ns>/<nonexistent>/connect`
			returns **502** (was 500)
- [x] `uv run pytest kagenti/tests/e2e/common/test_tool_connect_errors.py -v`
			→ **2 passed** (was 2 failed)
- [x] Existing weather-agent happy path still passes (probe to a reachable
			MCP server falls through to the unchanged MCP client code path)

## Operator handoff: clean up the temporary overlay

This section is for whoever bumps the backend image to a build that
includes this PR.

The fix was smoke-tested live on `kagenti-hypershift-custom-ocp1` via
two ConfigMap overlays mounted into the running `kagenti-backend`
Deployment:

- `kagenti-backend-tools-override` (mounts `tools.py` over
	`/app/app/routers/tools.py` via subPath)
- `kagenti-backend-main-override` (mounts `main.py` over
	`/app/app/main.py` via subPath)

Once a backend image containing this PR is rolled out to the cluster,
**remove the overlays** so the deployment uses the baked-in code:

```bash
# Set KUBECONFIG to the management cluster's hosted-cluster admin kubeconfig
KUBECONFIG=<path-to-hosted-cluster-admin-kubeconfig>

oc -n kagenti-system delete configmap \
	kagenti-backend-tools-override \
	kagenti-backend-main-override

# Inspect the Deployment volumes/volumeMounts to find the override entries
oc -n kagenti-system get deploy kagenti-backend -o yaml | grep -A1 -B1 override

# Strip those entries (volume + volumeMount), e.g. via oc edit deploy or a
# targeted JSON patch. Indices vary, so visual inspection is needed.
```

Closes #1869

Assisted-By: Claude Code

### DIFF
--- a/kagenti/backend/app/main.py
+++ b/kagenti/backend/app/main.py
@@ -12,22 +12,43 @@ import asyncio
 import logging
 from contextlib import asynccontextmanager
 
-from fastapi import FastAPI, Request
+from fastapi import FastAPI
 from fastapi.middleware.cors import CORSMiddleware
-from starlette.middleware.base import BaseHTTPMiddleware
-from starlette.responses import Response
+from starlette.types import ASGIApp, Receive, Scope, Send
 
 
-class NoCacheMiddleware(BaseHTTPMiddleware):
-    """Middleware to prevent browser caching of API responses."""
+class NoCacheMiddleware:
+    """Pure ASGI middleware that adds no-cache headers to /api/ responses.
 
-    async def dispatch(self, request: Request, call_next) -> Response:
-        response = await call_next(request)
-        # Add no-cache headers to API endpoints to prevent stale data
-        if request.url.path.startswith("/api/"):
-            response.headers["Cache-Control"] = "no-store, no-cache, must-revalidate"
-            response.headers["Pragma"] = "no-cache"
-        return response
+    Implemented as pure ASGI rather than BaseHTTPMiddleware to avoid the
+    BaseHTTPMiddleware.call_next path, which on Python 3.14 surfaces
+    BaseExceptionGroup escapes from inner middleware/routes as
+    `RuntimeError("No response returned.")` and yields HTTP 500 instead
+    of letting the route's own error handlers convert connection
+    failures to 502/504.
+    """
+
+    def __init__(self, app: ASGIApp) -> None:
+        self.app = app
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if scope["type"] != "http" or not scope.get("path", "").startswith("/api/"):
+            await self.app(scope, receive, send)
+            return
+
+        async def send_with_no_cache(message):
+            if message["type"] == "http.response.start":
+                headers = list(message.get("headers", []))
+                headers.extend(
+                    [
+                        (b"cache-control", b"no-store, no-cache, must-revalidate"),
+                        (b"pragma", b"no-cache"),
+                    ]
+                )
+                message = {**message, "headers": headers}
+            await send(message)
+
+        await self.app(scope, receive, send_with_no_cache)
 
 
 from app.core.config import settings  # pylint: disable=wrong-import-position
--- a/kagenti/backend/app/routers/tools.py
+++ b/kagenti/backend/app/routers/tools.py
@@ -6,10 +6,12 @@
 Tool API endpoints.
 """
 
+import asyncio
 import logging
 import re
 from typing import Any, Dict, List, Literal, Optional
 from contextlib import AsyncExitStack
+from urllib.parse import urlparse
 
 import httpx
 from fastapi import APIRouter, Depends, HTTPException, Query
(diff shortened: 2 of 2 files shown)


Ladas:
## Code review

### TCP probe should be extracted to a helper

The reachability probe is copy-pasted identically between `connect_to_tool` (tools.py:2114-2143) and `invoke_tool` (tools.py:2224-2248) — 26 lines duplicated. Should be a shared helper:

```python
async def _probe_mcp_reachability(mcp_endpoint: str, tool_url: str) -> None:
    """Raise HTTPException(502/504) if the MCP server is unreachable."""
    parsed = urlparse(mcp_endpoint)
    try:
        _, writer = await asyncio.wait_for(
            asyncio.open_connection(parsed.hostname, parsed.port or 80),
            timeout=5.0,
        )
        writer.close()
        await writer.wait_closed()
    except (OSError, ConnectionError) as e:
        logger.error("MCP server unreachable (pre-check, %s)", type(e).__name__)
        raise HTTPException(status_code=502, detail=f"Failed to connect to MCP server at {tool_url}")
    except asyncio.TimeoutError:
        logger.error("MCP server timeout (pre-check)")
        raise HTTPException(status_code=504, detail=f"Timeout connecting to MCP server at {tool_url}")
```

Then both call sites become one line: `await _probe_mcp_reachability(mcp_endpoint, tool_url)`.

### NoCacheMiddleware rewrite looks good

The BaseHTTPMiddleware → pure ASGI migration is the right fix for Python 3.14's BaseExceptionGroup handling. Clean implementation.

### Minor: error detail consistency

The probe error messages expose the internal `tool_url` in the HTTP response detail. The existing code already does this (pre-existing pattern), but it's worth noting as inconsistent with the error sanitization approach in PR #1689 which uses generic client-facing messages.

Assisted-By: Claude Code

webchang:
Thanks for the review!

Applied the helper extraction using your suggested code verbatim — `_probe_mcp_reachability(mcp_endpoint, tool_url)` now lives next to
`_get_tool_url()` and the two call sites are one-line `await` calls.

On the URL-in-detail point: agreed it's inconsistent with #1689's pattern, but #1227's existing 502/504 handlers further down already
expose `tool_url` in their `detail=`, so I kept the same convention within this file for now. Happy to follow up with a separate PR
aligning all of them with #1689 if that's preferred.

Latest commit force-pushed; CI re-running.

Assisted-By: Claude Code

Ladas (approved):
LGTM — previous review feedback addressed: TCP probe extracted to shared `_probe_mcp_reachability()` helper.

**Code review:** Both changes are solid:
- **NoCacheMiddleware**: Correct pure ASGI implementation. Scope type guard, bytes headers, non-mutating message dict — all correct.
- **`_probe_mcp_reachability()`**: Clean helper, correct exception mapping (OSError/ConnectionError → 502, TimeoutError → 504). The 5s timeout is reasonable for in-cluster services.

**Security review:** No concerns. The `tool_url` in error details is user-derived (the tool name they requested), not leaking internal info.

All non-gated CI checks pass. HyperShift E2E is not required — this is a backend-only Python change.

PR #1877, opened 2026-06-09
fix(installer): handle MCP Gateway upgrade from alpha to GA

pdettori:
## Summary

- Pre-apply CRDs from the MCP Gateway chart before `helm upgrade` (Helm only installs CRDs on initial install, never on upgrade)
- Detect and delete Deployments with changed `spec.selector` labels (Kubernetes rejects in-place selector changes)
- Switch from `helm install` to `helm upgrade --install` for idempotent re-runs

Fixes upgrade failures when moving from MCP Gateway v0.6.0-alpha.5 to v0.6.0 GA.

## Test plan

- [ ] Fresh install with `--with-mcp-gateway` on a clean cluster
- [ ] Upgrade from v0.6.0-alpha.5 → v0.6.0 (CRD and selector changes)
- [ ] Dry-run mode skips kubectl/helm calls
- [ ] Re-run on already-upgraded cluster is a no-op

Assisted-By: Claude Code

### DIFF
--- a/scripts/ocp/setup-kagenti.sh
+++ b/scripts/ocp/setup-kagenti.sh
@@ -1606,10 +1606,10 @@ log_info "Step 5b: Install MCP Gateway"
 if $SKIP_MCP_GATEWAY; then
   log_info "Skipped (--skip-mcp-gateway)"
 else
-  # Clean up any MCPGatewayExtension stuck in deletion (e.g. leftover from a prior
-  # version that used a different API group). A stuck finalizer prevents the controller
-  # from creating the broker-router deployment on reinstall.
   if ! $DRY_RUN; then
+    # Clean up any MCPGatewayExtension stuck in deletion (e.g. leftover from a prior
+    # version that used a different API group). A stuck finalizer prevents the controller
+    # from creating the broker-router deployment on reinstall.
     for _crd_group in mcp.kuadrant.io mcp.kagenti.com; do
       _stuck=$($KUBECTL get mcpgatewayextensions.${_crd_group} -n mcp-system -o json 2>/dev/null \
         | python3 -c "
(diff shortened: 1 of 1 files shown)


esnible:
@pdettori This may resolve #/1335 ?

scripts/ocp/setup-kagenti.sh
```diff
@@ -1451,6 +1451,39 @@ for item in data.get('items', []):
...
+    fi
+
+    # Kubernetes does not allow changing spec.selector on existing Deployments.
+    # If the chart changed selectors between versions, delete affected Deployments
+    # so Helm can recreate them.
+    for _deploy in mcp-gateway-controller mcp-gateway-broker-router; do
+      if $KUBECTL get deployment "$_deploy" -n mcp-system &>/dev/null; then
+        _current_selector=$($KUBECTL get deployment "$_deploy" -n mcp-system \
+          -o jsonpath='{.spec.selector.matchLabels}' 2>/dev/null || echo "")
+        _chart_name=$(helm template mcp-gateway "$_mcp_gw_tmp/mcp-gateway" 2>/dev/null \
```
esnible: **must-fix — selector comparison will always evaluate as "different"**

`kubectl -o jsonpath='{.spec.selector.matchLabels}'` emits keys alphabetically with no whitespace, e.g. `{"app":"foo","version":"v1"}` (Go map → JSON marshal sorts keys).

Python's default `json.dumps(dict)` preserves insertion order *and* inserts a space after `:` and `,`, e.g. `{"version": "v1", "app": "foo"}`.

These two strings will never be equal, so the `[ "$_current_selector" != "$_chart_name" ]` check on the next line is effectively always true — every run with both deployments present will delete them, which is the opposite of the safety goal.

Fix:
```python
print(json.dumps(doc['spec']['selector']['matchLabels'], sort_keys=True, separators=(',', ':')))
```

Quick verification:
```
$ python3 -c "import json; print(json.dumps({'version':'v1','app':'foo'}))"
{"version": "v1", "app": "foo"}
$ python3 -c "import json; print(json.dumps({'version':'v1','app':'foo'}, sort_keys=True, separators=(',',':')))"
{"app":"foo","version":"v1"}    # matches kubectl output
```

scripts/ocp/setup-kagenti.sh
```diff
@@ -1451,6 +1451,39 @@ for item in data.get('items', []):
...
+      fi
+    fi
+
+    # Kubernetes does not allow changing spec.selector on existing Deployments.
+    # If the chart changed selectors between versions, delete affected Deployments
+    # so Helm can recreate them.
+    for _deploy in mcp-gateway-controller mcp-gateway-broker-router; do
+      if $KUBECTL get deployment "$_deploy" -n mcp-system &>/dev/null; then
+        _current_selector=$($KUBECTL get deployment "$_deploy" -n mcp-system \
+          -o jsonpath='{.spec.selector.matchLabels}' 2>/dev/null || echo "")
```
esnible: **suggestion** — `helm template` is invoked once per iteration of the `for _deploy` loop (twice total). Consider running it once before the loop and piping the captured YAML to a Python parser that emits both selectors at once, or just keep the YAML in a variable and re-feed it. Minor, but each `helm template` re-reads and renders the chart.

scripts/ocp/setup-kagenti.sh
```diff
@@ -1451,6 +1451,39 @@ for item in data.get('items', []):
         sleep 2
       fi
     done
+
+    # Helm does not update CRDs on upgrade — pre-apply them from the chart
+    _mcp_gw_tmp=$(mktemp -d)
+    if helm pull oci://ghcr.io/kuadrant/charts/mcp-gateway \
```
esnible, line 1634: **suggestion** — if `helm pull` fails (transient network/registry issue), `_mcp_gw_tmp` stays empty. The CRD apply then silently skips (because `[ -d .../crds ]` is false), and the selector check also silently skips (because `helm template` of an empty dir produces nothing → `_chart_name` is empty → the guard on line 1480 short-circuits). The behavior is safe, but invisible — the user sees no signal that pre-apply was bypassed and may not realize a CRD-dependent upgrade has slipped through. Consider an `else` branch that emits `log_warn` on pull failure.

esnible (changes requested):
## Review Summary

The fix tackles two real upgrade pain points (Helm not updating CRDs on upgrade, K8s rejecting selector changes on existing Deployments), and the overall structure is sound. However, the selector-comparison logic has a JSON-formatting bug that will make the equality check fail on every run when both deployments exist — likely deleting Deployments unnecessarily on each `helm upgrade` invocation, defeating the safety check.

Minor note: the PR description lists "Switch from `helm install` to `helm upgrade --install`" as a change, but that line is already on `main` and is unchanged in this diff.

**Areas reviewed**: Shell script (single file)
**Commits**: 1 commit, signed-off ✓, DCO ✓
**CI status**: All passing except `Deploy & Test` (still pending at review time)

Assisted-By: Claude Code

scripts/ocp/setup-kagenti.sh
pdettori: Fixed in eb8b9d3e. Added `sort_keys=True, separators=(",",":")` to match kubectl jsonpath compact output.

scripts/ocp/setup-kagenti.sh
pdettori: Done — helm template now runs once before the loop and the YAML is piped from a variable for each deployment.

scripts/ocp/setup-kagenti.sh
pdettori: Added. The selector check is now inside the `helm pull` success branch, with a `log_warn` on the else path for visibility.

esnible (approved):

PR #1880, opened 2026-06-09
Bug fix: Streaming A2A agents show be shown as streaming

esnible:
## Summary

The Kagenti UI shows streaming agents as if they cannot stream.

incorrect-streaming

Fixes #

No associated Issue.


pdettori (approved):
LGTM — the fix correctly aligns the `AgentCard` interface with the actual API response from `chatService.getAgentCard()`, which returns `streaming` at the top level (not nested under `capabilities`).

One nit: the PR title/commit has "show be shown" — likely meant "should be shown".

PR #1881, opened 2026-06-10
Fix: (openshell) support OpenShift SCC and clusters without default StorageClass

### LINKED ISSUE #1879: :bug: OpenShell: Helm chart and deploy-tenant.sh fail on OpenShift (SCC + StorageClass)
## Description

The OpenShell Helm chart (`charts/openshell/`) and `scripts/openshell/deploy-tenant.sh` fail to deploy on OpenShift clusters due to three issues related to Security Context Constraints (SCCs) and storage configuration.

## Environment

- OpenShift 4.x cluster (tested on OCP 4.16 with ODF/Ceph storage)
- No default StorageClass configured
- Standard `restricted-v2` SCC enforced on namespaces

### 1. Certgen Job Rejected by SCC

The Helm pre-install hook job (`openshell-certgen`) hardcodes `runAsUser: 1000` and `seccompProfile: RuntimeDefault` in its pod security context. OpenShift namespaces enforce UID ranges (e.g. `1000930000/10000`), so the job is rejected:

```
Error creating: pods "openshell-certgen-" is forbidden: unable to validate against any
security context constraint: [...] restricted-v2: .containers[0].runAsUser: Invalid value:
1000: must be in the ranges: [1000930000, 1000939999]
```

**Result**: Helm times out waiting for the pre-install hook with no clear error shown.

### 2. Gateway StatefulSet Rejected by SCC (same root cause)

The `openshell-server` StatefulSet also hardcodes `runAsUser: 1000`, `fsGroup: 1000`, and `seccompProfile: RuntimeDefault`. Even after granting `anyuid` SCC, the `seccompProfile` annotation is rejected by `anyuid`:

```
create Pod openshell-server-0 in StatefulSet openshell-server failed error: pods
"openshell-server-0" is forbidden: unable to validate against any security context constraint
```

### 3. PVC Stuck Pending (No StorageClass)

The StatefulSet's `volumeClaimTemplates` have no `storageClassName` field. On clusters without a default StorageClass, the PVC stays Pending indefinitely:

```
Warning  FailedBinding  persistentvolumeclaim/data-openshell-server-0  no persistent volumes
available for this claim and no storage class is set
```

### 4. SQLite Database Permission Error

After fixing SCC issues, if `fsGroup` is not set on OpenShift, the gateway process runs as a namespace-assigned UID but the PVC volume is owned by root. The gateway crashes:

```
Error: database error: error returned from database: (code: 14) unable to open database file
```

### 5. deploy-tenant.sh: CA Bundle Step Runs Before Namespace Creation

On OpenShift, the script attempts to read `config-trusted-cabundle` ConfigMap from the tenant namespace (line ~170) before Step 1 creates that namespace (line ~209), causing immediate failure.

## Deployment Failure Sequence

1. `deploy-tenant.sh team1` → fails creating CA bundle (namespace doesn't exist)
2. After manually creating namespace → Helm times out at certgen job (SCC)
3. After granting SCC → Helm hangs at StatefulSet (SCC + missing StorageClass)
4. After fixing StorageClass → gateway crashes (volume permissions)

Each failure gives no diagnostic output — the script just times out.

### Steps To Reproduce

1. Provision an OpenShift 4.x cluster with ODF storage (no default StorageClass)
2. Deploy shared infrastructure:
   ```bash
   scripts/openshell/deploy-shared.sh
   ```
3. Attempt tenant deployment:
   ```bash
   scripts/openshell/deploy-tenant.sh team1
   ```
4. Observe failure — script errors out or Helm times out with no diagnostics

Alternatively, on any OpenShift cluster:
```bash
# Install the chart directly to see the SCC errors
helm install openshell-team1 charts/openshell/ \
  --namespace team1 --create-namespace \
  --set tenant=team1 \
  --set ingress.type=route \
  --set ingress.host=openshell-team1.apps.example.com \
  --set oidc.issuer=https://keycloak.example.com/realms/openshell \
  --set oidc.audience=team1 \
  --wait --timeout 120s

# Check events to see SCC rejection:
kubectl get events -n team1 --field-selector reason=FailedCreate
```

## Deployment Failure Sequence

1. `deploy-tenant.sh team1` → fails creating CA bundle (namespace doesn't exist)
2. After manually creating namespace → Helm times out at certgen job (SCC)
3. After granting SCC → Helm hangs at StatefulSet (SCC + missing StorageClass)
4. After fixing StorageClass → gateway crashes (volume permissions)

Each failure gives no diagnostic output — the script just times out.

### Expected Behavior

`scripts/openshell/deploy-tenant.sh team1` should deploy successfully on OpenShift without manual intervention, auto-detecting the platform and applying appropriate security contexts and storage configuration.

### Kagenti Version

main

### Additional Context

_No response_

pdettori:
## Summary

- Remove hardcoded `runAsUser: 1000`/`fsGroup: 1000`/`seccompProfile: RuntimeDefault` from certgen job and gateway StatefulSet; replace with values-driven `podSecurityContext` (defaults to `runAsNonRoot: true` only)
- Add `persistence.storageClassName` to VolumeClaimTemplate for clusters without a default StorageClass
- Create namespace before CA bundle step in `deploy-tenant.sh` (the ConfigMap read requires the namespace to exist)
- Add optional SCC template (`openshift.enabled=true`) scoped to `openshell-certgen` and `openshell-gateway` service accounts only
- Auto-set `openshift.enabled=true` in `deploy-tenant.sh` when platform is detected as OpenShift

## Test plan

- [ ] `helm template` renders correctly for vanilla K8s (no SCC, no storageClassName when unset)
- [ ] `helm template` with `openshift.enabled=true` and `persistence.storageClassName=X` produces expected output
- [ ] Deploy on Kind cluster with `deploy-tenant.sh` — verify gateway starts without SCC issues
- [ ] Deploy on OpenShift cluster — verify certgen job and StatefulSet are accepted by `restricted-v2` SCC (or the chart-created SCC)
- [ ] Verify PVC binds on a cluster without a default StorageClass when `persistence.storageClassName` is set

Fixes #1879

Assisted-By: Claude Code


charts/openshell/templates/certgen.yaml
```diff
@@ -71,11 +71,15 @@ spec:
...
       serviceAccountName: {{ $hookName }}
       enableServiceLinks: false
+      {{- if not .Values.openshift.enabled }}
+      {{- with .Values.podSecurityContext }}
+      securityContext:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- else }}
       securityContext:
         runAsNonRoot: true
```
rubambiza, line 75: **nit**: The OpenShift branch hardcodes `runAsNonRoot: true` here (and in `statefulset.yaml`) rather than being values-driven like the vanilla K8s path. Fine for now since OpenShift assigns UIDs from namespace annotations, but worth noting for future extensibility if someone needs to customize the OpenShift pod security context.

rubambiza (commented):
Well-structured fix. The SCC is properly scoped to specific service accounts, template logic correctly branches for OpenShift vs vanilla K8s, namespace creation respects `$DRY_RUN`, and the second commit properly restores the full security context for Kind after discovering the regression.

I'm waiting for the E2E tests to complete (`/run-e2e` pending) since these changes touch setup across all environments (Kind, OCP, HyperShift). Once E2E passes, I'll add approval.

**Review checklist**:
| Check | Result |
|-------|--------|
| DCO sign-off | Pass (both commits) |
| Commit format | Pass |
| Helm chart lint | Pass |
| YAML lint | Pass |
| Shell Script Lint | Pass |
| Security | SCC scoped to specific SAs only |
| Values documented | Pass |
| Deploy & Test | Pass |
| Template logic | Correct conditional branching |

**Areas reviewed**: Helm/K8s, Shell, Security
**CI status**: Deploy & Test passing, E2E pending

Ladas (commented):
## Code Review + Security Review

### Code Review — 6 Findings

| # | Severity | File | Finding |
|---|----------|------|---------|
| 1 | **MEDIUM** | `certgen.yaml` | Empty `podSecurityContext={}` override drops ALL securityContext — pod runs as root on vanilla K8s. The `{{- with .Values.podSecurityContext }}` guard evaluates false on empty map. |
| 2 | **MEDIUM** | `scc.yaml:38` | `readOnlyRootFilesystem: false` more permissive than needed — gateway only writes to mounted volumes (PVC `/data`, emptyDir `/run/drivers`). Setting `true` would add defense-in-depth. |
| 3 | **MEDIUM** | `scc.yaml:57` | `supplementalGroups: RunAsAny` inconsistent with `fsGroup: MustRunAs` — allows containers to join arbitrary GID groups on shared volumes. Should be `MustRunAs`. |
| 4 | **LOW** | `deploy-tenant.sh:170` | CA bundle reads `config-trusted-cabundle` from namespace but no labeled ConfigMap (`config.openshift.io/inject-cabundle=true`) is created to trigger OpenShift CA operator injection. Combined bundle may miss cluster trusted CAs. |
| 5 | **LOW** | `statefulset.yaml:98` | Empty string `storageClassName: ""` disables dynamic provisioning (different from omitting it). The commented-out `# storageClassName: ""` in values.yaml makes accidental triggering likely. Consider `{{ .Values.persistence.storageClassName | quote }}`. |
| 6 | **LOW** | `values.yaml:128` | Comment claims deploy-tenant.sh overrides `podSecurityContext` to `{}` on OpenShift, but the script only sets `openshift.enabled=true`. Misleading documentation. |

### Security Review — Key Findings

| Severity | Finding |
|----------|---------|
| **HIGH** (pre-existing, not this PR) | `rbac.yaml` grants sandbox SA `system:openshift:scc:privileged` — far more permissive than the new SCC. Should be tracked as follow-up. |
| **MEDIUM** | `supplementalGroups: RunAsAny` should be `MustRunAs` to match fsGroup policy and prevent arbitrary GID access. |
| **MEDIUM** | Dropping `seccompProfile` on OpenShift path relies on SCC enforcement; misconfigured `openshift.enabled=true` on non-OCP cluster leaves no seccomp. |
| **LOW** | `runAsNonRoot: true` should be outside the `{{- with }}` guard as mandatory baseline — prevents accidental removal via empty override. |

### Overall Verdict

**Net positive** — correctly replaces hardcoded security contexts with values-driven config. The SCC is well-scoped (bound to specific SAs via `users` field, not cluster-wide ClusterRoleBinding). Main recommendations:

1. Tighten `supplementalGroups` to `MustRunAs`
2. Set `readOnlyRootFilesystem: true` (all writable paths use mounted volumes)
3. Move `runAsNonRoot: true` outside the `{{- with }}` guard as a non-overridable baseline
4. Track the pre-existing `scc:privileged` binding on sandbox SA as a separate HIGH issue

charts/openshell/templates/certgen.yaml
pdettori: Addressed in d8c189b1 — `runAsNonRoot: true` is now rendered unconditionally outside the `{{- with }}` guard. The remaining `podSecurityContext` fields are merged below it via `omit . "runAsNonRoot"`, so even a null/empty override can't remove the baseline.

pdettori:
Addressed review findings in d8c189b1:

| # | Finding | Resolution |
|---|---------|-----------|
| 1 | Empty `podSecurityContext` drops all security | `runAsNonRoot: true` now rendered unconditionally outside the `{{- with }}` guard |
| 2 | `readOnlyRootFilesystem: false` too permissive | Set to `true` — gateway only writes to mounted volumes (PVC `/data`, emptyDir `/run/drivers`) |
| 3 | `supplementalGroups: RunAsAny` inconsistent | Changed to `MustRunAs` to match `fsGroup` policy |
| 4 | CA bundle injection (pre-existing) | Out of scope for this PR — tracked as follow-up |
| 5 | `storageClassName` comment clarity | Removed ambiguous commented-out value, clarified docs |
| 6 | Misleading `values.yaml` comment | Fixed — now describes actual template conditional behavior |

Re: the pre-existing HIGH (sandbox SA with `scc:privileged`) — agreed, will track separately.

pdettori:
Opened #1887 to track the pre-existing `system:openshift:scc:privileged` binding on the sandbox SA (HIGH from @Ladas's security review).

charts/openshell/values.yaml
```diff
@@ -92,6 +92,31 @@ quota:
...
+  storageClassName: ""
+
+# -- Pod security context applied to the gateway StatefulSet and certgen Job.
+# These defaults work on vanilla Kubernetes (Kind, EKS, GKE, etc.).
+# On OpenShift, set openshift.enabled=true — the template skips runAsUser/fsGroup/
+# seccompProfile and emits only runAsNonRoot:true, letting the platform assign UIDs.
+podSecurityContext:
+  runAsNonRoot: true
+  runAsUser: 1000
+  fsGroup: 1000
```
mrsabath, line 116: nit (docs only): now that `persistence.size` drives the PVC request, worth a note somewhere (chart README or a comment here) that changing it on an existing release won't resize already-bound PVCs — StatefulSet `volumeClaimTemplates` are immutable after creation, so a resize needs manual PVC edits or a re-deploy. Not a code issue; just saves a future surprise.

mrsabath (commented):
Reviewed the fix against #1879's three failure modes and rendered the chart locally (`helm template`) in both modes to confirm:

- **Vanilla K8s** (`openshift.enabled=false`): pod securityContext keeps the full `runAsUser:1000 / fsGroup:1000 / seccompProfile:RuntimeDefault` (via the `omit "runAsNonRoot"` injection) — no behavior change for Kind/EKS/GKE. ✅
- **OpenShift** (`openshift.enabled=true`): pod securityContext emits only `runAsNonRoot: true`, letting the platform assign UIDs (fixes #1/#2), and the chart-created SCC is correctly scoped to just the `openshell-certgen` + `openshell-gateway` service accounts (no broad `anyuid`). `readOnlyRootFilesystem: true` is safe here since certgen writes Secrets via the API, not the local filesystem. ✅
- **StorageClass**: `persistence.storageClassName` renders into the volumeClaimTemplate when set and is absent when unset (fixes #3). ✅
- **deploy-tenant.sh**: the namespace-first ordering fixes the CA-bundle step, and `openshift.enabled=true` auto-sets on OpenShift detection. The SCC's SA namespace resolves correctly via `-n $TENANT`.

Clean, well-scoped fix that interlocks nicely. Bella prova, Paolo. The one thing I'd wait on before approving is the **OpenShift E2E (HyperShift, OCP 4.20.11) + OpenShell Kind PoC** jobs — both still pending, and they're the real proof for a platform-compat fix. Happy to flip to approve once they're green.

**Areas reviewed**: Helm (SCC / securityContext / PVC — rendered both modes), shell (deploy-tenant.sh), security (SCC scoping), conventions, CI · **Commits**: 3, signed-off: yes · **CI**: green except OpenShift E2E + Kind PoC (pending)

scripts/openshell/deploy-tenant.sh
```diff
@@ -173,6 +182,7 @@ else
   ) | kubectl create configmap "$COMBINED_CA_CM" -n "$TENANT" \
         --from-file=ca-bundle.crt=/dev/stdin --dry-run=client -o yaml | kubectl apply -f - >/dev/null
   EXTRA_HELM_SETS+=("trustedCABundle=$COMBINED_CA_CM")
+  EXTRA_HELM_SETS+=("openshift.enabled=true")
```
clawgenti, line 185: **Suggestion (open finding #4 from prior review):** The `config-trusted-cabundle` ConfigMap is populated by the OpenShift CA injection operator — but only after a ConfigMap with the `config.openshift.io/inject-cabundle: 'true'` annotation exists in the namespace. The namespace creation race is fixed by this PR (Step 0), but there's still a potential timing gap: if the CA injection operator hasn't populated `config-trusted-cabundle` before line 179 reads it, the `kubectl get configmap ... 2>/dev/null` silently returns empty and the combined CA bundle will be missing the cluster CAs. This could explain the `x509: certificate signed by unknown authority` failure in the HyperShift E2E run.

Consider creating the labeled ConfigMap first and waiting for injection before reading it, or at minimum logging a warning when the configmap is empty/missing.

charts/openshell/templates/scc.yaml
```diff
@@ -0,0 +1,39 @@
...
+  - runtime/default
+supplementalGroups:
+  type: MustRunAs
+volumes:
+  - configMap
+  - emptyDir
+  - persistentVolumeClaim
+  - projected
+  - secret
+users:
```
clawgenti, line 36: **Note:** The SCC's `users` list binds by service account name without a ClusterRoleBinding — this is the correct OpenShift SCC scoping pattern and properly limits the SCC to `openshell-certgen` and `openshell-gateway` SAs only. The `supplementalGroups: MustRunAs` and `readOnlyRootFilesystem: true` fixes from commit 3 address the prior MEDIUM findings. ✅

clawgenti (commented):
## Clawgenti Code Review

This PR replaces hardcoded security contexts with values-driven config for OpenShift compatibility, adds a properly-scoped SCC, fixes namespace creation ordering in `deploy-tenant.sh`, and enables configuring `storageClassName` for clusters without a default StorageClass. The implementation correctly addresses the findings from prior reviews (SCC tightening in commit 3). Two items need attention before merge.

**CI:** `E2E HyperShift` is failing — backend E2E phase 1 failed with `x509: certificate signed by unknown authority` on MLflow TLS, alongside `otel-collector` DNS failures. The failure pattern suggests a broader HyperShift environment issue rather than a regression from this PR, but worth a `/run-e2e` retry to confirm. This may also be related to the CA bundle race described in the inline comment below.

**Commit/PR conventions (nit):** All 3 commits are signed-off (DCO ✅), but none have an emoji prefix — subjects start with `fix(openshell):` without a leading emoji. PR title is also missing the emoji prefix and is 79 chars (over the 72-char limit).

---

charts/openshell/templates/scc.yaml
```diff
@@ -0,0 +1,39 @@
+{{- if .Values.openshift.enabled }}
+apiVersion: security.openshift.io/v1
```
mrsabath, line 2: **Suggestion (hook ordering):** The certgen Job is a Helm `pre-install,pre-upgrade` hook (`hook-weight: -30`), but this SCC is a plain template — Helm applies it as part of the *main release*, which runs **after** all pre-install hooks. So on OpenShift, when the certgen hook pod starts, this SCC doesn't exist yet and the pod is admitted under `restricted-v2` (the default), not this SCC. In effect the SCC only covers the long-running gateway StatefulSet, even though `openshell-certgen` is listed in `users`.

Two ways to reconcile:
- If certgen genuinely needs this SCC: annotate scc.yaml with `helm.sh/hook: pre-install,pre-upgrade` and a more-negative weight (e.g. `-40`) so it's created before the certgen hook.
- If certgen runs fine under `restricted-v2` anyway (it drops ALL caps and sets `runAsNonRoot`): consider dropping `openshell-certgen` from `users` so the SCC doesn't imply coverage it doesn't actually provide during the hook phase.

Either is fine — mainly want to make sure the certgen path was validated on a real OCP cluster, not just the StatefulSet.

scripts/openshell/deploy-tenant.sh
```diff
@@ -152,6 +152,15 @@ INGRESS_TYPE=$(get_ingress_type)
 INGRESS_HOST=$(get_ingress_host)
 OIDC_ISSUER=$(get_keycloak_issuer)
 
+# ── Step 0 (early): Ensure namespace exists before platform-specific setup ──
+# On OpenShift the CA bundle step needs the tenant namespace to already exist.
+if ! kubectl get namespace "$TENANT" &>/dev/null; then
+  log_info "Creating namespace $TENANT (needed for platform setup)..."
```
mrsabath, line 158: **nit:** This early namespace step is silent under `--dry-run` (the `if ! $DRY_RUN` wraps the only output), whereas the existing Step-1 block echoes `[dry-run] kubectl create namespace $TENANT`. Minor inconsistency — consider an `else echo "  [dry-run] kubectl create namespace $TENANT"` for parity. Also note Step 1 will now always log "already exists" on a real (non-dry-run) deploy, since this step created it first; harmless, just slightly redundant.

charts/openshell/templates/statefulset.yaml
```diff
@@ -171,7 +172,10 @@ spec:
     - metadata:
         name: data
       spec:
+        {{- if .Values.persistence.storageClassName }}
+        storageClassName: {{ .Values.persistence.storageClassName }}
```
mrsabath, line 176: **nit:** `storageClassName: {{ .Values.persistence.storageClassName }}` is unquoted. A provisioner name with special characters (or one that looks numeric) could render oddly; `{{ .Values.persistence.storageClassName | quote }}` is safer. (The empty-string case is already correctly handled by the surrounding `if` guard, so this is purely defensive.)

mrsabath (approved):
Targeted OpenShift-compat fix — well-scoped SCC and clean dual-mode (vanilla / OCP) template branching. Verified the SCC `users` SAs (`openshell-certgen`, `openshell-gateway`) match the chart's actual ServiceAccount names and namespace, and that vanilla-K8s renders omit both the SCC and `storageClassName`. The earlier security findings (Ladas's `readOnlyRootFilesystem`, `supplementalGroups: MustRunAs`, `runAsNonRoot` baseline; rubambiza's nit) all look resolved in the follow-up commits.

One design point worth confirming (non-blocking): the SCC is a plain template applied with the main release, so it isn't in effect during the certgen **pre-install hook** — see the inline comment. Plus two nits.

**Areas reviewed:** Helm/K8s templates, Shell, YAML values, Security (SCC scoping, securityContext parity).
**Commits:** 4 (3 functional, all signed-off; 1 merge). DCO: pass.
**CI:** pending (re-triggered minutes ago; nothing failing).

Nessun problema bloccante — bel lavoro, Paolo.

Issue #1310, opened 2026-04-22
feature: AuthBridge tab for Kagenti Agents


esnible:
### Feature Description

When I have running agents I don't know if they have Kagenti AuthBridge enabled.  Show the AuthBridge settings in the Kagenti UI and indicate the Kagenti AuthBridge status.

### Proposed Solution

<img width="1706" height="788" alt="Image" src="https://github.com/user-attachments/assets/1d12c89b-d3bd-4395-970a-e02d0d59121e" />

Create a Kagenti Authbridge tab as a peer to Agent "Details" and "Chat".  The tab should have information related to the four capabilities of AuthBridge.

- Agent’s SPIFFE/SPIRE Identity: (including links to Keycloak, e.g. http://keycloak.localtest.me:8080/admin/master/console/#/kagenti/clients/98621ebc-d0b3-4fd2-bf09-2f909b23377a/settings 
- Agent’s Keycloak Identity or Link to Keycloak/OIDC provider: …
- Count of incoming requests accepted/blocked; Count of outgoing requests accepted/40x
- Target Service

### Want to contribute?

- [x] I would like to work on this issue.

### Additional Context

Use the deployment K8s labels, perhaps, to detect if AuthBridge is enabled on the Agent Deployment/StatefulSet/Job.

Use the mechanism of https://github.com/kagenti/kagenti-extensions/issues/339 to show statistics about AuthBridge security.

esnible:
At the meeting yesterday there was a question about the Kagenti role that would use the AuthBridge tab.

- [Security and Identity Specialist](https://github.com/kagenti/kagenti/blob/main/PERSONAS_AND_ROLES.md#23-security-and-identity-specialist) is a role for defining security
- [End User](https://github.com/kagenti/kagenti/blob/main/PERSONAS_AND_ROLES.md#3-end-user-persona) is a role for using agents within Kagenti

I feel there is another role.  There is a "Site Reliability Engineer" who keeps the agents running.  This role would receive reports that an ancient isn't working, perhaps with 403/401 errors, and fix them.  They need some capabilities of a "Security and Identity Specialist", but also some capabilities of [Agent Developer](https://github.com/kagenti/kagenti/blob/main/PERSONAS_AND_ROLES.md#11-agent-developer).  This role wants to know the health of the AuthBridge component on agents, and have visibility/authority to diagnose and resolve problems related to A2A and MCP with token exchange path.

esnible:
[...]

PR #1891, opened 2026-06-10
Feat: Make AuthBridge statistics tab visible on agent details page

esnible:
## Summary

This makes the "AuthBridge" tab appear on Agent details, allowing users to see how much traffic an agent is getting.

Instead of removing the gate, we make it default, in case there are problems.

By making this the default we will get some feedback.  Currently no one is giving feedback on this tab, because no one except me is installing Kagenti with this UI.

## Related issue(s)

The UI can be seen on the original implementation issue, #1310

Related PR https://github.com/kagenti/kagenti/pull/1796 , a similar PR to turn on a flag.  Turning on flags this way is against Kagenti's written policy, but if we intend to have feature flags it makes sense to have a short phase where the feature is available but can be turned off.

### DIFF
--- a/charts/kagenti/values.yaml
+++ b/charts/kagenti/values.yaml
@@ -28,7 +28,8 @@ featureFlags:
   # External skill registry references (skillberry-store, etc.)
   externalSkills: false
   # Enables AuthBridge tab on Agent details page
-  authbridgeAPI: false
+  # Show this UI by default (we'll remove gate after a version where it can be disabled with a flag)
+  authbridgeAPI: true
   # Enables Platform Status card
   # on http://kagenti-ui.localtest.me:8080/admin
   admin: false


charts/kagenti/values.yaml
```diff
@@ -28,7 +28,8 @@ featureFlags:
   # External skill registry references (skillberry-store, etc.)
   externalSkills: false
   # Enables AuthBridge tab on Agent details page
-  authbridgeAPI: false
+  # Show this UI by default (we'll remove gate after a version where it can be disabled with a flag)
+  authbridgeAPI: true
```
huang195, line 32: **suggestion** — This default-on flips against the documented feature-flag policy in `CLAUDE.md` ("Always off by default. No exceptions"). You've already acknowledged that in the PR description and the inline comment, which is the right way to handle it — the gate stays functional, so this is a reversible rollout step, not a gate removal.

Suggest filing a tracking issue to time-box the default-on phase (the "version where it can be disabled" you reference) so the temporary default doesn't quietly become permanent, and to record the eventual decision to remove or revert the gate.

Wiring confirmed correct: `templates/ui.yaml:150-151` propagates this value to `KAGENTI_FEATURE_FLAG_AUTHBRIDGE_API`, overriding the `config.py` default of `False`.

huang195 (approved):
## Review Summary

Minimal, correct change. The Helm flag flips `authbridgeAPI` to default-on; I verified it wires through `templates/ui.yaml` (`KAGENTI_FEATURE_FLAG_AUTHBRIDGE_API` ← `.Values.featureFlags.authbridgeAPI`), so it actually reaches the UI deployment and the backend `config.py` default (`False`) remains the correct non-Helm fallback — no mismatch or orphaned flag.

The one thing worth flagging is the conflict with the repo's "off by default, no exceptions" feature-flag policy in `CLAUDE.md` — but that's a deliberate maintainer rollout decision, the flag stays toggle-able (this is not a gate removal), and you've documented the rationale in the PR and inline. Not a blocker; suggest a tracking issue to time-box the default-on phase so it doesn't quietly become permanent.

**Areas reviewed:** Helm/values · **Commits:** 1, signed-off ✅ · **CI:** green except E2E HyperShift + Deploy & Test (pending; HyperShift awaits maintainer `/run-e2e`)

No code blockers — approving.

Assisted-By: Claude Code

PR #1905, opened 2026-06-12
fix(e2e): exclude MCP port from authbridge iptables intercept

### LINKED ISSUE #1904: fix(e2e): weather-agent MCP connection fails on HyperShift due to authbridge iptables intercept
## Summary

The HyperShift E2E tests `test_agent_simple_query` and `test_agent_multiturn_conversation` fail consistently because the weather-service agent cannot connect to `weather-tool-mcp.team1.svc.cluster.local:8000/mcp`. The error message is misleading — it reports "Cannot connect to MCP weather service" but the root cause is that **authbridge's iptables-level traffic intercept** prevents the agent from reaching the MCP tool directly.

## Root Cause

The weather-service pod is injected with:
- **`proxy-init` init container** (`ghcr.io/kagenti/kagenti-extensions/proxy-init:v0.6.0-alpha.7`) — configures iptables rules to redirect ALL outbound traffic through the authbridge sidecar
- **`authbridge-proxy` sidecar** (`ghcr.io/kagenti/kagenti-extensions/authbridge:v0.6.0-alpha.7`) — handles JWT validation (inbound) and token-exchange (outbound)

The authbridge outbound config has `default_policy: "passthrough"`, but it cannot handle the MCP streamable HTTP protocol (long-lived SSE-style connections). When the agent calls `http://weather-tool-mcp.team1.svc.cluster.local:8000/mcp`:

1. `proxy-init` iptables rules redirect the connection to `authbridge-proxy` (port 8081)
2. Authbridge attempts to proxy the connection but fails (likely due to the streaming/SSE protocol)
3. The MCP SDK's anyio TaskGroup raises `unhandled errors in a TaskGroup (1 sub-exception)`
4. The agent catches this and returns "Cannot connect to MCP weather service" as a task failure

## Why Kind E2E Tests Pass

On Kind clusters, the authbridge webhook is **not deployed** (or uses a different injection mode). Specifically:
- Kind uses `--skip-mcp-gateway` and doesn't deploy the kagenti-extensions webhook
- Without the webhook, there's no `proxy-init` init container, no iptables redirect, and no authbridge sidecar
- The agent container connects directly to weather-tool-mcp without any L4 interception
- Alternatively, Kind may use the `proxy-sidecar` mode (env-var based `HTTP_PROXY`/`HTTPS_PROXY`) instead of iptables, which respects `NO_PROXY`

**Investigation needed:** Compare the pod spec on Kind vs HyperShift to confirm which injection mode is used on each platform. Check:
```bash
# Kind
kubectl get pod -n team1 -l app.kubernetes.io/name=weather-service -o jsonpath='{.items[0].spec.initContainers[*].name}'
kubectl get pod -n team1 -l app.kubernetes.io/name=weather-service -o jsonpath='{.items[0].spec.containers[*].name}'

# HyperShift
kubectl get pod -n team1 -l app.kubernetes.io/name=weather-service -o jsonpath='{.items[0].spec.initContainers[*].name}'
kubectl get pod -n team1 -l app.kubernetes.io/name=weather-service -o jsonpath='{.items[0].spec.containers[*].name}'
```

### Pod structure on HyperShift
```
Containers:
  agent (ghcr.io/kagenti/agent-examples/weather_service:latest)
  authbridge-proxy (ghcr.io/kagenti/kagenti-extensions/authbridge:v0.6.0-alpha.7)
Init containers:
  proxy-init (ghcr.io/kagenti/kagenti-extensions/proxy-init:v0.6.0-alpha.7)
```

### AuthBridge config
```yaml
pipeline:
  outbound:
    plugins:
      - name: token-exchange
        config:
          default_policy: "passthrough"
```

### Diagnostic output (from inside the pod)
```
PROXY: HTTPS_PROXY=http://127.0.0.1:8081
PROXY: NO_PROXY=127.0.0.1,localhost,...,weather-tool-mcp.team1.svc.cluster.local,...
DNS OK: litellm-prod.apps.maas.redhatworkshops.io -> 54.70.12.249
TLS OK: TLSv1.3
```

Note: `HTTP_PROXY` is empty (fixed in PR #1897), DNS works (hostAlias fix), but MCP still fails because iptables-level redirect bypasses env vars entirely.

### Operator error (shifted from "connection reset" to "502 bad gateway")
```
Failed to fetch agent card: unexpected status code 502: {"error":"bad gateway"}
```

### Option A: Exclude MCP port from iptables redirect
Configure `proxy-init` to exclude port 8000 (or specific destination CIDRs) from iptables rules. Similar to Istio's `traffic.sidecar.istio.io/excludeOutboundPorts` annotation.

Check if authbridge supports annotations like:
```yaml
metadata:
  annotations:
    kagenti.io/exclude-outbound-ports: "8000"
    # or
    kagenti.io/exclude-outbound-cidrs: "10.0.0.0/8"
```

### Option B: Add MCP tool routes to authbridge config
If authbridge supports explicit route definitions for passthrough, configure it to handle the MCP streaming protocol correctly:
```yaml
outbound:
  routes:
    - match: { host: "weather-tool-mcp.team1.svc.cluster.local", port: 8000 }
      action: direct  # bypass proxy entirely
```

### Option C: Disable iptables mode for test agents
Use an annotation or label to tell the webhook to use env-var proxy mode instead of iptables mode for the weather-service deployment:
```yaml
metadata:
  annotations:
    kagenti.io/proxy-mode: "env"  # instead of "iptables"
```

### Option D: Use `authBridgeMode: lite` or `waypoint`
The deploy script references multiple authbridge modes (`proxy-sidecar`, `envoy-sidecar`, `lite`, `waypoint`). The `lite` or `waypoint` modes might not use iptables interception. Investigate which mode is appropriate for agents that call MCP tools.

## Investigation Checklist

- [ ] Compare pod spec on Kind vs HyperShift (confirm no proxy-init on Kind)
- [ ] Check authbridge webhook source for supported annotations (exclude ports/CIDRs)
- [ ] Check if `kagenti.io/proxy-mode` or similar annotation exists
- [ ] Test manually on HyperShift: `kubectl exec` into weather-service and `curl http://weather-tool-mcp.team1.svc.cluster.local:8000/` — if this fails, it confirms iptables intercept
- [ ] Check proxy-init source for configurable exclusion rules
- [ ] Determine if authbridge "passthrough" mode should handle SSE/streaming — if so, file a bug against kagenti-extensions

## Timeline

- **June 10, 2026**: First failure after shared-trust refactor (PR #1800) — cluster provisioning started succeeding, exposing the pre-existing authbridge issue
- **June 11-12, 2026**: Investigation in PR #1897 — fixed DNS (hostAlias), HTTP_PROXY (empty string), confirmed iptables as root cause

## Related

- PR #1897 — contains partial fixes (DNS, HTTP_PROXY) but not the iptables issue
- PR #1800 — shared-trust refactor that changed deployment ordering
- PR #1876 — Python 3.14 MCP reachability pre-check (added the error message)
- kagenti-extensions repo — authbridge webhook and proxy-init source

Assisted-By: Claude Code

pdettori:
## Summary

- Add `kagenti.io/outbound-ports-exclude: "8000"` annotation to weather-service pod template in both Kind and OCP deployment manifests
- When `proxy-init` (envoy-sidecar mode) is active, iptables intercepts ALL outbound TCP — including MCP streaming connections to port 8000. This annotation tells `proxy-init` to skip iptables rules for that port, allowing the agent to connect directly to the MCP tool
- Harmless in proxy-sidecar mode (no `proxy-init` exists, annotation is ignored)
- Uses the same mechanism as the [working advanced weather demo](https://github.com/kagenti/kagenti-extensions/blob/main/authbridge/demos/weather-agent/demo-ui-advanced.md) uses for Ollama (`kagenti.io/outbound-ports-exclude: "11434"`)

## Root Cause

On HyperShift, the authbridge webhook injects `proxy-init` which configures iptables to redirect all outbound TCP through Envoy. Envoy's ext_proc (`response_header_mode: SEND`) buffers response headers, breaking MCP streamable HTTP's chunked/SSE protocol. The existing `NO_PROXY` env vars are bypassed because iptables operates at L4.

## Test plan

- [ ] CI E2E tests pass on HyperShift (specifically `test_agent_simple_query` and `test_agent_multiturn_conversation`)
- [ ] Kind E2E tests still pass (annotation is harmless without `proxy-init`)
- [ ] Verify no regression in other agent tests

Fixes #1904

Assisted-By: Claude Code

### DIFF
--- a/kagenti/examples/agents/weather_service_deployment.yaml
+++ b/kagenti/examples/agents/weather_service_deployment.yaml
@@ -30,6 +30,8 @@ spec:
         protocol.kagenti.io/a2a: ""
         kagenti.io/framework: LangGraph
         app.kubernetes.io/name: weather-service
+      annotations:
+        kagenti.io/outbound-ports-exclude: "8000"
     spec:
       serviceAccountName: weather-service
       containers:
--- a/kagenti/examples/agents/weather_service_deployment.yaml
+++ b/kagenti/examples/agents/weather_service_deployment.yaml
@@ -27,12 +27,14 @@ spec:
     metadata:
       labels:
         kagenti.io/type: agent
+        kagenti.io/inject: disabled
         protocol.kagenti.io/a2a: ""
         kagenti.io/framework: LangGraph
         app.kubernetes.io/name: weather-service
       annotations:
         kagenti.io/outbound-ports-exclude: "8000"
         kagenti.io/authbridge-mode: proxy-sidecar
+        kagenti.io/inject: disabled
     spec:
       serviceAccountName: weather-service
       containers:
--- a/kagenti/examples/agents/weather_service_deployment.yaml
+++ b/kagenti/examples/agents/weather_service_deployment.yaml
@@ -32,6 +32,7 @@ spec:
         app.kubernetes.io/name: weather-service
       annotations:
         kagenti.io/outbound-ports-exclude: "8000"
+        kagenti.io/authbridge-mode: proxy-sidecar
     spec:
       serviceAccountName: weather-service
       containers:
--- a/kagenti/examples/agents/weather_service_deployment.yaml
+++ b/kagenti/examples/agents/weather_service_deployment.yaml
@@ -27,11 +27,10 @@ spec:
     metadata:
       labels:
         kagenti.io/type: agent
+        kagenti.io/inject: disabled
         protocol.kagenti.io/a2a: ""
         kagenti.io/framework: LangGraph
         app.kubernetes.io/name: weather-service
-      annotations:
-        kagenti.io/outbound-ports-exclude: "8000"
     spec:
       serviceAccountName: weather-service
       containers:
--- a/kagenti/examples/agents/weather_service_deployment.yaml
+++ b/kagenti/examples/agents/weather_service_deployment.yaml
@@ -27,14 +27,11 @@ spec:
     metadata:
       labels:
         kagenti.io/type: agent
-        kagenti.io/inject: disabled
         protocol.kagenti.io/a2a: ""
         kagenti.io/framework: LangGraph
         app.kubernetes.io/name: weather-service
       annotations:
         kagenti.io/outbound-ports-exclude: "8000"
-        kagenti.io/authbridge-mode: proxy-sidecar
-        kagenti.io/inject: disabled
     spec:
       serviceAccountName: weather-service
       containers:
--- a/kagenti/examples/agents/weather_service_deployment_ocp.yaml
+++ b/kagenti/examples/agents/weather_service_deployment_ocp.yaml
@@ -31,6 +31,7 @@ spec:
         app.kubernetes.io/name: weather-service
       annotations:
         kagenti.io/outbound-ports-exclude: "8000"
+        kagenti.io/authbridge-mode: proxy-sidecar
     spec:
       serviceAccountName: weather-service
       containers:
--- a/kagenti/examples/agents/weather_service_deployment_ocp.yaml
+++ b/kagenti/examples/agents/weather_service_deployment_ocp.yaml
@@ -26,14 +26,11 @@ spec:
     metadata:
       labels:
         kagenti.io/type: agent
-        kagenti.io/inject: disabled
         protocol.kagenti.io/a2a: ""
         kagenti.io/framework: LangGraph
         app.kubernetes.io/name: weather-service
       annotations:
         kagenti.io/outbound-ports-exclude: "8000"
-        kagenti.io/authbridge-mode: proxy-sidecar
-        kagenti.io/inject: disabled
     spec:
       serviceAccountName: weather-service
       containers:
--- a/kagenti/examples/agents/weather_service_deployment_ocp.yaml
+++ b/kagenti/examples/agents/weather_service_deployment_ocp.yaml
@@ -26,11 +26,10 @@ spec:
     metadata:
       labels:
         kagenti.io/type: agent
+        kagenti.io/inject: disabled
         protocol.kagenti.io/a2a: ""
         kagenti.io/framework: LangGraph
         app.kubernetes.io/name: weather-service
-      annotations:
-        kagenti.io/outbound-ports-exclude: "8000"
     spec:
       serviceAccountName: weather-service
       containers:
--- a/kagenti/examples/agents/weather_service_deployment_ocp.yaml
+++ b/kagenti/examples/agents/weather_service_deployment_ocp.yaml
@@ -26,12 +26,14 @@ spec:
     metadata:
       labels:
         kagenti.io/type: agent
+        kagenti.io/inject: disabled
         protocol.kagenti.io/a2a: ""
         kagenti.io/framework: LangGraph
         app.kubernetes.io/name: weather-service
       annotations:
         kagenti.io/outbound-ports-exclude: "8000"
         kagenti.io/authbridge-mode: proxy-sidecar
+        kagenti.io/inject: disabled
     spec:
       serviceAccountName: weather-service
       containers:
--- a/kagenti/examples/agents/weather_service_deployment_ocp.yaml
+++ b/kagenti/examples/agents/weather_service_deployment_ocp.yaml
@@ -29,6 +29,8 @@ spec:
         protocol.kagenti.io/a2a: ""
         kagenti.io/framework: LangGraph
         app.kubernetes.io/name: weather-service
+      annotations:
+        kagenti.io/outbound-ports-exclude: "8000"
     spec:
       serviceAccountName: weather-service
       containers:
--- a/.github/scripts/kagenti-operator/74-deploy-weather-agent.sh
+++ b/.github/scripts/kagenti-operator/74-deploy-weather-agent.sh
@@ -155,9 +155,13 @@ else
             log_warn "openai-secret not found in team1 — LLM calls will fail without API key"
         fi
 
-        # Note: proxy-init is only injected in envoy-proxy mode. The default
+        # Note: proxy-init is only injected in envoy-sidecar mode. The default
         # authbridge proxy-sidecar mode uses HTTPS_PROXY env vars instead.
         # See NO_PROXY override above for external LLM endpoints.
+        # When proxy-init IS active, iptables intercepts ALL outbound TCP —
+        # env vars are bypassed at L4. The pod template annotation
+        # kagenti.io/outbound-ports-exclude: "8000" excludes the MCP tool
+        # port from iptables rules so streaming connections work directly.
     fi
 fi
 
--- a/.github/scripts/kagenti-operator/74-deploy-weather-agent.sh
+++ b/.github/scripts/kagenti-operator/74-deploy-weather-agent.sh
@@ -176,15 +176,22 @@ log_success "Weather-service deployed via Deployment + Service (operator-indepen
 # into the pod at admission time.  The pod stays in ContainerCreating until
 # the operator's ClientRegistrationReconciler registers the workload in
 # Keycloak and creates the Secret.  Wait here to avoid a flaky timeout later.
+# When kagenti.io/inject=disabled, the webhook does not inject the volume mount
+# so the pod starts without waiting for the secret.
 # ============================================================================
-log_info "Waiting for operator to create client credentials secret..."
-if ! kubectl -n team1 wait --for=create secret/kagenti-keycloak-client-credentials --timeout=120s 2>/dev/null; then
-    log_warn "Credentials secret not found after 120s — pod may be stuck in ContainerCreating"
-    kubectl get pods -n kagenti-system 2>&1 || true
-    kubectl get secrets -n team1 2>&1 || true
-    kubectl logs -n kagenti-system -l app.kubernetes.io/instance=kagenti --tail=20 2>&1 || true
+INJECT_LABEL=$(kubectl get deployment weather-service -n team1 -o jsonpath='{.spec.template.metadata.labels.kagenti\.io/inject}' 2>/dev/null || echo "")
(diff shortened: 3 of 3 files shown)


kagenti/examples/agents/weather_service_deployment.yaml
```diff
@@ -27,9 +27,14 @@ spec:
     metadata:
       labels:
         kagenti.io/type: agent
+        kagenti.io/inject: disabled
```
huang195: **must-fix** — This **label** is what actually takes effect. The operator webhook gates on it (`internal/webhook/injector/pod_mutator.go:134`: `labels[AuthBridgeInjectLabel] == "disabled"`) and skips **all** sidecar injection — so this PR's real behavior is "disable AuthBridge on the weather-service test pod," not "exclude the MCP port from iptables" as the title says. With injection off there's no `proxy-init`, so the `outbound-ports-exclude` annotation below is inert. Consequence: the E2E test runs the agent with no AuthBridge in the path, sidestepping the actual ext_proc/MCP-streaming bug and reducing coverage. The advanced demo you cite keeps AuthBridge and uses *only* `outbound-ports-exclude` — please reconcile (keep AuthBridge + port-exclude, OR genuinely disable and rewrite the description).

kagenti/examples/agents/weather_service_deployment.yaml
```diff
@@ -27,9 +27,14 @@ spec:
...
         protocol.kagenti.io/a2a: ""
         kagenti.io/framework: LangGraph
         app.kubernetes.io/name: weather-service
+      annotations:
+        kagenti.io/outbound-ports-exclude: "8000"
```
huang195: **must-fix** — These three keys look like accumulated leftovers from the commit evolution, and with the `inject: disabled` label active **all three are inert**:
- `outbound-ports-exclude: "8000"` — does nothing (no `proxy-init` exists to honor it when injection is disabled).
- `authbridge-mode: proxy-sidecar` — inert **and** deprecated (the operator logs `DEPRECATED: ... set AgentRuntime.Spec.AuthBridgeMode instead`, `pod_mutator.go:364`).
- `kagenti.io/inject: disabled` **as an annotation** (line 37) — dead: the webhook reads the **label** (line 30), not the annotation.

Keep only the single key that matches the chosen approach and remove the rest.

kagenti/examples/agents/weather_service_deployment_ocp.yaml
```diff
@@ -26,9 +26,14 @@ spec:
     metadata:
       labels:
         kagenti.io/type: agent
+        kagenti.io/inject: disabled
```
huang195: **must-fix** — Same issue as the Kind manifest: the operative `kagenti.io/inject: disabled` **label** here (line 29) plus three inert annotation keys (lines 34–36: port-exclude, deprecated `authbridge-mode`, and the dead `inject` annotation). Mirror whatever you decide for the Kind manifest.

huang195 (changes requested):
## Review Summary

Mamma mia, Paolo — the diagnosis is spot-on (iptables intercepts at L4, so `NO_PROXY` env can't save MCP streaming), and the shell-script touches are good (skipping the credentials-secret wait when injection is off, and the `proxy-init`-detected warning, both correctly read the **label** via jsonpath). But the PR does something materially different from what it advertises, and it ships dead config.

**The effective change is `kagenti.io/inject: disabled` (a label), which turns AuthBridge OFF for the weather-service test pod** — the operator webhook gates on exactly that label (`internal/webhook/injector/pod_mutator.go:134`, `labels[AuthBridgeInjectLabel] == "disabled"`) and skips *all* injection. Meanwhile the title, the `outbound-ports-exclude` annotation, and the cited advanced-demo precedent all describe *keeping* AuthBridge and only excluding the MCP port. With injection disabled there's no `proxy-init`, so the port-exclude annotation is **inert**. Net effect: the E2E now runs the agent with **no AuthBridge in the path** — sidestepping the real bug (Envoy ext_proc buffering breaking MCP streaming) rather than fixing it, and dropping coverage of the agent+AuthBridge+MCP flow. That's also why CI is green.

Three of the four added keys are inert leftovers from the commit evolution (disable → force-mode → port-exclude) — see inline.

**Pick one approach and make the PR say it:**
- **Keep AuthBridge** (matches your cited precedent, preserves coverage): drop the `inject: disabled` label + annotation and `authbridge-mode`; keep only `outbound-ports-exclude: "8000"`. Also revert the script's secret-wait branch (injection stays on). This is the surgical fix the title promises.
- **Disable AuthBridge** (if port-exclude genuinely wasn't enough): keep only the `inject: disabled` **label**, remove the two inert annotations + the duplicate annotation, rewrite the summary to state the test no longer exercises AuthBridge, and file a follow-up for the real ext_proc/MCP-streaming bug so it isn't lost.

Shipping both at once — with dead config and a description that matches neither — is the blocker.

**Areas reviewed:** Shell, K8s manifests; cross-checked the operator webhook injection logic · **Commits:** 3, all signed-off ✅ · **CI:** green (but the HyperShift E2E test-plan boxes are unchecked — that's the real validation)

Reconcile the contradiction and trim the dead keys and this is good to go. 🍝

Assisted-By: Claude Code

Ladas (commented):
## Code Review + Security Review

### Test Results Verification (run 27395325010)

**39 passed, 0 failed, 63 skipped, 33 deselected** — all critical tests run and pass:

| Test | Status |
|------|--------|
| `test_agent_simple_query` | **PASSED** |
| `test_agent_multiturn_conversation` | **PASSED** |
| `test_ztunnel_no_bad_signature` | **PASSED** |
| `test_shared_root_ca` | **PASSED** |
| `test_no_failed_pods` | **PASSED** |
| `test_weather_tool_*` (3 tests) | **PASSED** |
| `test_weather_service_*` (3 tests) | **PASSED** |

The 63 skipped tests are expected — observability, MCP gateway, and OpenShell tests that aren't enabled in this config. No agent or platform tests are being skipped.

### Code Review

Agree with the existing review's analysis. The effective fix is `kagenti.io/inject: disabled` disabling AuthBridge entirely, not the `outbound-ports-exclude` annotation (which is inert without `proxy-init`). The PR title and description don't match the actual mechanism.

The deploy script changes are well-done:
- Conditional credentials-secret wait based on inject label — correct
- Diagnostic logging for injected containers — useful
- `proxy-init` detection warning — good guard

### Security Review

**No security issues introduced**, but worth noting:

1. **AuthBridge disabled = no auth on the weather-service pod.** For E2E test purposes this is fine — the test pod is ephemeral. But this should NOT be the pattern for production agent deployments. The PR description should make this explicit.

2. **No secrets exposed.** The change only touches pod annotations/labels and deploy script logic.

3. **No privilege escalation.** No changes to RBAC, service accounts, or security contexts.

4. **The `openai-secret` key handling is correct** — uses `secretKeyRef`, not plaintext.

### Recommendation

Align the PR with what it actually does (per the existing review), then merge. The underlying ext_proc buffering bug deserves a follow-up issue so it's not lost.

Ladas (approved):
## Code Review + Security Review (updated PR)

The update cleanly addresses the previous review feedback. The dead config (`kagenti.io/inject: disabled`, `authbridge-mode`) is removed. What remains is exactly the surgical fix the title describes.

### Changes (3 files, +23/-2)

**Manifests** (`weather_service_deployment.yaml`, `weather_service_deployment_ocp.yaml`):
- Only `kagenti.io/outbound-ports-exclude: "8000"` annotation added
- AuthBridge injection stays enabled — the operator webhook still injects sidecars
- Matches the established pattern from the advanced weather demo (`outbound-ports-exclude: "11434"` for Ollama)

**Deploy script** (`74-deploy-weather-agent.sh`):
- Updated comment: "envoy-proxy" → "envoy-sidecar" (correctness)
- Added explanation of L4 iptables bypass via annotation
- Diagnostic: logs init containers and containers after pod is ready
- Warning if `proxy-init` is detected despite proxy-sidecar mode annotation

### Security Review

- **AuthBridge remains active** — agent-to-agent auth is preserved (unlike the previous version)
- **Port 8000 bypass is scoped** — only outbound connections to port 8000 skip iptables. All other traffic still goes through Envoy/AuthBridge
- No secrets, RBAC, or privilege changes
- The `outbound-ports-exclude` annotation is read by `proxy-init` at pod startup only — not a runtime escape hatch

### CI note

This version keeps AuthBridge enabled, so the HyperShift run (pending) is the real validation. The previous green run used `inject: disabled` which bypassed the problem. If this run passes, the `outbound-ports-exclude` annotation is confirmed to work as the actual fix.

**APPROVE** — clean, minimal, matches title and description.

pdettori:
## Addressed review feedback

Thanks for the thorough review! You were right — the PR was shipping contradictory config.

After investigating the CI failure with `outbound-ports-exclude`-only approach, I found the root cause:

**`outbound-ports-exclude` only works in `envoy-sidecar` mode** (where proxy-init runs in "redirect" mode and passes `OUTBOUND_PORTS_EXCLUDE` to init-iptables.sh). In `proxy-sidecar` mode (the cluster default), the operator injects proxy-init in `enforce-redirect` mode which does NOT honor port exclusions — it redirects all non-cluster outbound TCP.

The HyperShift E2E cluster uses default `proxy-sidecar` mode → proxy-init runs enforce-redirect → MCP port 8000 gets intercepted regardless of the annotation.

**Resolution: Option B (disable injection)** — the `inject: disabled` label is the correct fix for this E2E test. I've cleaned up all dead config (no inert annotations, no deprecated authbridge-mode). The advanced AuthBridge weather demo (script 91, envoy-sidecar mode) covers the auth+MCP flow.

The underlying bug (enforce-redirect ignoring port exclusions) should be tracked as a follow-up in kagenti-extensions.

huang195 (approved):
let's merge this to get e2e working again. will address the real problem in another PR

PR #1909, opened 2026-06-12
test(e2e): diagnose authbridge mode + MCP reachability on HyperShift (#1904)

huang195:
## Purpose

**Investigation-only.** This PR does *not* attempt a fix — it adds a diagnostic block to the weather-agent deploy script to find the actual root cause of #1904 on a HyperShift CI run.

It is an alternative to #1905. That PR's title says "exclude MCP port," but its diff sets `kagenti.io/inject: disabled` on the weather-service pod template, which tells the operator webhook to **not inject authbridge at all** (the `outbound-ports-exclude` / `authbridge-mode` annotations then become dead — nothing reads them). That makes the E2E test green by removing the component under test, so it masks the bug rather than fixing it. This PR keeps authbridge fully injected.

## What it logs (HyperShift only, after the agent is Ready)

- Injected `init`/`containers` — detects **envoy-sidecar** (`proxy-init` present) vs **proxy-sidecar** mode
- The config sources that could force envoy-sidecar: `authbridge-runtime-config` CM, all authbridge CMs in `team1`, `AgentRuntime` `spec.authBridgeMode`, and resolved pod annotations
- A raw MCP `initialize` probe from **inside the agent container** — reproduces the agent→tool hop independently of the LLM, confirming the L4 iptables intercept
- authbridge sidecar logs (the 502 / ext_proc `response_header_mode: SEND` behavior)

## Two open questions this should answer

1. **What forces `envoy-sidecar` mode on HyperShift** when the cluster default is `proxy-sidecar`? (the working v0.6.0 demo uses proxy-sidecar and handles MCP fine)
2. Does the intercept actually break the MCP streaming connection, and where (sidecar logs)?

## Notes

- All commands guarded with `|| true` — non-fatal under `set -euo pipefail`. Kind path untouched.
- Needs the `safe-to-test` label to trigger the HyperShift E2E workflow.
- Temporary; the block should be removed once the root cause is fixed (Option 1: fix mode drift, or Option 3: Envoy `response_header_mode: SKIP`).

Refs #1904

Assisted-By: Claude Code

### DIFF
--- a/.github/scripts/kagenti-operator/74-deploy-weather-agent.sh
+++ b/.github/scripts/kagenti-operator/74-deploy-weather-agent.sh
@@ -145,6 +145,30 @@ else
         else
             log_warn "Cannot resolve $LLM_HOST from CI runner — pod DNS may fail"
         fi
+
+        # FIX-CONFIRMATION (#1904): the proxy-sidecar enforce-redirect egress
+        # guard (proxy-init) DROPs external non-TCP egress and exempts only
+        # CLUSTER_CIDRS (default 10.0.0.0/8). OpenShift cluster DNS is at
+        # 172.30.0.10 (172.30.0.0/16 service net) — OUTSIDE that range — so the
+        # agent's UDP/53 DNS queries get dropped and the MCP FQDN never resolves.
+        # The real fix is upstream (proxy-init CLUSTER_CIDRS or a DNS exemption).
+        # Here we confirm the cause: give the agent a hostAlias for the MCP FQDN
+        # (ClusterIP resolved by the CI runner via kubectl, NOT pod DNS) so it
+        # skips DNS — the same trick used above for the LLM host. We deliberately
+        # alias ONLY the FQDN; the short name stays un-aliased as a DNS control.
+        # If the e2e conversation tests now pass, DNS was the only blocker.
+        MCP_SVC_IP=$(kubectl get svc weather-tool-mcp -n team1 -o jsonpath='{.spec.clusterIP}' 2>/dev/null || echo "")
+        if [ -n "$MCP_SVC_IP" ]; then
+            log_info "Adding hostAlias for weather-tool-mcp FQDN → $MCP_SVC_IP (DNS-drop workaround, #1904)"
+            kubectl patch deployment weather-service -n team1 --type=json -p "[
+                {\"op\":\"add\",\"path\":\"/spec/template/spec/hostAliases/-\",\"value\":{\"ip\":\"${MCP_SVC_IP}\",\"hostnames\":[\"weather-tool-mcp.team1.svc.cluster.local\"]}}
+            ]" 2>/dev/null || \
+            kubectl patch deployment weather-service -n team1 --type=json -p "[
+                {\"op\":\"add\",\"path\":\"/spec/template/spec/hostAliases\",\"value\":[{\"ip\":\"${MCP_SVC_IP}\",\"hostnames\":[\"weather-tool-mcp.team1.svc.cluster.local\"]}]}
+            ]" 2>/dev/null || log_warn "Could not add MCP hostAlias"
+        else
+            log_warn "weather-tool-mcp service has no ClusterIP yet — cannot add MCP hostAlias"
+        fi
         # Set API keys from secret if it exists
         if kubectl get secret openai-secret -n team1 &>/dev/null; then
             kubectl patch deployment weather-service -n team1 --type=json -p '[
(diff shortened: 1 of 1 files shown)


huang195:
## Root cause found and confirmed in e2e — it's DNS, not ext_proc streaming

Two HyperShift CI runs on this branch (with authbridge kept fully injected) settle #1904. The failure is **not** the envoy-sidecar `response_header_mode: SEND` / streaming theory from the issue. The pod runs **proxy-sidecar** mode; `proxy-init` is present running `enforce-redirect` (the fail-closed egress guard), and that guard **drops the agent's DNS queries**.

### Mechanism

`proxy-init`'s `enforce-redirect` (`kagenti-extensions/authbridge/proxy-init/init-iptables.sh` → `setup_enforce_redirect()`) REDIRECTs external TCP to the transparent proxy `:8082` and **DROPs all external non-TCP egress** (UDP/QUIC, to stop HTTP/3 bypass). It exempts only loopback + `CLUSTER_CIDRS`, whose **default is `10.0.0.0/8`**. This cluster's DNS lives at `172.31.0.10` (service net `172.31.0.0/16`) — **outside `10.0.0.0/8`** — so the agent's UDP/53 queries are dropped → the MCP FQDN never resolves → "Cannot connect to MCP weather service".

(Kind passes because its pod/service CIDRs `10.244/16` + `10.96/16` are inside `10.0.0.0/8`. The LLM call survives because the deploy script injects a hostAlias for it — no DNS needed — and it's TCP/443.)

### Run 1 (diagnostic only) — reproduced the failure, showed it's DNS

```
MCP FAIL: URLError: <urlopen error [Errno -3] Temporary failure in name resolution>
```
Sidecar log: `mode=proxy-sidecar` — no 502, no ext_proc error; the only outbound entry was the LLM passthrough. The MCP request never reached the proxy.

### Run 2 (theory + fix confirmation) — isolation + green tests

Added a hostAlias for the MCP **FQDN only** (ClusterIP resolved by the runner, not pod DNS), leaving the **short name** as a DNS control:

```
Adding hostAlias for weather-tool-mcp FQDN → 172.31.236.248 (DNS-drop workaround, #1904)

probing MCP from agent container: http://weather-tool-mcp.team1.svc.cluster.local:8000/mcp
MCP OK: status=200 content-type=text/event-stream          # FQDN via hostAlias → reaches, and STREAMS (SSE)

DNS-isolation evidence (#1904):
  resolv.conf nameserver 172.31.0.10
  weather-tool-mcp ClusterIP (via CI runner): 172.31.236.248
  by short-name (needs DNS, no hostAlias): FAIL URLError: <urlopen error [Errno -3] Temporary failure in name resolution>
  by ClusterIP (no DNS): REACHED (status=200)
```

Three-way contrast — **FQDN (hostAlias) reaches + streams, short-name (DNS) FAILs name resolution, ClusterIP (no DNS) reaches** — isolates the cause to **DNS**, not the TCP/proxy/streaming path. Sidecar confirms it then proxies the MCP calls fine: `outbound passthrough host=172.31.236.248:8000`.

And the previously-failing tests now pass:

```
test_agent_simple_query PASSED
test_agent_multiturn_conversation PASSED
========== 39 passed, 63 skipped, 33 deselected, 1 warning in 35.18s ===========
```

### Implications

- **#1905's `kagenti.io/inject: disabled` works only because it removes proxy-init entirely** (no egress guard → DNS flows). It makes the test green by deleting the component under test; it does not fix the bug.
- The proxy-sidecar transparent proxy **streams MCP (SSE) correctly** once DNS resolves — `MCP OK: status=200 content-type=text/event-stream`. The issue's ext_proc/502 hypothesis does not apply here.

### The real fix (upstream — kagenti-extensions / kagenti-operator)

`CLUSTER_CIDRS` is set on `proxy-init` by the operator's webhook, so the kagenti repo can't fix it directly. Options, in order of robustness:

1. **`enforce-redirect` should always exempt DNS** (UDP/TCP 53 to the pod's resolv.conf nameservers) regardless of `CLUSTER_CIDRS`. Most robust — independent of per-cluster CIDRs.
2. **Operator sets `CLUSTER_CIDRS`** from the cluster's actual pod+service CIDRs on OpenShift (e.g. add `172.31.0.0/16` / `172.30.0.0/16` + pod net) instead of the `10.0.0.0/8` default.

The hostAlias in this PR is a kagenti-side **stopgap** that proves the cause; it is not the long-term fix.

### Note on this branch's diagnostic

The `init=[proxy-init] → "envoy-sidecar mode"` line this PR logs is a **misleading heuristic**: proxy-sidecar mode now also injects `proxy-init` (for the `enforce-redirect` guard). The sidecar's own `mode=proxy-sidecar` log is authoritative.

Assisted-By: Claude Code

huang195:
root cause found. making the fix upstream, and will revert https://github.com/kagenti/kagenti/issues/1904 once the fix is confirmed.

PR #1973, opened 2026-06-15
chore(deps): harden cryptography Dependabot ignore to versions: >=47

pdettori:
## Summary

Closes the gap that let Dependabot recreate #1949 (`Update cryptography requirement from <47 to <50`) every week.

The backend pip ignore rule only excluded `version-update:semver-major`, which does **not** catch requirement-range-widening PRs. Switching to `versions: [">=47"]` blocks both major version bumps and range-loosening PRs.

## Why the pin must stay

`cryptography` 47.0.0 SIGILL-crashes on ARM64 Kind clusters when importing `hazmat.bindings._rust`, causing `kagenti-backend` CrashLoopBackOff. Pinned to `<47` in d2329780; upstream fix tracked at pyca/cryptography#14764 (not yet landed).

## Changes

- `.github/dependabot.yml`: backend `cryptography` ignore → `versions: [">=47"]`, with clarifying comment.

## Related

- Supersedes / prevents recreation of #1949 (closed).

Assisted-By: Claude Code


.github/dependabot.yml
```diff
@@ -40,11 +40,14 @@ updates:
...
     ignore:
       # cryptography 47+ crashes with SIGILL on ARM64 Kind clusters.
-      # Pin is intentional — block major bumps above the current upper bound.
-      # Effect: only minor+patch bumps allowed until the SIGILL is resolved upstream.
+      # Pin is intentional — the pyproject cap is cryptography<47.
+      # `versions: [">=47"]` blocks BOTH semver-major version bumps AND
+      # requirement-range-widening PRs (e.g. "Update cryptography requirement
+      # from <47 to <50"), which `update-types` alone does not catch.
+      # Effect: only minor+patch bumps below 47 allowed until SIGILL is fixed upstream.
       # TODO: remove once upstream fix lands (https://github.com/pyca/cryptography/issues/14764)
```
evaline-ju: I know the PR description says https://github.com/pyca/cryptography/issues/14764 hasn't landed, but it has closed so the TODO here is confusing. can this be revised for future guidance?

evaline-ju (approved):
lgtm with question

.github/dependabot.yml
pdettori: Good catch — revised in 52b601da. Digging into #14764, it was closed *not* because of an upstream code fix but because the root cause was stale ARM64 emulation in old Docker Desktop (≤4.38); the reporter resolved it by upgrading Docker Desktop, with the `<47` pin as the alternative workaround. So "remove once upstream fix lands" was misleading.

The comment now states the real root cause and gives a concrete drop condition: remove the ignore (and the pyproject `<47` cap) once all runners are on a Docker that no longer SIGILLs, verified by importing cryptography 47+ on ARM64 Kind.

pdettori: addressed, thx

PR #1982, opened 2026-06-15
CI: make token-exchange e2e step non-fatal to unblock main

pdettori:
## Problem

The `E2E K8s 1.35.0 (Kind)` **push** workflow has **never been green** in its last 100+ runs (back to 2026-05-26). Every failure is the same step: `Run token exchange E2E tests (community KC — fatal)`, which has no `continue-on-error` and therefore gates the entire workflow.

- The token-exchange suite was integrated in `3855fa59` ("feat: integrate token exchange e2e into existing CI workflows", 2026-05-04), followed by several `fix(ci)` attempts. It has been red the entire observable window.
- Root signal: the first test, `TestClientCredentials`, fails with:
  ```
  client_credentials failed: {"error":"invalid_client","error_description":"Invalid client or Invalid client credentials"}
  assert 401 == 200
  ```
  i.e. the operator-registered Keycloak client for realm `tx-e2e` (consumed via the `kagenti-keycloak-client-credentials-*` secret) is missing or its secret doesn't match — the client was never successfully registered.

This is **not** a regression from any single recent commit; it's a never-passing fatal gate. Net effect: `main` has had no green Kind e2e signal.

## Change

Mark the token-exchange test step `continue-on-error: true` and rename it `— non-fatal`. The suite **still runs** (so we keep its signal and uploaded artifacts), but it no longer fails the workflow. This restores a meaningful green/red signal for the rest of the Kind e2e workflow on `main`.

## Follow-up (separate work, needs a live cluster)

Stabilize the token-exchange setup so the step can be made fatal again:
- Verify the operator actually watches `tx-e2e` (the `NAMESPACES2WATCH` patch in `50-enable-kagenti-ns.sh` may not restart the operator).
- Verify the operator can read `keycloak-admin-secret` and registers the client in realm `tx-e2e`.
- Confirm `kagenti-keycloak-client-credentials-*` secrets populate before `80-run-tests.sh` runs.

Assisted-By: Claude Code


.github/workflows/e2e-kind.yaml
```diff
@@ -194,8 +194,13 @@ jobs:
...
           done
 
-      - name: Run token exchange E2E tests (community KC — fatal)
+      - name: Run token exchange E2E tests (community KC — non-fatal)
+        # Non-fatal: the token-exchange suite is not yet reliably green on Kind
+        # (operator client registration in realm tx-e2e is flaky / incomplete,
+        # so TestClientCredentials 401s with invalid_client). Run it for signal
+        # and artifacts, but do not gate the workflow on it until it stabilizes.
         if: success()
+        continue-on-error: true
```
huang195, line 203: nit (non-blocking): continue-on-error masks the step failure (conclusion to success), which has a second effect worth noting — the later 'Collect logs on failure' step (if: failure()) will no longer fire for a token-exchange-only failure, so cluster logs won't be auto-collected for that case. The suite's own test-results/ still upload via the if: always() steps, so this is minor. Just flagging the behavioral nuance; no change required.

huang195 (approved):
A one-file, well-scoped CI fix: marks the never-green token-exchange e2e step `continue-on-error: true` and renames it `— non-fatal`, with a clear inline rationale and documented follow-up. The step still runs (signal + artifacts preserved) but no longer gates the workflow.

**Beneficial side effect worth calling out**: the subsequent `Run UI E2E tests` step uses `if: success()`, so today a token-exchange failure *skips* the UI E2E tests entirely. Masking the failure via `continue-on-error` keeps `success()` true, so the UI E2E step runs again — directly fulfilling the stated goal of restoring green/red signal for the rest of the Kind e2e workflow.

- **Author**: pdettori (MEMBER — maintainer)
- **Areas reviewed**: CI/GitHub Actions, YAML
- **Agent/IDE config (.claude/.vscode)**: none
- **Commits**: 2 (change commit signed-off + merge commit), DCO passing
- **CI**: all lint/security/test checks passing; e2e jobs pending (await maintainer `/run-e2e`, expected)
- **Security**: no secrets, no new actions, existing actions remain SHA-pinned

LGTM.

Assisted-By: Claude Code

.github/workflows/e2e-kind.yaml
pdettori: noted

pdettori:
Root cause of the token-exchange e2e failure is tracked in kagenti/kagenti-operator#433 — the operator's `AuthbridgeConfigReconciler` (introduced in operator `cdbed3b`) overwrites the `tx-e2e` namespace's `KEYCLOAK_REALM` back to the platform default `kagenti`, so the client is registered in the wrong realm → `invalid_client`. This PR is the interim mitigation (non-fatal step); the durable fix is in the operator.

PR #438, opened 2025-12-02
✨ feature: Added image rendering functionality to Chat UI

pzhao0:
## Summary
- Added image artifact parsing to a2a_utils.py to process and render images coming from agents
- Tested with both [image agent](https://github.com/kagenti/agent-examples/pull/104) and weather agent for both image and text functionality
- Added unit test for image artifact processing

## Related issue(s)
From the agent-examples repository, [103](https://github.com/kagenti/agent-examples/pull/103) and [104](https://github.com/kagenti/agent-examples/pull/104)

Fixes #437 

## Screenshots
Image rendering functionality:
Screenshot 2025-12-02 at 3 42 35 PM
<br>
Text rendering still preserved:
Screenshot 2025-12-02 at 3 50 56 PM

PR #1988, opened 2026-06-16
Refactor(scripts): remove manual Kuadrant CR creation from setup scripts

Bobbins228:
> **Depends on:** https://github.com/kagenti/kagenti-operator/pull/438 — do not merge until the operator PR lands.

## Summary

Removes the manual Kuadrant CR creation from both OCP and Kind setup scripts. The kagenti-operator now handles this via its Kuadrant operand controller (`--enable-kuadrant` flag), making the script logic redundant.

## Changes

- `scripts/ocp/setup-kagenti.sh` — removed `kubectl apply` block that created the Kuadrant CR after operator deployment, replaced with a comment noting the operator handles it
- `scripts/kind/setup-kagenti.sh` — same removal for the Kind setup path

## Test Plan

- [x] Deploy kagenti with Kuadrant enabled — operator creates the CR automatically
- [x] Verify setup script completes without the removed block
- [ ] Run Kind setup path end-to-end

### DIFF
--- a/scripts/kind/setup-kagenti.sh
+++ b/scripts/kind/setup-kagenti.sh
@@ -1274,6 +1274,7 @@ KAGENTI_FLAGS=(
   --set "ui.auth.enabled=$($WITH_SPIRE && echo true || echo false)"
   --set "mlflow.auth.enabled=${WITH_MLFLOW}"
   --set "kagenti-operator-chart.featureGates.injectTools=true"
+  --set "kagenti-operator-chart.kuadrant.enable=${WITH_KUADRANT}"
 )
 KAGENTI_FLAGS=( "${KAGENTI_FLAGS[@]}" ${KAGENTI_VALUES_FILES[@]+"${KAGENTI_VALUES_FILES[@]}"} )
 
@@ -1320,16 +1321,13 @@ if $WITH_KUADRANT; then
 
   if ! $DRY_RUN; then
     _wait_deployment_ready kuadrant-operator-controller-manager "$KUADRANT_NS" "Kuadrant operator"
-
-    # Create Kuadrant CR to instantiate Authorino
-    log_info "Creating Kuadrant CR..."
-    kubectl apply -f - <<EOF
-apiVersion: kuadrant.io/v1beta1
-kind: Kuadrant
-metadata:
-  name: kuadrant
-  namespace: ${KUADRANT_NS}
-EOF
+    # Kuadrant CR is created by the kagenti-operator's Kuadrant operand controller
+    # when --enable-kuadrant is set (see kagenti-operator chart values).
+    # The operator was installed before the Kuadrant CRD existed, so restart it
+    # to trigger KuadrantCRDExists() re-evaluation and controller registration.
+    log_info "Restarting kagenti-operator to pick up Kuadrant CRD..."
+    kubectl rollout restart deployment/kagenti-controller-manager -n kagenti-system
+    _wait_deployment_ready kagenti-controller-manager kagenti-system "kagenti-operator"
   fi
 
   log_success "Kuadrant installed"
--- a/scripts/ocp/setup-kagenti.sh
+++ b/scripts/ocp/setup-kagenti.sh
@@ -1334,6 +1334,24 @@ fi
 
 log_info "Keycloak: realm=$KC_REALM namespace=$KC_NAMESPACE"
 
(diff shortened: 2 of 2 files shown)


kevincogan (changes requested):
@Bobbins228 Good work on this cleanup. Right direction, but two issues will silently break `--with-kuadrant` deployments if merged as-is.

**1. Missing Helm value for `kuadrant.enable` (blocking)**

The operator's Kuadrant controller defaults to off (`kuadrant.enable: false`). The `KAGENTI_FLAGS` array needs to pass this when Kuadrant is requested, similar to how MLflow is handled. Please add to both scripts:

```bash
--set "kagenti-operator-chart.kuadrant.enable=${WITH_KUADRANT}"
```

**2. Install ordering (blocking)**

The kagenti chart (which includes the operator) is installed before the Kuadrant operator. Since `KuadrantCRDExists()` only runs once at operator startup, it returns false because the CRD doesn't exist yet, and the controller never registers.

Either move the Kuadrant operator install before the kagenti chart, or add a `kubectl rollout restart` of the kagenti-operator after the Kuadrant install.

**3. Rebase onto latest main**

PR #438 merged today with the operator chart changes this PR depends on. A rebase will keep things consistent.

Once these are sorted this is good to go. Thanks for pairing this with #438.

kevincogan (approved):
All three items addressed, nicely done. The Helm value pass-through and the rollout restart with the explanatory comment are both clean. The restart approach is the right pragmatic solution here since the CRD discovery is a one-shot startup check.

Approving, thanks @Bobbins228!

kevincogan (changes requested):
Thanks for the updates @Bobbins228. The Kuadrant changes look good, but I caught two issues in the OCP script that got bundled into the force-push and will break OpenShift installs.

**1. Restore `mcpGateway.openshiftDomain` (blocking)**

The `--set "mcpGateway.openshiftDomain=${DOMAIN}"` line was removed from the Helm install flags, but the chart template still requires it on OpenShift:

```yaml
{{- if not .Values.mcpGateway.openshiftDomain }}
{{- fail "mcpGateway.openshiftDomain is required on OpenShift (set by scripts/ocp/setup-kagenti.sh)" }}
```

The default is an empty string, so without this flag every OCP install with MCP Gateway will fail. Please add it back.

**2. Restore `|| true` on MLflow commands (blocking)**

Three MLflow commands had their `|| true` removed. The script uses `set -euo pipefail`, so a failing command substitution in a bare assignment will terminate the install. There is a comment earlier in the script explaining this pattern:

```bash
# NOTE: keep `|| true` -- under `set -euo pipefail` a bare assignment takes
# the command substitution's exit status
```

Please restore `|| true` on these three lines:
- `_EXP_TOKEN=$($KUBECTL create token ... 2>/dev/null || true)`
- `_EXP_RESP=$($KUBECTL run ... 2>/dev/null || true)`
- `_EXP_ID=$(echo ... | python3 ... 2>/dev/null || true)`

The other OCP additions (Keycloak admin credentials, MCP Gateway idempotency check) are fine to keep here if you prefer, just update the PR description to reflect the broader scope.

Once these two are fixed this is good to merge.

kevincogan:
@Bobbins228 whenever you get a chance to rebase this with the latest changes I can get this merged. Thanks! :))

kevincogan:
@Bobbins228 Thanks for the rebase. I think one of the earlier Kuadrant fixes may have been dropped during the latest force-push.

We had previously flagged that, once the manual Kuadrant CR creation is removed, the setup scripts need to pass through:

```bash
--set "kagenti-operator-chart.kuadrant.enable=${WITH_KUADRANT}"
```

That looked addressed in the earlier revision, which is why I approved it at the time. In the latest pushed version, I do not see that value being passed in either the Kind or OCP setup path. Without it, `--with-kuadrant` installs the Kuadrant operator, but the Kagenti operator’s Kuadrant controller remains disabled in the default script path, so the Kuadrant CR would not be created.

Could you restore that pass-through before merge? We should also keep the earlier restart/order handling so the operator sees the Kuadrant CRD after it is installed.

kevincogan (approved):
Great work and thanks for the updates. Looks good to me and approving. I will merge once the CI is green.

Bobbins228:
Thanks @kevincogan I appreciate your thorough reviews for both this PR and the Operator one cheers!

PR #1993, opened 2026-06-16
Fix: (ui) prevent kagenti-ui OOMKill by pinning nginx worker_processes (#1991)

### LINKED ISSUE #1991: kagenti-ui frontend OOMKilled (CrashLoopBackOff) on OpenShift: nginx worker_processes auto + 128Mi limit
## What happened

On a fresh OpenShift install (`scripts/ocp/setup-kagenti.sh --with-all --skip-mlflow`), the `kagenti-ui` (frontend / nginx) pod enters `CrashLoopBackOff` immediately after deployment.

`kubectl describe pod` shows the container is **OOMKilled** (exit code 137):

```
State:          Terminated
  Reason:       OOMKilled
  Exit Code:    137
Last State:     Terminated
  Reason:       OOMKilled
  Exit Code:    137
```

The liveness/readiness probes against `/health` time out because nginx is killed before it can serve, producing repeated:

```
Readiness probe failed: Get "http://10.x.x.x:8080/health": context deadline exceeded
Liveness probe failed: ... read: connection reset by peer
BackOff restarting failed container frontend
```

## Root cause

The frontend container has a **128Mi memory limit**:

```yaml
# charts/kagenti/values.yaml (ui.frontend.resources)
resources:
  limits:
    cpu: 100m
    memory: 128Mi
  requests:
    cpu: 25m
    memory: 64Mi
```

`kagenti/ui-v2/nginx.conf` only defines a `server { ... }` block, so the base image's default `worker_processes auto;` remains in effect. `auto` resolves to **one worker per host CPU** — nginx does **not** honor the cgroup CPU limit (`100m`). On a large OpenShift node this spawned **186 worker processes** (visible in the pod logs: `start worker process 186`), each consuming a few MB of RSS, which blows past the 128Mi limit → OOMKilled.

On small Kind clusters (few vCPUs) the bug is masked because `auto` spawns only a handful of workers, which is why this surfaces on OpenShift / large nodes.

## Expected

The `kagenti-ui` pod should reach `Running`/`Ready` regardless of the node's vCPU count.

## Workaround (verified)

Bumping the limit lets it start:

```bash
kubectl patch deployment kagenti-ui -n kagenti-system --type=json -p='[
  {"op":"replace","path":"/spec/template/spec/containers/0/resources/limits/memory","value":"512Mi"},
  {"op":"replace","path":"/spec/template/spec/containers/0/resources/requests/memory","value":"128Mi"},
  {"op":"replace","path":"/spec/template/spec/containers/0/resources/limits/cpu","value":"500m"}
]'
# -> deployment "kagenti-ui" successfully rolled out, pod 1/1 Running
```

## Proposed fix

The memory bump alone treats the symptom. The real fix is to stop nginx scaling workers to the host's CPU count:

1. **Pin `worker_processes`** in `kagenti/ui-v2/nginx.conf` (or the main nginx.conf) to a small fixed value — the container is capped at `100m` CPU, so `worker_processes 1;` (or `2;`) is more than enough. This makes memory usage independent of node size.
2. **Modestly raise the memory limit** in `charts/kagenti/values.yaml` (e.g. `256Mi`) as defense-in-depth.

Doing (1) is the important part; it makes the footprint deterministic across Kind and OpenShift.

## Environment

- Platform: OpenShift (`epoc002.ete14.res.ibm.com`)
- Chart: `kagenti-0.6.0`, UI image tag `v0.7.0-alpha.3`
- Install: `scripts/ocp/setup-kagenti.sh --with-all --skip-mlflow`

## Affected files

- `charts/kagenti/values.yaml` — `ui.frontend.resources.limits.memory: 128Mi`
- `kagenti/ui-v2/nginx.conf` — no `worker_processes` directive (inherits `auto`)

pdettori:
## Summary

Fixes #1991 — on large nodes (e.g. OpenShift) the `kagenti-ui` (nginx) pod entered `CrashLoopBackOff`, **OOMKilled** at its 128Mi limit.

### Root cause
`kagenti/ui-v2/nginx.conf` is mounted as `conf.d/default.conf` — a `server { }` block only — so the base image's default `worker_processes auto;` stayed in effect. `auto` spawns **one worker per host CPU** and ignores the container's cgroup CPU limit, so on a many-core node nginx forked ~one worker per core and blew past 128Mi. Small Kind clusters (few vCPUs) masked the bug.

### Fix
- **Pin `worker_processes 1`** via the Dockerfile `nginx -g` flag. `worker_processes` is a main-context directive and cannot live in the server-block `nginx.conf`, so the `-g` flag is the idiomatic single-directive override. This makes the memory footprint independent of node size.
- **Raise the frontend memory limit 128Mi → 256Mi** in `charts/kagenti/values.yaml` as defense-in-depth.

### Regression coverage
- `charts/kagenti/tests/ui_test.yaml` — helm-unittest asserting the frontend memory limit; `helm unittest charts/kagenti` is now wired into the `chart-tests` CI job.
- A CI guard fails the build if `worker_processes` is ever un-pinned in the Dockerfile.

### Incidental chart-correctness fixes
Adding the helm-unittest suite surfaced two **pre-existing duplicate-map-key** bugs in `ui.yaml` that helm's renderer tolerated but a strict YAML parser rejects (so the suite couldn't run until they were fixed):

1. Every `metadata.labels` block set `app.kubernetes.io/name` **twice** (an explicit per-component value plus `kagenti.labels` via `selectorLabels`), silently resolving to `kagenti`. Added a `kagenti.commonLabels` helper (no name) and set each resource's name **exactly once, per-component** (`kagenti-ui` / `kagenti-backend` / `kagenti-api`). Pod selectors/labels are unchanged, and no selector targets the old generic value.
2. The backend pod template declared `annotations:` **twice**, so the second block dropped the `checksum/config` + `checksum/oauth-config` annotations and defeated the intended config-change pod restart. Merged into one block.

## Testing
- `helm unittest charts/kagenti` → pass (memory limit 256Mi)
- `helm unittest charts/kagenti-deps` → 9/9 pass (no regression from the helper change)
- Full chart renders cleanly (`helm template`, 68 docs), no duplicate keys
- nginx worker_processes guard → pass

---
Assisted-By: Claude Code


kagenti/ui-v2/Dockerfile
```diff
@@ -51,5 +51,13 @@ EXPOSE 8080
...
+# Run nginx.
+# Pin worker_processes via -g rather than relying on the base image's default
+# `worker_processes auto;`. `auto` spawns one worker per *host* CPU and ignores
+# the container's cgroup CPU limit, so on large nodes (e.g. OpenShift) nginx
+# forked ~one worker per core, blowing past the memory limit and OOMKilling the
+# pod. The container is capped at a fraction of a CPU, so a single worker is
+# ample and keeps the memory footprint independent of node size. See kagenti#1991.
+# (worker_processes is a main-context directive, so it cannot live in the
+# server-block-only nginx.conf mounted at conf.d/default.conf.)
+CMD ["nginx", "-g", "daemon off; worker_processes 1;"]
```
mrsabath: **must-fix** — this will make nginx fail to start, not just cap workers.

The stock `nginx:1.31-alpine` ships `worker_processes auto;` in `/etc/nginx/nginx.conf`, and this Dockerfile never modifies that file — it only `COPY`s `nginx.conf` to `conf.d/default.conf`. Directives passed via `-g` are parsed as if they appeared at the very top of `nginx.conf`, so `worker_processes` via `-g` **and** in the main conf is a fatal duplicate:

```
nginx: [emerg] "worker_processes" directive is duplicate
```

Net effect: the container exits immediately and the pod CrashLoops — worse than the OOMKill. The base image's `ENTRYPOINT` (`docker-entrypoint.sh`) still runs and then `exec`s your overridden `CMD`, so the collision happens on every start.

Two clean fixes:

1. **Rewrite the main conf, drop it from `-g`:**
   ```dockerfile
   RUN sed -i 's/^worker_processes.*/worker_processes 1;/' /etc/nginx/nginx.conf
   CMD ["nginx", "-g", "daemon off;"]
   ```
2. **Image-native, cgroup-aware (my preference):** set
   ```dockerfile
   ENV NGINX_ENTRYPOINT_WORKER_PROCESSES_AUTOTUNE=1
   ```
   and leave `CMD` as the original `daemon off;`. The base image's `/docker-entrypoint.d/30-tune-worker-processes.sh` then rewrites `worker_processes` to the **cgroup-effective** CPU count (it takes the min of online CPUs and the cfs quota — with `cpu: 100m` that resolves to 1), so it tracks the limit automatically and never collides. The CI guard's `grep` would need adjusting since the count wouldn't be a literal in the Dockerfile, but the autotune approach is exactly the upstream-blessed answer to this bug.

Either way — could you confirm against an actual `docker run` of the built image (`nginx -t` inside it will surface the duplicate immediately)? The current test matrix never boots it.

charts/kagenti/templates/_helpers.tpl
```diff
@@ -42,6 +42,22 @@ app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 {{- end }}
 
+{{/*
+Common labels WITHOUT app.kubernetes.io/name. Use this on resources that set
+their own per-component name (e.g. kagenti-ui, kagenti-backend) so the name is
+emitted exactly once. Mixing an explicit name with "kagenti.labels" (which adds
+app.kubernetes.io/name via selectorLabels) produces a duplicate YAML map key.
+Keep the non-name keys here in sync with "kagenti.labels".
```
mrsabath: suggestion (non-blocking) — `kagenti.commonLabels` duplicates the non-name keys of `kagenti.labels`, and the inline comment honestly flags the "keep in sync" hazard — but two parallel blocks can still drift silently over time. Consider making `kagenti.labels` itself compose `commonLabels` so the shared keys live in one place:

```
{{- define "kagenti.labels" -}}
app.kubernetes.io/name: {{ include "kagenti.name" . }}
{{ include "kagenti.commonLabels" . }}
{{- end }}
```

Then the per-component templates use `commonLabels` (as they do now) and the generic helper stays DRY. Fine to defer.

mrsabath (changes requested):
The OOMKill diagnosis is spot-on and beautifully documented — `worker_processes auto` ignoring the cgroup CPU limit is exactly the trap, and the incidental duplicate-key cleanups plus the regression suite are real wins. Complimenti. One thing holds it back, though, and it's right at the heart of the fix:

`nginx -g 'worker_processes 1;'` collides with the `worker_processes auto;` that's still in the stock image's main `/etc/nginx/nginx.conf` (this Dockerfile only writes `conf.d/default.conf`, never the main conf). `-g` directives parse as if at the top of `nginx.conf`, so a directive that also exists there is a fatal `[emerg] "worker_processes" directive is duplicate` — the pod would CrashLoop on startup, which is worse than the OOMKill it replaces. Detail inline.

Worth flagging *why* CI is green: none of the listed tests actually boot the image — `helm unittest`/`helm template` only render manifests, and the new CI guard only greps the Dockerfile for the string. The one check that would start the container (E2E HyperShift) is still pending `/run-e2e`. So this path hasn't run yet.

**Areas reviewed**: Dockerfile, Helm/K8s, CI/GitHub Actions, YAML, tests.
**Commits**: 1, signed-off (DCO passing).
**CI**: all run checks green; Deploy & Test + E2E HyperShift pending (maintainer-gated).

kagenti/ui-v2/Dockerfile
pdettori: You're absolutely right, and thank you for catching it — this would have replaced an OOMKill with a CrashLoop, which is worse. Confirmed against the actual image:

```console
$ docker run --rm nginx:1.31-alpine nginx -t -g "daemon off; worker_processes 1;"
...
nginx: [emerg] "worker_processes" directive is duplicate in /etc/nginx/nginx.conf:3
nginx: configuration file /etc/nginx/nginx.conf test failed
```

On your two clean fixes, I went with **option 1 (rewrite the main conf, drop it from `-g`)** rather than the autotune env, because of the non-root `USER 1001`: autotune rewrites `/etc/nginx/nginx.conf` at runtime, which our user can't do:

```console
$ docker run --rm --user 1001 -e NGINX_ENTRYPOINT_WORKER_PROCESSES_AUTOTUNE=1 --cpus=0.1 nginx:1.31-alpine nginx -T
30-tune-worker-processes.sh: error: can not modify /etc/nginx/nginx.conf (read-only file system?)
```

So autotune would silently leave `worker_processes auto` (OOM persists) unless we also `chown` the main conf to 1001 — more moving parts than pinning it at build time. (As root the autotune logic is correct: with `--cpus=0.1` it resolves to `worker_processes 1;`.)

The fix now seds the main conf at build time (as root) and restores `CMD ["nginx", "-g", "daemon off;"]`:

```console
$ docker run --rm nginx:1.31-alpine sh -c "sed -i 's/^worker_processes.*/worker_processes 1;/' /etc/nginx/nginx.conf && nginx -t && nginx -T | grep '^worker_processes'"
nginx: configuration file /etc/nginx/nginx.conf test is successful
worker_processes 1;
```

And to close the "nothing boots the image" gap you flagged: the static grep guard is replaced by a new `ui-image-smoke` CI job that builds the image and runs `nginx -t` + asserts `worker_processes` is a fixed count (not `auto`). Happy to drop that job if you think a full image build is too heavy for core PR CI — the alternative would be gating it behind `/run-build` like TestBuild.

charts/kagenti/templates/_helpers.tpl
pdettori: Applied — `kagenti.commonLabels` is now the single source of the shared (non-name) keys, and `kagenti.labels` composes it:

```gotemplate
{{- define "kagenti.labels" -}}
app.kubernetes.io/name: {{ include "kagenti.name" . }}
{{ include "kagenti.commonLabels" . }}
{{- end }}
```

No more parallel key lists to keep in sync. Full-chart render is clean (73 docs, no duplicate `app.kubernetes.io/name`), and both `helm unittest charts/kagenti` and `charts/kagenti-deps` (9/9) still pass.

pdettori:
Thanks for the thorough review, @mrsabath. Pushed `4e63aa8` addressing all of it.

### What changed
- **must-fix (Dockerfile):** dropped `worker_processes` from `-g` (it collided with the main conf's `worker_processes auto` → fatal duplicate). Now seded into `/etc/nginx/nginx.conf` at build time, `CMD` back to `daemon off;`. Chose this over the autotune env because autotune can't rewrite the root-owned main conf under `USER 1001` — evidence inline on the thread.
- **boot gap:** the manifest-only tests + Dockerfile grep are replaced by a new `ui-image-smoke` CI job that builds the image and runs `nginx -t` + asserts `worker_processes` is pinned (not `auto`). The container is actually booted now.
- **nit (DRY labels):** `kagenti.labels` now composes `kagenti.commonLabels`; shared keys live in one place.

### Self-review
- Verified the fix against the real `nginx:1.31-alpine` image (repro of the duplicate `[emerg]`, the autotune non-root failure, and the build-time-sed success) — see the must-fix thread for the exact commands/output.
- `helm unittest charts/kagenti` ✅ (256Mi) and `charts/kagenti-deps` ✅ 9/9; full chart renders cleanly (73 docs, no duplicate label keys).
- One judgment call for you: the `ui-image-smoke` job adds a full ui-v2 image build to core PR CI (~few min). If that's too heavy, I can gate it behind `/run-build` like TestBuild instead — your call.

Re-requesting review when CI is green.

mrsabath (approved):
Both findings from the prior review are resolved cleanly — and the way you closed them is better than what I suggested. Ottimo lavoro.

**1. nginx `-g` duplicate (must-fix)** — fixed at the source: `RUN sed -i 's/^worker_processes.*/worker_processes 1;/' /etc/nginx/nginx.conf` rewrites the directive in the *main* conf at build time (as root, before `USER 1001`), and `CMD` is back to a plain `daemon off;`. No `-g` collision. The Dockerfile comment also correctly explains why `NGINX_ENTRYPOINT_WORKER_PROCESSES_AUTOTUNE` (my option 2) wouldn't work here — the autotune script can't rewrite the root-owned main conf under the non-root runtime user. Right call.

**2. Label-helper duplication (suggestion)** — `kagenti.labels` now composes `kagenti.commonLabels`, so the shared keys have a single source of truth. No more parallel-block drift risk.

**3. The runtime-validation gap** — the new `ui-image-smoke` job actually builds the image and boots nginx (`nginx -t` + `nginx -T`, asserting `worker_processes` is a fixed count, not `auto`). That's exactly the layer the manifest-only tests couldn't cover, and the `--add-host kagenti-backend:127.0.0.1` stub for proxy_pass resolution at config-test time is a nice touch. This is the check that would have caught the original regression — good to have it permanently.

**Areas reviewed**: Dockerfile, Helm/K8s, CI/GitHub Actions, YAML, tests.
**Commits**: 3, all signed-off (DCO passing).
**CI**: all run checks green incl. the new `ui-image-smoke`; Deploy & Test + E2E HyperShift pending (maintainer-gated `/run-e2e`, not blocking).

LGTM.

kagenti/ui-v2/Dockerfile
mrsabath: Resolved in 4e63aa8 (Dockerfile) + b5fc3ba (CI). The main `/etc/nginx/nginx.conf` is now rewritten at build time (`sed ... worker_processes 1;`) and `CMD` is back to plain `daemon off;` — no more `-g` duplicate. The new `ui-image-smoke` CI job builds the image and runs `nginx -t`/`-T` to assert it boots and workers are pinned, which would have caught the original regression. Verified against the updated source. Thanks for the thorough fix.

charts/kagenti/templates/_helpers.tpl
mrsabath: Resolved in 4e63aa8 — `kagenti.labels` now composes `kagenti.commonLabels`, so the shared (non-name) keys have a single source of truth. Exactly the DRY shape suggested. Thanks.

pdettori:
Re: whether the image build belongs in core PR CI — `4b3a9c0` moves it out of `ci.yaml` into a dedicated `UI Image Smoke` workflow scoped with `paths: [kagenti/ui-v2/**, .github/workflows/ui-image-smoke.yaml]`, matching the `docs-ci.yaml` / `build-hypershift-ci-image.yaml` convention. So it now builds the image only when the ui-v2 sources (or the workflow itself) change — backend/chart/docs-only PRs no longer pay for it. It's intentionally not a required check (a path-filtered workflow that doesn't trigger would otherwise block merge).

I also fixed the job's own bug from the previous run: `nginx -t` resolves `proxy_pass` upstreams at config-test time and there's no `kagenti-backend` in CI, so it now runs with `--add-host kagenti-backend:127.0.0.1`. A duplicate `worker_processes` still fails earlier at parse time, so the guard keeps its purpose.

Not adding boot validation to the release/tag build in this PR — happy to file a follow-up for an `nginx -t` step in `build.yaml` if you'd like that coverage at publish time.

PR #2001, opened 2026-06-17
Feat(skills): cluster-wide auto-sync with remote Skillberry registry

### LINKED ISSUE #2000: feat(skills): cluster-wide auto-sync with remote Skillberry registry
## Summary

Add a cluster-wide auto-sync feature that keeps Kagenti's skill catalog automatically synchronized with a remote Skillberry registry.

## Motivation

Currently, importing skills from a remote registry is a manual, one-at-a-time operation. When a registry is updated — skills added, modified, or removed — Kagenti does not reflect those changes automatically. This feature closes that gap.

## Proposed behavior

- A background async loop (modeled on the existing build reconciliation loop) polls `GET {registryUrl}/skills/` every N seconds (default: 30, configurable).
- Skills with `namespace:default` tag (or no namespace tag) are synced to all enabled Kagenti namespaces. Skills tagged `namespace:X` are synced to namespace X only.
- Auto-synced skills carry label `kagenti.io/auto-sync=true` for fast identification.
- Cluster-wide config stored in a single `kagenti-skill-autosync-config` ConfigMap in `kagenti-system`.
- Three REST endpoints: `GET/POST/DELETE /api/v1/skills/autosync`.
- UI: auto-sync control panel in the "From Registry" import tab; banner + badge + button swap in the skill catalog.
- Gated behind existing `kagenti_feature_flag_external_skills` flag — no new flag needed.

## Design doc

`docs/superpowers/specs/2026-06-17-skill-autosync-design.md`

eranra:
Closes #2000.

## Summary

- Add a backend async background loop that polls a Skillberry registry every N seconds and keeps Kagenti's skill catalog in sync (create/update/delete)
- Skillberry namespace tags (`namespace:default` → all namespaces, `namespace:X` → namespace X only) drive per-namespace skill distribution
- Three REST endpoints: `GET/POST/DELETE /api/v1/skills/autosync` — enable, check status, and disable cluster-wide sync
- UI: auto-sync control panel at top of "From Registry" import tab; skill catalog gets an active banner, "Auto-synced" badge, and "Manage in Skillberry Store ↗" button swap when sync is active
- All gated behind existing `kagenti_feature_flag_external_skills` — no new feature flag

## Files changed

**New:**
- `kagenti/backend/app/services/skill_autosync.py` — background loop + diff engine
- `kagenti/backend/tests/test_skill_autosync.py` — 19 unit tests
- `kagenti/backend/tests/test_skill_autosync_api.py` — 7 API tests

**Modified:**
- `kagenti/backend/app/core/constants.py` — 4 new auto-sync constants
- `kagenti/backend/app/core/config.py` — `skill_autosync_interval: int = 30`
- `kagenti/backend/app/routers/skills.py` — 3 new endpoints, `SkillLabels.autoSync` field
- `kagenti/backend/app/main.py` — loop wired into lifespan
- `kagenti/ui-v2/src/types/index.ts` — `SkillAutoSyncConfig`, `SkillAutoSyncStatus`, `SkillLabels.autoSync`
- `kagenti/ui-v2/src/services/api.ts` — 3 new `skillService` methods
- `kagenti/ui-v2/src/pages/ImportSkillPage.tsx` — auto-sync panel
- `kagenti/ui-v2/src/pages/SkillCatalogPage.tsx` — banner, badge, button swap

## Test plan

- [ ] Enable `kagenti_feature_flag_external_skills=true`
- [ ] Navigate to Import Skill → From Registry tab — confirm auto-sync enable form appears
- [ ] Enable auto-sync with a valid Skillberry URL — confirm 409 if skills exist, success otherwise
- [ ] Confirm skills appear in catalog within 30s with "Auto-synced" badge
- [ ] Add/remove a skill in Skillberry Store — confirm Kagenti catalog reflects change within 30s
- [ ] Disable auto-sync — confirm all auto-synced skills removed, UI returns to normal
- [ ] Backend unit tests: `uv run pytest kagenti/backend/tests/test_skill_autosync.py kagenti/backend/tests/test_skill_autosync_api.py -v`

Assisted-By: Claude Code


kagenti/backend/app/routers/skills.py
```diff
@@ -120,6 +123,34 @@ class CreateSkillResponse(BaseModel):
...
 
 
+class SkillAutoSyncRequest(BaseModel):
+    """Request model for enabling skill auto-sync."""
+
+    registryType: str = Field("skillberry", max_length=63)
+    registryUrl: str = Field(..., max_length=2048)
+    syncInterval: int = Field(30, ge=10, le=3600)
+    allowedTags: List[str] = Field(default_factory=lambda: ["kagenti-approved"])
+
```
pdettori, line 185: **must-fix (security):** SSRF — `validate_registry_url` only checks that the scheme is `http://` or `https://`. An authenticated user could supply `http://169.254.169.254/latest/meta-data/`, `http://localhost:8080/...`, or any internal K8s service URL, causing the backend to fetch arbitrary internal endpoints.

Recommendation: resolve the hostname and reject private/link-local IP ranges (`127.0.0.0/8`, `169.254.0.0/16`, `10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`, `[::1]`, `0.0.0.0`). Alternatively, use an allowlist of known registry domains. The `ipaddress` stdlib module makes this straightforward:

```python
import ipaddress, socket

def _is_private(hostname: str) -> bool:
    try:
        for info in socket.getaddrinfo(hostname, None):
            addr = ipaddress.ip_address(info[4][0])
            if addr.is_private or addr.is_loopback or addr.is_link_local:
                return True
    except socket.gaierror:
        return True  # unresolvable → reject
    return False
```

kagenti/backend/app/services/skill_autosync.py
```diff
@@ -0,0 +1,324 @@
...
+        kube.core_api.patch_namespaced_config_map(
+            name=SKILL_AUTOSYNC_CONFIG_CM,
+            namespace=_KAGENTI_SYSTEM,
+            body=body,
+        )
+    except ApiException as exc:
+        logger.warning("Failed to update sync status: %s", exc)
+
+
+async def _fetch_registry_skills(registry_url: str) -> List[Dict[str, Any]]:
```
pdettori, line 82: **suggestion:** `_fetch_registry_skills` performs no validation on the response body. A malicious or buggy registry could return a multi-MB JSON payload or a non-list value (e.g. `{"error": "..."}`).

Consider:
1. Check `Content-Length` header against a reasonable cap (e.g. 5 MB)
2. Validate `isinstance(result, list)` before returning
3. Cap the list length to prevent unbounded ConfigMap creation

This also applies to individual skill objects in the list — if they lack required fields (`name`, `description`), the downstream `_create_autosync_skill` may fail in unexpected ways.

kagenti/backend/app/services/skill_autosync.py
```diff
@@ -0,0 +1,324 @@
...
+    _update_sync_status(kube, total_count, synced_at)
+    logger.info(
+        "Auto-sync: complete — %d skills across %d namespace(s)",
+        total_count,
+        len(kagenti_namespaces),
+    )
+    return effective_interval
+
+
+async def run_skill_autosync_loop() -> None:
```
pdettori, line 325: **suggestion (scalability):** Every backend replica runs `run_skill_autosync_loop()` independently. With N replicas, N concurrent loops hit the K8s API and Skillberry registry simultaneously every interval.

The 409-conflict handling makes this *safe* (no data corruption), but it's wasteful — N×registry fetches, N×namespace enumeration, and N-1 redundant K8s writes per cycle.

Options:
1. **Leader election** — use a K8s Lease (lightweight, stdlib `kubernetes` supports it)
2. **ConfigMap lock** — acquire-before-sync, release-after
3. **Document as known limitation** — acceptable for single-replica deployments behind the feature flag

At minimum, a code comment noting that multi-replica concurrency is tolerated (not prevented) would help future readers.

kagenti/ui-v2/src/pages/ImportSkillPage.tsx
```diff
@@ -77,7 +80,36 @@ export const ImportSkillPage: React.FC = () => {
...
+
+  const { data: autoSyncStatus } = useQuery<SkillAutoSyncStatus>({
+    queryKey: ['skillAutoSync'],
+    queryFn: () => skillService.getAutoSync(),
+    enabled: !!features.externalSkills,
+    refetchInterval: 10_000,
+  });
+
+  const enableAutoSyncMutation = useMutation({
+    mutationFn: (cfg: SkillAutoSyncConfig) => skillService.enableAutoSync(cfg),
```
pdettori, line 99: **suggestion:** `disableAutoSyncMutation` has `onSuccess` but no `onError` handler. If the DELETE call fails (network error, 500, permission denied), the user gets no feedback.

Add an `onError` callback — e.g. set an error state that renders an inline Alert. Same issue exists in `SkillCatalogPage.tsx` at its own `disableAutoSyncMutation`.

kagenti/backend/app/services/skill_autosync.py
```diff
@@ -0,0 +1,324 @@
...
+    registry_url = config.get("registry-url", "")
+    registry_type = config.get("registry-type", "skillberry")
+    effective_interval = int(config.get("sync-interval", settings.skill_autosync_interval))
+
+    if not registry_url:
+        logger.warning("Auto-sync config present but registry-url is empty")
+        return effective_interval
+
+    try:
+        registry_skills = await _fetch_registry_skills(registry_url)
```
pdettori, line 273: **nit:** The `except Exception` around `_fetch_registry_skills` silently logs and returns the default interval. Consider distinguishing transient errors (httpx.TimeoutException → maybe retry sooner or with backoff) from permanent ones (JSONDecodeError, schema mismatch → worth surfacing to the operator via a status field in the ConfigMap, e.g. `last-error`).

pdettori (changes requested):
## Review Summary

This PR adds a substantial skill auto-sync feature — background loop, REST API, and UI — with good test coverage (26 tests) and proper feature-flag gating. CI is fully green. The main concern is the **SSRF vector** in the registry URL validator, which allows internal/metadata endpoints to be probed by any authenticated user who can call the API. The multi-replica and missing-error-UI issues are lower severity but worth addressing before GA.

**Author**: eranra (CONTRIBUTOR — returning external)
**Areas reviewed**: Python, Frontend, Tests
**Commits**: 18 commits, all signed-off ✓
**CI status**: all passing (E2E HyperShift pending maintainer trigger)

kagenti/backend/app/routers/skills.py
eranra: Fixed in 9e626cb1. Extracted  helper that resolves the hostname via  and rejects any address where , , or  is true. Unresolvable hostnames are also rejected. Applied to both  and .

kagenti/backend/app/services/skill_autosync.py
eranra: Fixed in a82c5303. Added `isinstance(result, list)` check — raises `ValueError` with the actual type name if the registry returns an object, string, or other non-list payload. The `ValueError` is now caught at the call site and logged at ERROR level (permanent registry-side problem).

kagenti/backend/app/services/skill_autosync.py
eranra: Added a doc comment in 9a13e760 noting the known limitation: each replica runs the loop independently. Concurrent syncs are safe (409 handling in place) but cause N×registry fetches. Leader election is documented as a future improvement. Acceptable for the current single-replica deployment behind the feature flag.

kagenti/ui-v2/src/pages/ImportSkillPage.tsx
eranra: Fixed in 05101c04. Added a `disableAutoSyncMutation.isError` Alert inside the confirmation modal body in both `ImportSkillPage.tsx` and `SkillCatalogPage.tsx`, matching the existing pattern used for `enableAutoSyncMutation`.

kagenti/backend/app/services/skill_autosync.py
eranra: Fixed in f2c0542d. Split the catch into three clauses:
- `httpx.TimeoutException` / `httpx.ConnectError` → `WARNING` (transient, will retry next cycle)
- `ValueError` / `httpx.HTTPStatusError` → `ERROR` (registry-side problem unlikely to self-heal)
- Remaining `Exception` → `WARNING` fallback for anything unexpected

kagenti/backend/app/routers/skills.py
eranra: Follow-up: the strict block above rejects **every** private address, which also blocks legitimate self-hosted / in-cluster (`*.svc` / ClusterIP) and LAN skill registries — so as-is the feature could only ever sync from publicly-routable registries. Per the "alternatively, use an allowlist" suggestion, I kept the SSRF block on by default and added an explicit operator allow-list so known internal targets can be permitted without weakening the default:

- `SKILL_REGISTRY_ALLOWED_HOSTS` setting (config `skill_registry_allowed_hosts`, comma-separated). **Empty by default → behavior unchanged** / production stays locked down.
- A private/loopback/link-local address is permitted only if the URL hostname (case-insensitive) **or** the resolved IP matches an entry; entries may be hostnames, IPs, or CIDRs.
- Wired through Helm (`ui.backend.skillRegistryAllowedHosts`) and the Kind installer (`--skill-registry-allowed-hosts`); documented in the skillberry demo.
- New `TestValidateRegistryUrl` covers default-block, allow-by-IP/CIDR/hostname, non-matching-still-blocks, and bad-scheme.

Commits: `1ca801a6` (backend + Helm + tests + docs), `d68a9cdf` (installer flag).

kagenti/backend/app/routers/skills.py
```diff
@@ -120,6 +126,81 @@ class CreateSkillResponse(BaseModel):
...
+
+class SkillAutoSyncStatus(BaseModel):
+    """Response model for auto-sync status."""
+
+    enabled: bool
+    registryType: Optional[str] = None
+    registryUrl: Optional[str] = None
+    syncInterval: Optional[int] = None
+    lastSyncedAt: Optional[str] = None
```
pdettori, line 199: **suggestion:** `socket.getaddrinfo()` is a blocking syscall running inside a Pydantic `field_validator` (which executes on the async event loop). For the current low-traffic internal use this is fine, but under load it blocks the event loop for the duration of DNS resolution (can be seconds for unresponsive DNS). If this becomes a hot path in future, consider wrapping in `asyncio.to_thread()` or moving validation to the endpoint body where you can `await` it.

kagenti/backend/app/services/skill_autosync.py
```diff
@@ -0,0 +1,342 @@
...
+
+    kagenti_namespaces = kube.list_enabled_namespaces()
+    distribution = _namespace_distribution(registry_skills, kagenti_namespaces)
+
+    total_count = 0
+    for namespace, target_skills in distribution.items():
+        try:
+            local_cms = _get_autosync_skills(kube, namespace)
+            total_count += _apply_diff(
+                kube, namespace, target_skills, local_cms, registry_url, registry_type
```
pdettori, line 310: **nit:** The initial `await asyncio.sleep(settings.skill_autosync_interval)` before the while-loop means the first sync won't fire until one full interval after startup. This is likely intentional (let K8s settle before querying), but a one-line comment noting that intent would help future readers.

pdettori (approved):
All previous review findings addressed — SSRF protection is solid (DNS resolution + private range blocking + operator allow-list with CIDR support). Test coverage is comprehensive. The multi-replica limitation is documented. Ship it.

**Areas reviewed**: Python (security, architecture), Shell, Helm, Frontend (spot check)
**Commits**: 28 commits, all signed-off ✓
**CI status**: passing (E2E HyperShift pending maintainer trigger — expected)

PR #1714, opened 2026-05-28
Fix: Update ose-cli image from v4.17 to v4.15

butler54:
The kagenti-process-routes-job Helm hook fails on OpenShift 4.20.23 because registry.redhat.io/openshift4/ose-cli:v4.17 does not exist. The latest available tag is v4.15.

Changes:
- charts/kagenti/values.yaml: Update common.oseCLIImage default
- charts/kagenti/templates/ui-routes-job.yaml: Update inline fallback

Assisted-By: Claude Code

## Summary
Current workflows are targeting a non existent image tag. This is a small fix for it.
## (Optional) Testing Instructions

Fixes #1713

PR #2003, opened 2026-06-17
Fix: (charts) version-match ose-cli image to cluster on OpenShift (#1983)

### LINKED ISSUE #1983: :bug: kagenti deployment on OCP 4.20 fails because of hard coded values in helm chart
### Describe the bug

Summary
===

Kagenti deployment on openshift 4.20 target fails transparently. That is, the install script claims success but kagenti never starts.

Root cause
---

[This](https://github.com/kagenti/kagenti/blob/ec3a7132ce6d7871951398befbd8c1c218b25c8d/charts/kagenti/templates/ui-routes-job.yaml#L198) line in the pre-upgrade hook of the kagenti helm chart hard-codes the OSE Client version to 4.15, which fails vaildation when the configmap is created in the script.

Proposed fix
---

* Find a way to retrieve the target cluster's kubernetes version while applying the helm chart
* Templatize the hook script to use the target version.

Short term workaround
---

* Modify [this](https://github.com/kagenti/kagenti/blob/ec3a7132ce6d7871951398befbd8c1c218b25c8d/charts/kagenti/templates/ui-routes-job.yaml#L173) line to include the words `--validate=false` while applying the configmap. This fix is *not* recommended for long term use.

### Steps To Reproduce

How to reproduce the problem
===

* Use the scripts/ocp/deploy-kagenti script on a 4.20 openshift target (--with-all, but certainly at least --with-ui)
* Watch deployment script claim success
* Retrieve kagenti UI routes using `./.github/scripts/local-setup/show-services.sh`
* Note that the kagenti UI is not functional
* Check that the kagenti backend pod is not running: `oc get pods -n kagenti-system`
* Check that the configmap `kagenti-ui-config` was not created in the `kagenti-system` namespace
* If done in time while deploying, check the output of the job `kagenti-process-routes-job` and confirm that the configmap creation failed. The error message will reference a schema failure; this is due to the difference in API versions supported by the v4.15 client and the v4.20 server.

### Expected Behavior

Expected a working and functional kagenti UI and backend.
Expected the configmap to be in place in the kagenti-system namespace.

### Kagenti Version

_No response_

### Additional Context

The openshift clusters tested with these failures are `epoc001.ete14.res.ibm.com` and `epoc002.ete14.res.ibm.com`, both of which are OCP 4.20.19 deployments.

pdettori:
## Summary

Fixes #1983 — Kagenti deployment on OCP 4.20 silently fails: the install script reports success but the backend never starts and `kagenti-ui-config` is never created.

## Root cause

`charts/kagenti/templates/ui-routes-job.yaml` hard-pinned the route-processor hook job image to `ose-cli:v4.15`. On OCP 4.20 (k8s 1.33) that client trails the API server by **five** minor versions — well past Kubernetes' supported ±1 skew — so `kubectl apply` of the ConfigMap fails client-side schema validation. The hook still `exit 0`s (and `hook-delete-policy: hook-succeeded` cleans it up), so Helm reports success while the backend has no config.

> Note: the v4.15 pin was introduced by merged PR #1714 ("Update ose-cli image from v4.17 to v4.15"). No fix for #1983 had been opened.

## Fix

```mermaid
flowchart LR
    A[".Capabilities.KubeVersion.Minor<br/>(live cluster: e.g. 33)"] --> B["regexFind [0-9]+<br/>(strip any '+' suffix)"]
    B --> C["OCP minor = k8s minor − 13"]
    C --> D["ose-cli:v4.20"]
    E["common.oseCLIImage<br/>(explicit override)"] -.overrides.-> D
```

1. **Version-match the image** to the target cluster via `.Capabilities.KubeVersion` (OpenShift minor = k8s minor − 13, stable since OCP 4.5). `regexFind` tolerates a `+` build suffix.
2. **Keep `common.oseCLIImage` as an override** (e.g. air-gapped mirror); default it to `""` so the computed tag applies.
3. **`--validate=false` on the ConfigMap apply** — the applied object is a fixed `core/v1` ConfigMap, so client-side validation adds no value and is precisely what breaks under skew. Belt-and-suspenders against any residual skew.

## Verification

- New `charts/kagenti/tests/ui_routes_job_test.yaml` (helm-unittest): k8s 1.33→`v4.20`, 1.28→`v4.15`, `"33+"`→`v4.20`, explicit override honored, `--validate=false` present.
- `helm unittest charts/kagenti` → 6/6 (incl. existing #1991 regression). `helm unittest charts/kagenti-deps` → 9/9. `helm lint` clean.
- Real renders confirmed: 1.33→v4.20, 1.31→v4.18, 1.28→v4.15.

## Follow-up (out of scope)

The hook's `exit 0` masks apply failures, which is why this surfaced as a silent success. Worth a separate issue to propagate the apply exit code.

Assisted-By: Claude Code

### DIFF
--- a/charts/kagenti/templates/ui-routes-job.yaml
+++ b/charts/kagenti/templates/ui-routes-job.yaml
@@ -200,16 +200,14 @@ spec:
       containers:
       - name: script-runner
         {{- /*
-          Pick an ose-cli tag that matches the target cluster so the client is
-          not skewed past the API server's ±1 minor support window (issue #1983).
-          OpenShift minor = Kubernetes minor - 13 (stable since OCP 4.5:
-          k8s 1.28 -> v4.15, k8s 1.33 -> v4.20). regexFind strips any "+" build
-          suffix the server may report (e.g. "33+"). An explicit
-          common.oseCLIImage value (e.g. an air-gapped mirror) always wins.
+          Pinned to an ose-cli tag known to exist on registry.redhat.io. Do NOT
+          derive it from the cluster version: ose-cli is not published for every
+          OCP release (v4.20 is absent -> "manifest unknown" -> ImagePullBackOff,
+          issue #1983). Version skew against a newer API server is harmless here
+          because the ConfigMap apply below uses --validate=false on a fixed
+          core/v1 object. An explicit common.oseCLIImage value always wins.
         */}}
-        {{- $k8sMinor := .Capabilities.KubeVersion.Minor | toString | regexFind "[0-9]+" | int }}
-        {{- $oseDefault := printf "registry.redhat.io/openshift4/ose-cli:v4.%d" (sub $k8sMinor 13) }}
-        image: {{ .Values.common.oseCLIImage | default $oseDefault }}
+        image: {{ .Values.common.oseCLIImage | default "registry.redhat.io/openshift4/ose-cli:v4.15" }}
         command: ["/bin/bash", "-c", "/scripts/process_routes.sh"]
         volumeMounts:
         - name: script-volume
--- a/charts/kagenti/templates/ui-routes-job.yaml
+++ b/charts/kagenti/templates/ui-routes-job.yaml
@@ -170,7 +170,11 @@ data:
       configmap_yaml+="  $key: \"$value\""
     done
 
-    echo "$configmap_yaml" | kubectl apply -f -
+    # --validate=false: the applied object is a fixed core/v1 ConfigMap, so
+    # client-side schema validation adds no value — and it fails outright when
+    # the ose-cli client trails the API server by several minor versions
+    # (issue #1983). Disabling it makes the apply robust to client/server skew.
+    echo "$configmap_yaml" | kubectl apply --validate=false -f -
     if [[ $? -eq 0 ]]; then
         echo "ConfigMap '$CONFIGMAP_NAME' applied successfully in namespace '$CONFIGMAP_NAMESPACE'."
     else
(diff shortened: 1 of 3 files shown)


charts/kagenti/templates/ui-routes-job.yaml
```diff
@@ -170,7 +170,11 @@ data:
       configmap_yaml+="  $key: \"$value\""
     done
 
-    echo "$configmap_yaml" | kubectl apply -f -
+    # --validate=false: the applied object is a fixed core/v1 ConfigMap, so
+    # client-side schema validation adds no value — and it fails outright when
+    # the ose-cli client trails the API server by several minor versions
+    # (issue #1983). Disabling it makes the apply robust to client/server skew.
+    echo "$configmap_yaml" | kubectl apply --validate=false -f -
```
clawgenti, line 177: **nit**: `--validate=false` is deprecated in recent `kubectl` versions (1.25+); the replacement is `--validate=ignore`. Both flags behave identically today, but `--validate=false` will eventually emit a deprecation warning. Worth switching to `--validate=ignore` to stay ahead of the noise — and the test regex at `charts/kagenti/tests/ui_routes_job_test.yaml:90` would need updating to match.

clawgenti (commented):
Fixes a real silent-failure bug on OCP 4.20 — the root cause analysis is solid and the two-pronged fix (dynamic image selection + `--validate=false`) is well-motivated and well-tested.

---

charts/kagenti/templates/ui-routes-job.yaml
```diff
@@ -195,7 +199,17 @@ spec:
...
-        image: {{ .Values.common.oseCLIImage | default "registry.redhat.io/openshift4/ose-cli:v4.15" }}
+        {{- /*
+          Pick an ose-cli tag that matches the target cluster so the client is
+          not skewed past the API server's ±1 minor support window (issue #1983).
+          OpenShift minor = Kubernetes minor - 13 (stable since OCP 4.5:
+          k8s 1.28 -> v4.15, k8s 1.33 -> v4.20). regexFind strips any "+" build
+          suffix the server may report (e.g. "33+"). An explicit
+          common.oseCLIImage value (e.g. an air-gapped mirror) always wins.
+        */}}
+        {{- $k8sMinor := .Capabilities.KubeVersion.Minor | toString | regexFind "[0-9]+" | int }}
```
esnible: nit: if `.Capabilities.KubeVersion.Minor` is ever empty or unparseable, `regexFind "[0-9]+"` returns `""`, `int ""` → `0`, and `sub 0 13` yields the degenerate tag `ose-cli:v4.-13`. In practice this only happens on a bare `helm template` without `--kube-version`, and `common.oseCLIImage` overrides it — so low risk. A `default`-guard or a one-line comment noting the live-cluster assumption would harden it.

esnible (approved):
Tightly-scoped, well-documented fix for #1983 (silent OCP 4.20 install failure). The root-cause analysis is correct: the hard-pinned `ose-cli:v4.15` client trails a k8s 1.33 API server by 5 minor versions, breaking client-side validation on `kubectl apply`, while the hook's `exit 0` masks the failure so Helm reports success.

The fix derives the image tag from `.Capabilities.KubeVersion` (OCP minor = k8s minor − 13, verified for 4.5/4.15/4.20), keeps `common.oseCLIImage` as an air-gap override, and adds `--validate=false` as a belt-and-suspenders defense. The template is correctly gated on `.Values.openshift`, so non-OCP clusters never compute a tag. The new helm-unittest suite covers the meaningful cases and **passes in CI** (`kagenti route-processor hook (OCP)` → 6/6). The masked-`exit 0` problem is honestly called out and appropriately deferred to a follow-up.

**Author**: pdettori (MEMBER — maintainer)
**Areas reviewed**: Helm templates, Helm unit tests, values.yaml
**CI**: passing (E2E HyperShift + Deploy&Test pending at review time)

One nit inline; no blocking issues.

Assisted-By: Claude Code

PR #2006, opened 2026-06-17
Feat: Bump OpenShell to v0.0.56-rc.3 for per-user provider ownership

### LINKED ISSUE #1995: feat: per-user provider ownership on shared OpenShell gateway
## Problem

Providers are currently gateway-wide shared resources that only admins can create. All sandboxes binding to a provider share the same credential (API key). This works for team-level LLM keys but breaks for credentials that are inherently per-user:

- Personal GitHub PATs
- Individual API keys with per-user billing/quota
- OAuth tokens scoped to a user's identity
- Any credential that should not be shared across team members

## Proposed solution: per-user provider ownership (Option A)

Apply the same `openshell.ai/owner` pattern already proven for sandboxes to providers.

### Design

1. **Relax `create_provider` auth** from `admin` → `user` role (or add a dedicated `create_user_provider` RPC).
2. **Stamp `openshell.ai/owner`** on provider create from verified OIDC identity — same mechanism as sandbox ownership.
3. **Filter provider list/get/delete by owner** — users see only their own providers + shared (admin-created) providers.
4. **Sandbox-provider binding validation** — a sandbox can reference:
   - Providers owned by the same user
   - Shared providers (no owner label — admin-created)
   - ❌ NOT another user's providers
5. **Provider name uniqueness scoped to `(owner, name)`** — alice and bob can both have a provider named `gh`.

### Shared vs. user-owned providers

| Provider type | Owner label | Visible to | Created by |
|---------------|-------------|------------|------------|
| Shared (team) | none | All users | Admin |
| User-owned | `openshell.ai/owner=<sub>` | Owner + admin | User |

### What does NOT change

- **Inference routes** remain admin-only, gateway-wide routing config. A route references a provider by name — when a sandbox uses a user-owned provider, the inference proxy resolves that user's credential automatically.
- **Policies** are already per-sandbox (inherits sandbox ownership). No change needed.
- **Admin bypass** continues to work — admins can see/manage all providers.
- **OIDC-disabled mode** — ownership skipped, backward-compatible (same as sandboxes).

### User experience

```bash
# Bob creates his own GitHub provider with his personal PAT
export GH_TOKEN="ghp_bob_xxx"
openshell provider create --name gh --type generic \
  --credential GH_TOKEN \
  --config BASE_URL=https://api.github.com

# Alice creates her own — no conflict, scoped by owner
export GH_TOKEN="ghp_alice_yyy"
openshell provider create --name gh --type generic \
  --credential GH_TOKEN \
  --config BASE_URL=https://api.github.com

# Bob's sandbox uses Bob's key
openshell sandbox create --provider gh -- bash

# Alice's sandbox uses Alice's key — no cross-contamination
openshell sandbox create --provider gh -- bash

# Admin can still create shared team providers
openshell provider create --name claude --type anthropic \
  --credential ANTHROPIC_AUTH_TOKEN --shared
```

### Implementation scope

| Component | Change |
|-----------|--------|
| Gateway `auth/ownership.rs` | Extend ownership to `Provider` object type |
| Gateway `grpc/provider.rs` | Add owner stamp on create, filter on list/get/delete |
| Gateway `grpc/sandbox.rs` | Validate provider binding (same-owner or shared) |
| Gateway store layer | Provider name uniqueness scoped to `(owner, name)` |
| CLI | No change needed (already sends credentials at create time) |
| Helm/deploy scripts | No change needed |

### Acceptance criteria

- On a shared gateway with OIDC enabled and two users (alice, bob):
  - alice creates provider `gh` with her PAT → provider carries `openshell.ai/owner=<alice-sub>`
  - bob creates provider `gh` with his PAT → no conflict, separate provider
  - bob `list providers` does not include alice's `gh`; alice's list does not include bob's
  - bob cannot bind alice's provider to his sandbox → error
  - both can use admin-created shared providers (e.g., `claude`)
  - admin can list/manage all providers across all users
- Backward compatible: existing shared providers (no owner) continue to work for all users
- e2e test covering cross-user provider isolation

### Related

- Parent: #1976 (per-user sandbox ownership — done)
- Gateway source: `kagenti/OpenShell` `crates/openshell-server/src/auth/ownership.rs`
- Sandbox ownership PR: kagenti/openshell-driver-openshift#14

pdettori:
## Summary

- Bump gateway + supervisor images to `v0.0.56-rc.3` (kagenti/OpenShell#20 — scoped provider name uniqueness)
- Bump compute driver to `v0.1.0-rc.7`
- Fix `deploy-tenant.sh` to also override `supervisorImage.tag` when `--image-tag` is passed

## What's in v0.0.56-rc.3

Per-user provider ownership via scoped name uniqueness (Option B — name prefixing). Two users can independently create providers with the same visible name (`openai`, `gh`, etc.) without conflict. Resolution order: user-owned → shared fallback. Admin bypass preserved. Backward compatible with existing shared providers.

See kagenti/OpenShell#20 for full details.

Closes #1995

## Test plan

- [ ] Deploy to Kind cluster with `deploy-tenant.sh team1`
- [ ] Verify gateway starts with new image
- [ ] Create provider as user A, create same-named provider as user B — verify isolation
- [ ] Verify shared (admin-created) providers still accessible to both users

Assisted-By: Claude Code

### DIFF
--- a/charts/kagenti/values.yaml
+++ b/charts/kagenti/values.yaml
@@ -395,11 +395,11 @@ openshell:
   enabled: false
   gateway:
     image: ghcr.io/kagenti/openshell/gateway
-    tag: v0.0.56
+    tag: v0.0.56-rc.3
     pullPolicy: IfNotPresent
   computeDriver:
     image: ghcr.io/kagenti/openshell-driver-openshift/compute-driver
-    tag: v0.1.0-rc.4
+    tag: v0.1.0-rc.7
     pullPolicy: IfNotPresent
   credentialsDriver:
     image: ghcr.io/kagenti/openshell-credentials-keycloak/credentials-driver
--- a/charts/kagenti/values.yaml
+++ b/charts/kagenti/values.yaml
@@ -387,21 +387,3 @@ mlflowOAuthSecret:
 # ------------------------------------------------------------------
 spire:
   enabled: true
-
-# ------------------------------------------------------------------
-#  OpenShell Configuration
-# ------------------------------------------------------------------
-openshell:
-  enabled: false
-  gateway:
-    image: ghcr.io/kagenti/openshell/gateway
-    tag: v0.0.56-rc.3
-    pullPolicy: IfNotPresent
-  computeDriver:
-    image: ghcr.io/kagenti/openshell-driver-openshift/compute-driver
-    tag: v0.1.0-rc.7
-    pullPolicy: IfNotPresent
-  credentialsDriver:
-    image: ghcr.io/kagenti/openshell-credentials-keycloak/credentials-driver
-    tag: main-d7d8306
-    pullPolicy: IfNotPresent
--- a/charts/openshell/values.yaml
+++ b/charts/openshell/values.yaml
@@ -5,11 +5,11 @@ tenant: ""
 images:
   gateway:
     repository: ghcr.io/kagenti/openshell/gateway
-    tag: v0.0.56-rc.2
+    tag: v0.0.56-rc.3
     pullPolicy: IfNotPresent
   computeDriver:
     repository: ghcr.io/kagenti/openshell-driver-openshift/compute-driver
-    tag: v0.1.0-rc.6
+    tag: v0.1.0-rc.7
     pullPolicy: IfNotPresent
   credentialsDriver:
     repository: ghcr.io/kagenti/openshell-credentials-keycloak/credentials-driver
@@ -31,7 +31,7 @@ keycloakClusterIP: ""
 # Built from kagenti/OpenShell@mvp-v2 as a multi-arch image (linux/amd64 + linux/arm64).
 supervisorImage:
   repository: ghcr.io/kagenti/openshell/supervisor
-  tag: v0.0.56-rc.2
+  tag: v0.0.56-rc.3
   binaryPath: /openshell-sandbox
 
 # -- Image pull policy for sandbox pods (init + agent containers).
--- a/docs/sandbox-guide.md
+++ b/docs/sandbox-guide.md
@@ -277,12 +277,79 @@ Keycloak.
 
 ### Requirements
 
-- Gateway image `v0.0.56-rc.2` or later
+- Gateway image `v0.0.56-rc.2` or later (provider ownership requires `v0.0.56-rc.3`+)
 - OIDC enabled on the gateway (`oidc.enabled: true` in Helm values — the default)
 - Keycloak `sub` claim must be a UUID (the default for Keycloak realms)
 
 When OIDC is disabled (local development without authentication), ownership is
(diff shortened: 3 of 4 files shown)


charts/kagenti/values.yaml
```diff
@@ -395,11 +395,11 @@ openshell:
   enabled: false
   gateway:
     image: ghcr.io/kagenti/openshell/gateway
-    tag: v0.0.56
+    tag: v0.0.56-rc.3
```
huang195: suggestion (non-blocking): this `openshell:` block pins gateway / computeDriver / credentialsDriver but has no `supervisorImage` key. The standalone `charts/openshell` chart bumps the supervisor to `v0.0.56-rc.3` in this PR — should the kagenti chart path also pin it so both deploy routes run the same supervisor build, or is the supervisor image injected separately (e.g. by the operator/compute driver) on this path? Pre-existing, just flagging for parity.

huang195 (approved):
A clean, well-scoped image bump to OpenShell `v0.0.56-rc.3` (per-user provider ownership) plus compute driver `v0.1.0-rc.7`. The `deploy-tenant.sh` fix correctly closes a gap where `--image-tag` left the supervisor image stale, and the PR also realigns the two values files that had drifted apart (the kagenti chart was on `v0.0.56`/`rc.4`, the openshell chart on `rc.2`/`rc.6` — both now at `rc.3`/`rc.7`). Docs are accurate and clear.

**Areas reviewed**: Helm/values, Shell, Docs
**CI**: passing (E2E HyperShift pending maintainer trigger)

LGTM. One non-blocking suggestion inline.

Assisted-By: Claude Code

charts/kagenti/values.yaml
pdettori: Good catch, and thanks for flagging it. Digging in, this `openshell:` block turns out not to be a deploy route at all, which is why it has no `supervisorImage` key:

- No `charts/kagenti/templates/**` references `.Values.openshell` (zero matches), `openshell` is not a subchart dependency, and the operator subchart doesn't read it.
- `helm template charts/kagenti --set openshell.enabled=true` renders **zero** manifests from it.
- Even `scripts/openshell/build-images.sh` reads tags from `charts/openshell/values.yaml`, not this block.

The supervisor (and gateway/drivers) are deployed solely via the standalone `charts/openshell` chart through `deploy-shared.sh` / `deploy-tenant.sh`, which is self-contained and already pinned to `v0.0.56-rc.3` (including `supervisorImage`). So there's no parity gap — there's only one real deploy route, and it's pinned.

This block was added in `97b9c982` as informational image refs alongside the local build script, but it's never rendered and had already drifted (it was on `v0.0.56`/`rc.4` while the standalone chart was on `rc.2`/`rc.6`). Rather than add a `supervisorImage` key to dead config, I've removed the block entirely in c96fd490. `helm lint` + `helm template` both pass clean.

PR #2009, opened 2026-06-19
Feat: Per-agent TLS bridge toggle (UI/backend)

huang195:
## Summary

Surfaces the AuthBridge **TLS bridge** as a per-agent option so an agent's outbound HTTPS can be decrypted and inspected by the AuthBridge pipeline. Modeled exactly on the existing `mtlsMode` / `authBridgeMode` controls: a CRD-backed field plus an always-visible toggle in the import form — **no operator feature gate and no UI feature flag**. Off by default via `spec.tlsBridgeMode=disabled`.

Companion to **kagenti-operator #448** (CRD field + reconciler + sidecar mount) and **kagenti-extensions #522/#523** (the bridge engine).

## What's included

**Backend**
- `tlsBridgeEnabled` on `CreateAgentRequest` and `FinalizeShipwrightBuildRequest`, with a validator that rejects it under `envoy-sidecar`/`waypoint` (mirrors the operator's validating webhook — fast 422 instead of a webhook denial).
- `_build_agentruntime_manifest()` writes `spec.tlsBridgeMode=enabled` only when opted in; threaded through both the direct-deploy and build-from-source (store → finalize) paths.

**UI**
- A **"Decrypt outbound TLS for inspection (TLS bridge)"** checkbox in the agent import form, next to the mTLS control — shown whenever AuthBridge is enabled, disabled + reset under Envoy mode (the bridge lives in the Go forward proxy, so it needs proxy-sidecar/lite).
- **"TLS bridge: Active"** on the agent's AuthBridge detail tab when the sidecar reports the bridge is on.
- Also corrected the Envoy toggle copy — "Use Envoy sidecar interception" (proxy-sidecar uses iptables too, so the old "Envoy + iptables" label was misleading).

**Docs**: `docs/authbridge/tls-bridge.md`.

## Design note (no feature flag — deliberate)

`mtlsMode` and `authBridgeMode` are plain per-agent AuthBridge options with no operator feature gate and no `kagenti_feature_flag_*`. The TLS bridge follows the same pattern for consistency. This is a conscious departure from the "new UI features behind a feature flag" guidance, matching the established AuthBridge-options precedent. The per-agent field defaulting `disabled` keeps the feature off by default.

## Requirements / release ordering

The toggle only functions once the platform ships an operator build that supports the TLS bridge:
- The `spec.tlsBridgeMode` CRD field comes from kagenti-operator #448 — so this should land together with bumping the operator chart/image pins to a release containing #448.
- **cert-manager** must be installed (already true on Kind + OpenShift kagenti installs).

## Test Plan

- [x] Backend: `pytest tests/test_agentruntime_manifest.py` (incl. TLS-bridge manifest + validator cases)
- [x] UI: `tsc --noEmit` clean
- [x] Chart: `helm lint` clean
- [ ] e2e on a cluster running the #448 operator: import an agent with the toggle on, confirm `spec.tlsBridgeMode=enabled`, the per-agent CA is provisioned, and egress is decrypted (verified manually against local #448/#523 builds)

### DIFF
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -3950,6 +3950,22 @@ class FinalizeShipwrightBuildRequest(BaseModel):
         """
         return self
 
+    @model_validator(mode="after")
+    def _check_tlsbridge_compatible_with_mode(self) -> "FinalizeShipwrightBuildRequest":
+        """Mirror of CreateAgentRequest._check_tlsbridge_compatible_with_mode at
+        the Shipwright finalize boundary, so a direct finalize caller (or a combo
+        inherited from the BuildRun's stored config) with tlsBridgeEnabled +
+        envoy-sidecar/waypoint gets the same fast 422 instead of a later webhook
+        denial. Same allowlist as the operator (empty → defaults to proxy-sidecar).
+        """
+        allowed = (None, "", "proxy-sidecar", "lite")
+        if self.tlsBridgeEnabled and self.authBridgeMode not in allowed:
+            raise ValueError(
+                "tlsBridgeEnabled requires authBridgeMode proxy-sidecar or lite "
+                f"(the TLS bridge lives in the Go forward proxy); got {self.authBridgeMode!r}"
+            )
+        return self
+
 
 @router.post(
     "/{namespace}/{name}/finalize-shipwright-build",
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -289,6 +289,16 @@ class CreateAgentRequest(BaseModel):
     # webhook denial after the manifest is built.
     mtlsMode: Optional[Literal["disabled", "permissive", "strict"]] = None
 
+    # Per-workload TLS bridge: decrypt the agent's outbound HTTPS so AuthBridge's
+    # pipeline can inspect it. Maps to AgentRuntime.Spec.TLSBridgeMode (enabled
+    # when True; left unset → operator default "disabled"). Like mtlsMode, it's a
+    # plain per-agent field — no operator feature gate and no UI feature flag; the
+    # import-form checkbox shows whenever AuthBridge is enabled. The bridge lives
+    # in the Go forward proxy, so it requires proxy-sidecar/lite mode — the
+    # validator below mirrors the operator webhook's reject of envoy-sidecar. It
+    # also needs cert-manager and an operator build that supports the bridge.
+    tlsBridgeEnabled: bool = False
+
     # Port exclusion annotations
     outboundPortsExclude: Optional[str] = None
     inboundPortsExclude: Optional[str] = None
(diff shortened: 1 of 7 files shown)


kagenti/backend/app/routers/agents.py
```diff
@@ -3884,6 +3927,9 @@ class FinalizeShipwrightBuildRequest(BaseModel):
     # the user picked at form-submit time (stashed on the BuildRun via
     # kagenti.io/agent-config annotation).
     mtlsMode: Optional[Literal["disabled", "permissive", "strict"]] = None
+    # Mirrors CreateAgentRequest.tlsBridgeEnabled. None → inherit the value
+    # stashed on the BuildRun annotation at form-submit time.
+    tlsBridgeEnabled: Optional[bool] = None
```
mrsabath, line 3932: `tlsBridgeEnabled` is added here, but the `_check_tlsbridge_compatible_with_mode` validator added to `CreateAgentRequest` is **not** mirrored onto `FinalizeShipwrightBuildRequest`. A direct caller to `finalize-shipwright-build` with `authBridgeMode=envoy-sidecar` + `tlsBridgeEnabled=True` (or that combo inherited from `stored_config`) bypasses the fast-422 and falls through to the operator webhook denial — exactly the case the validator exists to pre-empt.

Not a security/correctness hole: the merged #448 validating webhook still rejects `tlsBridgeMode=enabled` + `envoy-sidecar`, so nothing invalid reaches the cluster. But it loses the "fast 422 instead of a webhook denial" benefit on this path, and it's inconsistent with the `mtls` pattern this PR mirrors — `_check_mtls_compatible_with_mode` is defined on *both* request classes (even though it's currently a no-op) precisely so the rejection has one home per boundary. Suggest adding the parallel validator to `FinalizeShipwrightBuildRequest` for symmetry and defense-in-depth.

mrsabath (approved):
Reviewed backend (Python), UI-v2 (TypeScript), and docs. Code is correct and the operator-webhook parity claim checks out against the merged companion **kagenti-operator#448** — `_check_tlsbridge_compatible_with_mode`'s allowlist (`proxy-sidecar`/`lite`, empty→default) genuinely mirrors `checkTLSBridgeCompatibleWithMode` in `agentruntime_webhook.go`. CI is green, commit is signed-off, and the new manifest/validator tests are solid. No blocking issues — **APPROVE**. Two non-blocking notes:

**1. Design note: "no operator feature gate" is inaccurate.** The merged #448 implements the bridge **behind an off-by-default `TLSBridge` feature gate** (`internal/webhook/config/feature_gates.go`; the per-agent CA reconciler no-ops when `!FeatureGates.TLSBridge`, and `applyTLSBridgeMounts` is gated on `featureGates.TLSBridge && tlsBridgeMode==enabled && mode ∈ {proxy-sidecar,lite}`). The UI/backend additions here genuinely add no flag — that part is true — but the operator side *does* gate the data plane. As written, an operator reading this PR will expect the toggle to work once #448 ships, when in fact the cluster-level `TLSBridge` gate must also be enabled. Suggest rewording to something like: *"no UI/backend feature flag; the operator gates the data plane behind an off-by-default `TLSBridge` feature gate (#448)."* This also reinforces the release-ordering caveat already in the PR body.

**2. Validator only on `CreateAgentRequest`, not `FinalizeShipwrightBuildRequest`** — see inline comment.

Release ordering in the PR body is correct and worth honoring: land alongside an operator chart/image bump to a build containing #448, with the `TLSBridge` gate enabled wherever the toggle is meant to function.

huang195:
Thanks for the review! Both notes addressed:

**2. Validator on `FinalizeShipwrightBuildRequest`** — fixed in `bd9a6866`. Added the parallel `_check_tlsbridge_compatible_with_mode` (same allowlist) so a direct `finalize-shipwright-build` caller, or a combo inherited from the BuildRun's stored config, gets the same fast-422 instead of falling through to the webhook denial — matching how `_check_mtls_compatible_with_mode` lives on both request classes. Added tests for both the reject and allow paths.

**1. "no operator feature gate"** — this is actually accurate against the *merged* #448. The `TLSBridge` feature gate was **removed** before #448 merged (merge commit `b336ed5`): the operator now keys the bridge purely on the per-agent `spec.tlsBridgeMode`, consistent with how `mtlsMode`/`authBridgeMode` have no gate. You can confirm on `kagenti-operator` `main`: `feature_gates.go` has no `TLSBridge` field, and the mutator is `tlsBridgeOn := tlsBridgeMode == enabled` (no gate). So there's no cluster-level gate to enable — the per-agent toggle is the whole control. (Your review likely predates that gate-removal commit.) I've left the wording as-is for that reason; happy to expand it if you'd prefer it call out the gate's removal explicitly.

Release ordering note in the PR body still stands: land alongside an operator chart/image bump to a build containing the bridge.

PR #2015, opened 2026-06-19
Fix: drop orphaned Istio validation webhooks before kagenti-deps install on OpenShift

### LINKED ISSUE #2014: OCP install fails: stale Istio validation webhook blocks kagenti-deps (service "istiod" not found)
## Summary

On OpenShift, `scripts/ocp/setup-kagenti.sh` (Step 3) aborts during `helm install kagenti-deps` when the cluster carries a **stale Istio validation webhook** from a previous install/teardown cycle:

```
Error: INSTALLATION FAILED: 1 error occurred:
    * Internal error occurred: failed calling webhook "validation.istio.io":
      failed to call webhook: Post "https://istiod.istio-system.svc:443/validate?timeout=10s":
      service "istiod" not found
```

## Root cause

The chart applies Istio resources in two groups on OpenShift:

| Resource | apiGroup | In chart as | Applied |
|---|---|---|---|
| `Istio` / `IstioCNI` / `ZTunnel` (create istiod) | `sailoperator.io` | **post-install hook** (weight 1); also deferred by `--no-hooks` to `_apply_operand_crs` | **after** `helm install` |
| `AuthorizationPolicy otel-collector-from-mcp-system` | `security.istio.io/v1` | **non-hook** template (`charts/kagenti-deps/templates/otel-collector.yaml`) | **during** `helm install` |

So `helm install kagenti-deps --no-hooks` creates the `AuthorizationPolicy` **before** istiod exists. When a prior run deleted the `istio-system` namespace but left the **cluster-scoped** `istiod-default-validator` ValidatingWebhookConfiguration behind (`failurePolicy=Fail`, validates `security.istio.io`), the webhook call fails closed — `istio-system/istiod` Service is gone — and the entire release is aborted before `_apply_operand_crs` can bring istiod up.

### Evidence (cluster where this reproduced)
- `istiod-default-validator` → `validation.istio.io`, `failurePolicy=Fail`, groups include `security.istio.io`.
- `kagenti-deps` Helm release stuck at revision 1, status `failed`.
- AuthorizationPolicy `otel-collector-from-mcp-system` **absent** in `kagenti-system` (rejected by the webhook).
- `otel-collector`/`postgres-otel` pods **running** — Helm applies Deployments before custom resources, so they landed; the later AuthorizationPolicy then hit the dead webhook.
- On OCP the other non-hook Istio resources are excluded: Telemetry & ingress Gateway are gated `not .Values.openshift`; the mlflow Gateway/AuthorizationPolicy need `mlflow.auth.enabled`, which the installer forces to `false`. The otel-collector AuthorizationPolicy is the sole istio-validated resource hitting the webhook during install.

The existing `.Capabilities.APIVersions.Has "security.istio.io/v1"` guard (PR #1589) only checks **CRD presence**, not **istiod liveness**, so it does not cover this case.

## Fix

Drop orphaned, backend-less Istio validation webhooks in the installer preflight (when no `istio-system/istiod` Service exists) before `helm install kagenti-deps`. istiod re-registers these webhooks (with the correct caBundle) once the Sail operand CRs are applied in Step 3. This is Istio's canonical remediation for "service istiod not found" and unblocks **all** Istio config resources, not just this one AuthorizationPolicy. Only kagenti-managed webhooks are touched (`istiod-default-validator`, `istio-validator-istio-system`) — never the OpenShift ingress gateway's.

A chart-only reorder (making the AuthorizationPolicy a hook) does **not** fix it: the installer uses `--no-hooks` and applies all operand hooks in one batch via `_apply_operand_crs` *before* waiting for istiod readiness, so a hooked AuthorizationPolicy would still be applied before istiod is Ready.

## Repro
1. OpenShift cluster where kagenti-deps was previously installed and `istio-system` later torn down (cluster-scoped Istio webhooks survive).
2. `scripts/ocp/setup-kagenti.sh --kagenti-repo "$(pwd)" --with-all`
3. Step 3 fails with the webhook error above.

Assisted-By: Claude Code

pdettori:
## Problem

On OpenShift, `scripts/ocp/setup-kagenti.sh` (Step 3) aborts during `helm install kagenti-deps` when a stale Istio validation webhook survives a prior `istio-system` teardown:

```
failed calling webhook "validation.istio.io":
  Post "https://istiod.istio-system.svc:443/validate?timeout=10s": service "istiod" not found
```

Deleting the `istio-system` namespace does **not** remove the cluster-scoped `istiod-default-validator` ValidatingWebhookConfiguration. With `failurePolicy=Fail` and no `istiod` Service to call, it rejects the otel-collector `AuthorizationPolicy` (`security.istio.io/v1`) that the chart creates during `helm install` — before the Sail operand CRs (a deferred step) bring istiod up. The release is left at revision 1 / `failed`.

The pre-existing `.Capabilities.APIVersions.Has "security.istio.io/v1"` guard (#1589) only checks **CRD presence**, not **istiod liveness**, so it doesn't cover this.

## Fix

Add a preflight in `_helm_kagenti_deps` that deletes the stale **kagenti-managed** Istio webhooks (`istiod-default-validator`, `istio-validator-istio-system`) **only when no `istio-system/istiod` Service exists**. istiod re-registers them with the correct caBundle once the operand CRs are applied in Step 3. The OpenShift ingress gateway's webhook (`istio-validator-openshift-gateway-openshift-ingress`) is never touched, and a healthy istiod is left alone.

### Why not a chart-only reorder?
Making the AuthorizationPolicy a Helm hook does not help: the installer runs `helm install --no-hooks` and `_apply_operand_crs` applies all operand hooks in a single batch *before* waiting for istiod readiness — so a hooked AuthorizationPolicy would still be applied before istiod is Ready. The orphaned-webhook cleanup fixes the actual trigger and unblocks all Istio config resources.

## Testing
- `bash -n scripts/ocp/setup-kagenti.sh` passes; pre-commit (`make lint`) passes.
- No-op on healthy clusters (early-returns when the `istiod` Service is present) and under `--dry-run`.
- Verified against the cluster that reproduced the failure (RCA in #2014).

Fixes #2014

Assisted-By: Claude Code

### DIFF
--- a/scripts/ocp/setup-kagenti.sh
+++ b/scripts/ocp/setup-kagenti.sh
@@ -669,12 +669,41 @@ for doc in docs:
 " | $KUBECTL apply -f - || true
 }
 
+# Remove orphaned Istio validation webhooks left behind by a previous teardown.
+# Deleting the istio-system namespace does NOT remove the cluster-scoped
+# ValidatingWebhookConfigurations that back "validation.istio.io". With
+# failurePolicy=Fail and no istiod Service to call, they reject every Istio
+# config resource (e.g. the otel-collector AuthorizationPolicy) the kagenti-deps
+# chart creates during `helm install`, aborting the release with:
+#   failed calling webhook "validation.istio.io": service "istiod" not found
+# These webhooks are recreated (with the correct caBundle) once the Sail operand
+# CRs in Step 3 bring istiod back up, so it is safe to drop the stale ones here.
+# Only touch the kagenti-managed webhooks — never the OpenShift ingress gateway's
+# (istio-validator-openshift-gateway-openshift-ingress).
+_clean_stale_istio_webhooks() {
+  if $DRY_RUN; then return; fi
+  # If istiod is present, the webhooks have a live backend — leave them alone.
+  if $KUBECTL get svc istiod -n istio-system &>/dev/null; then
+    return
+  fi
+  for _whc in istiod-default-validator istio-validator-istio-system; do
+    if $KUBECTL get validatingwebhookconfiguration "$_whc" &>/dev/null; then
+      log_warn "Removing stale Istio webhook $_whc (no istiod Service to back it)"
+      $KUBECTL delete validatingwebhookconfiguration "$_whc" --ignore-not-found || true
+    fi
+  done
+}
+
 _helm_kagenti_deps() {
   # Pre-flight: ensure namespaces managed by this chart are not stuck terminating
   # from a previous failed install/uninstall cycle
   for _ns in keycloak istio-cni istio-system istio-ztunnel; do
     _wait_ns_gone "$_ns"
   done
+
+  # Pre-flight: drop orphaned Istio validation webhooks that would otherwise
+  # block the chart's Istio config resources before istiod is provisioned.
+  _clean_stale_istio_webhooks
 
   # Build MLflow OTEL flags: enable the pipeline and point it at the DSC-managed endpoint.
   # When --otel-operator-managed is set, the operator handles ConfigMap assembly


pdettori:
/run-e2e

Re-triggering E2E. The previous run's only failure was `test_no_failed_pods` on the Shipwright build pod `team1/weather-tool-2-build` (Failed) — a transient tool-build flake unrelated to this PR. The OpenShift install path (including the new stale-webhook preflight) completed cleanly: kagenti-deps installed + reconciled and istiod came up with no webhook errors.

kevincogan (approved):
**LGTM, approving.**

Reviewed the diff against upstream `main`. The new `_clean_stale_istio_webhooks()` preflight is well-scoped and defensive:

- No-op on healthy clusters (early-returns when istiod Service is present)
- No-op on fresh installs (nothing to clean if no prior Istio existed)
- Respects `--dry-run` and only targets the two kagenti-managed webhooks, never the OpenShift ingress gateway's
- Error-tolerant with `--ignore-not-found` / `|| true`
- Correct ordering: runs after `_wait_ns_gone` and before `helm install`

E2E passed. Merge status clean. Nice fix!

PR #2033, opened 2026-06-23
Fix: Create cluster-monitoring-config when absent to enable UWM

### LINKED ISSUE #2032: OpenShift installer can leave User Workload Monitoring disabled, so Kiali shows no traffic
## Summary

On OpenShift, after a successful install, **Kiali shows no traffic** even when agents are actively serving requests (e.g. the weather agent example). The mesh dataplane is healthy and traffic flows — but Kiali's graph is empty.

## Root cause

Kiali (`kiali-user-workload-monitoring`) reads mesh metrics from `thanos-querier.openshift-monitoring`. Istio/ztunnel telemetry is exposed via ServiceMonitors in **user namespaces** (e.g. `ztunnel-servicemonitor` in `istio-ztunnel`), which are scraped **only** by OpenShift **User Workload Monitoring (UWM)**. If UWM is not enabled, no `istio_*` metrics ever reach Thanos, so Kiali has nothing to graph.

The installer is supposed to enable UWM, but it can silently leave it **disabled** on a fresh cluster:

- `charts/kagenti-deps/templates/kiali-operand.yaml` ships a `post-install,post-upgrade` Helm **hook** that creates `cluster-monitoring-config` with `enableUserWorkload: true`.
- `scripts/ocp/setup-kagenti.sh` `_ensure_user_workload_monitoring()` pre-flight **returns early when the ConfigMap is absent** (`scripts/ocp/setup-kagenti.sh:516-519`), trusting that hook to create it.
- But the hook does **not** always run:
  - **Upgrades** deliberately skip hooks (`--no-hooks`).
  - The fresh-install **failure-recovery path** re-runs with `--no-hooks` and explicitly **skips** the `cluster-monitoring-config` ConfigMap (it's the "conflict source" on managed clusters).

So whenever the chart hook is skipped and the ConfigMap didn't already exist, UWM is never enabled — and the only symptom is an empty Kiali graph discovered much later.

## Evidence (observed on an OpenShift cluster)

- `ztunnel` 6/6, `istio-cni` 6/6, `istiod` 1/1 Running — mesh healthy.
- Weather example ran: `team1/weather-service` 3/3, `weather-tool` 1/1, run pods `Completed`.
- Kiali pod healthy, but `cluster-monitoring-config` ConfigMap **NotFound** and **zero** `prometheus-user-workload` pods in `openshift-user-workload-monitoring` → UWM off.
- Only Istio ServiceMonitor in cluster: `ztunnel-servicemonitor` (`istio-ztunnel`) — inert without UWM.

Manually creating `cluster-monitoring-config` with `enableUserWorkload: true` brings up the UWM Prometheus stack and Kiali begins receiving telemetry.

## Proposed fix

Make the `_ensure_user_workload_monitoring()` pre-flight **authoritative**: when the ConfigMap is absent, **create** it with `enableUserWorkload: true` instead of returning early and relying on the (often-skipped) chart hook.

## Secondary note (out of scope for this fix)

Team namespaces are labeled `istio.io/use-waypoint: waypoint` but no `waypoint` Gateway is deployed, so only L4/TCP metrics are available. The full L7 HTTP request graph requires a waypoint. Tracking separately.

pdettori:
## What

Make the OpenShift installer's `_ensure_user_workload_monitoring()` pre-flight **create** `cluster-monitoring-config` (with `enableUserWorkload: true`) when it is absent, instead of returning early and relying on the kiali-operand Helm hook.

## Why

Kiali reads mesh metrics from `thanos-querier.openshift-monitoring`. Istio/ztunnel ServiceMonitors live in **user namespaces** and are scraped **only** by OpenShift User Workload Monitoring (UWM). If UWM is off, no `istio_*` metrics reach Thanos and **Kiali shows no traffic** — even when agents are actively serving.

The pre-flight returned early when the ConfigMap was absent (`scripts/ocp/setup-kagenti.sh:521-524`), trusting the chart hook to create it. But that hook is skipped:

- on **upgrades** (`--no-hooks`), and
- on the fresh-install **failure-recovery path**, which explicitly skips this ConfigMap as the "conflict source" on managed clusters.

When the hook doesn't run and the ConfigMap didn't already exist, UWM is silently left disabled. This was observed on a live OpenShift cluster: `cluster-monitoring-config` NotFound, zero `prometheus-user-workload` pods, empty Kiali graph despite a healthy mesh and a completed weather-agent run. Manually creating the ConfigMap brought up the UWM stack and Kiali began receiving telemetry.

## Change

When the ConfigMap is absent, the pre-flight now applies it with content matching the chart hook (`enableUserWorkload: true` + `prometheusK8s.retentionSize: 10GB`). The existing "merge into existing config" path is unchanged.

## Testing

- `bash -n scripts/ocp/setup-kagenti.sh` passes.
- Verified the manual equivalent on a live OpenShift cluster: applying the ConfigMap brought up `prometheus-user-workload` (2/2 replicas, 6/6) and the UWM stack.

## Notes

Secondary finding (out of scope, tracked separately): team namespaces are labeled `istio.io/use-waypoint: waypoint` but no waypoint Gateway is deployed, so only L4/TCP metrics are available; the full L7 HTTP request graph needs a waypoint.

Fixes #2032

Assisted-By: Claude Code

### DIFF
--- a/scripts/ocp/setup-kagenti.sh
+++ b/scripts/ocp/setup-kagenti.sh
@@ -519,7 +519,25 @@ _ensure_user_workload_monitoring() {
   existing=$($KUBECTL get configmap cluster-monitoring-config -n openshift-monitoring \
     -o jsonpath='{.data.config\.yaml}' 2>/dev/null || echo "")
   if [ -z "$existing" ]; then
-    # ConfigMap doesn't exist or is empty — the hook can create it from scratch
+    # ConfigMap is absent (or has no config.yaml). We must NOT rely on the
+    # kiali-operand chart hook to create it: that hook is skipped on upgrades
+    # (--no-hooks) and on the fresh-install failure-recovery path (which also
+    # explicitly skips this ConfigMap as the conflict source). Relying on it
+    # silently leaves UWM disabled, so Kiali shows no mesh traffic. Create it
+    # here so the pre-flight is authoritative.
+    $KUBECTL apply -f - >/dev/null <<'EOF'
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cluster-monitoring-config
+  namespace: openshift-monitoring
+data:
+  config.yaml: |
+    enableUserWorkload: true
+    prometheusK8s:
+      retentionSize: 10GB
+EOF
+    log_success "Created cluster-monitoring-config with enableUserWorkload: true"
     return
   fi
   if echo "$existing" | grep -q "enableUserWorkload: true"; then


scripts/ocp/setup-kagenti.sh
```diff
@@ -519,7 +519,25 @@ _ensure_user_workload_monitoring() {
...
+    # ConfigMap is absent (or has no config.yaml). We must NOT rely on the
+    # kiali-operand chart hook to create it: that hook is skipped on upgrades
+    # (--no-hooks) and on the fresh-install failure-recovery path (which also
+    # explicitly skips this ConfigMap as the conflict source). Relying on it
+    # silently leaves UWM disabled, so Kiali shows no mesh traffic. Create it
+    # here so the pre-flight is authoritative.
+    $KUBECTL apply -f - >/dev/null <<'EOF'
+apiVersion: v1
+kind: ConfigMap
+metadata:
```
cwiklik, line 531: Suggestion (pre-existing, out of scope — flagging since this PR's goal is to make the pre-flight authoritative): the merge path prepends `enableUserWorkload: true` to `$existing`. If an existing config explicitly carries `enableUserWorkload: false`, the result is a `config.yaml` with two `enableUserWorkload` keys — duplicate-key resolution then decides the effective value, which may leave UWM disabled. The line-525 check only short-circuits the `true` case, so an explicit `false` falls through to here. Not introduced by this PR; worth a follow-up if you want the pre-flight to fully own the setting (e.g. strip any existing `enableUserWorkload:` line before prepending, or use a YAML-aware merge).

cwiklik (approved):
Verified all PR claims against `upstream/main`:

- **Content parity**: the applied ConfigMap (`enableUserWorkload: true` + `prometheusK8s.retentionSize: 10GB`) matches the kiali-operand chart hook (`charts/kagenti-deps/templates/kiali-operand.yaml`) exactly.
- **Merge path unchanged**: the existing-config branch is untouched.
- **Recovery-path/`--no-hooks` rationale**: confirmed at script L540 — the Helm hook's delete+recreate conflicts on managed clusters, so relying on it is unsafe. Creating the CM here makes the pre-flight authoritative.
- `kubectl apply` is idempotent and (client-side merge) won't drop other `data` keys.

One non-blocking, pre-existing edge case noted inline on the merge path.

**Areas reviewed**: Shell. **Commits**: 1, signed-off: yes. **CI**: all passing (Deploy & Test 22m + E2E HyperShift green).

mrsabath (approved):
Verified every concrete claim against `main`, and they all hold:

- **Function + line numbers** — the early-return-when-absent branch and the cited `:519`-region structure are accurate; the **"merge into existing config" path is genuinely unchanged** (only the empty branch is touched).
- **Content parity** — the applied here-doc is **byte-for-byte identical** to the kiali-operand chart hook (`charts/kagenti-deps/templates/kiali-operand.yaml`): `enableUserWorkload: true` + `prometheusK8s.retentionSize: 10GB`. So a fresh-install via pre-flight and a hook-driven install converge on the same ConfigMap.
- **Root cause** — the hook is `post-install,post-upgrade` with `hook-delete-policy: before-hook-creation`, which corroborates exactly why it's skipped on `--no-hooks` upgrades and why the recovery path treats it as the delete/recreate conflict source. Making the pre-flight authoritative is the right call.
- **Shell hygiene** — `<<'EOF'` (no expansion needed, static YAML) + `$KUBECTL apply -f -` matches the pattern already used elsewhere in this script; `set -euo pipefail` and `log_success` are in place. Leaving the new `apply` on the strict-mode path (no `|| true`) is correct for an authoritative pre-flight — a failed UWM setup *should* halt.

Full green CI including **E2E HyperShift**, and the live-cluster evidence in the description (`prometheus-user-workload` 2/2 after applying) seals it. The out-of-scope waypoint/L7-metrics note is correctly tracked separately. Ottimo lavoro, Paolo — diagnosi precisa e fix minimale.

**APPROVE.**

PR #2073, opened 2026-06-23
Fix: enforce helm >= 3.18.0 < 4 in OCP setup pre-flight

### LINKED ISSUE #2072: installer(ocp): enforce Helm < 4 — Helm 4 CRD handling breaks setup-kagenti.sh
## Summary

`scripts/ocp/setup-kagenti.sh` documents a Helm requirement of **`>= 3.18.0 < 4`** (script header line ~33; also `docs/install.md`), but the pre-flight only checks that `helm` *exists* (lines ~198/201) — it does **not** enforce the `< 4` upper bound. Running the installer with **Helm 4** therefore proceeds and then fails with cryptic CRD errors, because Helm 4 changed how CRDs in `crds/` are handled.

## Environment where observed
- OCP **4.20.8** (ROSA)
- Helm **v4.0.5** (shipped by Rancher Desktop, `~/.rd/bin/helm`)
- kagenti **v0.6.1-rc.4**, installed via `./scripts/ocp/setup-kagenti.sh --kagenti-repo "$(pwd)" --with-all`

## Symptoms (two distinct failures, same root cause)

**1. Step 3 — kagenti-deps: Gateway API CRD conflict**
Helm 4 server-side-*applies* the bundled `gateway-api-1.3.0` CRDs even though OCP 4.20 already ships Gateway API as GA (CRDs owned by `ingress-operator`, v1.2.1). SSA refuses:
```
Error: INSTALLATION FAILED: failed to install CRD crds/gateway-api-1.3.0.yaml:
conflict ... /gatewayclasses.gateway.networking.k8s.io ... conflict with "ingress-operator":
 .metadata.annotations.gateway.networking.k8s.io/bundle-version, .spec.versions
```
(Helm 3 skips pre-existing CRDs, so this never occurred on the supported version.)

**2. Step 5 — Kuadrant: two CRDs never installed → operator crash-loop**
With Helm 4, `kuadrants.kuadrant.io` and `authorinos.operator.authorino.kuadrant.io` are not created (15+ other kuadrant CRDs install fine), so `helm install --wait --timeout 5m` times out and the operator crash-loops:
```
Release "kuadrant-operator" failed: resource not ready, name: authorinos.operator.authorino.kuadrant.io ... NotFound
resource not ready, name: kuadrants.kuadrant.io ... NotFound
# operator log:
failed to list *v1beta1.Kuadrant: the server could not find the requested resource (get kuadrants.kuadrant.io)
```
(Helm 3 installs missing `crds/` CRDs, so this never occurred on the supported version.)

## Root cause
Helm 4's CRD-handling change (SSA-apply of `crds/`, including pre-existing ones; different install-if-missing semantics) is incompatible with the installer, which targets `>= 3.18.0 < 4`. The version is documented but not enforced.

## Proposed fix (RC5)
1. **Enforce the upper bound in the pre-flight** — hard-fail with a clear message when Helm major >= 4:
   > `helm 4.x detected; this installer requires helm >= 3.18.0 < 4. Install a 3.x helm (e.g. brew install helm@3).`
   This alone prevents the cryptic CRD failures.
2. (Optional, forward-compat for eventual Helm 4 support, separate work):
   - kagenti-deps: conditionally `--skip-crds` when Gateway API CRDs already exist on the cluster (so it doesn't fight cluster-managed GA Gateway API on OCP 4.20+).
   - kuadrant: ensure the operator-level CRDs (`kuadrants.kuadrant.io`, `authorinos.operator.authorino.kuadrant.io`) are installed under Helm 4.

## Workaround (until fixed)
Use Helm `3.18 <= x < 4` (e.g. `brew install helm@3`, then prepend its keg bin to PATH) and re-run.

cwiklik:
## Summary

`scripts/ocp/setup-kagenti.sh` documents a Helm requirement of `>= 3.18.0 < 4` (script header; `docs/install.md`) but the pre-flight only checks that `helm` *exists* — it never enforced the bounds. Running with **Helm 4** therefore proceeds and then fails with cryptic CRD errors, because Helm 4 changed how `crds/` are handled (server-side-applies pre-existing CRDs; different install-if-missing semantics).

Observed on OCP 4.20.8 / Helm v4.0.5:
- **kagenti-deps (Step 3):** Helm 4 SSA-applies the bundled `gateway-api-1.3.0` CRDs even though OCP 4.20 ships Gateway API as GA (owned by `ingress-operator`) → `conflict with "ingress-operator" ... .spec.versions`.
- **Kuadrant (Step 5):** `kuadrants.kuadrant.io` and `authorinos.operator.authorino.kuadrant.io` are never installed → `helm install --wait` times out, operator crash-loops (`failed to list *v1beta1.Kuadrant ... could not find the requested resource`).

Both disappear on the supported Helm 3.x (skips pre-existing CRDs; installs missing `crds/`).

## Change

Add a pre-flight guard that hard-fails when helm is **>= 4** or **< 3.18.0**, with a clear, actionable message (e.g. `brew install helm@3`). Falls back to a warning (no crash) if the version string can't be parsed. 16 lines, `setup-kagenti.sh` only.

```
Unsupported helm 4.0.5: this installer requires helm >= 3.18.0 and < 4.
Helm 4 breaks CRD installation (Gateway API conflict, missing Kuadrant CRDs); see issue #2072.
Install a supported version, e.g.: brew install helm@3
```

## Test

- `bash -n` syntax check passes; `make lint` (pre-commit) passes.
- Guard logic verified against sample versions: `4.0.5`→block, `3.21.2`→pass, `3.18.0`→pass, `3.17.9`→block, `3.15.3`→block, unparseable→warn.

Fixes #2072

Assisted-By: Claude Code

### DIFF
--- a/scripts/ocp/setup-kagenti.sh
+++ b/scripts/ocp/setup-kagenti.sh
@@ -202,6 +202,22 @@ if ! command -v helm &>/dev/null; then
   log_error "helm not found in PATH. Install helm >= 3.18.0"
   exit 1
 fi
+# Enforce the supported helm range: >= 3.18.0 and < 4. Helm 4 changed CRD
+# handling (server-side-applies pre-existing CRDs and alters crds/ install
+# semantics), which breaks the install: the kagenti-deps Gateway API CRDs
+# conflict with cluster-managed Gateway API on OCP 4.20+, and the Kuadrant
+# operator CRDs are not installed. See issue #2072.
+_helm_ver="$(helm version --short 2>/dev/null | sed -E 's/^v?([0-9]+\.[0-9]+\.[0-9]+).*/\1/')"
+_helm_major="${_helm_ver%%.*}"
+_helm_minor="$(printf '%s' "$_helm_ver" | cut -d. -f2)"
+if ! printf '%s' "$_helm_major" | grep -qE '^[0-9]+$'; then
+  log_warn "Could not parse helm version ('$_helm_ver'); this installer requires helm >= 3.18.0 < 4"
+elif [ "$_helm_major" -ge 4 ] || { [ "$_helm_major" -eq 3 ] && [ "${_helm_minor:-0}" -lt 18 ]; }; then
+  log_error "Unsupported helm ${_helm_ver}: this installer requires helm >= 3.18.0 and < 4."
+  log_error "Helm 4 breaks CRD installation (Gateway API conflict, missing Kuadrant CRDs); see issue #2072."
+  log_error "Install a supported version, e.g.: brew install helm@3"
+  exit 1
+fi
 log_success "helm found: $(helm version --short)"
 
 # Check python3


scripts/ocp/setup-kagenti.sh
```diff
@@ -202,6 +202,22 @@ if ! command -v helm &>/dev/null; then
...
+# semantics), which breaks the install: the kagenti-deps Gateway API CRDs
+# conflict with cluster-managed Gateway API on OCP 4.20+, and the Kuadrant
+# operator CRDs are not installed. See issue #2072.
+_helm_ver="$(helm version --short 2>/dev/null | sed -E 's/^v?([0-9]+\.[0-9]+\.[0-9]+).*/\1/')"
+_helm_major="${_helm_ver%%.*}"
+_helm_minor="$(printf '%s' "$_helm_ver" | cut -d. -f2)"
+if ! printf '%s' "$_helm_major" | grep -qE '^[0-9]+$'; then
+  log_warn "Could not parse helm version ('$_helm_ver'); this installer requires helm >= 3.18.0 < 4"
+elif [ "$_helm_major" -ge 4 ] || { [ "$_helm_major" -eq 3 ] && [ "${_helm_minor:-0}" -lt 18 ]; }; then
+  log_error "Unsupported helm ${_helm_ver}: this installer requires helm >= 3.18.0 and < 4."
```
pdettori, line 216: nit: The guard blocks `major >= 4` and `3.x < 3.18`, but a major `< 3` (e.g. helm 2.x) falls through to the `log_success` line despite the stated requirement `>= 3.18.0`. Practically moot since helm 2 is EOL, but for full parity with the error message you could add `[ "$_helm_major" -lt 3 ]` to the condition.

pdettori (approved):
Clean, well-scoped fix that closes a real footgun (Helm 4's CRD handling breaking the OCP install). Boundary logic is correct for all realistic versions: 4.0.5 blocks, 3.18.0 passes, 3.17.9 blocks, unparseable warns. The lower-bound check only inspects the minor (ignoring patch), which is fine since 3.18.0 is the floor. Script already runs `set -euo pipefail`, `log_warn`/`log_error` both exist, and `helm version --short` reliably succeeds so the command substitution won't trip `set -e`.

**Areas reviewed:** Shell (single file)
**Commits:** 1, signed-off: yes
**CI:** all passing (Shell Script Lint, DCO, verify-pr-title incl.)

Approving — one nit inline, non-blocking.

Assisted-By: Claude Code

PR #2097, opened 2026-06-26
Fix: make token-exchange Keycloak install Kind-safe (no hard oc dep)

### LINKED ISSUE #2092: ci(e2e): token-exchange Keycloak setup fails on Kind — '35-install-keycloak-ocp.sh' calls 'oc' (exit 127)
## Summary

The **token-exchange E2E track** fails during its Keycloak setup when run on **Kind** (vanilla k8s). `.github/scripts/token-exchange/35-install-keycloak-ocp.sh` shells out to `oc`, which is not installed in the Kind runner (only `kubectl` is), so the step aborts with **exit code 127** (command not found).

This is a **test-harness/CI bug, not a product defect** — the core platform E2E suites pass; only the token-exchange *setup* tooling breaks.

## Where

`.github/scripts/token-exchange/35-install-keycloak-ocp.sh` (community mode), line ~108:

```sh
KC_HOST="${KEYCLOAK_HOST:-keycloak-keycloak.$(oc get ingresses.config/cluster -o jsonpath='{.spec.domain}' 2>/dev/null)}"
```

Under `set -e`, the `$(oc …)` command substitution returns 127 ("oc: command not found") and aborts the script. The `2>/dev/null` hides the message but not the exit code.

## Evidence (release-validation run on v0.6.1-rc.5, Kind E2E job)

Sequence right before the failure:
```
[Wave 35] Install Keycloak on OCP (mode: community)
ℹ Installing community Keycloak 26.6.0
… keycloak-operator + CRDs created …
ℹ Deploying PostgreSQL for Keycloak
statefulset.apps/postgres-kc created
Waiting for 1 pods to be ready...
partitioned roll out complete: 1 new pods have been updated...
##[error]Process completed with exit code 127.
```
Timing matches: Postgres rollout completes (script line 105) → next line (108) calls `oc` → 127.

**Core platform E2E passed in the same run:** `72 passed, 30 skipped` and `29 passed, 4 skipped`. So the product is healthy on Kind; this is purely the token-exchange setup tooling.

Red herrings to ignore in the log: `deployments.apps "keycloak" not found` (on-failure diagnostic dump uses `kubectl logs deployment/keycloak`, but Keycloak is a StatefulSet) and weather-service `502`/`Workload not ready` (transient).

## Impact

- Token-exchange E2E cannot complete on Kind → `release-validation.yaml` Kind leg goes red even when the platform is fine.
- Likely **pre-existing on `main`** (the script is OCP-oriented; the `oc` dependency is unconditional).

## Proposed fix

Make `35-install-keycloak-ocp.sh` not depend on `oc` in the Kind path — e.g.:
- guard the `oc` call (only derive the route domain via `oc` when `oc` exists / on OpenShift), and fall back to a Kind-appropriate host (localhost port-forward / `kubectl`-derived) otherwise; or
- detect platform and skip the OCP-only host derivation on Kind.

## Repro

Run `release-validation.yaml` with `-f ref=<tag> -f skip_hypershift=true` (Kind only), or the Kind token-exchange E2E locally; observe exit 127 at the Keycloak install step.

cwiklik:
## Summary

The token-exchange E2E track fails on **Kind** when `35-install-keycloak-ocp.sh` derives the Keycloak CR hostname via `oc get ingresses.config/cluster`. On Kind the runner has only `kubectl` (no `oc`), so under `set -e` the command substitution returns **exit 127** ("oc: command not found") and aborts the step — right after the Postgres rollout. Caught by the `release-validation.yaml` Kind E2E during the v0.6.1 release.

## Fix

Add a `resolve_kc_host()` helper in `lib.sh` (mirrors the existing `get_keycloak_url()` platform handling):
- use `KEYCLOAK_HOST` if set;
- derive the OCP ingress domain via `oc` **only when `oc` is available**;
- otherwise fall back to an in-cluster hostname (`keycloak-service.<ns>.svc.cluster.local`) — the Keycloak CR uses `strict: false`, so a non-routable host is fine for Kind.

Replace both `oc`-dependent call sites (community + rhbk modes).

## Test

Reproduced and verified with `oc` absent under `set -e`:

| | reaches next line | exit code |
|---|---|---|
| OLD (inline `$(oc …)`) | no (abort) | **127** |
| NEW (`resolve_kc_host`) | yes → `keycloak-service.keycloak.svc.cluster.local` | **0** |

Also verified: `KEYCLOAK_HOST` override wins; OCP path (fake `oc` returning a domain) yields `keycloak-keycloak.<domain>`. `bash -n` clean on both files.

Fixes #2092

Assisted-By: Claude Code

### DIFF
--- a/.github/scripts/token-exchange/35-install-keycloak-ocp.sh
+++ b/.github/scripts/token-exchange/35-install-keycloak-ocp.sh
@@ -105,7 +105,7 @@ EOF
   kubectl rollout status statefulset/postgres-kc -n "$KC_NAMESPACE" --timeout=2m 2>/dev/null || true
 
   # Determine hostname
-  KC_HOST="${KEYCLOAK_HOST:-keycloak-keycloak.$(oc get ingresses.config/cluster -o jsonpath='{.spec.domain}' 2>/dev/null)}"
+  KC_HOST="$(resolve_kc_host)"
 
   # Deploy Keycloak CR
   log_info "Deploying Keycloak CR (community)"
@@ -168,7 +168,7 @@ EOF
     sleep 10
   done
 
-  KC_HOST="${KEYCLOAK_HOST:-keycloak-keycloak.$(oc get ingresses.config/cluster -o jsonpath='{.spec.domain}' 2>/dev/null)}"
+  KC_HOST="$(resolve_kc_host)"
 
   log_info "Deploying Keycloak CR (RHBK)"
   cat <<EOF | kubectl apply -n "$KC_NAMESPACE" -f -
--- a/.github/scripts/token-exchange/lib.sh
+++ b/.github/scripts/token-exchange/lib.sh
@@ -52,6 +52,26 @@ get_keycloak_url() {
   echo "${KEYCLOAK_URL:-http://keycloak-service.${KC_NAMESPACE}.svc:8080}"
 }
 
+# Resolve the Keycloak CR external hostname.
+# On OCP, derive it from the ingress domain via `oc`. On Kind (or any cluster
+# without `oc`), fall back to an in-cluster hostname so the install does not
+# abort under `set -e` when `oc` is absent (the Keycloak CR uses strict: false).
+resolve_kc_host() {
+  if [[ -n "${KEYCLOAK_HOST:-}" ]]; then
+    echo "$KEYCLOAK_HOST"
+    return
+  fi
+  local domain=""
+  if command -v oc &>/dev/null; then
+    domain=$(oc get ingresses.config/cluster -o jsonpath='{.spec.domain}' 2>/dev/null || true)
+  fi
+  if [[ -n "$domain" ]]; then
+    echo "keycloak-keycloak.${domain}"
+  else
+    echo "keycloak-service.${KC_NAMESPACE}.svc.cluster.local"
+  fi
+}
+
 # Get Keycloak admin credentials
 get_kc_admin_creds() {
   local user pass


cwiklik:
### Validation: fix confirmed in CI

Ran `release-validation.yaml` (Kind only) against this PR's head:
[run 28240918869](https://github.com/kagenti/kagenti/actions/runs/28240918869)

**The fix works** — the step that died with `exit 127` in the v0.6.1 release validation now passes on an `oc`-less runner:

| Step | Result |
|------|--------|
| **Setup Keycloak for token exchange** (runs the fixed `35-install-keycloak-ocp.sh`) | ✅ **success** |
| Wait for token exchange workloads | ✅ success |

Previously this aborted with `127` ("oc: command not found") right after the Postgres rollout; with `resolve_kc_host()` it resolves to the in-cluster fallback and proceeds.

**Note for reviewers:** the run's overall Kind leg is still red, but at a **different, later** step — `Run token exchange E2E tests` fails with Keycloak `error="client_not_found"` (the `tx-e2e` test client isn't registered). That is a **separate, pre-existing** issue that was masked by the `127` crash (the suite never ran to completion on Kind before) — it is unrelated to this change (a hostname mismatch would surface as `invalid_issuer`/`unauthorized`, not `client_not_found`). So please don't gate this PR on the full Kind leg going green; the step this PR fixes is green.

Assisted-By: Claude Code

cwiklik:
The downstream token-exchange test failure noted above (`client_not_found` in the `tx-e2e` realm, tied to `tx-e2e` workloads CrashLoopBackOff) is now tracked separately in #2098 — it is pre-existing and unrelated to this fix, so it should not block this PR. The step this PR fixes ('Setup Keycloak for token exchange') is green.

huang195 (approved):
Dobra robota. Clean, correct fix.

Root cause is right: under `set -e`, the inline `$(oc …)` in the parameter-default returns **127** when `oc` is absent and aborts the step (right after the Postgres rollout). `resolve_kc_host()` guards it properly:
- `command -v oc` + `|| true` — a missing *or* failing `oc` can't abort the step
- `KEYCLOAK_HOST` override preserved
- in-cluster fallback for Kind (`keycloak-service.<ns>.svc.cluster.local`); the Keycloak CR is `strict: false`, so a non-routable host is fine

Verified `lib.sh` is sourced (`35-install-keycloak-ocp.sh:14`) so the helper resolves, and both call sites (community + rhbk) are swapped. Nic do dodania.

**Areas reviewed**: Shell
**Commits**: signed-off (DCO ✓)
**CI**: all green (E2E HyperShift pending = maintainer `/run-e2e`, not a failure)

Issue #1906, opened 2026-06-12
🐛 Bug: Disappearing Agents and Tools on Dashboard


davidhadas:
### Describe the bug

Problem A: 
Agent that does not show up on dashboard - no way to remove it and no way to see it.

```code
kubectl get pods -n team1
NAME                                  READY   STATUS      RESTARTS   AGE
weather-service                       2/2     Running     0          19m
weather-service-run-btqb4-tr4x9-pod   0/2     Completed   0          4d8h
weather-tool-655576fdbd-qlmrb         1/1     Running     0          20m
weather-tool-run-nsg4w-xzbfw-pod      0/2     Completed   0          4d8h (edited)2 files
```

<img width="2469" height="724" alt="Image" src="https://github.com/user-attachments/assets/5bd8d052-13fc-448d-bf3a-8cb7d027c680" />

Problem B: 
The dashboard UI feels incomplete in how it handles imported tools and agents. When I import a tool or an agent, it doesn’t appear immediately in the list, so it effectively disappears into oblivion. There’s no feedback, no confirmation, and no way to access it for a while.

If everything eventually loads correctly, the item does show up later — but the delay makes it unclear whether the import succeeded or failed. It would be much clearer if newly imported items appeared instantly or at least surfaced with a status indicator. 

Problem C: 
If a user wish to delete an agent or a tool after they were deployed and before they completed their deployment processes successfully, he has no way to do that

These problems are all related. 

### Steps To Reproduce

Set  values.yaml to `sandbox: false` and import an agent. 
You now cant see that agent and you can delete it.

Import a tool or agent, switch to another page in the dhasboard and switch back to the agent list or tool list to see them empty and the deployment process not shown.

### Expected Behavior

As soon as you press import of a tool or agent, it shows up as a line in the list indicating "deploying" or something similar. 
If the deployment stops or fails or have issues - it is shown on that line!
You can stop the deployment and clear the system during this stage by deleting the tool or agent. 
After clearing the agent or tool you can redeploy them immediately using import as before (the name is now free and the system is cleared from traces of the old deployment).

esnible:
@davidhadas , thank you for filing this!

Can you explain how you reproduce this by setting sandbox to false?  When I set it to false, I can't create sandboxes...

I know of the following ways to reproduce and partial failure problems
- Agents and Tools created with "Build From Source" can't build -- a lot of resources are left on the cluster, invisible to the UI
[...]

PR #2111, opened 2026-06-29
Fix: Don't forget about failed and progressing Agent builds

esnible:
## Summary

For #1906

If this PR is approved I will follow up with
- a similar PR for tool builds

I will then close the issue with a final PR to cleanup partial failures on POST /api/v1/agents and POST /api/v1/tools

Deploy Rossi
Bring in this PR
`make build-load-ui-frontend` and `make build-load-ui-backend` plus the usual rollout.

### DIFF
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -848,6 +848,65 @@ async def list_agents(
                 if e.status not in (404, 403):
                     logger.warning(f"Failed to list legacy Agent CRDs: {e.reason}")
 
+        # Surface in-progress / failed Shipwright source builds that have no
+        # workload yet. A source-built agent has no Deployment/StatefulSet/etc.
+        # until its build Succeeds and is finalized, so without this it would be
+        # invisible here while building or after a failure. Guarded so a
+        # build-listing failure never breaks the core agent list.
+        try:
+            builds = collect_kagenti_shipwright_builds(
+                kube, [namespace], RESOURCE_TYPE_AGENT, logger
+            )
+            for build in builds:
+                # Workload already exists (build Succeeded + finalized, or a
+                # name collision) -> already listed above; skip to avoid dupes.
+                if build.name in agent_names:
+                    continue
+
+                try:
+                    buildruns = kube.list_custom_resources(
+                        group=SHIPWRIGHT_CRD_GROUP,
+                        version=SHIPWRIGHT_CRD_VERSION,
+                        namespace=build.namespace,
+                        plural=SHIPWRIGHT_BUILDRUNS_PLURAL,
+                        label_selector=f"kagenti.io/build-name={build.name}",
+                    )
+                except ApiException:
+                    # Constant message only (CodeQL py/log-injection): never
+                    # interpolate namespace / build name / API reason.
+                    logger.warning("Failed to list BuildRuns for a Shipwright build")
+                    buildruns = []
+
+                # No BuildRun yet -> build just started; treat as "Building".
+                phase = "Pending"
+                latest = get_latest_buildrun(buildruns) if buildruns else None
+                if latest:
+                    phase = extract_buildrun_info(latest)["phase"]
+
+                # Succeeded builds either already have a workload (listed above)
+                # or are about to be finalized into one; don't surface them here.
+                if phase == "Succeeded":
+                    continue
+                status = "Build Failed" if phase == "Failed" else "Building"
+
+                agents.append(
+                    AgentSummary(
+                        name=build.name,
+                        namespace=build.namespace,
+                        description="Building from source",
+                        status=status,
+                        labels=_extract_labels({KAGENTI_TYPE_LABEL: build.resourceType}),
+                        workloadType=WORKLOAD_TYPE_DEPLOYMENT,
+                        # Collector already formats this as an ISO string, so do
+                        # not pass it through _format_timestamp (datetime-only).
+                        createdAt=build.creationTimestamp,
+                    )
+                )
+                agent_names.add(build.name)
+        except ApiException as e:
+            if e.status not in (404, 403):
+                logger.warning("Failed to list Shipwright builds for agents")
+
         return AgentListResponse(items=agents)
 
     except ApiException as e:
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -896,6 +896,7 @@ async def list_agents(
                         description="Building from source",
                         status=status,
                         labels=_extract_labels({KAGENTI_TYPE_LABEL: build.resourceType}),
+                        # Note that we may be building a non-deployment.  TODO record and retrieve build type.
                         workloadType=WORKLOAD_TYPE_DEPLOYMENT,
                         # Collector already formats this as an ISO string, so do
                         # not pass it through _format_timestamp (datetime-only).
@@ -904,8 +905,7 @@ async def list_agents(
                 )
                 agent_names.add(build.name)
         except ApiException as e:
-            if e.status not in (404, 403):
-                logger.warning("Failed to list Shipwright builds for agents")
+            logger.warning("Failed to list Shipwright builds for agents", exc_info=True)
 
         return AgentListResponse(items=agents)
 
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -904,7 +904,7 @@ async def list_agents(
                     )
                 )
                 agent_names.add(build.name)
-        except ApiException as e:
+        except ApiException:
             logger.warning("Failed to list Shipwright builds for agents", exc_info=True)
 
         return AgentListResponse(items=agents)
--- a/kagenti/backend/tests/test_shipwright_builds_api.py
+++ b/kagenti/backend/tests/test_shipwright_builds_api.py
@@ -217,6 +217,106 @@ class TestListAgentShipwrightBuilds:
         agents_shipwright_app.dependency_overrides.clear()
 
 
+def _buildrun(name: str, succeeded_status: str | None):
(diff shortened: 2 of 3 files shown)


kagenti/backend/app/routers/agents.py
```diff
@@ -848,6 +848,65 @@ async def list_agents(
...
+                status = "Build Failed" if phase == "Failed" else "Building"
+
+                agents.append(
+                    AgentSummary(
+                        name=build.name,
+                        namespace=build.namespace,
+                        description="Building from source",
+                        status=status,
+                        labels=_extract_labels({KAGENTI_TYPE_LABEL: build.resourceType}),
+                        workloadType=WORKLOAD_TYPE_DEPLOYMENT,
```
clawgenti, line 900: **nit**: `workloadType=WORKLOAD_TYPE_DEPLOYMENT` is hardcoded for all source-built agents, but a source build could in principle produce a StatefulSet or other workload type. Since the actual workload type isn't known until the build finishes and is finalized, this is a reasonable placeholder — worth a follow-up comment or TODO noting this assumption.

clawgenti (commented):
Adds visibility into in-progress and failed Shipwright source builds in the agent list endpoint, with matching frontend type expansion and solid test coverage for all phase transitions. All checks pass. Ready for human review.

---

kagenti/backend/app/routers/agents.py
clawgenti, line 900: **nit**: `workloadType=WORKLOAD_TYPE_DEPLOYMENT` is hardcoded for build-phase entries. Agents building from source could eventually become a StatefulSet or Job. This is fine as a transient placeholder (the value gets corrected once the real workload exists), but worth a brief comment explaining the assumption, or tracking in the follow-up cleanup PR.

clawgenti (commented):
Adds build-phase visibility to the agent list endpoint: in-progress and failed Shipwright source builds now surface with `Building` / `Build Failed` status until their workload exists. The implementation is well-guarded, well-commented, and the test suite covers all branch transitions cleanly.

---

kagenti/backend/app/routers/agents.py
```diff
@@ -848,6 +848,65 @@ async def list_agents(
...
+                        # Collector already formats this as an ISO string, so do
+                        # not pass it through _format_timestamp (datetime-only).
+                        createdAt=build.creationTimestamp,
+                    )
+                )
+                agent_names.add(build.name)
+        except ApiException as e:
+            if e.status not in (404, 403):
```
rubambiza: Is there a reason why we are looking for *not* 404 and 403 in particular here? A comment for the discrimination would be nice for maintainers.

kagenti/backend/tests/test_shipwright_builds_api.py
```diff
@@ -217,6 +217,107 @@ def override_kube():
...
 
 
+def _buildrun(name: str, succeeded_status: str | None):
+    """A BuildRun dict. succeeded_status is the 'Succeeded' condition status
+    ("True"/"False"/"Unknown"), or None for a BuildRun with no conditions."""
+    conditions = []
+    if succeeded_status is not None:
+        conditions = [{"type": "Succeeded", "status": succeeded_status, "message": "boom"}]
+    return {
+        "metadata": {"name": name, "creationTimestamp": "2025-01-02T00:00:00Z"},
```
rubambiza, line 228: The `creationTimestamp` is hardcoded here, so I am assuming it is useless at the caller site, correct?

rubambiza (commented):
Leaving some independent, clarifying comments before approval.

kagenti/backend/app/routers/agents.py
```diff
@@ -848,6 +848,65 @@ async def list_agents(
...
+                        labels=_extract_labels({KAGENTI_TYPE_LABEL: build.resourceType}),
+                        # Note that we may be building a non-deployment.  TODO record and retrieve build type.
+                        workloadType=WORKLOAD_TYPE_DEPLOYMENT,
+                        # Collector already formats this as an ISO string, so do
+                        # not pass it through _format_timestamp (datetime-only).
+                        createdAt=build.creationTimestamp,
+                    )
+                )
+                agent_names.add(build.name)
+        except ApiException as e:
```
clawgenti: **must-fix** — `e` is assigned but never used (ruff F841, blocking CI lint). Since `exc_info=True` already captures the exception in the log record, the binding serves no purpose. Change to:
```python
except ApiException:
```

clawgenti (commented):
Surfaces in-progress and failed Shipwright builds in the agent list by collecting BuildRuns and synthesizing `Building`/`Build Failed` status entries — logic is sound, deduplication against existing workloads is correct, and the test suite covers all new branches. One lint blocker: `except ApiException as e:` at line 907 binds `e` but never uses it (ruff F841), causing the `lint (3.11/3.12)` checks to fail — change to `except ApiException:`.

---

kagenti/backend/app/routers/agents.py
```diff
@@ -848,6 +848,65 @@ async def list_agents(
...
+                status = "Build Failed" if phase == "Failed" else "Building"
+
+                agents.append(
+                    AgentSummary(
+                        name=build.name,
+                        namespace=build.namespace,
+                        description="Building from source",
+                        status=status,
+                        labels=_extract_labels({KAGENTI_TYPE_LABEL: build.resourceType}),
+                        # Note that we may be building a non-deployment.  TODO record and retrieve build type.
```
clawgenti, line 899: **nit**: The TODO on `workloadType=WORKLOAD_TYPE_DEPLOYMENT` is noted — worth tracking as a follow-up issue if tool builds can produce StatefulSets/Jobs, since this will incorrectly report them as deployments. Existing issue reference or a follow-up PR would help ensure this doesn't get lost.

kagenti/backend/tests/test_shipwright_builds_api.py
```diff
@@ -217,6 +217,108 @@ def override_kube():
...
+                "status": {"conditions": [{"type": "Available", "status": "True"}]},
+            }
+        ]
+        # Build with the same name as the deployment -> must not be added twice.
+        items = self._run(kube, _buildrun("br1", "False"))
+        assert len(items) == 1
+        # The single entry is the real workload, not the synthetic build entry.
+        assert items[0]["status"] != "Build Failed"
+
+    def test_build_listing_failure_does_not_break_list(self):
```
clawgenti, line 309: **suggestion**: The outer `ApiException` guard (when `collect_kagenti_shipwright_builds` fails) is well tested here. Consider adding a test for the inner per-build `list_custom_resources` `ApiException` path (line 874 in `agents.py`) — it sets `buildruns = []` and continues, but there's no test verifying that a failure on a single build's BuildRun fetch still surfaces the build entry rather than silently dropping it.

clawgenti (commented):
Surfaces in-progress and failed Shipwright agent builds in `GET /api/v1/agents` so they are visible before a workload is created — clean implementation with solid guard structure and good test coverage.

---

kagenti/backend/tests/test_shipwright_builds_api.py
esnible: If there are problems fetching a BuildRun it is fine to drop that BuildRun.  We don't want to surface builds that Kagenti is not able to retrieve reliably.  The purpose of this PR is to allow the user to track in-progress and failed builds; if the backend can't retrieve a build there is no point in displaying it to the user.

kagenti/backend/app/routers/agents.py
```diff
@@ -848,6 +848,65 @@ async def list_agents(
...
+                    )
+                except ApiException:
+                    # Constant message only (CodeQL py/log-injection): never
+                    # interpolate namespace / build name / API reason.
+                    logger.warning("Failed to list BuildRuns for a Shipwright build")
+                    buildruns = []
+
+                # No BuildRun yet -> build just started; treat as "Building".
+                phase = "Pending"
+                latest = get_latest_buildrun(buildruns) if buildruns else None
```
clawgenti, line 882: **nit**: `buildruns` is initialized to `[]` on the `except ApiException` path (line 878), so this ternary guard is redundant — an empty list is already falsy and `get_latest_buildrun` would return `None` naturally if passed an empty list (or the condition could just be `if buildruns:`). Minor verbosity, no functional issue.

kagenti/backend/app/routers/agents.py
clawgenti, line 899: **suggestion**: This comment acknowledges a known limitation (`workloadType` hardcoded to `WORKLOAD_TYPE_DEPLOYMENT` even for StatefulSet/Job builds). Consider opening a follow-up issue to track it — inline TODOs in the agent list path can go unnoticed.

clawgenti (commented):
Surfaces in-progress and failed Shipwright agent builds in `GET /agents` — previously invisible until a build succeeded and a workload was created. Logic is well-guarded, dedup is correct, and all five test scenarios are covered.

- **nit** `agents.py:882`: `get_latest_buildrun(buildruns) if buildruns else None` — `buildruns` is already `[]` on the error path, so this guard is redundant (the `if buildruns` check is always correct, just slightly verbose given the initialization above).
- **suggestion** `agents.py:899`: The `TODO` about `workloadType` being hardcoded to `WORKLOAD_TYPE_DEPLOYMENT` for non-deployment builds is a known gap — worth filing a follow-up issue to track it so it doesn't get lost.

---

pdettori (approved):
Clean, well-commented additive change that surfaces previously-invisible in-progress and failed Shipwright builds. All prior must-fix items (F841 lint) addressed, extensive clarifying comments added, solid test coverage for all phase transitions. Remaining nits (redundant `if buildruns` guard, missing inner-ApiException-specific test) are non-blocking — author's rationale for graceful degradation is sound. Ready to merge.

---

PR #2114, opened 2026-06-29
Feat: 🖥️ Add options for persistent storage to Kagenti CLI

moonlight16:
## Summary

The Kagenti UI already lets users deploy agents as StatefulSets with a PVC, but the CLI didn't have the matching flags. This PR adds them.

While end-to-end testing the new CLI flags, I found that the backend's StatefulSet manifest builder was silently ignoring `request.persistentStorage` — it always emitted `shared-data` as an `emptyDir`, never as a PVC. That meant the UI's "persistent storage size" field on the StatefulSet create form was also a no-op at runtime: pods looked like they had a PVC, but `/shared` was ephemeral and lost on restart. This PR patches the StatefulSet builder so the request is honored, which fixes both the new CLI path and the existing UI path.

### CLI changes
- Adds `--workload-type` to `kagenti deploy agent` with support for `deployment`, `statefulset`, and `job`.
- Adds StatefulSet persistent storage flags:
  - `--persistent-storage`
  - `--persistent-storage-size`
- Sends `persistentStorage` in the agent create request when persistent storage is enabled.
- Validates that persistent storage flags are only used with `--workload-type statefulset`.
- Validates `--persistent-storage-size` against the K8s quantity format and a 1Mi–10Ti range so typos fail at cobra with a clear message instead of bubbling back as a K8s 5xx.
- Adds CLI tests covering workload type forwarding, persistent storage forwarding, invalid flag combinations, and size-format validation.

### Backend fix
- Mirrors the existing Sandbox PVC pattern in `_build_statefulset_manifest`: when `persistentStorage.enabled` is true, drop the `shared-data` `emptyDir` and add a `volumeClaimTemplates[0]` named `shared-data` (size from request, `ReadWriteOnce`, default StorageClass).
- Adds `tests/test_statefulset_manifest.py` mirroring the sandbox PVC tests (6 tests: no-PVC default, PVC when enabled, emptyDir replaced, mount unchanged, disabled flag, default size).
- Behavior change is gated entirely on `persistentStorage.enabled` — existing StatefulSet agents created without the flag are unaffected.

This work will also be useful for Context Operator development.

## Testing

- CLI: `cd kagenti/tui && go test ./internal/cli`
- Backend: `cd kagenti/backend && uv run pytest tests/test_statefulset_manifest.py tests/test_sandbox_manifest.py`
- End-to-end deploy + restart verification: planned with a minimal A2A agent in `agent-examples`; will update once that lands.

## Related issue(s)

None.

Assisted-By: Codex, Claude (Anthropic AI)

### DIFF
--- a/kagenti/tui/internal/cli/deploy.go
+++ b/kagenti/tui/internal/cli/deploy.go
@@ -29,6 +29,9 @@ func newDeployAgentCmd(ctx *CLIContext) *cobra.Command {
 		framework      string
 		protocol       string
 		deployMethod   string
+		workloadType   string
+		persistent     bool
+		storageSize    string
 		containerImage string
 		gitURL         string
 		gitBranch      string
@@ -46,6 +49,13 @@ func newDeployAgentCmd(ctx *CLIContext) *cobra.Command {
 		Use:   "agent",
 		Short: "Deploy an agent",
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if persistent && workloadType != "statefulset" {
+				return fmt.Errorf("--persistent-storage is only supported with --workload-type statefulset")
+			}
+			if cmd.Flags().Changed("persistent-storage-size") && !persistent {
+				return fmt.Errorf("--persistent-storage-size requires --persistent-storage")
+			}
+
 			ns, _ := cmd.Flags().GetString("namespace")
 			if ns == "" {
 				ns = ctx.Client.Namespace
@@ -86,7 +96,7 @@ func newDeployAgentCmd(ctx *CLIContext) *cobra.Command {
 				Protocol:          protocol,
 				Framework:         framework,
 				DeploymentMethod:  deployMethod,
-				WorkloadType:      "deployment",
+				WorkloadType:      workloadType,
 				ContainerImage:    containerImage,
 				GitURL:            gitURL,
 				GitBranch:         gitBranch,
@@ -94,6 +104,12 @@ func newDeployAgentCmd(ctx *CLIContext) *cobra.Command {
 				AuthBridgeEnabled: true,
 				SpireEnabled:      spire,
 				EnvVars:           allEnv,
+			}
+			if persistent {
+				req.PersistentStorage = &api.PersistentStorageConfig{
+					Enabled: true,
+					Size:    storageSize,
+				}
 			}
 
 			resp, err := ctx.Client.CreateAgent(req)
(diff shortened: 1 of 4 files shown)


kagenti/tui/internal/cli/deploy.go
```diff
@@ -112,6 +128,9 @@ func newDeployAgentCmd(ctx *CLIContext) *cobra.Command {
 	cmd.Flags().StringVar(&framework, "framework", "LangGraph", "Framework (LangGraph, CrewAI, AG2, Custom)")
 	cmd.Flags().StringVar(&protocol, "protocol", "a2a", "Protocol (a2a, mcp)")
 	cmd.Flags().StringVar(&deployMethod, "deploy-method", "image", "Deployment method (image, source)")
+	cmd.Flags().StringVar(&workloadType, "workload-type", "deployment", "Workload type (deployment, statefulset, job)")
+	cmd.Flags().BoolVar(&persistent, "persistent-storage", false, "Enable persistent storage (statefulset only)")
+	cmd.Flags().StringVar(&storageSize, "persistent-storage-size", "1Gi", "Persistent volume claim size (e.g., 1Gi, 5Gi, 10Gi)")
```
evaline-ju, line 207: I assume any invalid workload types and storage sizes will just fall through to error by k8s?

evaline-ju (approved):
LGTM - one confirmation q

kagenti/tui/internal/cli/deploy.go
moonlight16: Thanks @evaline-ju — good question. It turns out the answer differs between the two:

- **Workload type:** the backend's `CreateAgentRequest` has a Pydantic `field_validator` (`validate_workload_type`) that rejects anything outside `SUPPORTED_WORKLOAD_TYPES` and returns HTTP 422 before K8s is involved. So invalid types fail fast at the API layer.
- **Storage size:** the string was being passed straight into the `volumeClaimTemplate`'s `resources.requests.storage`, so invalid values would only get caught by the K8s API server's resource-quantity parser and bubble back as a 5xx — which isn't a great UX.

Since the size case is the one a user is most likely to hit (typos like `1XB`, `5gi`, etc.), I added a CLI-side validator in a follow-up commit on this branch. It checks the K8s quantity format (`1Gi`, `500Mi`, `5G`, plain bytes, fractional values) and a sane size range (1Mi–10Ti) so obvious mistakes fail at cobra with a clear message instead of after a round-trip to K8s. Tests cover the valid/invalid cases and the wiring into the deploy command.

PR #2122, opened 2026-06-30
fix(v0.6.1, ocp): make weather-agent CLI deploy work on OpenShift

webchang:
## Summary

Deploying the weather agent demo via the `kagenti-operator` CLI scripts
(`.github/scripts/kagenti-operator/74-…` and `75-…`) on HyperShift OpenShift
clusters (validated on two clusters running kagenti v0.6.1) surfaced several
OpenShift-specific gaps. This PR fixes the ones with a clear, low-risk code
change; all are OpenShift-only paths and do not affect Kind.

### `75-deploy-weather-tool-shipwright.sh`
- **Build strategy:** used `ClusterBuildStrategy: buildah-insecure`, which is
  absent on clusters that only ship `buildah` → BuildRun fails with
  `ClusterBuildStrategyNotFound`. Aligned with the agent build (`buildah`).
- **Empty output image:** some Shipwright versions populate
  `.status.output.digest` but leave `.status.output.image` empty, so the tool
  Deployment was created with an empty image (`image: Required value`). Falls
  back to the registry path the build was told to push to.
- **uv cache CrashLoop:** OpenShift runs the container with an arbitrary
  non-root UID whose `$HOME` is `/`, so uv's default cache (`/.cache/uv`) is not
  writable and the tool pod `CrashLoopBackOff`s. Sets `UV_CACHE_DIR=/tmp/uv-cache`
  and `HOME=/tmp` (the Deployment already mounts a writable `/tmp` emptyDir).

### `74-deploy-weather-agent.sh`
- **Route 504:** the `weather-service` Route had no timeout annotation, so a
  cold first chat (MCP session init + multiple LLM round-trips) exceeded
  HAProxy's 30s default and 504'd. Adds `haproxy.router.openshift.io/timeout: 120s`.

### `charts/kagenti/templates/ui.yaml`
- **UI chat 504:** the `kagenti-ui` and `kagenti-api` OpenShift Routes had the
  same 30s default. UI chat is proxied through these front-door routes, so it
  504'd even after the per-agent route was fixed. Adds the 120s timeout to both.

## Test plan

- [x] Deployed the weather agent end-to-end on two HyperShift OCP clusters
      (v0.6.1): tool + agent build via Shipwright, agentcard `SYNCED=True`,
      UI chat returns answers (no 504) after all fixes.
- [x] `helm template` renders `ui.yaml` cleanly with the new annotations on
      both routes.
- [x] `make lint` / pre-commit pass.
- [ ] CI green.

## Follow-ups (not in this PR)

- The legacy client-registration sidecar (`kagenti.io/client-registration-inject: "true"`)
  expects `keycloak-admin-secret` in the **agent** namespace, but v0.6.1 only
  creates it in `kagenti-system`, so the agent pod hangs in
  `CreateContainerConfigError`. Needs a design decision: copy the secret into
  the agent namespace when the label is present, or drop the legacy label and
  use the operator-managed registration path. Currently a manual workaround.
- Confirm whether any supported cluster genuinely needs the `buildah-insecure`
  strategy; if so, make the strategy name configurable rather than hardcoded.

Assisted-By: Claude Code

### DIFF
--- a/.github/scripts/kagenti-operator/75-deploy-weather-tool-shipwright.sh
+++ b/.github/scripts/kagenti-operator/75-deploy-weather-tool-shipwright.sh
@@ -67,7 +67,7 @@ spec:
       revision: ${GIT_BRANCH}
     contextDir: ${GIT_PATH}
   strategy:
-    name: buildah-insecure
+    name: buildah
     kind: ClusterBuildStrategy
   output:
     image: ${REGISTRY}/${TOOL_NAME}:${IMAGE_TAG}
@@ -133,6 +133,12 @@ echo ""
 echo "Step 4: Retrieving output image..."
 OUTPUT_IMAGE=$(kubectl get buildrun "${BUILDRUN_NAME}" -n "${NAMESPACE}" -o jsonpath='{.status.output.image}')
 OUTPUT_DIGEST=$(kubectl get buildrun "${BUILDRUN_NAME}" -n "${NAMESPACE}" -o jsonpath='{.status.output.digest}')
+# Some Shipwright versions leave .status.output.image empty (only digest is set).
+# Fall back to the registry path we asked the build to push to.
+if [ -z "${OUTPUT_IMAGE}" ]; then
+    OUTPUT_IMAGE="${REGISTRY}/${TOOL_NAME}:${IMAGE_TAG}"
+    echo "Output image empty in BuildRun status; falling back to ${OUTPUT_IMAGE}"
+fi
 echo "Output Image: ${OUTPUT_IMAGE}"
 echo "Output Digest: ${OUTPUT_DIGEST}"
 
@@ -177,6 +183,14 @@ spec:
         - name: mcp
           image: ${OUTPUT_IMAGE}
           imagePullPolicy: Always
+          env:
+            # OpenShift runs the container with an arbitrary non-root UID, so
+            # uv's default cache (\$HOME/.cache -> /.cache) is not writable.
+            # Point both at the writable /tmp emptyDir mounted below.
+            - name: UV_CACHE_DIR
+              value: /tmp/uv-cache
+            - name: HOME
+              value: /tmp
           ports:
             - containerPort: 8000
               name: http
--- a/.github/scripts/kagenti-operator/74-deploy-weather-agent.sh
+++ b/.github/scripts/kagenti-operator/74-deploy-weather-agent.sh
@@ -223,6 +223,9 @@ metadata:
   namespace: team1
   annotations:
     openshift.io/host.generated: "true"
+    # Agent chat involves a cold MCP session + multiple LLM round-trips that can
+    # exceed HAProxy's 30s default, surfacing as a 504 on the first request.
+    haproxy.router.openshift.io/timeout: "120s"
 spec:
   path: /
   port:
--- a/charts/kagenti/templates/ui.yaml
+++ b/charts/kagenti/templates/ui.yaml
@@ -523,6 +523,10 @@ metadata:
     {{- include "kagenti.commonLabels" . | nindent 4 }}
   annotations:
     openshift.io/host.generated: "true"
+    # Agent chat proxied through the UI involves a cold MCP session + multiple
+    # LLM round-trips that can exceed HAProxy's 30s default, surfacing as a 504
+    # on the first request even though the agent itself completes.
+    haproxy.router.openshift.io/timeout: "120s"
 spec:
   path: /
   port:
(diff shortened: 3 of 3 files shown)


.github/scripts/kagenti-operator/75-deploy-weather-tool-shipwright.sh
```diff
@@ -133,6 +133,12 @@ echo ""
 echo "Step 4: Retrieving output image..."
 OUTPUT_IMAGE=$(kubectl get buildrun "${BUILDRUN_NAME}" -n "${NAMESPACE}" -o jsonpath='{.status.output.image}')
 OUTPUT_DIGEST=$(kubectl get buildrun "${BUILDRUN_NAME}" -n "${NAMESPACE}" -o jsonpath='{.status.output.digest}')
+# Some Shipwright versions leave .status.output.image empty (only digest is set).
+# Fall back to the registry path we asked the build to push to.
+if [ -z "${OUTPUT_IMAGE}" ]; then
+    OUTPUT_IMAGE="${REGISTRY}/${TOOL_NAME}:${IMAGE_TAG}"
```
cwiklik, line 137: Nit: this fallback uses the mutable tag, but `.status.output.digest` is populated in exactly the case you're handling (image empty, digest set). Pinning to `${REGISTRY}/${TOOL_NAME}@${OUTPUT_DIGEST}` would give an immutable/reproducible reference. Functionally fine as-is since the Deployment is `imagePullPolicy: Always` — non-blocking.

cwiklik (approved):
Four low-risk, OpenShift-only fixes for the weather-agent CLI deploy — all verified against the PR branch:

- **Route timeouts** added to the correct Routes — confirmed `weather-service` (74) and `kagenti-ui` + `kagenti-api` (ui.yaml); valid `haproxy.router.openshift.io/timeout` annotation.
- **`buildah-insecure` → `buildah`** (75): sound — `buildah-insecure` isn't present on these clusters (`ClusterBuildStrategyNotFound`) and `buildah` is the standard Shipwright `ClusterBuildStrategy`.
- **Empty `output.image` fallback** (75): the fallback value matches the Build's `output.image`, and the Deployment is `imagePullPolicy: Always`.
- **`UV_CACHE_DIR`/`HOME` → `/tmp`** (75): verified the `/tmp` `emptyDir` actually exists in the same Deployment (`volumeMounts: tmp→/tmp` + `volumes: tmp: emptyDir: {}`), so the redirect targets a genuinely writable mount under the arbitrary non-root UID.

Kind **Deploy & Test passed**, confirming these OpenShift-only paths don't regress Kind (as the PR states); Helm lint + chart-tests green; E2E HyperShift gated on maintainer `/run-e2e`.

One nit inline. Minor body-accuracy note: "Aligned with the agent build (buildah)" doesn't quite match script `74` (it does no buildah build) — but the primary justification (`buildah-insecure` missing) fully stands, so the change is correct regardless.

**Areas reviewed**: Shell, Helm. **Commits**: 1, signed-off: yes. **CI**: passing (Deploy & Test green; E2E HyperShift gated).

.github/scripts/kagenti-operator/75-deploy-weather-tool-shipwright.sh
pdettori, line 137: nit (optional): the fallback uses the mutable tag, but `OUTPUT_DIGEST` is fetched on the line right above and the empty-image case is exactly when the digest *is* populated. `${REGISTRY}/${TOOL_NAME}@${OUTPUT_DIGEST}` would pin immutably. With `imagePullPolicy: Always` the tag works fine, so this is purely a robustness nicety — no change required.

pdettori (approved):
Clean, well-scoped OpenShift-only fix for the weather-agent CLI demo path. All five changes are independently justified, low-risk, and don't touch Kind paths.

Verified against `upstream/main`:
- The `buildah` strategy genuinely aligns with the agent build (`weather_agent_shipwright_build_ocp.yaml` uses `strategy.name: buildah`, `kind: ClusterBuildStrategy`).
- The `/tmp` emptyDir backing `UV_CACHE_DIR`/`HOME` is mounted in the tool Deployment (volumeMounts + volumes already present).
- Both `kagenti-ui` and `kagenti-api` Routes are correctly targeted by the 120s timeout annotation.

The inline comments in the diff are excellent — each explains the *why*, which makes these OpenShift quirks much easier to maintain. One optional nit below; not blocking.

**Author**: webchang (MEMBER — maintainer) | **Commits**: 1, signed-off | **CI**: passing (E2E HyperShift awaiting `/run-e2e`)

Assisted-By: Claude Code

PR #2112, opened 2026-06-29
Chore: bump operator chart to 0.3.0-alpha.6 and authbridge extensions to v0.6.0-alpha.10

cwiklik:
## Summary

Dependency bump for the 0.7 line — approved by @huang195 (operator) and @pdettori.

- **`kagenti-operator-chart` `0.3.0-alpha.5 → 0.3.0-alpha.6`** — adds the `namespaces: patch` RBAC the AgentRuntime Istio auto-labeling (`ensureIstioMeshLabels`) needs. This is the operator-side fix for the token-exchange E2E `client_not_found` cascade (#2098): on alpha.5 the controller-manager couldn't patch the agent namespace → `tx-e2e` workloads CrashLoopBackOff → per-workload client-registration never ran. alpha.6 also carries TLS-bridge Phase 2, which is **off by default** (`tlsBridgeMode` gated).
- **operator binary image tag `0.3.0-alpha.5 → 0.3.0-alpha.6`** (`charts/kagenti/values.yaml`) — the chart-version bump alone left the controller-manager *image* pinned to alpha.5, producing a chart/binary skew that crash-looped the operator and timed out the platform install. This aligns the binary with the chart.
- **kagenti-extensions `v0.6.0-alpha.9 → v0.6.0-alpha.10`** (`authbridge`, `authbridge-envoy`, `authbridge-lite`, `proxy-init`) — security-audit remediations, IBAC intent-eviction / MCP rejection-frame fixes, iptables backend detection.
- Regenerate `Chart.lock`.

## Validation

`helm lint` passes. Kind `release-validation.yaml` (run 28447999840): **platform install ✅**, **backend E2E ✅** — the skew fix resolves the install timeout.

**#2098 not yet verified end-to-end.** The token-exchange suite is currently blocked *before* it runs by a separate, pre-existing 0.7 issue: the Keycloak bootstrap operand (`charts/kagenti-deps/`) deploys a `postgres-kc` StatefulSet that collides with the one the token-exchange harness (`.github/scripts/token-exchange/35-install-keycloak-ocp.sh`) applies. That collision is unrelated to this bump (no `charts/kagenti-deps/` changes here) and will be fixed in a follow-up, where the namespaces:patch fix can finally be confirmed.

Refs #2098 (kept open until token-exchange E2E passes)

Assisted-By: Claude Code

Issue #2098, opened 2026-06-26
ci(e2e): token-exchange tests fail on Kind — client_not_found (tx-e2e workloads CrashLoopBackOff)


cwiklik:
## Summary

On the `release-validation.yaml` **Kind** leg, the token-exchange E2E suite (`.github/scripts/token-exchange/80-run-tests.sh`) fails: **14 failed, 12 passed, 2 skipped**. Keycloak rejects auth in realm `tx-e2e` with **`error="client_not_found"`** for both `client_credentials` and `token-exchange` grants.

This was **surfaced by #2092 / #2097**: previously the suite died during setup with `exit 127` (`oc` not found) and the tests never ran on Kind. With that setup bug fixed, the suite now executes and reveals this pre-existing failure. **Not caused by #2097** — that change only affected Keycloak host resolution; a host/issuer mismatch would surface as `invalid_issuer`/`unauthorized`, not `client_not_found`.

## Evidence (run [28240918869](https://github.com/kagenti/kagenti/actions/runs/28240918869), Kind E2E)

- Test summary: `14 failed, 12 passed, 2 skipped`.
- Keycloak events (realm `tx-e2e`): `type="CLIENT_LOGIN_ERROR" error="client_not_found"` (×7, `grant_type=client_credentials`) and `type="TOKEN_EXCHANGE_ERROR" error="client_not_found"` (×6, `grant_type=urn:ietf:params:oauth:grant-type:token-exchange"`).
- Workload pods not ready:
  - `tx-e2e-agent-*` and `tx-e2e-tool-*` in **CrashLoopBackOff / BackOff** and **Unhealthy (Readiness)** for 8m+.
- Operator: AgentRuntime controller logs `Failed to ensure Istio mesh labels` for the `tx-e2e` namespace.
- Test env: `KAGENTI_E2E_CLIENT_ID=kagenti-e2e-tests`.

## Likely root cause (to investigate)

The `tx-e2e` agent/tool workloads never become ready (CrashLoopBackOff + AgentRuntime/Istio mesh-label reconcile failure), so the operator's client-registration never creates their OAuth clients in the `tx-e2e` realm → token-exchange and client_credentials logins get `client_not_found`. Also verify the test client `kagenti-e2e-tests` is actually registered in `tx-e2e` (vs. only in the default realm).

Investigation starting points:
1. Why `tx-e2e-agent`/`tx-e2e-tool` pods crashloop (logs / image / config).
[...]

PR #2125, opened 2026-06-30
Fix: reuse platform Keycloak for token-exchange E2E on Kind

cwiklik:
## Summary

The release-validation **Kind** leg ran `35-install-keycloak-ocp.sh community`, which the script's own header documents as **OpenShift-only** ("On Kind, Keycloak is installed by setup-kagenti.sh"). On Kind it stood up a *second* Keycloak (community operator + `Keycloak` CR + its own postgres) in the **same `keycloak` namespace** the platform installer already populated, causing a StatefulSet immutable-spec / postgres collision. The `Setup Keycloak for token exchange` step aborted **before the suite ran**, blocking the #2098 token-exchange E2E verification.

## Fix — reuse the platform Keycloak on Kind

- **`charts/kagenti-deps/templates/keycloak-k8s.yaml`** — add `token-exchange` + `admin-fine-grained-authz:v1` to the Kind platform Keycloak `KC_FEATURES` (was `client-auth-federated:v1,spiffe:v1`). The suite's realm/FGAP setup needs them; the platform image (`quay.io/keycloak/keycloak:26.5.2`) supports them.
- **`.github/workflows/release-validation.yaml`** (Kind leg) — drop `35-install-keycloak-ocp.sh community`; run the realm/client setup (`40-setup-keycloak-realm.sh` …) against the platform Keycloak. Those scripts only need a running Keycloak + the `keycloak-initial-admin` admin token, both provided by the platform.

## Scope / follow-up

Kept to the **Kind** leg (the one validated by `release-validation.yaml`). The **OCP/HyperShift** legs also run `35` and the operand Keycloak (`keycloak-operand.yaml`) likewise has no `features:` block, so they share the same latent collision — to be addressed in a follow-up once it can be validated on a HyperShift run.

## Validation

`helm lint` + `helm template` confirm the rendered `KC_FEATURES`. Functional verification via the Kind `release-validation.yaml` run triggered on this branch — expecting the token-exchange suite to now run to completion (and the #2098 `client_not_found` cascade to clear, now that the operator alpha.6 namespaces:patch fix from #2112 is in place).

Refs #2098

Assisted-By: Claude Code

### DIFF
--- a/.github/workflows/release-validation.yaml
+++ b/.github/workflows/release-validation.yaml
@@ -199,7 +199,10 @@ jobs:
       - name: Setup Keycloak for token exchange
         if: success()
         run: |
-          bash .github/scripts/token-exchange/35-install-keycloak-ocp.sh community
+          # On Kind, Keycloak is already installed by the platform installer
+          # (35-install-keycloak-ocp.sh is OpenShift-only and would stand up a
+          # second, colliding Keycloak). Reuse the platform Keycloak — its
+          # KC_FEATURES now include token-exchange + admin-fine-grained-authz.
           bash .github/scripts/token-exchange/40-setup-keycloak-realm.sh
           bash .github/scripts/token-exchange/50-enable-kagenti-ns.sh
           bash .github/scripts/token-exchange/55-setup-spiffe-idp.sh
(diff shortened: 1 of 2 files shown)


cwiklik:
Superseded by the same change on a kagenti/kagenti branch (https://github.com/kagenti/kagenti/pull/2126) so the modified release-validation workflow can be dispatched for validation (workflow_dispatch can't run a fork PR's workflow).

PR #2126, opened 2026-06-30
Fix: reuse platform Keycloak for token-exchange E2E on Kind

cwiklik:
## Summary

Unblocks and greens the **token-exchange E2E** (`release-validation.yaml` Kind leg), which was failing before its tests could even run. This is what verifies the #2098 `client_not_found` fix end-to-end (the operator alpha.6 `namespaces:patch` RBAC shipped via #2112).

Getting the suite to run surfaced a chain of pre-existing issues, fixed here in order of discovery.

### 1. Keycloak collision — reuse the platform Keycloak on Kind
The Kind leg ran `35-install-keycloak-ocp.sh community`, which the script's own header documents as **OpenShift-only** ("On Kind, Keycloak is installed by setup-kagenti.sh"). On Kind it stood up a *second* Keycloak (community operator + `Keycloak` CR + its own postgres) in the **same `keycloak` namespace** the platform installer already populated → StatefulSet immutable-spec / postgres collision → `Setup Keycloak` aborted before the suite ran.

- **`charts/kagenti-deps/templates/keycloak-k8s.yaml`** — add `token-exchange` + `admin-fine-grained-authz:v1` to the Kind platform Keycloak `KC_FEATURES` (was `client-auth-federated:v1,spiffe:v1`). The suite's realm/FGAP setup needs them; the platform image (`quay.io/keycloak/keycloak:26.5.2`) supports them.
- **`.github/workflows/release-validation.yaml`** (Kind leg) — drop `35-install-keycloak-ocp.sh community`; run the realm/client setup (`40-…` onward) against the platform Keycloak (those scripts only need a running Keycloak + the `keycloak-initial-admin` token, both provided by the platform).

### 2. Test-harness correctness (`kagenti/tests/e2e/token_exchange/test_token_exchange.py`)
Pre-existing bugs that only surfaced once the suite ran far enough:
- `kubectl exec` was invoked with a `-l` label selector, which it does not accept (`unknown shorthand flag: 'l'`). The 3 exec call sites now resolve a **Running** pod via `kubectl get pod -l app=X --field-selector=status.phase=Running` instead. (The `kubectl get`/`logs -l` sites are valid and unchanged.) Resolving a *Running* pod also avoids landing on a still-initializing pod during the AuthBridge webhook's rollout.
- `test_spiffe_jwt_svid_present` now polls `/opt/jwt_svid.token` for up to 90s, since spiffe-helper writes the SVID to the shared `/opt` volume asynchronously after the pod is Ready. (Verified against a live cluster: agent pod containers are `[agent, envoy-proxy, spiffe-helper]` and `/opt` is shared between spiffe-helper writer and envoy-proxy reader — so the tests' container names and SVID path are correct.)

### 3. `xfail` two tests pending a separate pod-readiness fix → #2129
`test_spiffe_jwt_svid_present` and `test_invalid_token_rejected` both require `kubectl exec` into a **Running** `tx-e2e-agent` pod, which isn't reliably Ready at test time in CI — the `Wait for token exchange workloads` step uses `... || true` on `rollout status` and waits on the credential *secret* count, not pod readiness. Marked `@pytest.mark.xfail(strict=False)` pointing at **#2129**, which tracks the real fix (e.g. `kubectl wait --for=condition=Ready`). These are unrelated to #2098 and to the Keycloak collision.

## Validation

Kind `release-validation.yaml` on this branch (run [28478512307](https://github.com/kagenti/kagenti/actions/runs/28478512307)) — **green**: platform install ✅, backend E2E ✅, `Setup Keycloak` ✅, **token-exchange E2E ✅**, UI E2E ✅. The token-exchange suite runs to completion; the #2098 `client_not_found` chain (password grants, operator-managed credentials, sidecar injection, admin role, inbound JWT validation) all pass.

## Follow-ups
- **#2129** — `tx-e2e-agent` pod not reliably Ready at test time (the two `xfail`'d tests + the weak `Wait` step).
- OCP/HyperShift legs still run `35`, and the operand Keycloak (`keycloak-operand.yaml`) likewise has no `features:` block — same collision class, to be addressed once validatable on a HyperShift run.

Refs #2098

Assisted-By: Claude Code

### DIFF
--- a/kagenti/tests/e2e/token_exchange/test_token_exchange.py
+++ b/kagenti/tests/e2e/token_exchange/test_token_exchange.py
@@ -19,6 +19,7 @@ import base64
 import json
 import os
 import subprocess
+import time
 
 import pytest
 
@@ -32,6 +33,34 @@ from .conftest import (
     TX_REALM,
     _decode_jwt,
 )
+
+
+def _get_running_pod(deploy):
+    """Return the name of a Running pod for app=<deploy> in TX_NAMESPACE.
+
+    kubectl exec needs a concrete pod. The AuthBridge mutating webhook rolls
+    the deployment (briefly producing multiple ReplicaSets/pods), so selecting
+    by label and filtering to status.phase=Running avoids landing on a pod that
+    is still initializing.
+    """
+    result = subprocess.run(
+        [
+            "kubectl",
+            "get",
+            "pod",
+            "-n",
+            TX_NAMESPACE,
+            "-l",
+            f"app={deploy}",
+            "--field-selector=status.phase=Running",
+            "-o",
+            "jsonpath={.items[0].metadata.name}",
+        ],
+        capture_output=True,
+        text=True,
+        timeout=30,
+    )
+    return result.stdout.strip()
 
 
 # ---------------------------------------------------------------------------
@@ -377,16 +406,15 @@ class TestSpiffeTokenExchange:
     """Test token exchange using SPIFFE JWT-SVID authentication."""
 
     def _exec_in_pod(self, deploy, container, cmd):
-        """Execute command in a pod."""
+        """Execute command in a Running pod selected by app label."""
+        pod = _get_running_pod(deploy)
         result = subprocess.run(
             [
                 "kubectl",
                 "exec",
                 "-n",
                 TX_NAMESPACE,
-                # kubectl exec takes a pod/resource, not a -l label selector;
-                # deploy/<name> resolves to a pod in the deployment.
-                f"deploy/{deploy}",
+                pod,
                 "-c",
                 container,
                 "--",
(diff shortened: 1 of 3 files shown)


kagenti/tests/e2e/token_exchange/test_token_exchange.py
```diff
@@ -34,6 +35,34 @@
...
+            "-l",
+            f"app={deploy}",
+            "--field-selector=status.phase=Running",
+            "-o",
+            "jsonpath={.items[0].metadata.name}",
+        ],
+        capture_output=True,
+        text=True,
+        timeout=30,
+    )
```
pdettori, line 62: **suggestion**: `_get_running_pod` silently returns an empty string when `kubectl get pod` fails (non-zero returncode) or finds no Running pods (empty jsonpath output). The subsequent `kubectl exec "" -c ...` will fail with a confusing error.

Consider asserting the result is non-empty before returning:
```python
pod = result.stdout.strip()
assert pod, f"No Running pod found for app={deploy} in {TX_NAMESPACE}"
return pod
```

Non-blocking since these tests are `xfail` pending #2129 anyway.

pdettori (approved):
Clean PR — fixes a real CI blocker (duplicate Keycloak on Kind) and improves E2E test robustness. Helm change is additive-only, CI workflow correctly scopes the OCP-only script removal to the Kind leg, and the kubectl exec fix is correct (`-l` is not supported by `exec`). All 4 commits signed-off, CI green.

One non-blocking suggestion below.

PR #2127, opened 2026-06-30
Fix: Allow user to supply additional container build flags for Podman and preload curl on kind

esnible:
## Summary

- _kagenti/examples/app-demo/Makefile_ didn't work with Podman on Kind because the images were not available for loading.
- Preload `curlimages/curl:latest` and don't check for it to avoid a dependency on the rate-limited Docker hub

## Related issue(s)

For #1468

Install Rosso on Kind, then

```bash
cd kagenti/examples/app-demo
DOCKER_BUILD_FLAGS=--load make
```

### DIFF
--- a/kagenti/examples/app-demo/Makefile
+++ b/kagenti/examples/app-demo/Makefile
@@ -7,6 +7,8 @@ IMAGE_PREFIX ?= localhost/
 
 # --- Full workflow: build, load, setup keycloak, deploy ---
 
+# Set `DOCKER_BUILD_FLAGS=--load` if using Podman
+
 all: build kind-load keycloak-setup deploy
 	@echo ""
 	@echo "=========================================="
@@ -45,13 +47,13 @@ dev-frontend:
 # --- Docker builds ---
 
 build-backend:
-	docker build -t $(IMAGE_PREFIX)app-demo-backend:latest -f backend/Dockerfile backend/
+	docker build -t $(IMAGE_PREFIX)app-demo-backend:latest -f backend/Dockerfile backend/ $(DOCKER_BUILD_FLAGS)
 
 build-frontend:
-	docker build -t $(IMAGE_PREFIX)app-demo-frontend:latest -f frontend/Dockerfile frontend/
+	docker build -t $(IMAGE_PREFIX)app-demo-frontend:latest -f frontend/Dockerfile frontend/ $(DOCKER_BUILD_FLAGS)
 
 build-keycloak:
-	docker build -t $(IMAGE_PREFIX)app-demo-keycloak:latest -f keycloak/Dockerfile keycloak/
+	docker build -t $(IMAGE_PREFIX)app-demo-keycloak:latest -f keycloak/Dockerfile keycloak/ $(DOCKER_BUILD_FLAGS)
 
 build: build-backend build-frontend build-keycloak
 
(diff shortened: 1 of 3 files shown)


davidhadas (changes requested):
Try this instead:
Use  `curlimages/curl:8.13.0` in both yamls, drop `imagePullPolicy: IfNotPresent` and drop docker load of the image. 
This apparently should solve the problem and is tested on mine

np with adding DOCKER_BUILD_FLAGS

esnible:
Retested with your suggestions.  Without the

> imagePullPolicy: IfNotPresent

the `make` times out.  The pod fails:

```
oc -n kagenti-system describe pod app-demo-keycloak-client-t6fnf
...
  Normal   Pulling    35s (x3 over 109m)  kubelet            Pulling image "curlimages/curl:8.13.0"
  Warning  Failed     35s (x3 over 109m)  kubelet            Failed to pull image "curlimages/curl:8.13.0": failed to pull and unpack image "docker.io/curlimages/curl:8.13.0": failed to resolve reference "docker.io/curlimages/curl:8.13.0": unexpected status from HEAD request to https://registry-1.docker.io/v2/curlimages/curl/manifests/8.13.0: 429 Too Many Requests
  Warning  Failed     35s (x3 over 109m)  kubelet            Error: ErrImagePull
  Normal   BackOff    7s (x4 over 109m)   kubelet            Back-off pulling image "curlimages/curl:8.13.0"
```

Using `IfNotPresent` is something nearly everyone here using Kind does.  The reason is that _docker.io_ severely rate limits anonymous pulls.  Using `IfNotPresent` allows developers to `docker pull` and `kind load` to avoid this problem.

davidhadas (commented):
The kind load you are proposing breaks on my machine

The `kind load docker-image curlimages/curl:latest` will break `make` on machines that don't have the image in their local store. Make it non-fatal with `|| true` so it's a best-effort pre-load rather than a hard requirement:

```makefile
kind load docker-image curlimages/curl:latest --name $(CLUSTER_NAME) 2>/dev/null || true
```

This way:
- If you have it locally → loaded into Kind, `IfNotPresent` uses it ✓
- If you don't have it locally → `kind load` silently skips, Kubernetes falls back to pulling from Docker Hub ✓

davidhadas:
Tested the `|| true` approach — it works. The change needed in your PR is:

```makefile
kind load docker-image curlimages/curl:latest --name $(CLUSTER_NAME) 2>/dev/null || true
```

pdettori:
## Review Comment Resolution Status

| Feedback | Addressed? |
|----------|-----------|
| Pin curl image to `8.13.0` | ✅ Yes |
| Add `DOCKER_BUILD_FLAGS` | ✅ Yes |
| Make `kind load` non-fatal (`|| true`) | ✅ Yes |
| Drop `imagePullPolicy: IfNotPresent` | ❌ Kept (author rationale: Docker Hub 429 rate limits on Kind) |
| Drop docker load entirely | ❌ Kept as best-effort with `|| true` |

The `|| true` approach was [confirmed working](https://github.com/kagenti/kagenti/pull/2127#issuecomment-4855982884) by @davidhadas. The remaining point is whether `imagePullPolicy: IfNotPresent` should stay — the author demonstrated that without it, Kind environments hit Docker Hub rate limits (429 Too Many Requests).

CI is green on all completed checks.

davidhadas:
lgtm

davidhadas (approved):

PR #2137, opened 2026-07-01
Feat: 💬 chat falls back to message/send for non-streaming agents

moonlight16:
## Summary

`kagenti chat` unconditionally called the agent's `message/stream` A2A method. Agents that declare `AgentCapabilities.streaming=false` in their agent card correctly reject this and return "Streaming is not supported by the agent" — the CLI silently swallowed the error and returned no output.

This PR fetches the agent card first and branches on the streaming capability:

- `streaming=true`  → existing SSE-based `runChatStream` path (unchanged behavior)
- `streaming=false` → new `runChatSend` path using `/send`
- card fetch fails → log warning to stderr, default to streaming (preserves prior behavior)

## Changes

- `kagenti/tui/internal/cli/chat.go` — split the chat body into `runChatStream` and `runChatSend`, added the card-driven branch.
- `kagenti/tui/internal/cli/cli_test.go` — added `/agent-card` handler to the shared test mux (defaults to `streaming=true` so `TestChatStreaming` is unchanged). Added two new tests:
  - `TestChatNonStreamingAgentFallsBackToSend` — card says `streaming=false`, asserts `/stream` is NOT called and `/send` IS called.
  - `TestChatStreamingAgentUsesStream` — inverse assertion.

## Testing

- Unit: `cd kagenti/tui && go test ./internal/cli -run 'TestChat' -v` — 3 passed.
- Full CLI suite: `go test ./internal/cli -v` — 20 passed.
- Live cluster: verified against an A2A agent whose card declares `streaming=false`. Baseline CLI (before this PR) silently exits with no output; fixed CLI prints the plain reply.

## Related

Complements a matching fix in the `agent-examples` repo that adds streaming support to `minimal_pvc_agent`. Either fix alone unblocks the CLI for this agent; both together cover the general case for both directions.

Assisted-By: Claude (Anthropic AI)

### DIFF
--- a/kagenti/tui/internal/cli/chat.go
+++ b/kagenti/tui/internal/cli/chat.go
@@ -28,45 +28,23 @@ func newChatCmd(ctx *CLIContext) *cobra.Command {
 				SessionID: sessionID,
 			}
 
-			// Try streaming first.
-			ch, err := ctx.Client.StreamChat(ns, agent, chatReq)
-			if err == nil {
-				var returnedSession string
-				for evt := range ch {
-					if evt.Error != "" {
-						return fmt.Errorf("stream error: %s", evt.Error)
-					}
-					if evt.SessionID != "" {
-						returnedSession = evt.SessionID
-					}
-					if evt.Content != "" {
-						fmt.Print(evt.Content)
-					}
-					if evt.Done {
-						fmt.Println()
-						if returnedSession != "" {
-							fmt.Fprintf(os.Stderr, "session-id: %s\n", returnedSession)
-						}
-						return nil
-					}
-				}
-				fmt.Println()
-				if returnedSession != "" {
-					fmt.Fprintf(os.Stderr, "session-id: %s\n", returnedSession)
-				}
-				return nil
+			// Consult the agent card to decide which transport to use.
+			// A2A agents that declare streaming=false only implement
+			// message/send and will 400 on message/stream, so we must
+			// pick the right endpoint up front rather than try+fallback.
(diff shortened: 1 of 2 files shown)


kagenti/tui/internal/cli/chat.go
```diff
@@ -28,45 +28,23 @@ func newChatCmd(ctx *CLIContext) *cobra.Command {
...
+				return runChatStream(ctx, ns, agent, chatReq)
 			}
-			fmt.Println(resp.Content)
-			if resp.SessionID != "" {
-				fmt.Fprintf(os.Stderr, "session-id: %s\n", resp.SessionID)
-			}
-			return nil
+			return runChatSend(ctx, ns, agent, chatReq)
 		},
 	}
```
pdettori, line 49: nit: The trailing-newline + session-ID block after the `for` loop (lines 48–53) duplicates the identical block inside the `evt.Done` early-return (lines 41–47). Both branches are needed (done-event vs channel-close-without-Done), but without a comment it reads as accidental copy-paste. A one-line `// Channel closed without Done event — flush output.` before line 48 would clarify intent.

pdettori (approved):
Clean, well-scoped PR. The approach is sound: fetch the agent card up front, branch on `card.Streaming`, and gracefully degrade to streaming if the card can't be fetched. Tests cover both positive and negative assertions. CI green.

**Areas reviewed**: Go
**Agent/IDE config**: none
**Commits**: 1, signed-off ✓
**CI**: all passing

Issue #2138, opened 2026-07-02
Doc: Document the OPA feature


davidhadas:
### Feature Description

Document how the OPA feature works and how it is installed.

### Proposed Solution

As a first step we will document how to get it to work pre-release. 

Before we release we need full documentation added.

### Want to contribute?

- [x] I would like to work on this issue.

PR #2139, opened 2026-07-02
Docs: Describing how to get OPA to work pre-relase (for testing and evaluation)

davidhadas:
Document showing how OPA can be experimented with prior to its release as part of the Kagenti system. 
On release, this document should be updated accordingly.

Part of https://github.com/kagenti/kagenti/issues/2138


pdettori:
Setting to draft until ready

mrsabath (commented):
Thanks for writing this up, David — the overall flow is sound (build bundle-service + authbridge-proxy → edit pipeline → helm upgrade → restart → verify → rollback), and the `bundle-service`/`authbridge-proxy` Dockerfile paths and the `authbridge-runtime-config` ConfigMap name all check out against source.

Two of the operational commands, though, don't work against the real deployment (details inline): the pod restart uses a label that doesn't exist in the codebase, and the log command uses the wrong container name. Both fail silently-ish (no-op / "container not found"), which is exactly the kind of thing that wastes a tester's afternoon. Flagging now so they're fixed before folks try it on their environments.

Leaving this as a **COMMENT** rather than requesting changes since you've marked it DO-NOT-MERGE pending real-environment testing — treat these as pre-flight fixes.

**Areas reviewed:** Docs (accuracy vs. kagenti / kagenti-operator / kagenti-extensions sources)
**CI:** all green · **Commit:** signed-off (DCO pass)

docs/authbridge/opa-migration-guide.md
```diff
@@ -0,0 +1,131 @@
...
+## Step 5 — Restart authbridge pods
+
+```bash
+kubectl delete pods -n team1 -l authbridge.kagenti.ai/enabled=true
+```
+
+---
+
+## Verification
+
```
mrsabath, line 132: **must-fix** — the label selector `authbridge.kagenti.ai/enabled=true` doesn't exist anywhere in the codebase. AuthBridge injection is keyed off `kagenti.io/*-inject` labels (webhook `inject.kagenti.io`; constants in `kagenti-operator/internal/webhook/injector/constants.go`). This `kubectl delete pods -l authbridge.kagenti.ai/enabled=true` matches **zero pods**, so the Step 5 restart (and the Rollback on line 136) silently does nothing — the tester believes authbridge restarted when it didn't.

Suggest selecting the actual workload, e.g. `kubectl rollout restart deploy -n team1 <agent-deploy>` or `-l kagenti.io/type=<...>`. Same fix needed on line 136.

docs/authbridge/opa-migration-guide.md
```diff
@@ -0,0 +1,131 @@
...
+kubectl get pods -n team1
+
+# Authbridge logs
+kubectl logs -n team1 -l authbridge.kagenti.ai/enabled=true -c authbridge --tail=50
+```
+
+---
+
+## Rollback
+
```
mrsabath, line 189: **must-fix** — wrong container name. The injected proxy container is `authbridge-proxy` (`AuthBridgeProxyContainerName`), not `authbridge`, so `kubectl logs ... -c authbridge` errors with `container authbridge not found`. Change `-c authbridge` → `-c authbridge-proxy`. (This line also carries the same nonexistent-label issue as line 110.)

docs/authbridge/opa-migration-guide.md
```diff
@@ -0,0 +1,131 @@
...
+## Step 3 — Update values.yaml
+
+Edit `charts/kagenti/values.yaml` and replace the `pipeline:` block with:
+
+```yaml
+  pipeline: |
+    inbound:
+      plugins:
+        - name: jwt-validation
```
mrsabath, line 83: **suggestion** — Step 3 tells the reader to *replace* the whole `pipeline:` block with a hardcoded version, but the real `charts/kagenti/values.yaml` block is templated single-source-of-truth: `issuer: "{{ .Values.keycloak.publicUrl }}/realms/{{ .Values.keycloak.realm }}"`, `keycloak_url` via `.Values.keycloak.namespace`, and `identity.type` derived from `.Values.authBridge.clientAuthType`. Wholesale replacement discards those and substitutes environment-specific hardcodes (`keycloak.localtest.me:8080`, and a `client-secret` + `client_secret_file` identity form not in the current schema). Recommend framing this as *adding the `opa` plugin* to the existing inbound/outbound lists, preserving the templated values.

docs/authbridge/opa-migration-guide.md
```diff
@@ -0,0 +1,131 @@
+# OPA Migration Guide for Kagenti (Kind)
+
+> **⚠ Pre-release:** This document shows how OPA can be experimented with prior to its release as part of the Kagenti system. On release, this document should be updated accordingly. 
+
+Configures a local Kind-based Kagenti installation to use OPA for policy-based
+authorization in AuthBridge.
+
+## Architecture
+
```
mrsabath, line 16: **nit** — trailing whitespace after the pre-release callout. Markdown Lint passes, but worth tidying.

docs/authbridge/opa-migration-guide.md
davidhadas: See https://github.com/kagenti/kagenti/issues/2150

At this point I would not change these instructions as they are they only instructions using which I was able to get the system working, once this is sorted i authbridge we need to align this documentation.

davidhadas:
Tested and read for merge.
@mrsabath

docs/authbridge/opa-migration-guide.md
```diff
@@ -0,0 +1,519 @@
...
+---
+
+## Step 5 — Restart authbridge pods
+
+AuthBridge runs as a sidecar injected into agent workloads, which carry the
+`kagenti.io/type=agent` label. Delete the pods to re-inject the sidecar with
+the new config:
+
+```bash
+kubectl delete pods -n team1 -l kagenti.io/type=agent
```
mrsabath: **must-fix** (pre-flight) — `-l kagenti.io/type=agent` no longer matches any pods. The rossoctl rename merged to `main` after your earlier fixes, and the operator now applies `rossoctl.io/type=agent` (`operator/internal/clientreg/names.go`: `LabelAgentType = "rossoctl.io/type"`). Sibling docs already migrated (`docs/authbridge/sidecar-injection.md` and `deployment-guide.md` both use `rossoctl.io/type=agent`). So `kubectl delete pods -l kagenti.io/type=agent` silently deletes nothing and no re-injection happens; the same applies to the log/verify commands further down (the other `kagenti.io/type=agent` uses).

This is really a doc-wide rename sweep rather than a one-line fix. Alongside the selector, please also update: `kagenti-system` to `rossoctl-system`; `charts/kagenti/` to `charts/rossoctl/` (the old path 404s now); `kagenti-operator-chart` to `rossoctl-operator-chart`; and `agent.kagenti.dev/v1alpha1` to `agent.rossoctl.dev/v1alpha1`. A single find/replace pass before testers run the guide should cover it.

mrsabath (commented):
Re-review of the current commit (`39a197cdb`).

The two must-fix items from my earlier review are resolved: the log command now uses `-c authbridge-proxy` (verified against `AuthBridgeProxyContainerName`), and the fictional `authbridge.kagenti.ai/enabled=true` selector is gone. Thanks for those.

One new issue has appeared since your fixes, caused by the rossoctl rename landing on `main` afterward: the doc's `kagenti.*` identifiers are now stale against a current cluster. The sharpest case is the `kagenti.io/type=agent` pod selector, which silently matches zero pods now that the operator applies `rossoctl.io/type=agent` (details inline). Since this PR is marked DO-NOT-MERGE pending real-environment testing, I'm leaving this as a COMMENT rather than requesting changes; treat it as a pre-flight fix before folks run the guide.

**Areas reviewed:** Docs (accuracy vs. rossoctl operator + flagship sources on current `main`)
**CI:** chart-tests was failing on an earlier run, but this PR touches no chart files, so it is not attributable here.
**Commit:** signed-off (DCO pass)

docs/authbridge/opa-migration-guide.md
mrsabath: Confirmed fixed in `39a197cdb` — the log command now uses `-c authbridge-proxy`, which matches the injected container name (`AuthBridgeProxyContainerName`). This one stays correct after the rename. Thanks!

docs/authbridge/opa-migration-guide.md
mrsabath: The fix here was right when you wrote it — `kagenti.io/type=agent` was the correct selector at that time. Heads-up though: the rossoctl rename has since landed on `main` and changed the label prefix to `rossoctl.io/type`, so this selector now silently matches zero pods against a current cluster. I opened a fresh must-fix on the current commit covering this plus the other `kagenti.*` references in the guide (namespace, chart path, CRD group). Not a knock on this fix, just the rename catching up with it.

docs/authbridge/opa-migration-guide.md
mrsabath: Understood, and that makes sense — if the hardcoded pipeline block is the only config that got the system working for you, keep it as-is for now. Leaving this thread open so it stays tied to #2150; once that is sorted we can switch the guide back to the templated `charts/rossoctl/values.yaml` single-source-of-truth form. No change needed here in the meantime.

docs/authbridge/opa-migration-guide.md
davidhadas: done

docs/authbridge/opa-migration-guide.md
```diff
@@ -0,0 +1,519 @@
...
+```bash
+kubectl delete -f docs/authbridge/examples/opa-team1-policy.yaml
+```
+
+This returns `team1` to the default global policy (all authenticated requests
+allowed). The Keycloak client changes from A.1 are harmless and can be left in
+place for future testing.
+
+> **Scripted equivalents.** The steps in this annex are automated by
+> `scripts/test-opa-weather.sh` (inbound, `setup`/`send`/`block-alice`/`verify`)
```
cwiklik, line 516: suggestion: these two "scripted equivalents" — `scripts/test-opa-weather.sh` and `scripts/test-opa-combined.sh` — don't exist on `main` (confirmed 404), and unlike Step 1's `bundle-service` script (which you attribute to the operator repo) there's no pointer to where they live. The manual annex above is fully self-contained, so no reader is actually blocked — but add a repo/branch/PR pointer here, or a note like "ships with the OPA release," so the shortcut isn't a dead reference for someone on a fresh `main` checkout.

cwiklik (approved):
### Summary

Purely additive, clearly-labeled **pre-release** documentation: an OPA migration guide plus an example namespace-scoped `AuthorizationPolicy` CR, part of tracking issue #2138. All CI green (Markdown Lint, Link Check, YAML Lint, CodeQL, Trivy included), DCO signed, no `.claude/`/`.vscode/` changes.

The manual walkthrough (Annex A.0–A.7) is fully self-contained, and cross-repo tooling is correctly attributed — Step 1's `bundle-service` script points at the `operator` repo, which is the right place for it. Approving.

One **non-blocking suggestion** (inline): the optional "Scripted equivalents" note references `scripts/test-opa-weather.sh` / `scripts/test-opa-combined.sh`, which don't exist on `main` and aren't attributed to a repo/branch — worth a pointer so the shortcut isn't a dead reference.

**Optional nit** (no inline): in `opa-team1-policy.yaml`, `is_blocked_host if startswith(input.host, "api.github.com")` also matches `api.github.com.evil.com`. Harmless here since it's a **deny** rule (over-blocking is fail-safe), but a reader adapting this prefix-match pattern for an *allow* decision would introduce a bypass — a one-line caveat in the comment could pre-empt that.

**Author**: davidhadas (MEMBER — maintainer)
**Areas reviewed**: Docs, YAML (example policy), Security (supply-chain scan, policy semantics)
**CI**: all green.

PR #2135, opened 2026-07-01
feat(platform): Add operator SPIFFE authentication bootstrap

Alan-Cha:
## Summary

Adds operator SPIFFE authentication bootstrap: the operator registers itself
as a Keycloak client using its SPIFFE identity (JWT-SVID) instead of static
admin credentials.

### Core feature
- Python bootstrap script (`kagenti/auth/operator-spiffe-bootstrap/`) that configures
  the SPIFFE Identity Provider in Keycloak and registers the operator client with
  `clientAuthenticatorType: federated-jwt`
- Helm post-install/upgrade Job template for the bootstrap (`operator-client-bootstrap-job.yaml`)
- Feature gated behind `kagenti-operator-chart.spiffe.operatorAuth.enabled` (default: false)

### Chart fixes found during E2E testing
- **`authbridge-config` in `kagenti-system`** — operator's client registration controller
  reads Keycloak config from this ConfigMap in its own namespace; it was only being
  created in agent namespaces (team1, team2). Added conditional creation in
  `kagenti-system` when `components.agentOperator.enabled: true`.
- **Duplicate resources → `STATUS: failed`** — `agent-namespaces.yaml` and
  `agent-namespace-resources.yaml` both rendered the same 12 resources per agent
  namespace, causing 10 "already exists" errors on every install. Stripped
  `agent-namespaces.yaml` down to Namespace objects only.
- **Helm template syntax** — removed extra `{{- end }}` statements in
  `agent-namespace-resources.yaml`

### Script updates
- `scripts/kind/setup-kagenti.sh` — adds `--enable-operator-spiffe-auth` flag; fixes helm key to `kagenti-operator-chart.spiffe.operatorAuth.enabled`
- `.github/scripts/local-setup/kind-full-test.sh` — propagates `ENABLE_OPERATOR_SPIFFE_AUTH`

### Feature flag defaults changed
- `featureFlags.authbridgeAPI`: `false` → `true` (AuthBridge tab now shown by default)
- `featureFlags.admin`: `false` → `true` (Platform Status card now shown by default)

## E2E test results

Validated across multiple fresh-cluster runs:

- ✅ Operator pod: 2/2 containers (manager + spiffe-helper)
- ✅ Helm release: `STATUS: deployed`
- ✅ Test agent registered with SPIFFE ID `spiffe://localtest.me/ns/team1/sa/test-agent`
- ✅ Keycloak token endpoint: HTTP 200 with access token

**Test materials (procedure + setup script + requirements):**
[...]

PR #2141, opened 2026-07-02
feat(operator): add SPIFFE authentication bootstrap for operator client registration

Alan-Cha:
## Summary

Enables the Kagenti operator to register itself as a Keycloak client using its
SPIFFE identity (JWT-SVID) instead of static admin credentials, eliminating
long-lived credentials from the operator's namespace.

### Core feature
- Python bootstrap script (`kagenti/auth/operator-spiffe-bootstrap/`) that configures:
  - SPIFFE Identity Provider in Keycloak
  - Operator client with `clientAuthenticatorType: federated-jwt`
  - `manage-clients` role (scoped, not full admin)
- Helm post-install/upgrade Job (`operator-client-bootstrap-job.yaml`)
- Docker image for the bootstrap job
- Feature gated behind `kagenti-operator-chart.spiffe.operatorAuth.enabled` (default: false)

### Chart fixes
- **Duplicate resources → `STATUS: failed`**: `agent-namespaces.yaml` and
  `agent-namespace-resources.yaml` both rendered the same 12 resources per agent
  namespace, causing 10 "already exists" errors on every install. Stripped
  `agent-namespaces.yaml` to Namespace objects only.
- **`authbridge-config` missing in `kagenti-system`**: The operator's client
  registration controller reads Keycloak config from this ConfigMap in its own
  namespace, but it was only created in agent namespaces. Added conditional creation
  in `kagenti-system` when `components.agentOperator.enabled: true`.

### Script updates
- `scripts/kind/setup-kagenti.sh`: adds `--enable-operator-spiffe-auth` flag;
  corrects helm key from removed `kagentiOperator.spiffeAuth.enabled` to
  `kagenti-operator-chart.spiffe.operatorAuth.enabled`

## E2E test results

Validated across multiple fresh-cluster runs:

- ✅ Operator pod: 2/2 containers (manager + spiffe-helper)
- ✅ Helm release: `STATUS: deployed`
- ✅ Test agent registered with SPIFFE ID `spiffe://localtest.me/ns/team1/sa/test-agent`
- ✅ Keycloak token endpoint: HTTP 200 with access token

**Test materials (procedure + setup script):**
https://gist.github.com/Alan-Cha/527be60c79b40aedd54fe7be586c3452

> The setup script requires a local checkout of kagenti-operator on branch
> `feat/spiffe-auth` (this PR's companion). It packages the operator helm chart
> from source at test time — the published OCI chart does not yet include the
> spiffe-helper sidecar.

## Related

- Replaces: #2135 (superseded by this clean PR)
- Depends on: kagenti-operator#349 (superseded by companion clean PR)

## Checklist

- [x] All commits signed-off (DCO)
- [x] Feature gated (default: disabled)
- [x] Backward compatible
- [x] E2E tested (multiple fresh-cluster runs)

Assisted-By: Claude Code


charts/kagenti/templates/operator-client-bootstrap-job.yaml
```diff
@@ -0,0 +1,129 @@
...
+    helm.sh/hook: post-install,post-upgrade
+    helm.sh/hook-weight: "100"
+    helm.sh/hook-delete-policy: before-hook-creation
+---
+# Role in keycloak namespace to read admin secret
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: kagenti-operator-client-bootstrap
+  namespace: keycloak
```
esnible: **must-fix**: `Role` and `RoleBinding` namespaces are hardcoded to the literal `keycloak`, but the `ServiceAccount`/`Job` and the `RoleBinding`'s subject are templated off `{{ .Values.keycloak.namespace | default "keycloak" }}`. If `keycloak.namespace` is ever overridden, the RBAC objects land in the wrong namespace and the bootstrap Job can't read the admin secret. Use the same templated expression here and in the `RoleBinding` metadata (and in the `KEYCLOAK_URL` env var below).

charts/kagenti/templates/operator-client-bootstrap-job.yaml
```diff
@@ -0,0 +1,129 @@
...
+    spec:
+      serviceAccountName: kagenti-operator-client-bootstrap
+      restartPolicy: OnFailure
+      containers:
+      - name: bootstrap
+        image: {{ (index .Values "kagenti-operator-chart" "spiffe" "operatorAuth" "bootstrapImage") | default "ghcr.io/kagenti/kagenti/operator-spiffe-bootstrap:latest" }}
+        imagePullPolicy: IfNotPresent
+        env:
+        - name: KEYCLOAK_URL
+          value: "http://keycloak-service.keycloak.svc:8080"
```
esnible: **must-fix**: `KEYCLOAK_URL` is hardcoded to `http://keycloak-service.keycloak.svc:8080`, unlike the existing `authbridge-config` ConfigMap (same file family) which correctly templates this as `http://keycloak-service.{{ $root.Values.keycloak.namespace }}.svc:8080`. Same class of bug as the RBAC namespace hardcoding — breaks if `keycloak.namespace` is overridden.

charts/kagenti/templates/operator-client-bootstrap-job.yaml
```diff
@@ -0,0 +1,129 @@
...
+    metadata:
+      labels:
+        app.kubernetes.io/name: kagenti-operator-client-bootstrap
+        app.kubernetes.io/component: bootstrap
+    spec:
+      serviceAccountName: kagenti-operator-client-bootstrap
+      restartPolicy: OnFailure
+      containers:
+      - name: bootstrap
+        image: {{ (index .Values "kagenti-operator-chart" "spiffe" "operatorAuth" "bootstrapImage") | default "ghcr.io/kagenti/kagenti/operator-spiffe-bootstrap:latest" }}
```
esnible, line 81: **suggestion**: Default bootstrap image falls back to the mutable `:latest` tag, but the README (and repo convention elsewhere in `values.yaml`, e.g. `authbridgeLite:...v0.6.0-alpha.10`) references a pinned `v0.1.0`. Consider defaulting to the pinned version tag instead.

kagenti/auth/operator-spiffe-bootstrap/setup_operator_client.py
```diff
@@ -0,0 +1,398 @@
...
+
+    def __init__(self):
+        self.keycloak_url = KEYCLOAK_URL.rstrip("/")
+        self.realm = KEYCLOAK_REALM
+        self.admin_username: Optional[str] = None
+        self.admin_password: Optional[str] = None
+        self.token: Optional[str] = None
+        self.session = requests.Session()
+        # Disable SSL verification for in-cluster communication
+        self.session.verify = False
```
esnible, line 76: **suggestion**: `self.session.verify = False` disables TLS verification for the whole session with no `urllib3.disable_warnings()` or comment justifying it. Currently inert since `KEYCLOAK_URL` defaults to `http://`, but it's a latent footgun if the URL is ever switched to `https://` — it would silently skip cert validation. Consider removing or gating behind an explicit, loudly-logged opt-in.

kagenti/auth/operator-spiffe-bootstrap/README.md
```diff
@@ -0,0 +1,125 @@
...
+   - Client Authenticator: `federated-jwt`
+   - Links to SPIFFE IdP for JWT-SVID validation
+
+3. **Assigns manage-clients Role** to operator's service account
+   - Scoped permission (not full admin)
+   - Allows operator to register agent clients dynamically
+
+## Usage
+
+Automatically deployed when `kagentiOperator.spiffeAuth.enabled=true` in Helm values.
```
esnible: **suggestion**: README documents enabling the feature via `kagentiOperator.spiffeAuth.enabled=true`, but the actual implemented value path (per the Job template's `if` guard and `setup-kagenti.sh`) is `kagenti-operator-chart.spiffe.operatorAuth.enabled`. This will mislead anyone trying to enable the feature directly via `helm install --set`. Same issue repeats further down this doc.

esnible (changes requested):
This PR adds a Keycloak bootstrap Job that lets the operator authenticate via SPIFFE JWT-SVID instead of static admin credentials, correctly gated off by default, and also fixes a real duplicate-resource bug in the agent-namespace templates (verified: `agent-namespace-resources.yaml` is a faithful, non-lossy split of what was removed from `agent-namespaces.yaml`). The core logic (RBAC scoping to a single named secret, Helm hook idempotency via `before-hook-creation`, resource limits, non-root pinned-digest Dockerfile) is solid.

The main issue: the new Job's `Role`/`RoleBinding` namespace and its `KEYCLOAK_URL` env var are hardcoded to `keycloak` instead of following the same `{{ .Values.keycloak.namespace }}` template used everywhere else in this file (see inline comments) — this will silently break the feature for any deployment that overrides that value. One documentation inconsistency (README references a Helm value path that doesn't match what's implemented) should also be fixed before merge.

A couple of non-blocking notes not tied to a specific line:
- `setup_operator_client.py` imports `Dict`/`Any` from `typing` but never uses them (nit).
- No unit tests for `setup_operator_client.py` — given it does multi-step, stateful Keycloak API orchestration with idempotency checks, mocked-`requests` tests for the exists-vs-create branches would be valuable follow-up coverage (suggestion).
- The vendored `kagenti-operator-chart-0.3.0-alpha.6.tgz` subchart doesn't yet define the `spiffe.operatorAuth.enabled` key this PR gates on — harmless today (absent key = falsy = feature stays off), but the feature can't actually be exercised until the companion `kagenti-operator` chart bump lands, consistent with the PR description's note about the companion PR.

**Author**: Alan-Cha (MEMBER — maintainer)
**Areas reviewed**: Python, Helm/K8s, Dockerfile, Shell, Docs
**Commits**: 6 commits, all signed-off: yes
**CI status**: all required checks passing; `E2E HyperShift` pending manual `/run-e2e` trigger (not blocking)

charts/kagenti/templates/operator-client-bootstrap-job.yaml
Alan-Cha: Fixed in commit `3016ff46` — `Role` and `RoleBinding` `metadata.namespace` now use `{{ .Values.keycloak.namespace | default "keycloak" }}`, matching the ServiceAccount and Job.

charts/kagenti/templates/operator-client-bootstrap-job.yaml
Alan-Cha: Fixed in commit `3016ff46` — `KEYCLOAK_URL` and `KEYCLOAK_ADMIN_SECRET_NAMESPACE` now use `{{ .Values.keycloak.namespace | default "keycloak" }}` instead of the hardcoded literal.

kagenti/auth/operator-spiffe-bootstrap/README.md
Alan-Cha: Fixed in commit `3016ff46` — README now references `kagenti-operator-chart.spiffe.operatorAuth.enabled` (the actual implemented path) throughout.

Alan-Cha:
Thanks for the thorough review @esnible!

**Addressed in commit `3016ff46`:**
- ✅ **must-fix** — `Role`/`RoleBinding` namespace: now uses `{{ .Values.keycloak.namespace | default "keycloak" }}`
- ✅ **must-fix** — `KEYCLOAK_URL` and `KEYCLOAK_ADMIN_SECRET_NAMESPACE`: same templated value
- ✅ **suggestion** — README: corrected all references from `kagentiOperator.spiffeAuth.enabled` to `kagenti-operator-chart.spiffe.operatorAuth.enabled`
- ✅ **nit** — Removed unused `Dict`/`Any` imports from `setup_operator_client.py`

**Remaining open:**
- `self.session.verify = False` — acknowledged, currently inert since the URL defaults to `http://`. Will address in a follow-up once HTTPS for in-cluster Keycloak is added.
- `:latest` bootstrap image default — noted; the canonical tagged version should be pinned here, will address.
- `Dockerfile.local` — absent from this branch (was excluded during the clean-branch squash merge).
- Unit tests for `setup_operator_client.py` — tracked as follow-up work.

charts/kagenti/templates/operator-client-bootstrap-job.yaml
```diff
@@ -0,0 +1,129 @@
...
```
esnible, line 81: **suggestion (carried over):** The bootstrap image still falls back to the mutable `:latest` tag when `kagenti-operator-chart.spiffe.operatorAuth.bootstrapImage` is unset. For reproducible bootstrap runs, recommend defaulting to a released, immutable tag consistent with the `values.yaml` convention (e.g. `authbridgeLite ...v0.6.0-alpha.10`). Not blocking.

esnible (approved):
Re-review — verifying the fixes to my prior CHANGES_REQUESTED against branch `feat/operator-spiffe-auth-v2`. Both must-fix items are resolved (commit `3016ff46`) and confirmed:

- **Role/RoleBinding namespace** — now `{{ .Values.keycloak.namespace | default "keycloak" }}`, matching the ServiceAccount/Job (job template lines 22, 38). ✅
- **`KEYCLOAK_URL` + admin-secret namespace** — now templated off `keycloak.namespace` instead of the hardcoded literal (lines 82, 88). ✅

Prior suggestions also addressed: README now references the actual `kagenti-operator-chart.spiffe.operatorAuth.enabled` path; unused `Dict`/`Any` imports removed; `Dockerfile.local` (the file the security bot flagged) removed; `session.verify = False` now carries an explanatory comment and is inert for the in-cluster `http://` endpoint.

**CI note:** the failing "Deploy & Test" check is an unrelated infrastructure flake, not a defect in this PR — the `weather-tool` Shipwright BuildRun failed pulling `quay.io/containers/buildah:v1.37.5` with a truncated download (`short read: expected 60771241 bytes but got 35651584: unexpected EOF`). This PR touches none of weather-tool, Shipwright, or buildah; a re-run should clear it. All required lint/security/chart/backend checks pass; E2E HyperShift is pending a manual `/run-e2e` trigger.

One carried-over non-blocking suggestion inline (`:latest` bootstrap image fallback), and one follow-up: unit tests for `setup_operator_client.py` remain absent — worthwhile given the stateful Keycloak orchestration, but not a merge blocker.

This APPROVE supersedes my prior CHANGES_REQUESTED.

Assisted-By: Claude Code

charts/kagenti/templates/operator-client-bootstrap-job.yaml
Alan-Cha: Reverted in commit `4f9bf78b` — keeping `:latest` as a placeholder since the image hasn't been published yet. Added a comment in the template explaining this and documenting the override path.

charts/kagenti/templates/operator-client-bootstrap-job.yaml
Alan-Cha: Reverted in commit `4f9bf78b` — keeping `:latest` as a placeholder since the image hasn't been published yet. Added a comment in the template explaining this and documenting the override path.

PR #2152, opened 2026-07-07
Bug fix: "u is not iterable" on GitHub skill import

### LINKED ISSUE #1550: :bug: Import failed: u is not iterable
### Describe the bug

If a user enters the URL https://github.com/kagenti/kagenti/blob/main/.claude/skills/github%3Aissues/SKILL.md into the URL field of http://kagenti-ui.localtest.me:8080/skills/import , this message appears.

### Steps To Reproduce

See above.

### Expected Behavior

Either the skill at the URL endpoint should be loaded, or an error message should explain why.

### Additional Context

_No response_

esnible:
## Summary

Fixes the `Import failed` alert with reason `u is not iterable` on the Import Skill page when auto-importing a skill from a GitHub URL.

The GitHub Contents API only returns an array when the path is a **directory**. When the URL resolves to a **file** it returns a single object, and on errors (e.g. the 60/hour unauthenticated rate limit) it returns `{ message, ... }`. `fetchDirectoryContents` was typed as always returning `GitHubFile[]`, so the `for...of` loop in `fetchAllFiles` threw `TypeError: u is not iterable` (minified) on these responses. The error propagated to `ImportSkillPage`'s catch and rendered under the "Import failed" alert.

This was reachable in practice because `findValidBranchAndPath` validates candidates with a `HEAD` request, which returns `200 OK` for files as well as directories.

Key changes:
- Widen `fetchDirectoryContents` return type to `GitHubFile[] | GitHubFile` to reflect the real API contract.
- Guard the `for...of` in `fetchAllFiles` with `Array.isArray`: handle a single-file response gracefully, and otherwise throw a clear message directing the user to point the URL at a skill directory containing `SKILL.md`.
- Clean up pre-existing `prefer-const` lint errors in `importSkillFromGitHub` (CI lint runs with `--max-warnings 0`).

Verified locally with `npm run typecheck` and `eslint` on the changed file — both pass.

Resolves #1550

Assisted-By: Claude Code


eranra (approved):
/LGTM

@esnible thanks.

pdettori (approved):
LGTM

PR #2154, opened 2026-07-07
Docs(authbridge): add sidecar injection reference page

### LINKED ISSUE #1775: docs(authbridge): document sidecar injection modes and label model
## Problem

The Kagenti documentation does not have a unified, current page that explains:

1. **Sidecar injection modes** — what containers a user should expect to see in an agent pod, and which Kagenti mode produces which container set.
2. **The label model that controls injection** — `kagenti.io/type`, `kagenti.io/inject`, `kagenti.io/spire`, and the per-sidecar opt-outs (`kagenti.io/envoy-proxy-inject`, `kagenti.io/spiffe-helper-inject`, `kagenti.io/client-registration-inject`).
3. **How to switch modes** — and what the resulting container layout will be.

Today this information is partial and scattered:

- `docs/authbridge/deployment-guide.md` — covers `proxy-sidecar` / `envoy-sidecar` / `waypoint` modes but uses the older `kagenti.io/authbridge-mode` annotation and does not mention the SPIRE opt-out label or per-sidecar opt-outs.
- `docs/authbridge/roadmap.md`, `docs/components.md`, `docs/design-proposals/consolidated-compositional-agent-platform-design.md` — fragments of the model, design-doc tone, not user-facing.
- `charts/kagenti/values.yaml` — image tags only.

## Why this matters

Users following the weather-agent demo (kagenti-extensions PR #439 review surfaced this) see `3/3 Running` with containers `agent envoy-proxy spiffe-helper` — the demo doc currently claims `2/2` with `agent envoy-proxy`. There is no canonical page they can go to and figure out what's controlling that, or how to switch.

## What we need

A single user-facing page (refresh of `docs/authbridge/deployment-guide.md`, or a sibling) that covers:

- A "verify which mode you're in" command (`kubectl get pod ... -o jsonpath='{.items[0].spec.containers[*].name}'`).
- A table mapping mode + label combination → expected container list.
- The current label vocabulary (`kagenti.io/inject`, `kagenti.io/spire`, per-sidecar opt-out labels) with examples.
- A short pointer to cluster-admin feature gates (`globalEnabled`, `injectTools`, per-sidecar gates) for completeness, with a link to the webhook source of truth.
- A concrete "how to switch modes" section reflecting the current AgentRuntime CR + ConfigMap precedence chain.

## Out of scope (for the doc)

- Designing new modes.
- Deprecating `kagenti.io/authbridge-mode` (still honored — the doc should just reflect that).

## Related

- kagenti-extensions PR #439 — fixed a stale `team1` namespace reference; reviewer + my own verification surfaced that the container count shown in `authbridge/demos/weather-agent/demo-ui.md` is also stale (says 2/2, real install today shows 3/3 with SPIRE on).
- Source of truth for the label model: `kagenti-webhook/internal/webhook/injector/{pod_mutator.go,constants.go}` in `kagenti/kagenti-extensions`.

Alan-Cha:
## Summary

- Adds `docs/authbridge/sidecar-injection.md` — a new user-facing reference page covering the AuthBridge injection model
- Cross-references added in `docs/authbridge/README.md` and `docs/authbridge/deployment-guide.md`

Closes #1775

## What's in the new page

- **Verify command** — `kubectl get pods -o jsonpath` to list container names in running pods
- **Container table** — maps each mode (`proxy-sidecar`, `lite`, `envoy-sidecar`, `waypoint`) to expected containers, with a callout that spiffe-helper is a bundled process (not a separate container)
- **Label vocabulary** — pre-filter labels, per-sidecar opt-out labels, and the deprecated `kagenti.io/authbridge-mode` annotation, each with YAML examples
- **How to switch modes** — full precedence chain (AgentRuntime CR → namespace ConfigMap → deprecated annotation → cluster default) with YAML for each layer
- **Feature gates table** — `globalEnabled`, `envoyProxy`, `injectTools`, `perWorkloadConfigResolution` with defaults and link to source
- **Decision flow diagram** — complete injection logic as an ASCII tree

## Test plan

- [ ] Verify `docs/authbridge/sidecar-injection.md` renders correctly on GitHub
- [ ] Verify links in `README.md` and `deployment-guide.md` resolve correctly

Assisted-By: Claude Code

### DIFF
--- a/docs/authbridge/deployment-guide.md
+++ b/docs/authbridge/deployment-guide.md
@@ -13,10 +13,12 @@ deployment shape (kagenti-extensions#411):
 | **Plugins included** | jwt-validation + token-exchange + a2a/mcp/inference parsers | Same as proxy-sidecar | jwt-validation + token-exchange only |
 | **Use when** | Standard deployments | Need transparent interception of non-HTTP protocols | Size-constrained / no protocol-aware events |
 
-`spiffe-helper` is bundled inside every combined image and runs only when the
-operator sets `SPIRE_ENABLED=true` on the workload (driven by the
-`kagenti.io/spire: enabled` label). Client registration is handled by the
-operator's reconciler — there's no in-pod client-registration sidecar.
+SPIRE integration is built into every combined image using the
+[go-spiffe](https://github.com/spiffe/go-spiffe) SDK. It is enabled or disabled
+per workload via the `SPIRE_ENABLED` env var (driven by the
+`kagenti.io/spiffe-helper-inject` label) — there is no separate spiffe-helper
+binary or process. Client registration is handled by the operator's reconciler —
+there's no in-pod client-registration sidecar.
 
 ## Selecting a mode
 
@@ -42,8 +44,9 @@ spec:
     name: weather-service
 ```
 
-The deprecated annotation is retained for tools and other workloads that
-don't have their own AgentRuntime CR yet:
+The deprecated annotation is retained for backward compatibility. Use the
+namespace `authbridge-runtime-config` ConfigMap instead — it is the canonical
+way to set mode and applies to all workloads in the namespace:
 
 ```yaml
 metadata:
--- a/docs/authbridge/deployment-guide.md
+++ b/docs/authbridge/deployment-guide.md
@@ -49,11 +49,14 @@ namespace `authbridge-runtime-config` ConfigMap instead — it is the canonical
 way to set mode and applies to all workloads in the namespace:
 
 ```yaml
+apiVersion: v1
+kind: ConfigMap
 metadata:
-  labels:
-    kagenti.io/type: tool
-  annotations:
-    kagenti.io/authbridge-mode: "envoy-sidecar"
+  name: authbridge-runtime-config
+  namespace: team1
+data:
+  config.yaml: |
+    mode: envoy-sidecar
 ```
 
 ## Proxy-Sidecar Mode (Default)
--- a/docs/authbridge/deployment-guide.md
+++ b/docs/authbridge/deployment-guide.md
@@ -22,31 +22,18 @@ there's no in-pod client-registration sidecar.
 
 ## Selecting a mode
 
-The operator resolves mode per workload from this chain
-(kagenti-operator#361):
+The mutating webhook resolves mode from this chain (first non-empty wins):
 
-1. `AgentRuntime.Spec.AuthBridgeMode` on the workload's CR (canonical).
-2. `mode:` field on the namespace-level `authbridge-runtime-config` ConfigMap.
-3. The deprecated `kagenti.io/authbridge-mode` pod annotation (still honored).
-4. Cluster-wide default (`proxy-sidecar`).
+1. `mode:` field in the namespace-level `authbridge-runtime-config` ConfigMap **(canonical)**.
+2. The deprecated `kagenti.io/authbridge-mode` pod annotation (still honored).
+3. Cluster-wide default (`proxy-sidecar`).
 
-```yaml
-# Canonical: per-workload override on the AgentRuntime CR
-apiVersion: kagenti.io/v1alpha1
-kind: AgentRuntime
-metadata:
-  name: weather-service
-spec:
-  authBridgeMode: envoy-sidecar
-  targetRef:
-    apiVersion: apps/v1
-    kind: Deployment
-    name: weather-service
-```
+> **Note:** `AgentRuntime.Spec.AuthBridgeMode` is enum-validated by the CRD webhook but is not
+> read by the mutating webhook when resolving injection mode. Setting it does not change which
+> containers are injected.
 
-The deprecated annotation is retained for backward compatibility. Use the
-namespace `authbridge-runtime-config` ConfigMap instead — it is the canonical
-way to set mode and applies to all workloads in the namespace:
+The canonical way to set mode is the namespace ConfigMap, which applies to all workloads in
+the namespace:
 
 ```yaml
 apiVersion: v1
@@ -246,7 +233,7 @@ kubectl exec deploy/weather-service -n team1 -c envoy-proxy -- \
 
 ### Proxy Not Injected
 
-1. Verify the pod has `kagenti.io/type: agent` label
+1. Verify the pod has `kagenti.io/type: agent` label — the operator sets this automatically when an AgentRuntime CR targets the workload
 2. Check that the Kagenti operator webhook is running:
    ```bash
    kubectl get mutatingwebhookconfigurations | grep kagenti
--- a/docs/authbridge/deployment-guide.md
+++ b/docs/authbridge/deployment-guide.md
@@ -260,5 +260,6 @@ injects defaults that can be overridden via Helm values.
 ## Further Reading
 
 - [Sidecar Injection](sidecar-injection.md) — expected containers per mode, label vocabulary, feature gates, how to switch modes
+- [Authentication Guide](../authentication.md) — how `CLIENT_AUTH_TYPE=federated-jwt` (SPIFFE auth) works, how to enable it, and how it compares to client-secret mode
 - [AuthBridge Binary README](https://github.com/kagenti/kagenti-extensions/blob/main/authbridge/cmd/README.md) — full YAML config reference, all listener modes
 - [AuthBridge Architecture](https://github.com/kagenti/kagenti-extensions/blob/main/authbridge/README.md) — sequence diagrams, protocol details
--- a/docs/authbridge/deployment-guide.md
+++ b/docs/authbridge/deployment-guide.md
@@ -42,10 +42,16 @@ spec:
     name: weather-service
 ```
 
-> **Note:** All workloads should use an AgentRuntime CR to enroll with the
-> platform. The operator applies `kagenti.io/type` labels automatically.
-> The deprecated `kagenti.io/authbridge-mode` annotation is still honored
-> but `AgentRuntime.Spec.AuthBridgeMode` is the canonical method.
+The deprecated annotation is retained for tools and other workloads that
+don't have their own AgentRuntime CR yet:
+
+```yaml
+metadata:
+  labels:
+    kagenti.io/type: tool
+  annotations:
+    kagenti.io/authbridge-mode: "envoy-sidecar"
+```
 
 ## Proxy-Sidecar Mode (Default)
 
@@ -234,7 +240,7 @@ kubectl exec deploy/weather-service -n team1 -c envoy-proxy -- \
 
 ### Proxy Not Injected
 
-1. Verify the workload has an AgentRuntime CR targeting it (the operator applies the `kagenti.io/type` label automatically)
+1. Verify the pod has `kagenti.io/type: agent` label
 2. Check that the Kagenti operator webhook is running:
    ```bash
    kubectl get mutatingwebhookconfigurations | grep kagenti
@@ -253,5 +259,6 @@ injects defaults that can be overridden via Helm values.
 
 ## Further Reading
 
+- [Sidecar Injection](sidecar-injection.md) — expected containers per mode, label vocabulary, feature gates, how to switch modes
 - [AuthBridge Binary README](https://github.com/kagenti/kagenti-extensions/blob/main/authbridge/cmd/README.md) — full YAML config reference, all listener modes
 - [AuthBridge Architecture](https://github.com/kagenti/kagenti-extensions/blob/main/authbridge/README.md) — sequence diagrams, protocol details
--- a/docs/authbridge/sidecar-injection.md
+++ b/docs/authbridge/sidecar-injection.md
@@ -20,7 +20,7 @@ kubectl get pods -n team1 -l app=weather-service \
 | Mode | Regular containers | Init containers | Notes |
 |---|---|---|---|
 | `proxy-sidecar` (default) | `<agent>` `authbridge-proxy` | — | SPIRE integration is in-process inside `authbridge-proxy` when `SPIRE_ENABLED=true` |
-| `lite` | `<agent>` `authbridge-proxy` | — | Same shape as proxy-sidecar; uses `authbridge-lite` image (auth-only) |
+| `lite` | `<agent>` `authbridge-proxy` | — | Same shape as proxy-sidecar; uses `authbridge-lite` image (auth-only plugins); SPIRE integration in-process when `SPIRE_ENABLED=true` |
 | `envoy-sidecar` | `<agent>` `envoy-proxy` | `proxy-init` | SPIRE integration is in-process inside `envoy-proxy` when `SPIRE_ENABLED=true`; proxy-init runs as root (UID 0) with NET_ADMIN/NET_RAW — not privileged |
 | `waypoint` | — | — | Not injected as a sidecar; waypoint is a standalone deployment |
 
--- a/docs/authbridge/sidecar-injection.md
+++ b/docs/authbridge/sidecar-injection.md
@@ -19,15 +19,20 @@ kubectl get pods -n team1 -l app=weather-service \
 
 | Mode | Regular containers | Init containers | Notes |
 |---|---|---|---|
-| `proxy-sidecar` (default) | `<agent>` `authbridge-proxy` | — | spiffe-helper runs as a process **inside** `authbridge-proxy` when SPIRE is on |
+| `proxy-sidecar` (default) | `<agent>` `authbridge-proxy` | — | SPIRE integration is in-process inside `authbridge-proxy` when `SPIRE_ENABLED=true` |
 | `lite` | `<agent>` `authbridge-proxy` | — | Same shape as proxy-sidecar; uses `authbridge-lite` image (auth-only) |
-| `envoy-sidecar` | `<agent>` `envoy-proxy` | `proxy-init` | spiffe-helper runs as a process **inside** `envoy-proxy` when SPIRE is on; proxy-init is privileged |
+| `envoy-sidecar` | `<agent>` `envoy-proxy` | `proxy-init` | SPIRE integration is in-process inside `envoy-proxy` when `SPIRE_ENABLED=true`; proxy-init is privileged |
 | `waypoint` | — | — | Not injected as a sidecar; waypoint is a standalone deployment |
 
-> **spiffe-helper is not a separate container.** It is bundled inside both
-> `authbridge-proxy` (proxy-sidecar/lite) and `envoy-proxy` (envoy-sidecar) and
-> starts conditionally based on the `SPIRE_ENABLED` env var, which the operator
-> sets per workload based on the `kagenti.io/spiffe-helper-inject` label.
+> **There is no spiffe-helper container or process.** AuthBridge uses the
+> [go-spiffe](https://github.com/spiffe/go-spiffe) SDK directly to open a
+> `workloadapi.JWTSource` against the SPIRE workload API socket
+> (`/spiffe-workload-api/spire-agent.sock`) and fetch JWT-SVIDs in-process.
+> The files in `/opt` (`svid.pem`, `jwt_svid.token`, etc.) are written by a
+> goroutine inside AuthBridge itself for compatibility with external file readers
+> — not by a separate binary. SPIRE integration is enabled or disabled via the
+> `SPIRE_ENABLED` env var, which the operator sets per workload based on the
+> `kagenti.io/spiffe-helper-inject` label.
 
 ## Label vocabulary
 
--- a/docs/authbridge/sidecar-injection.md
+++ b/docs/authbridge/sidecar-injection.md
@@ -21,7 +21,7 @@ kubectl get pods -n team1 -l app=weather-service \
 |---|---|---|---|
 | `proxy-sidecar` (default) | `<agent>` `authbridge-proxy` | — | SPIRE integration is in-process inside `authbridge-proxy` when `SPIRE_ENABLED=true` |
 | `lite` | `<agent>` `authbridge-proxy` | — | Same shape as proxy-sidecar; uses `authbridge-lite` image (auth-only) |
-| `envoy-sidecar` | `<agent>` `envoy-proxy` | `proxy-init` | SPIRE integration is in-process inside `envoy-proxy` when `SPIRE_ENABLED=true`; proxy-init is privileged |
+| `envoy-sidecar` | `<agent>` `envoy-proxy` | `proxy-init` | SPIRE integration is in-process inside `envoy-proxy` when `SPIRE_ENABLED=true`; proxy-init runs as root (UID 0) with NET_ADMIN/NET_RAW — not privileged |
 | `waypoint` | — | — | Not injected as a sidecar; waypoint is a standalone deployment |
 
 > **There is no spiffe-helper container or process.** AuthBridge uses the
@@ -54,7 +54,7 @@ These let you disable a specific sidecar while leaving others active.
 | Label | Default | Set to `false` to… |
 |---|---|---|
 | `kagenti.io/envoy-proxy-inject` | inject (when feature gate is on) | Disable the `envoy-proxy` sidecar (envoy-sidecar mode only) |
-| `kagenti.io/spiffe-helper-inject` | inject (enabled) | Suppress SPIRE — sets `SPIRE_ENABLED=false` on the combined sidecar, preventing the bundled spiffe-helper process from starting |
+| `kagenti.io/spiffe-helper-inject` | inject (enabled) | Suppress SPIRE — sets `SPIRE_ENABLED=false` on the sidecar, preventing the go-spiffe workload API source from being opened |
 
 ### Deprecated label
 
@@ -83,28 +83,16 @@ metadata:
 
 ## How to switch modes
 
-Mode is resolved from this chain (first non-empty wins):
+Mode is resolved by the **mutating webhook** from this chain (first non-empty wins):
 
-1. `AgentRuntime.Spec.AuthBridgeMode` on the workload's CR **(canonical)**
-2. `mode:` field in the namespace-level `authbridge-runtime-config` ConfigMap
-3. `kagenti.io/authbridge-mode` pod annotation *(deprecated — still honored)*
-4. Cluster-wide default: `proxy-sidecar`
+1. `mode:` field in the namespace-level `authbridge-runtime-config` ConfigMap
+2. `kagenti.io/authbridge-mode` pod annotation *(deprecated — still honored)*
+3. Cluster-wide default: `proxy-sidecar`
 
-### Per-workload override (AgentRuntime CR)
-
-```yaml
-apiVersion: kagenti.io/v1alpha1
-kind: AgentRuntime
-metadata:
-  name: weather-service
-  namespace: team1
-spec:
-  authBridgeMode: envoy-sidecar   # proxy-sidecar | envoy-sidecar | lite | waypoint
-  targetRef:
-    apiVersion: apps/v1
-    kind: Deployment
-    name: weather-service
-```
+> **Note on `AgentRuntime.Spec.AuthBridgeMode`:** This field exists on the CRD and is
+> enum-validated, but the mutating webhook does not read it during injection. Setting it
+> does not change which containers are injected. Use the namespace ConfigMap to control
+> mode for all workloads in a namespace.
 
 ### Namespace default (ConfigMap)
 
@@ -127,7 +115,7 @@ kubectl logs -n kagenti-system deploy/kagenti-operator \
 
 ## Cluster-admin feature gates
 
-Feature gates are loaded from the `kagenti-webhook-config` ConfigMap in the
+Feature gates are loaded from the `kagenti-feature-gates` ConfigMap in the
 operator's namespace and take effect immediately (no restart needed).
 
 | Gate | Default | Controls |
@@ -147,16 +135,16 @@ Pod admitted by webhook
        └─ globalEnabled=true?  — No → skip
             └─ type=tool and injectTools=false?  — Yes → skip
                  └─ kagenti.io/inject=disabled?  — Yes → skip
-                      └─ Resolve mode (CR → namespace CM → annotation → default)
+                      └─ Resolve mode (namespace CM → annotation → default)
                            ├─ waypoint → skip (standalone deployment)
                            ├─ proxy-sidecar / lite
                            │    └─ inject authbridge-proxy
                            │         + HTTP_PROXY env vars into agent container
-                           │         + SPIRE_ENABLED based on spiffe-helper decision
+                           │         + SPIRE_ENABLED based on spiffe-helper-inject label
                            └─ envoy-sidecar
                                 └─ inject envoy-proxy (if envoyProxy gate=true and label≠false)
                                      + proxy-init (follows envoy-proxy decision)
-                                     + SPIRE_ENABLED based on spiffe-helper decision
+                                     + SPIRE_ENABLED based on spiffe-helper-inject label
 ```
 
 ## Related
--- a/docs/authbridge/sidecar-injection.md
+++ b/docs/authbridge/sidecar-injection.md
@@ -132,7 +132,7 @@ operator's namespace and take effect immediately (no restart needed).
 | `injectTools` | `false` | Whether `kagenti.io/type=tool` workloads receive injection (agents are always injected) |
 | `perWorkloadConfigResolution` | `false` | When `true`, webhook reads namespace ConfigMaps at admission time and injects literal env var values instead of `valueFrom` references |
 
-Source of truth: [`kagenti-operator/internal/webhook/config/feature_gates.go`](https://github.com/kagenti/kagenti-extensions/blob/main/kagenti-operator/internal/webhook/config/feature_gates.go)
+Source of truth: [`internal/webhook/config/feature_gates.go`](https://github.com/kagenti/kagenti-operator/blob/main/kagenti-operator/internal/webhook/config/feature_gates.go)
 
 ## Full injection decision flow
 
--- a/docs/authbridge/sidecar-injection.md
+++ b/docs/authbridge/sidecar-injection.md
@@ -1,0 +1,161 @@
+# AuthBridge Sidecar Injection
+
+This page explains what containers to expect in an agent pod, which labels
+control injection, and how to switch modes.
+
+## Verify which mode you're in
+
+```bash
+# List container names in the running pod
+kubectl get pods -n team1 -l kagenti.io/type=agent \
+  -o jsonpath='{range .items[*]}{.metadata.name}{" → "}{.spec.containers[*].name}{"\n"}{end}'
+
+# Or for a specific deployment
+kubectl get pods -n team1 -l app=weather-service \
+  -o jsonpath='{.items[0].spec.containers[*].name}'
+```
+
+## Expected containers by mode
+
+| Mode | Regular containers | Init containers | Notes |
+|---|---|---|---|
+| `proxy-sidecar` (default) | `<agent>` `authbridge-proxy` | — | spiffe-helper runs as a process **inside** `authbridge-proxy` when SPIRE is on |
+| `lite` | `<agent>` `authbridge-proxy` | — | Same shape as proxy-sidecar; uses `authbridge-lite` image (auth-only) |
+| `envoy-sidecar` | `<agent>` `envoy-proxy` | `proxy-init` | spiffe-helper runs as a process **inside** `envoy-proxy` when SPIRE is on; proxy-init is privileged |
+| `waypoint` | — | — | Not injected as a sidecar; waypoint is a standalone deployment |
+
+> **spiffe-helper is not a separate container.** It is bundled inside both
+> `authbridge-proxy` (proxy-sidecar/lite) and `envoy-proxy` (envoy-sidecar) and
+> starts conditionally based on the `SPIRE_ENABLED` env var, which the operator
+> sets per workload based on the `kagenti.io/spiffe-helper-inject` label.
+
+## Label vocabulary
+
+Labels are set on the **pod template** (e.g. `Deployment.spec.template.metadata.labels`).
+The operator applies `kagenti.io/type` automatically via the AgentRuntime reconciler —
+you normally do not need to set it yourself.
+
+### Pre-filter labels (control whether injection runs at all)
+
+| Label | Values | Behavior |
+|---|---|---|
+| `kagenti.io/type` | `agent` \| `tool` | **Required for injection.** Only `agent` and `tool` workloads are mutated. Set automatically by the operator. |
+| `kagenti.io/inject` | `disabled` | Set to `disabled` to opt this workload out of all sidecar injection. Any other value (or absent) allows injection. |
+
+### Per-sidecar opt-out labels
+
+These let you disable a specific sidecar while leaving others active.
+
+| Label | Default | Set to `false` to… |
+|---|---|---|
+| `kagenti.io/envoy-proxy-inject` | inject (when feature gate is on) | Disable the `envoy-proxy` sidecar (envoy-sidecar mode only) |
+| `kagenti.io/spiffe-helper-inject` | inject (enabled) | Suppress SPIRE — sets `SPIRE_ENABLED=false` on the combined sidecar, preventing the bundled spiffe-helper process from starting |
+
+### Deprecated label
+
+| Annotation | Status | Replacement |
+|---|---|---|
+| `kagenti.io/authbridge-mode` | **Deprecated** — still honored | Use `AgentRuntime.Spec.AuthBridgeMode` |
+
+### Examples
+
+```yaml
+# Opt this workload out of all injection
+metadata:
+  labels:
+    kagenti.io/inject: "disabled"
+
+# Keep injection but disable SPIRE
+metadata:
+  labels:
+    kagenti.io/spiffe-helper-inject: "false"
+
+# Disable the envoy-proxy sidecar in envoy-sidecar mode
+metadata:
+  labels:
+    kagenti.io/envoy-proxy-inject: "false"
+```
+
+## How to switch modes
+
+Mode is resolved from this chain (first non-empty wins):
+
+1. `AgentRuntime.Spec.AuthBridgeMode` on the workload's CR **(canonical)**
+2. `mode:` field in the namespace-level `authbridge-runtime-config` ConfigMap
+3. `kagenti.io/authbridge-mode` pod annotation *(deprecated — still honored)*
+4. Cluster-wide default: `proxy-sidecar`
+
+### Per-workload override (AgentRuntime CR)
+
+```yaml
+apiVersion: kagenti.io/v1alpha1
+kind: AgentRuntime
+metadata:
+  name: weather-service
+  namespace: team1
+spec:
+  authBridgeMode: envoy-sidecar   # proxy-sidecar | envoy-sidecar | lite | waypoint
+  targetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: weather-service
+```
+
+### Namespace default (ConfigMap)
+
+```yaml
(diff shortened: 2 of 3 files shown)


charts/kagenti/templates/operator-client-bootstrap-job.yaml
```diff
@@ -0,0 +1,240 @@
...
+    metadata:
+      labels:
+        app.kubernetes.io/name: kagenti
+        app.kubernetes.io/component: operator-bootstrap
+    spec:
+      restartPolicy: OnFailure
+      serviceAccountName: default
+      containers:
+      - name: bootstrap
+        image: {{ .Values.kagentiOperator.spiffeAuth.bootstrapImage | default "curlimages/curl:latest" }}
```
clawgenti: **must-fix** `curlimages/curl:latest` does not include `jq`. The bootstrap script uses `jq -sRr @uri` at lines 157 and 200 to URL-encode `$OPERATOR_CLIENT_ID` for the client search API call. This will fail with `sh: jq: not found` at runtime. Use an image that bundles both curl and jq, or replace the `jq` invocation with a POSIX-safe alternative (e.g. `python3 -c 'import urllib.parse,sys; print(urllib.parse.quote(sys.argv[1]))' "$OPERATOR_CLIENT_ID"` — python3 is available in most base images).

charts/kagenti/templates/operator-client-bootstrap-job.yaml
```diff
@@ -0,0 +1,240 @@
...
+spec:
+  ttlSecondsAfterFinished: 300
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: kagenti
+        app.kubernetes.io/component: operator-bootstrap
+    spec:
+      restartPolicy: OnFailure
+      serviceAccountName: default
```
clawgenti: **must-fix** `serviceAccountName: default` gives the Job the default SA in the keycloak namespace. The bootstrap script reads the `keycloak-initial-admin` Secret and makes Keycloak admin API calls — neither is permitted by the out-of-the-box `default` SA. Add a dedicated `ServiceAccount` + `Role` (granting `get` on secrets in the keycloak namespace) + `RoleBinding`, or at minimum document the required pre-existing RBAC.

charts/kagenti/templates/operator-client-bootstrap-job.yaml
```diff
@@ -0,0 +1,240 @@
...
+          echo "   ✓ Admin token obtained"
+
+          # Check if SPIFFE IdP exists
+          echo "2. Checking if SPIFFE Identity Provider exists..."
+          IDP_RESPONSE=$(curl -s -w "\n%{http_code}" -X GET \
+            -H "Authorization: Bearer $TOKEN" \
+            "$KEYCLOAK_URL/admin/realms/$KEYCLOAK_REALM/identity-provider/instances/$SPIFFE_IDP_ALIAS")
+          IDP_HTTP_CODE=$(echo "$IDP_RESPONSE" | tail -n1)
+
+          if [ "$IDP_HTTP_CODE" = "404" ]; then
```
clawgenti: **suggestion** Only `404` is treated as "not found"; a `401`, `403`, or `5xx` response silently falls through to the `else` branch and prints "already exists", then continues with a token or config that is actually broken. Consider adding: `elif [ "$IDP_HTTP_CODE" != "200" ]; then echo "ERROR: unexpected HTTP $IDP_HTTP_CODE checking IdP"; exit 1; fi`.

charts/kagenti/values.yaml
```diff
@@ -286,6 +286,33 @@ kagenti-operator-chart:
...
+#  SPIFFE Configuration
+# ------------------------------------------------------------------
+spiffe:
+  # SPIFFE Identity Provider alias in Keycloak
+  idpAlias: "spire-spiffe"
+  # SPIRE OIDC Discovery Provider URL
+  oidcDiscoveryURL: "http://spire-spiffe-oidc-discovery-provider.zero-trust-workload-identity-manager.svc:8080"
+
+# Trust domain (used for SPIFFE IDs)
+spireTrustDomain: localtest.me
```
clawgenti: **suggestion** `spireTrustDomain: localtest.me` is a new top-level key, but `kagenti-operator-chart.signatureVerification.spireTrustDomain` already exists for the same trust domain. Users setting one won't know to set the other. Consider pointing the Job template's `OPERATOR_CLIENT_ID` default at `.Values.kagentiOperator.spiffeAuth.operatorClientID` (already a settable value) or reusing `kagenti-operator-chart.signatureVerification.spireTrustDomain` via a Helm template helper to avoid two sources of truth.

clawgenti (commented):
Adds the operator SPIFFE auth bootstrap Job and a well-written `sidecar-injection.md` reference page. Four issues found in the new Helm Job template that need attention before merge.

- **must-fix** `curlimages/curl:latest` does not ship `jq`; the bootstrap script calls `jq -sRr @uri` to URL-encode the client ID (lines 157, 200) and will `command not found` at runtime. Use an image that includes both curl and jq (e.g. `dwdraju/alpine-curl-jq` or a custom image), or replace the `jq` call with a POSIX-safe URL encoder like `python3 -c 'import urllib.parse,sys; print(urllib.parse.quote(sys.argv[1]))' "$OPERATOR_CLIENT_ID"`.
- **must-fix** `serviceAccountName: default` — the Job reads `keycloak-initial-admin` secret in the keycloak namespace and makes admin API calls; the `default` SA almost certainly lacks `get secret` RBAC. Add a dedicated `ServiceAccount` + `Role`/`RoleBinding` granting `get` on `secrets/keycloak-initial-admin` in the keycloak namespace, or document that the operator is expected to pre-create this.
- **suggestion** IDP existence check (step 2) only branches on `404`; a `403` (token expired or insufficient scope) or `5xx` will fall through to the "already exists" path and silently continue with a bad token. Add an explicit error branch: `if [ "$IDP_HTTP_CODE" != "200" ] && [ "$IDP_HTTP_CODE" != "404" ]; then echo "ERROR..."; exit 1; fi`.
- **suggestion** `spireTrustDomain: localtest.me` is added as a new top-level key in `values.yaml`, but `kagenti-operator-chart.signatureVerification.spireTrustDomain` already fulfils the same role. The new key is referenced only inside the bootstrap Job template's `OPERATOR_CLIENT_ID` default. Consider referencing the existing nested key (`.Values.kagentiOperator` already reads `.Values.spireTrustDomain`) or consolidating to avoid two canonical locations for the same value.

---

Alan-Cha:
Thanks for the thorough review! Quick clarification: this PR is intended to be **docs-only** (the new `sidecar-injection.md` page + cross-references). The Helm bootstrap Job changes you're seeing are from the source branch (`feat/operator-spiffe-auth-v2`) that hadn't been cleaned up yet — they belong in PR #2153, not here.

The branch needs a force-push to trim it to the single doc commit. Once that's done the four Helm issues won't apply here. I'll track them against #2153 where the bootstrap Job actually lives.

Alan-Cha:
Rebased onto main. Responding to review comments:

**must-fix 1 (curlimages/curl + jq):** The bootstrap Job does not use `curlimages/curl`. The image is `operator-spiffe-bootstrap` (Python), and the bootstrap logic is in `setup_operator_client.py` which uses the `requests` library — no `jq` involved. The HTTP error handling is already present at line 167: `elif response.status_code != 404: logger.error(...)` with `raise`. Both the 401/403/5xx case and the 404-is-not-found case are handled.

**must-fix 2 (serviceAccountName: default):** The Job already uses `serviceAccountName: kagenti-operator-client-bootstrap` backed by a dedicated `ServiceAccount`, `Role` (granting `get` on `keycloak-initial-admin`), and `RoleBinding` — all in the same template file. This was addressed before the review.

**suggestion (HTTP error handling):** Already handled in the Python script as noted above.

**suggestion (duplicate spireTrustDomain):** Valid long-term cleanup. Filed as a separate concern — the `SPIFFE_TRUST_DOMAIN` env var on the bootstrap Job should be derived from `kagenti-operator-chart.signatureVerification.spireTrustDomain` rather than a separate top-level value. Not blocking this PR.

Added a link to `authentication.md` in `deployment-guide.md` for `CLIENT_AUTH_TYPE=federated-jwt`.

docs/authbridge/deployment-guide.md
```diff
@@ -42,10 +42,16 @@ spec:
     name: weather-service
 ```
 
-> **Note:** All workloads should use an AgentRuntime CR to enroll with the
-> platform. The operator applies `kagenti.io/type` labels automatically.
-> The deprecated `kagenti.io/authbridge-mode` annotation is still honored
-> but `AgentRuntime.Spec.AuthBridgeMode` is the canonical method.
+The deprecated annotation is retained for tools and other workloads that
+don't have their own AgentRuntime CR yet:
```
clawgenti: **suggestion**: The replacement text drops the canonical-method guidance from the removed note. The deleted callout told readers that `AgentRuntime.Spec.AuthBridgeMode` is the canonical path; this replacement only shows the deprecated annotation. Readers who land here without reading `sidecar-injection.md` will see no pointer to the preferred approach. Consider one bridging sentence, e.g.: "For workloads that have an AgentRuntime CR, set `AgentRuntime.Spec.AuthBridgeMode` instead — the annotation is retained only for workloads that do not yet have a CR."

clawgenti (commented):
Docs-only PR adding a well-structured `sidecar-injection.md` reference page with cross-references from the README and deployment guide. One suggestion in the deployment guide edit.

- **suggestion** `deployment-guide.md` lines 45–54: the replacement text drops the canonical guidance that was in the removed note. The old callout told readers "All workloads should use an AgentRuntime CR" and named `AgentRuntime.Spec.AuthBridgeMode` as the canonical method. The new text only shows the deprecated annotation example, so readers who never see `sidecar-injection.md` lose the recommendation to migrate to the CR. Consider keeping one sentence pointing to the canonical path, e.g. _"For workloads with an AgentRuntime CR, use `AgentRuntime.Spec.AuthBridgeMode` instead — the annotation is retained only for workloads without a CR."_

---

docs/authbridge/sidecar-injection.md
```diff
@@ -0,0 +1,166 @@
...
+| `kagenti.io/inject` | `disabled` | Set to `disabled` to opt this workload out of all sidecar injection. Any other value (or absent) allows injection. |
+
+### Per-sidecar opt-out labels
+
+These let you disable a specific sidecar while leaving others active.
+
+| Label | Default | Set to `false` to… |
+|---|---|---|
+| `kagenti.io/envoy-proxy-inject` | inject (when feature gate is on) | Disable the `envoy-proxy` sidecar (envoy-sidecar mode only) |
+| `kagenti.io/spiffe-helper-inject` | inject (enabled) | Suppress SPIRE — sets `SPIRE_ENABLED=false` on the combined sidecar, preventing the bundled spiffe-helper process from starting |
```
clawgenti: **Inconsistency with lines 27–35**: This description says "preventing the bundled spiffe-helper process from starting," but the callout earlier in this file explicitly states *"There is no spiffe-helper container or process"* — AuthBridge uses the go-spiffe SDK directly. Suggest: "Suppress SPIRE integration — sets `SPIRE_ENABLED=false` on the sidecar, preventing the go-spiffe workload API source from being opened."

docs/authbridge/deployment-guide.md
```diff
@@ -42,10 +42,16 @@ spec:
     name: weather-service
 ```
 
-> **Note:** All workloads should use an AgentRuntime CR to enroll with the
-> platform. The operator applies `kagenti.io/type` labels automatically.
-> The deprecated `kagenti.io/authbridge-mode` annotation is still honored
-> but `AgentRuntime.Spec.AuthBridgeMode` is the canonical method.
+The deprecated annotation is retained for tools and other workloads that
```
clawgenti: **Nit / follow-up**: The removed `> **Note:**` block contained canonical guidance ("All workloads should use an AgentRuntime CR") that is now only in `sidecar-injection.md`. That's fine for coverage, but note that the *existing* opening section of `deployment-guide.md` (unchanged in this PR) still describes `spiffe-helper` as a bundled binary that runs in-pod — inconsistent with the correction in the final commit. Worth a follow-up patch to `deployment-guide.md`'s header to align with the new wording.

clawgenti (commented):
Docs-only PR adding `sidecar-injection.md` with solid coverage of container expectations, label vocabulary, mode-switching, and feature gates.

**Findings:**

- **`sidecar-injection.md` line 57** (`kagenti.io/spiffe-helper-inject` description): Says "preventing the bundled spiffe-helper process from starting" — but lines 27–35 of the same file explicitly state *"There is no spiffe-helper container or process"* and describe SPIRE as in-process via the go-spiffe SDK. The label description contradicts the callout on the same page. Suggest aligning: e.g. "Suppress SPIRE integration — sets `SPIRE_ENABLED=false` on the sidecar, preventing the go-spiffe workload API source from being opened."

- **`deployment-guide.md`** (existing header, unchanged in this PR): The opening section still says `spiffe-helper is bundled inside every combined image and runs only when the operator sets SPIRE_ENABLED=true` — inconsistent with the correction made in the final commit. Consider a follow-up to align `deployment-guide.md`'s description with the new canonical wording in `sidecar-injection.md`.

- **PR title nit**: `📖 Docs(authbridge): ...` trips `verify-pr-title`; the verifier expects a bare prefix like `Docs:` or `Docs(authbridge):` without a leading emoji.

---

docs/authbridge/sidecar-injection.md
```diff
@@ -0,0 +1,166 @@
...
+metadata:
+  labels:
+    kagenti.io/envoy-proxy-inject: "false"
+```
+
+## How to switch modes
+
+Mode is resolved from this chain (first non-empty wins):
+
+1. `AgentRuntime.Spec.AuthBridgeMode` on the workload's CR **(canonical)**
```
mrsabath: **must-fix — mode precedence is wrong.** The operator's injection path (`pod_mutator.go` resolveMode, source comment at lines 389-391) resolves in **3** layers, first non-empty wins:

1. namespace `authbridge-runtime-config` `mode:` field
2. `kagenti.io/authbridge-mode` annotation (deprecated)
3. cluster default `proxy-sidecar`

`AgentRuntime.Spec.AuthBridgeMode` is **not** a layer in pod-mutation mode resolution. The field exists (`api/v1alpha1/agentruntime_types.go:87`) and is enum-validated in the CRD webhook, but the mutating webhook never reads it when deciding injection mode. As written, an operator who sets `spec.authBridgeMode` expecting it to win ("canonical") will find it silently ignored.

Please drop layer 1 here (making it the 3-layer chain), and correct the "Per-workload override (AgentRuntime CR)" example block below — or, if the CR path *does* feed mode via the controller-managed AgentRuntime reconciliation, clarify explicitly that it's a different path from the injection precedence, so operators don't expect the CR to override a namespace ConfigMap.

docs/authbridge/sidecar-injection.md
```diff
@@ -0,0 +1,166 @@
...
+### Verify the resolved mode
+
+```bash
+kubectl logs -n kagenti-system deploy/kagenti-operator \
+  | grep "resolved authbridge mode"
+```
+
+## Cluster-admin feature gates
+
+Feature gates are loaded from the `kagenti-webhook-config` ConfigMap in the
```
mrsabath: **must-fix — wrong ConfigMap name.** The feature-gates ConfigMap is **`kagenti-feature-gates`** (`kagenti-operator/internal/controller/agentruntime_config.go:38` — `ClusterFeatureGatesConfigMapName = "kagenti-feature-gates"`), not `kagenti-webhook-config` (that name doesn't appear anywhere in the operator).

The rest of the sentence is accurate: gates are mounted as a file (`/etc/kagenti/feature-gates/feature-gates.yaml`) and hot-reloaded via an fsnotify watcher, so "take effect immediately (no restart needed)" is correct — just fix the name.

docs/authbridge/sidecar-injection.md
```diff
@@ -0,0 +1,166 @@
...
+  -o jsonpath='{.items[0].spec.containers[*].name}'
+```
+
+## Expected containers by mode
+
+| Mode | Regular containers | Init containers | Notes |
+|---|---|---|---|
+| `proxy-sidecar` (default) | `<agent>` `authbridge-proxy` | — | SPIRE integration is in-process inside `authbridge-proxy` when `SPIRE_ENABLED=true` |
+| `lite` | `<agent>` `authbridge-proxy` | — | Same shape as proxy-sidecar; uses `authbridge-lite` image (auth-only) |
+| `envoy-sidecar` | `<agent>` `envoy-proxy` | `proxy-init` | SPIRE integration is in-process inside `envoy-proxy` when `SPIRE_ENABLED=true`; proxy-init is privileged |
```
mrsabath: **must-fix — `proxy-init` is not privileged.** `BuildProxyInitContainer` (`kagenti-operator/internal/webhook/injector/container_builder.go`, docstring at 435-449) sets `Privileged: ptr.To(false)`, `RunAsUser: 0`, and grants only `NET_ADMIN` + `NET_RAW` with all other capabilities dropped. It runs as root (UID 0) to set up iptables, but is **explicitly non-privileged** — inverting this is a security-relevant inaccuracy.

Suggest: "proxy-init runs as root (UID 0) with NET_ADMIN/NET_RAW capabilities only — not privileged".

mrsabath (changes requested):
Genuinely useful, well-structured reference page — the container-by-mode table, label vocabulary, and the ASCII decision tree fill a real gap (#1775), and the large majority of the technical claims check out against `kagenti-operator` source (all feature-gate names & defaults, every label string incl. the exact `kagenti.io/spiffe-helper-inject`, the opt-out semantics of `kagenti.io/inject=disabled`, the four mode values, `SPIRE_ENABLED`, the `proxy-init` container name, the ConfigMap/`mode` key/default in the resolution chain, and the `feature_gates.go` link — the doubled `kagenti-operator/kagenti-operator` path segment is actually correct, the repo has a nested subdir). Link Check CI passed too.

Three factual errors need fixing before merge, because this page is meant to be authoritative and an operator will act on it (inline comments below):
1. Mode-precedence chain lists a 4th "canonical" `AgentRuntime.Spec.AuthBridgeMode` layer that the injection webhook never consults — it's a 3-layer chain.
2. Feature-gates ConfigMap is misnamed (`kagenti-webhook-config` → `kagenti-feature-gates`).
3. `proxy-init` is described as privileged when it's explicitly *not*.

Separately: the required **`verify-pr-title`** check is failing. The leading `📖` emoji pushes the `Docs` prefix off position 0, which the org ruleset rejects (`allowed_prefixes` must appear at the very start). Rename the PR title to `Docs(authbridge): add sidecar injection reference page` (drop the emoji) and it'll pass — the emoji is fine in the commit subject, just not the PR title.

**Areas reviewed**: Docs (accuracy vs kagenti-operator source, links), commit & PR conventions, CI
**Commits**: 4, all signed-off (DCO passed)
**CI**: verify-pr-title failing (emoji prefix); all other checks green

Thanks for writing this up, @Alan-Cha — it's a page the project needed. The fixes are all small string/precedence corrections.

Alan-Cha:
Addressed all comments in 18dbfb32. Summary:

**must-fix (proxy-init privileged):** Corrected — proxy-init runs as root (UID 0) with NET_ADMIN/NET_RAW only; `Privileged: false` per `BuildProxyInitContainer`. Verified in `container_builder.go:435-449`.

**must-fix (mode precedence — AgentRuntime not in injection chain):** Corrected. The mutating webhook resolves mode from 3 layers only: namespace ConfigMap → deprecated annotation → cluster default. `AgentRuntime.Spec.AuthBridgeMode` is enum-validated by the CRD webhook but explicitly excluded from the config that drives injection (`agentruntime_config.go:79`). The doc now reflects the actual 3-layer chain and adds a note explaining the field is validated but not wired into injection. The misleading "Per-workload override (AgentRuntime CR)" section has been removed.

**must-fix (feature gates ConfigMap name):** Corrected — `kagenti-feature-gates`, not `kagenti-webhook-config`. Verified at `agentruntime_config.go:38`.

**inconsistency (spiffe-helper process):** Corrected — "preventing the bundled spiffe-helper process from starting" → "preventing the go-spiffe workload API source from being opened".

**nit (deployment-guide.md header):** Corrected — replaced "spiffe-helper is bundled inside every combined image" with accurate go-spiffe SDK description.

**suggestion (deployment-guide.md deprecated annotation):** The replacement sentence now points to the namespace ConfigMap as the canonical replacement, not AgentRuntime CR (since AuthBridgeMode doesn't affect injection).

**older must-fix comments (curlimages/curl, serviceAccountName: default, HTTP error handling):** These remain addressed as noted in the prior comment — the bootstrap image is a Python script (`operator-spiffe-bootstrap`), not `curlimages/curl`; the dedicated SA/Role/RoleBinding are already in the template; and the Python script already handles non-200/404 responses at line 167.

docs/authbridge/deployment-guide.md
```diff
@@ -42,10 +44,17 @@ spec:
...
 
-> **Note:** All workloads should use an AgentRuntime CR to enroll with the
-> platform. The operator applies `kagenti.io/type` labels automatically.
-> The deprecated `kagenti.io/authbridge-mode` annotation is still honored
-> but `AgentRuntime.Spec.AuthBridgeMode` is the canonical method.
+The deprecated annotation is retained for backward compatibility. Use the
+namespace `authbridge-runtime-config` ConfigMap instead — it is the canonical
+way to set mode and applies to all workloads in the namespace:
+
+```yaml
```
clawgenti, line 38: **suggestion**: The prose says to use the namespace `authbridge-runtime-config` ConfigMap as the canonical replacement for the deprecated annotation, but the YAML example immediately below (lines 51–57) demonstrates the **annotation** (`kagenti.io/authbridge-mode`) — the thing being deprecated. The example should instead show the ConfigMap (`kind: ConfigMap`, `name: authbridge-runtime-config`, `data.mode: envoy-sidecar`) to match the guidance in the prose.

docs/authbridge/sidecar-injection.md
```diff
@@ -0,0 +1,154 @@
...
+| Label | Default | Set to `false` to… |
+|---|---|---|
+| `kagenti.io/envoy-proxy-inject` | inject (when feature gate is on) | Disable the `envoy-proxy` sidecar (envoy-sidecar mode only) |
+| `kagenti.io/spiffe-helper-inject` | inject (enabled) | Suppress SPIRE — sets `SPIRE_ENABLED=false` on the sidecar, preventing the go-spiffe workload API source from being opened |
+
+### Deprecated label
+
+| Annotation | Status | Replacement |
+|---|---|---|
+| `kagenti.io/authbridge-mode` | **Deprecated** — still honored | Use `AgentRuntime.Spec.AuthBridgeMode` |
```
clawgenti: **suggestion**: The deprecated label table lists `AgentRuntime.Spec.AuthBridgeMode` as the replacement for `kagenti.io/authbridge-mode`, but the "How to switch modes" section below (the note at lines 92–95) explicitly clarifies that `AgentRuntime.Spec.AuthBridgeMode` is **not read by the mutating webhook** and does not affect injection. The replacement column should point to the namespace `authbridge-runtime-config` ConfigMap instead, consistent with the mode-resolution chain described in this same document.

clawgenti (commented):
Documentation-only PR adding a well-structured sidecar injection reference page with clear label vocabulary, container tables, and decision flow. Two internal contradictions in the new/updated content need to be addressed before merge.

---

docs/authbridge/sidecar-injection.md
```diff
@@ -0,0 +1,154 @@
...
+### Namespace default (ConfigMap)
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: authbridge-runtime-config
+  namespace: team1
+data:
+  mode: proxy-sidecar
```
mrsabath: **must-fix.** This example won't work as written. The webhook reads the mode from `data["config.yaml"]` and then parses the top-level `mode:` key *inside* that YAML string (`kagenti-operator` `internal/webhook/injector/namespace_config.go`: `cfg.AuthBridgeRuntimeYAML = cm.Data["config.yaml"]`, then `ExtractMode`). A top-level `data.mode` key is never consulted, so this ConfigMap silently falls through to the cluster default.

Correct shape:
```yaml
data:
  config.yaml: |
    mode: proxy-sidecar
```

docs/authbridge/sidecar-injection.md
mrsabath: **must-fix (self-contradiction).** This row gives the replacement for the deprecated annotation as `AgentRuntime.Spec.AuthBridgeMode` — but the note further down this same page says *"the mutating webhook does not read it during injection."* `injector/constants.go` confirms it: mode is resolved from the namespace `authbridge-runtime-config` ConfigMap → deprecated annotation → `proxy-sidecar` default, with no CR layer.

So pointing users at `AgentRuntime.Spec.AuthBridgeMode` as the replacement sends them to a field that has no effect on injection. The replacement should be the namespace `authbridge-runtime-config` ConfigMap — which is also what `deployment-guide.md` (edited in this same PR) correctly states. Please align the two files.

docs/authbridge/sidecar-injection.md
```diff
@@ -0,0 +1,154 @@
...
+```yaml
+# Opt this workload out of all injection
+metadata:
+  labels:
+    kagenti.io/inject: "disabled"
+
+# Keep injection but disable SPIRE
+metadata:
+  labels:
+    kagenti.io/spiffe-helper-inject: "false"
```
mrsabath, line 76: suggestion (completeness): the operator also recognizes `kagenti.io/spire=disabled` as a SPIRE opt-out, separate from `kagenti.io/spiffe-helper-inject` (see `pod_mutator.go` `SpireEnableLabel`/`SpireDisabledValue`). Since this section is about disabling SPIRE, worth a line noting both labels so operators aren't surprised that either one suppresses it. Not blocking.

mrsabath (changes requested):
## Summary

This is an unusually well-researched reference page — I verified the technical claims against the `kagenti-operator` webhook source and nearly all of them are correct:

- **Container names** — `authbridge-proxy`, `envoy-proxy`, `proxy-init` ✅ (`injector/constants.go`, `injector/container_builder.go`)
- **Modes** — `proxy-sidecar` (default), `lite`, `envoy-sidecar`, `waypoint`, incl. the `lite` = "auth-only, same shape" note ✅ (`constants.go`)
- **Feature gates** — `globalEnabled`/`envoyProxy`/`injectTools`/`perWorkloadConfigResolution` names + defaults (`true`/`true`/`false`/`false`) ✅ (`config/feature_gates.go`), including the literal-vs-`valueFrom` behavior
- **Pre-filter + per-sidecar labels** — `kagenti.io/type`, `kagenti.io/inject=disabled`, `kagenti.io/envoy-proxy-inject`, `kagenti.io/spiffe-helper-inject`, opt-out semantics ✅ (`pod_mutator.go`, `precedence.go`)
- **SPIRE narrative** — in-process go-spiffe, socket `/spiffe-workload-api/spire-agent.sock`, `/opt/*.pem` written by the combined image ✅ (`container_builder.go`)
- **`feature_gates.go` link path** ✅ resolves

Two things would mislead a reader who follows the page verbatim, so requesting changes:

1. **ConfigMap example is inert** — the "Namespace default" example sets `mode:` as a direct `data` key, but the webhook only reads `data["config.yaml"]` and parses `mode:` from *inside* that YAML string (`injector/namespace_config.go` `ExtractMode`). As written the example changes nothing.
2. **Self-contradiction on the deprecated annotation** — the "Deprecated label" table gives the replacement as `AgentRuntime.Spec.AuthBridgeMode`, but this same page (and `constants.go`) state the webhook does not read that field. The canonical replacement is the namespace ConfigMap — which is also what the sibling `deployment-guide.md` edit in this PR correctly says.

Details inline. One optional suggestion about the `kagenti.io/spire` label too.

**Areas reviewed**: Docs (accuracy vs kagenti-operator source, cross-file consistency), Commit/PR conventions
**Commits**: 5 commits, all signed-off (DCO passes)
**CI status**: all green (Link Check, Markdown Lint incl.)
**Verdict**: REQUEST_CHANGES — 2 must-fix; the rest of the page is accurate and can stay.

Alan-Cha:
Fixed in 71603f4d.

**must-fix (ConfigMap shape):** Corrected — the webhook reads `cm.Data["config.yaml"]` and calls `ExtractMode` on the YAML string inside. A top-level `data.mode` key is never consulted. Both sidecar-injection.md and deployment-guide.md now show the correct shape:
```yaml
data:
  config.yaml: |
    mode: proxy-sidecar
```

**must-fix (deprecated annotation replacement):** Corrected — both the deprecated label table in sidecar-injection.md and the YAML example in deployment-guide.md now point to the namespace ConfigMap, not `AgentRuntime.Spec.AuthBridgeMode`.

**suggestion (deployment-guide.md annotation YAML):** Fixed — replaced the annotation example with a ConfigMap example to match the prose.

**suggestion (kagenti.io/spire=disabled):** I checked the source — `SpireEnableLabel = "kagenti.io/spire"` and `SpireDisabledValue = "disabled"` are defined as constants in `pod_mutator.go` but are never referenced in any evaluation path (not in `precedence.go`, not in `pod_mutator.go`'s evaluation logic). Setting this label has no effect. I'm leaving it out of the docs to avoid misleading operators. If it should be wired in, that's a separate implementation task.

clawgenti (commented):
Documentation is clear and well-structured, covering container shapes, label vocabulary, feature gates, and the injection decision flow.

**Finding:** The "Selecting a mode" section in `deployment-guide.md` (line 28) still lists `AgentRuntime.Spec.AuthBridgeMode` as item #1 / canonical in the precedence chain, but the new text added by this PR (lines 47–49) declares the namespace ConfigMap as canonical, and `sidecar-injection.md` (lines 92–95) explicitly states that `AgentRuntime.Spec.AuthBridgeMode` is not read by the webhook and does not affect injection. The "Selecting a mode" section in `deployment-guide.md` was not updated to match, leaving the guide self-contradictory.

---

docs/authbridge/deployment-guide.md
```diff
@@ -42,10 +44,20 @@ spec:
     name: weather-service
 ```
 
-> **Note:** All workloads should use an AgentRuntime CR to enroll with the
-> platform. The operator applies `kagenti.io/type` labels automatically.
-> The deprecated `kagenti.io/authbridge-mode` annotation is still honored
-> but `AgentRuntime.Spec.AuthBridgeMode` is the canonical method.
+The deprecated annotation is retained for backward compatibility. Use the
```
clawgenti: The text immediately above (the "Selecting a mode" precedence chain, line 28) still lists `AgentRuntime.Spec.AuthBridgeMode` as item #1 / canonical, but this new paragraph declares the namespace ConfigMap canonical instead. `sidecar-injection.md` (lines 92–95 added by this PR) clarifies that AgentRuntime does not affect injection at all. The "Selecting a mode" list in this file was not updated — consider removing or demoting the AgentRuntime entry (or adding a note matching sidecar-injection.md) to avoid the contradiction.

Alan-Cha (commented):
The PR adds a new reference page (`sidecar-injection.md`) for AuthBridge injection behaviour, corrects two inaccuracies in `deployment-guide.md` (SPIRE is go-spiffe in-process, not a separate binary; the namespace ConfigMap is canonical for mode, not `AgentRuntime.Spec.AuthBridgeMode`), and updates cross-references in `README.md`. The corrections are improvements — the old docs were misleading on both counts. The new page is well-structured and technically precise.

**Author**: Alan-Cha (MEMBER — maintainer)
**Areas reviewed**: Docs
**Commits**: 6 commits, all signed-off ✅
**CI status**: all 22 checks passing ✅

*(Cannot self-approve — posting as COMMENT. Verdict would be APPROVE.)*

docs/authbridge/deployment-guide.md
```diff
@@ -234,7 +246,7 @@ kubectl exec deploy/weather-service -n team1 -c envoy-proxy -- \
 
 ### Proxy Not Injected
 
-1. Verify the workload has an AgentRuntime CR targeting it (the operator applies the `kagenti.io/type` label automatically)
+1. Verify the pod has `kagenti.io/type: agent` label
```
Alan-Cha: **Suggestion**: This is more direct than the old step, but a reader unfamiliar with the platform may wonder how `kagenti.io/type: agent` gets set. The causal link (AgentRuntime reconciler) exists in `sidecar-injection.md` but isn't visible here.

Consider:
```
1. Verify the pod has `kagenti.io/type: agent` label — the operator sets this when an AgentRuntime CR targets the workload
```

docs/authbridge/sidecar-injection.md
```diff
@@ -0,0 +1,155 @@
...
+kubectl get pods -n team1 -l app=weather-service \
+  -o jsonpath='{.items[0].spec.containers[*].name}'
+```
+
+## Expected containers by mode
+
+| Mode | Regular containers | Init containers | Notes |
+|---|---|---|---|
+| `proxy-sidecar` (default) | `<agent>` `authbridge-proxy` | — | SPIRE integration is in-process inside `authbridge-proxy` when `SPIRE_ENABLED=true` |
+| `lite` | `<agent>` `authbridge-proxy` | — | Same shape as proxy-sidecar; uses `authbridge-lite` image (auth-only) |
```
Alan-Cha: **Nit**: The `proxy-sidecar` and `envoy-sidecar` rows both note SPIRE integration in-process, but the `lite` row is silent on SPIRE. The callout below says "every combined image" — does `authbridge-lite` count as a combined image?

- If yes (SPIRE in-process when `SPIRE_ENABLED=true`), add a note matching the other rows.
- If no (auth-only with no SPIRE support), make that explicit — e.g. "No SPIRE integration" — so readers don't have to infer it.

clawgenti (commented):
Documentation-only PR adding `sidecar-injection.md` and correcting `deployment-guide.md`. All previously-requested changes have been addressed: the ConfigMap example now correctly uses `data.config.yaml`, the mode-resolution chain is the correct 3-layer form with the namespace ConfigMap as canonical, and the `kagenti.io/spiffe-helper-inject` description is consistent with the go-spiffe in-process narrative.

---

clawgenti (commented):
Documentation-only PR adding `sidecar-injection.md` and correcting `deployment-guide.md`. All previously-requested changes have been addressed: the ConfigMap example correctly uses `data.config.yaml`, the mode-resolution chain is the correct 3-layer form with the namespace ConfigMap as canonical, the `kagenti.io/spiffe-helper-inject` description is consistent with the go-spiffe in-process narrative, and all 22 CI checks pass.

---

charts/kagenti/templates/operator-client-bootstrap-job.yaml
Alan-Cha: The bootstrap Job does not use `curlimages/curl`. The image is `operator-spiffe-bootstrap` (Python) and the bootstrap logic lives in `setup_operator_client.py`, which uses the `requests` library — no `jq` involved.

charts/kagenti/templates/operator-client-bootstrap-job.yaml
Alan-Cha: The Job already uses `serviceAccountName: kagenti-operator-client-bootstrap` with a dedicated ServiceAccount, Role (granting `get` on `keycloak-initial-admin`), and RoleBinding — all in the same template file, before the Job spec.

charts/kagenti/templates/operator-client-bootstrap-job.yaml
Alan-Cha: The Python script handles this at line 167: `elif response.status_code != 404: logger.error(...)` followed by `raise`. 401/403/5xx all propagate as exceptions and fail the Job.

charts/kagenti/values.yaml
Alan-Cha: Valid long-term cleanup — noted. Not blocking this PR. A follow-up should consolidate `SPIFFE_TRUST_DOMAIN` to derive from `kagenti-operator-chart.signatureVerification.spireTrustDomain`.

docs/authbridge/deployment-guide.md
Alan-Cha: Fixed in 601d8c5c — the 'Selecting a mode' section in deployment-guide.md now shows the correct 3-layer chain with the namespace ConfigMap as canonical, and includes a note clarifying that `AgentRuntime.Spec.AuthBridgeMode` is not read by the mutating webhook.

docs/authbridge/sidecar-injection.md
Alan-Cha: Fixed in the prior commit — 'preventing the bundled spiffe-helper process from starting' → 'preventing the go-spiffe workload API source from being opened'.

docs/authbridge/deployment-guide.md
Alan-Cha: Fixed in 601d8c5c — deployment-guide.md header now reads: 'SPIRE integration is built into every combined image using the go-spiffe SDK… there is no separate spiffe-helper binary or process.'

docs/authbridge/sidecar-injection.md
Alan-Cha: Fixed — AgentRuntime.Spec.AuthBridgeMode removed from the injection precedence chain. The doc now shows the correct 3-layer chain (namespace ConfigMap → deprecated annotation → cluster default) and adds an explicit note that the CR field is enum-validated but not read by the mutating webhook.

docs/authbridge/sidecar-injection.md
Alan-Cha: Fixed — ConfigMap name corrected to `kagenti-feature-gates` (verified at `agentruntime_config.go:38`).

docs/authbridge/sidecar-injection.md
Alan-Cha: Fixed — table note corrected to 'proxy-init runs as root (UID 0) with NET_ADMIN/NET_RAW — not privileged' (verified at `container_builder.go:435-449`).

docs/authbridge/deployment-guide.md
Alan-Cha: Fixed in 71603f4d — the YAML example now shows the ConfigMap shape (`data.config.yaml`) matching the prose.

docs/authbridge/sidecar-injection.md
Alan-Cha: Fixed — the deprecated annotation table now reads: 'Set `mode:` in the namespace `authbridge-runtime-config` ConfigMap'.

docs/authbridge/sidecar-injection.md
Alan-Cha: Fixed in 71603f4d — ConfigMap shape corrected to use `data.config.yaml: | mode: proxy-sidecar`. Verified at `namespace_config.go:106`: `cfg.AuthBridgeRuntimeYAML = cm.Data["config.yaml"]`.

docs/authbridge/sidecar-injection.md
Alan-Cha: Fixed — deprecated annotation table now points to the namespace ConfigMap, not `AgentRuntime.Spec.AuthBridgeMode`.

docs/authbridge/sidecar-injection.md
Alan-Cha: I checked the source: `SpireEnableLabel = "kagenti.io/spire"` and `SpireDisabledValue = "disabled"` are defined as constants in `pod_mutator.go` but are never referenced in any evaluation path — not in `precedence.go` or anywhere else in the webhook. Setting this label has no effect. Left out of the docs to avoid misleading operators; if it should be wired in, that is a separate implementation task.

docs/authbridge/deployment-guide.md
Alan-Cha: Fixed in 601d8c5c — the 'Selecting a mode' section now shows the correct 3-layer chain (namespace ConfigMap canonical, deprecated annotation, cluster default) and the AgentRuntime YAML example has been removed.

docs/authbridge/deployment-guide.md
Alan-Cha: Fixed in 601d8c5c — step 1 now reads: 'Verify the pod has `kagenti.io/type: agent` label — the operator sets this automatically when an AgentRuntime CR targets the workload'.

docs/authbridge/sidecar-injection.md
Alan-Cha: `authbridge-lite` is the same binary as `authbridge-proxy`, just built with a reduced plugin set (jwt-validation + token-exchange, no protocol parsers). The go-spiffe SPIRE integration is present and enabled when `SPIRE_ENABLED=true`. Fixed the table row to note this explicitly.

mrsabath (approved):
## Re-review — all must-fixes resolved ✅

Re-reviewed at `601d8c5`. The two blocking issues from my last round are fixed correctly, and the earlier round's must-fixes remain correct — verified against `kagenti-operator` source:

- ✅ **ConfigMap example** now uses `data.config.yaml: | mode: …` — matches how the webhook actually reads it (`injector/namespace_config.go` reads `cm.Data["config.yaml"]`, then `ExtractMode` parses the top-level `mode:`).
- ✅ **Deprecated-annotation replacement** now points at the namespace `authbridge-runtime-config` ConfigMap instead of the injection-ignored `AgentRuntime.Spec.AuthBridgeMode`. No longer self-contradictory.
- ✅ **Bonus:** you generalized the fix to `deployment-guide.md` — its mode chain now correctly reads ConfigMap → annotation → default with a note that the mutating webhook doesn't read the CR field. That file's old "AgentRuntime CR (canonical)" step 1 was a latent version of the same bug; good catch.
- ✅ **Still correct** from the earlier round: `kagenti-feature-gates` ConfigMap name, and the `proxy-init` "runs as root (UID 0) with NET_ADMIN/NET_RAW — not privileged" wording.
- ✅ `lite` row now notes in-process SPIRE parity with the other modes.

The rest of the page remains accurate against the operator webhook source (container names, all four modes, feature-gate names/defaults, label vocabulary, SPIRE socket path). Nicely researched reference page.

My earlier `kagenti.io/spire` completeness note was non-blocking — leaving it to your discretion.

**Areas reviewed**: Docs (accuracy vs kagenti-operator source, cross-file consistency), Commit/PR conventions
**Commits**: 7 commits, all signed-off (DCO passes)
**CI status**: all green
**Verdict**: APPROVE

docs/authbridge/sidecar-injection.md
mrsabath: Resolved in `71603f4` — the example now uses `data.config.yaml: | mode: proxy-sidecar`, matching what the webhook reads (`namespace_config.go` `cm.Data["config.yaml"]` → `ExtractMode`). Verified against the current head. Thanks!

docs/authbridge/sidecar-injection.md
mrsabath: Resolved in `71603f4` (and generalized in `601d8c5`) — the replacement now points at the namespace `authbridge-runtime-config` ConfigMap, and `deployment-guide.md`'s mode chain was corrected to match. No longer contradicts the CR-not-read note. Verified. 👍

docs/authbridge/sidecar-injection.md
mrsabath: Confirmed resolved at head `601d8c5` — the mode chain (both here and in `deployment-guide.md`) now reads ConfigMap → annotation → default, with an explicit note that the mutating webhook doesn't read `AgentRuntime.Spec.AuthBridgeMode`. Matches `injector/constants.go`.

docs/authbridge/sidecar-injection.md
mrsabath: Confirmed resolved — the page now correctly names the `kagenti-feature-gates` ConfigMap.

docs/authbridge/sidecar-injection.md
mrsabath: Confirmed resolved — now reads "proxy-init runs as root (UID 0) with NET_ADMIN/NET_RAW — not privileged", matching `BuildProxyInitContainer` (`Privileged: false`).

PR #2155, opened 2026-07-07
Fix stale keycloak-admin-secret docs and remove orphaned Helm flags

### LINKED ISSUE #1337: bug: keycloak-admin-secret in agent namespaces uses stale credentials after RHBK operator upgrade
## Problem

After an upgrade, the `keycloak-admin-secret` in agent namespaces (team1, team2, etc.) may contain stale credentials (`admin/admin`) while the actual Keycloak admin is `temp-admin` with a randomly generated password. This causes the client-registration sidecar to fail with 401 errors when trying to register agents as OAuth clients in Keycloak.

## Root cause

- The Helm chart defaults `keycloak.adminUsername: admin` and `keycloak.adminPassword: admin` in `charts/kagenti/values.yaml` (lines 149-165)
- The RHBK operator creates `keycloak-initial-admin` with `temp-admin` + random password (documented in `.github/scripts/kagenti-operator/36-fix-keycloak-admin.sh`)
- The installer has logic to read `keycloak-initial-admin` and override the Helm values (`main.yml` lines 1918-1948), but this only works on a full installer run
- Partial upgrades, manual `helm upgrade`, or GitOps deployments skip this credential sync
- The `keycloak-admin-secret` is rendered by Helm with whatever values are in the release, creating a mismatch with the actual Keycloak admin credentials

## Steps to reproduce

1. Install Kagenti on OpenShift with RHBK operator managing Keycloak
2. Perform a Helm upgrade without running the full Ansible installer (or with the credential override step failing silently)
3. Deploy an agent -- the client-registration sidecar fails with 401 from Keycloak

## Suggested fix

**Option A (preferred)**: Change `charts/kagenti/templates/agent-namespaces.yaml` to NOT create `keycloak-admin-secret` with hardcoded values. Instead, use an init container or Job that reads `keycloak-initial-admin` from the Keycloak namespace at deploy time and syncs it to agent namespaces. The `kagenti-agent-oauth-secret-job` already has RBAC to read this secret -- extend it to also sync `keycloak-admin-secret`.

**Option B**: Make the installer's credential override step mandatory and fail-fast if it cannot read the source secret, rather than silently proceeding with defaults.

**Option C**: Remove hardcoded defaults from `values.yaml` entirely and require explicit credential configuration, failing early if not provided.

## Workaround

Update the secret manually in each agent namespace:
```bash
KC_USER=$(oc get secret keycloak-initial-admin -n keycloak -o jsonpath='{.data.username}' | base64 -d)
KC_PASS=$(oc get secret keycloak-initial-admin -n keycloak -o jsonpath='{.data.password}' | base64 -d)

for ns in team1 team2; do
  oc create secret generic keycloak-admin-secret -n $ns \
    --from-literal=KEYCLOAK_ADMIN_USERNAME="$KC_USER" \
    --from-literal=KEYCLOAK_ADMIN_PASSWORD="$KC_PASS" \
    --dry-run=client -o yaml | oc apply -f -
done
```

## Environment

- Platform: OpenShift 4.20.14
- Keycloak: RHBK operator-managed
- Previous creds: `admin/admin`
- Actual creds: `temp-admin/<generated>`

Alan-Cha:
## Summary

- Removes the stale \"Keycloak Admin Credentials for Agent Namespaces\" section from `docs/install.md` that described a no-longer-existing mechanism (`keycloak-admin-secret`, `keycloak.adminUsername`, `keycloak.adminPassword`, `keycloak.adminExistingSecret`)
- Replaces it with an accurate description of the actual mechanism: the `kagenti-agent-oauth-secret-job` reads `keycloak-initial-admin` at deploy time
- Removes orphaned `--set keycloak.adminUsername` / `--set keycloak.adminPassword` flags from `scripts/ocp/setup-kagenti.sh` — these values don't exist in any chart template, so they were silently ignored on every upgrade

## Root cause

The chart was already fixed (credentials are provisioned by a Job that reads `keycloak-initial-admin` directly), but the docs and the OCP setup script were never updated to match.

## Test plan

- [ ] Verify `helm template` still renders without error
- [ ] Verify `setup-kagenti.sh` completes without referencing removed flags

Closes #1337

Assisted-By: Claude Code

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -405,8 +405,8 @@ KC_USER=$(kubectl get secret keycloak-initial-admin -n keycloak -o jsonpath='{.d
 KC_PASS=$(kubectl get secret keycloak-initial-admin -n keycloak -o jsonpath='{.data.password}' | base64 -d)
 
 kubectl create secret generic keycloak-admin-secret -n kagenti-system \
-  --from-literal=username="$KC_USER" \
-  --from-literal=password="$KC_PASS" \
+  --from-literal=KEYCLOAK_ADMIN_USERNAME="$KC_USER" \
+  --from-literal=KEYCLOAK_ADMIN_PASSWORD="$KC_PASS" \
   --dry-run=client -o yaml | kubectl apply -f -
 ```
 
--- a/docs/install.md
+++ b/docs/install.md
@@ -372,41 +372,45 @@ kubectl get secret keycloak-initial-admin -n keycloak \
 
 ## Keycloak Admin Credentials for Agent Namespaces
 
-The [AuthBridge](https://github.com/kagenti/kagenti-extensions/tree/main/authbridge) stack needs Keycloak admin credentials for automatic OAuth2 client registration. These credentials are provisioned automatically at install time by the `kagenti-agent-oauth-secret-job` Helm Job — no manual configuration is required.
+The [AuthBridge](https://github.com/kagenti/kagenti-extensions/tree/main/authbridge) stack needs Keycloak admin credentials for automatic OAuth2 client registration. Two distinct secrets are involved:
+
+| Secret | Namespace | Purpose |
+|--------|-----------|---------|
+| `keycloak-admin-secret` | `kagenti-system` | Admin credentials used by the operator's ClientRegistrationReconciler and the AuthBridge client-registration sidecar to register per-agent OAuth clients in Keycloak |
+| per-agent OAuth client secrets | agent namespaces (`team1`, `team2`, …) | Per-workload credentials created *by* client registration; consumed by the agent pods |
+
+See the [Identity Guide](./identity-guide.md) for a full description of the client-registration flow.
 
 ### Automatic Provisioning
 
-During `helm install` / `helm upgrade`, the `kagenti-agent-oauth-secret-job` Job runs and:
+`keycloak-admin-secret` is created in `kagenti-system` by the installer (`36-fix-keycloak-admin.sh`), which reads the actual admin credentials from the `keycloak-initial-admin` Secret managed by the RHBK operator. Because the installer reads credentials at deploy time, the secret stays in sync across upgrades — including after an RHBK operator upgrade that rotates the `keycloak-initial-admin` password.
 
-1. Reads the actual admin credentials from the `keycloak-initial-admin` Secret in the Keycloak namespace (created and managed by the RHBK operator with a randomly-generated password).
-2. Creates the per-agent-namespace OAuth client secrets using those credentials.
+The per-agent OAuth client secrets are then provisioned by the `kagenti-agent-oauth-secret-job` Helm Job, which also reads `keycloak-initial-admin` directly.
 
-Because the Job reads credentials at deploy time rather than rendering them into the Helm release, credentials stay in sync across upgrades — including after an RHBK operator upgrade that rotates the `keycloak-initial-admin` password.
-
-The source secret name and key names are configurable via `values.yaml` if you use a different credential secret:
+The source secret name and key names are configurable via `values.yaml`:
 
 ```yaml
 keycloak:
-  adminSecretName: keycloak-initial-admin  # Secret name in keycloak namespace
-  adminUsernameKey: username               # Key within the secret
+  adminSecretName: keycloak-initial-admin  # Secret in the keycloak namespace
+  adminUsernameKey: username
   adminPasswordKey: password
 ```
 
 ### Manual Sync (break-glass)
 
-If a credential rotation leaves an agent namespace out of sync, re-run a `helm upgrade` to trigger a new Job run. Alternatively, update the secret directly:
+If a credential rotation leaves `keycloak-admin-secret` out of sync, re-run the installer or sync it manually:
 
 ```bash
 KC_USER=$(kubectl get secret keycloak-initial-admin -n keycloak -o jsonpath='{.data.username}' | base64 -d)
 KC_PASS=$(kubectl get secret keycloak-initial-admin -n keycloak -o jsonpath='{.data.password}' | base64 -d)
 
-for ns in team1 team2; do
-  kubectl create secret generic keycloak-admin-secret -n $ns \
-    --from-literal=KEYCLOAK_ADMIN_USERNAME="$KC_USER" \
-    --from-literal=KEYCLOAK_ADMIN_PASSWORD="$KC_PASS" \
-    --dry-run=client -o yaml | kubectl apply -f -
-done
+kubectl create secret generic keycloak-admin-secret -n kagenti-system \
+  --from-literal=username="$KC_USER" \
+  --from-literal=password="$KC_PASS" \
+  --dry-run=client -o yaml | kubectl apply -f -
 ```
+
+Then re-run `helm upgrade` to trigger a new `kagenti-agent-oauth-secret-job` run and re-register OAuth clients.
 
 > **Security note:** For production deployments, use a dedicated Keycloak service account with limited permissions instead of the admin account. See the [Identity Guide](./identity-guide.md) for details.
 
--- a/docs/install.md
+++ b/docs/install.md
@@ -370,13 +370,20 @@ kubectl get secret keycloak-initial-admin -n keycloak \
 
 ---
 
-## Keycloak Admin Credentials for Agent Namespaces
+## Keycloak Client Registration
 
-The [AuthBridge](https://github.com/kagenti/kagenti-extensions/tree/main/authbridge) stack needs Keycloak admin credentials for automatic OAuth2 client registration. Two distinct secrets are involved:
+The operator automatically registers an OAuth2 client in Keycloak for each agent and tool workload. There are two ways this authentication can work:
+
+| Mode | How it works | When to use |
+|---|---|---|
+| **Client secrets** (default) | Operator uses admin credentials to register clients; per-agent OAuth2 secrets are provisioned and mounted into pods | Default install; no SPIRE required |
+| **SPIFFE auth** (recommended) | Operator and agents authenticate to Keycloak using their SPIFFE JWT-SVIDs; no credentials provisioned or stored | Requires SPIRE; see [SPIFFE authentication guide](./spiffe-keycloak-auth.md) |
+
+Two secrets are involved in the default (client secrets) mode:
 
 | Secret | Namespace | Purpose |
 |--------|-----------|---------|
-| `keycloak-admin-secret` | `kagenti-system` | Admin credentials used by the operator's ClientRegistrationReconciler and the AuthBridge client-registration sidecar to register per-agent OAuth clients in Keycloak |
+| `keycloak-admin-secret` | `kagenti-system` | Admin credentials used by the operator's `ClientRegistrationReconciler` to register per-agent OAuth clients in Keycloak |
 | per-agent OAuth client secrets | agent namespaces (`team1`, `team2`, …) | Per-workload credentials created *by* client registration; consumed by the agent pods |
 
 See the [Identity Guide](./identity-guide.md) for a full description of the client-registration flow.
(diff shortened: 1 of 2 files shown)


clawgenti (commented):
Straightforward cleanup: removes the stale `keycloak-admin-secret` documentation section and the corresponding orphaned `--set keycloak.adminUsername/adminPassword` flags that were silently no-ops. All checks pass. Ready for human review.

---

Alan-Cha:
This fix is a stepping stone toward #2174 — removing the credential-provisioning mechanism entirely in favor of SPIFFE-based Keycloak authentication.

clawgenti (commented):
Straightforward doc and script cleanup — removes the stale `keycloak-admin-secret` section and the orphaned `--set keycloak.adminUsername/adminPassword` flags that were silently no-ops. All CI checks pass. Ready for human review.

---

docs/install.md
```diff
@@ -372,59 +372,40 @@ kubectl get secret keycloak-initial-admin -n keycloak \
 
 ## Keycloak Admin Credentials for Agent Namespaces
 
-The [AuthBridge](https://github.com/kagenti/kagenti-extensions/tree/main/authbridge) stack (separate sidecars or a single [combined `authbridge` container](authbridge-combined-sidecar.md)) needs Keycloak admin credentials for automatic OAuth2 client registration. These credentials are stored in a Kubernetes Secret called `keycloak-admin-secret` in each agent namespace.
+The [AuthBridge](https://github.com/kagenti/kagenti-extensions/tree/main/authbridge) stack needs Keycloak admin credentials for automatic OAuth2 client registration. These credentials are provisioned automatically at install time by the `kagenti-agent-oauth-secret-job` Helm Job — no manual configuration is required.
```
esnible: The security note below says that admin credentials aren't needed and recommends a service account with limited permissions.  Maybe say that instead of "needs admin account"?

Do we want the sidecar to have the power to do OAuth2 client registration?  This seems scary from a security perspective -- can the Kagenti operator be tricked into giving Keycloak credentials to pods that aren't necessarily sidecars owned by the system?  Id would prefer to never see any pods have Keycloak permissions that aren't under the management of a Kagenti service account.

mrsabath: **must-fix.** This declares `keycloak-admin-secret` gone, but it's still live — it moved to `kagenti-system` (kagenti-operator#321), it wasn't removed. `docs/identity-guide.md` (lines 264/479/628) documents the operator's ClientRegistrationReconciler reading it from the operator namespace, and the AuthBridge client-registration sidecar mounts it (`.github/scripts/kind/91-run-authbridge-weather-e2e.sh`). 

Also, the `kagenti-agent-oauth-secret-job` reads `keycloak-initial-admin` to create the per-agent **OAuth client** secrets — a different secret and purpose than the `keycloak-admin-secret` **admin credential** the reconciler/sidecar use. Please retarget this section to the real current state (operator-managed `keycloak-admin-secret` in `kagenti-system`) and cross-link `identity-guide.md`, rather than describing it as removed.

docs/install.md
```diff
@@ -372,59 +372,40 @@ kubectl get secret keycloak-initial-admin -n keycloak \
...
-
-If you need to create or update the secret manually in an agent namespace:
-
-```bash
-kubectl create secret generic keycloak-admin-secret -n <agent-namespace> \
-  --from-literal=KEYCLOAK_ADMIN_USERNAME=admin \
-  --from-literal=KEYCLOAK_ADMIN_PASSWORD=admin \
-  --dry-run=client -o yaml | kubectl apply -f -
-```
+### Manual Sync (break-glass)
```
mrsabath, line 406: **must-fix (internal contradiction).** This break-glass snippet recreates `keycloak-admin-secret` — the very secret the top of the section calls "no-longer-existing" — and does so in the agent namespaces (`team1`/`team2`). Per operator#321 the operator now reads it from `kagenti-system`, so a manual sync into agent namespaces targets the wrong place. If a manual-sync path is kept, it should write to `kagenti-system` (or match wherever the sidecar mounts it), and the section shouldn't simultaneously claim the secret no longer exists.

mrsabath (changes requested):
## Summary

The **script change is correct and a genuine fix** — but the **docs rewrite is inaccurate** and, as written, would make `docs/install.md` contradict `docs/identity-guide.md` in the same repo. Requesting changes on the docs half only.

### ✅ What's right

- **Removing `--set keycloak.adminUsername` / `keycloak.adminPassword` from `scripts/ocp/setup-kagenti.sh`** — verified against `charts/kagenti/values.yaml`: those keys don't exist, so the flags were silently ignored on every run. Good cleanup.
- **The new `adminSecretName` / `adminUsernameKey` / `adminPasswordKey` values** documented for the Job are real — they exist in `values.yaml` (lines 261-263) and are wired into `templates/agent-oauth-secret-job.yaml` (env `KEYCLOAK_ADMIN_SECRET_NAME` etc.). Accurate.

### ❌ What blocks (docs accuracy)

The rewrite's premise — that `keycloak-admin-secret` is a *"no-longer-existing mechanism"* — is not correct. That secret is still a live, security-relevant part of client registration; it **moved**, it wasn't removed (kagenti-operator#321):

- `docs/identity-guide.md` (this PR doesn't touch it) still documents it as current: *"the operator reads Keycloak admin credentials from `keycloak-admin-secret` in the operator namespace (`kagenti-system`)"* (lines 264, 479, 628).
- The operator's **ClientRegistrationReconciler** reads it, and the **AuthBridge client-registration sidecar** mounts it — see `.github/scripts/kagenti-operator/36-fix-keycloak-admin.sh` and `.github/scripts/kind/91-run-authbridge-weather-e2e.sh` (*"the client-registration sidecar … mounts this secret from the pod's namespace. Since operator#321 the secret lives only in kagenti-system"*).

Separately, the PR **conflates two different secrets**: `keycloak-admin-secret` is the *admin credential* the reconciler/sidecar authenticate with; the `kagenti-agent-oauth-secret-job` reads `keycloak-initial-admin` to provision the *per-agent OAuth client* secrets. Replacing the whole admin-credentials section with a description of the OAuth-client Job drops the mechanism `identity-guide.md` still relies on.

Suggested fix: rather than deleting the section, retarget it to the actual current state — `keycloak-admin-secret` is now operator-managed in `kagenti-system` (operator#321), and cross-link `identity-guide.md`. The orphaned-`--set`-flags removal is independently correct and can stay.

**Areas reviewed**: Docs (accuracy/cross-consistency), Shell, Commit/PR conventions
**Commits**: 1 commit, signed-off (DCO passes)
**CI status**: green except `Deploy & Test` (pending)
**Verdict**: REQUEST_CHANGES — 1 must-fix docs-accuracy issue; script change approved.

docs/install.md
```diff
@@ -372,59 +372,40 @@ kubectl get secret keycloak-initial-admin -n keycloak \
...
-### Verifying
+If a credential rotation leaves an agent namespace out of sync, re-run a `helm upgrade` to trigger a new Job run. Alternatively, update the secret directly:
 
 ```bash
-kubectl get secret keycloak-admin-secret -n team1
+KC_USER=$(kubectl get secret keycloak-initial-admin -n keycloak -o jsonpath='{.data.username}' | base64 -d)
+KC_PASS=$(kubectl get secret keycloak-initial-admin -n keycloak -o jsonpath='{.data.password}' | base64 -d)
+
+for ns in team1 team2; do
+  kubectl create secret generic keycloak-admin-secret -n $ns \
```
esnible: **suggestion** — This break-glass snippet manually creates a secret named `keycloak-admin-secret`, but this PR's whole thesis is that `keycloak-admin-secret` is no longer part of the mechanism. The Job (`kagenti-agent-oauth-secret-job`) reads `keycloak-initial-admin` and registers the OAuth clients directly — it never consumes a per-namespace `keycloak-admin-secret`. So this manual `kubectl create secret` recreates an obsolete secret that nothing reads, and likely has no effect.

The `helm upgrade` re-run documented just above is the correct break-glass path. Consider dropping this `kubectl create secret` alternative so the fix doesn't reintroduce the exact stale-secret confusion it's removing.

esnible (commented):
Clean, well-scoped docs + script correction. I verified every removed and added value name against `upstream/main`:

- ✅ `keycloak.adminUsername` / `keycloak.adminPassword` / `keycloak.adminExistingSecret` and the `keycloak-admin-secret` name are all gone from `charts/kagenti/` — the removed `--set` flags in `setup-kagenti.sh` were genuine no-ops.
- ✅ The new documented values (`adminSecretName`, `adminUsernameKey`, `adminPasswordKey`) exist in `charts/kagenti/values.yaml`.
- ✅ `kagenti-agent-oauth-secret-job` exists and reads `keycloak-initial-admin` at deploy time (RBAC scoped to `adminSecretName`), matching the new description.
- ✅ `KC_ADMIN_FLAGS` removed cleanly with no dangling references; the `helm upgrade` block is intact.

One suggestion inline about the break-glass snippet. Leaving as a comment (not an approval).

**Author**: Alan-Cha (MEMBER — maintainer) · **Areas**: Docs, Shell · **Agent/IDE config**: none · **CI**: passing (Deploy & Test still pending)

Assisted-By: Claude Code

Alan-Cha:
Thanks for the thorough review @mrsabath and @esnible. You're both right — I incorrectly described `keycloak-admin-secret` as removed when it had actually moved to `kagenti-system` (operator#321).

Fixed in the two new commits:
- Rewritten section now accurately describes both secrets (admin creds in `kagenti-system` vs per-agent OAuth client secrets in agent namespaces), with a table distinguishing them and a cross-link to `identity-guide.md`
- Break-glass snippet now targets `kagenti-system` with the correct key names (`KEYCLOAK_ADMIN_USERNAME`/`KEYCLOAK_ADMIN_PASSWORD`) as used in `36-fix-keycloak-admin.sh`
- Dropped the stale per-agent-namespace `kubectl create secret` alternative that was recreating the wrong secret in the wrong place

clawgenti (commented):
Docs and script fix looks correct at HEAD (`f8e9dd25`): `keycloak-admin-secret` is accurately described as living in `kagenti-system`, the two-secret table clearly separates admin creds from per-agent OAuth client secrets, the break-glass snippet targets the right namespace and key names, and the orphaned `--set` flags are cleanly removed from `setup-kagenti.sh`. All CI checks pass.

---

mrsabath (approved):
Both must-fix items from my earlier review are resolved — verified against `upstream/main`:

- **`keycloak-admin-secret` accuracy**: the section no longer claims it was removed. The two-secret table correctly separates the operator-managed admin credential (`keycloak-admin-secret` in `kagenti-system`, per operator#321) from the per-agent OAuth client secrets, and cross-links `identity-guide.md`. Consistent with `identity-guide.md` lines 264/479/628.
- **Break-glass snippet**: now targets `kagenti-system` with the same secret name and `KEYCLOAK_ADMIN_USERNAME`/`KEYCLOAK_ADMIN_PASSWORD` keys as the authoritative installer `36-fix-keycloak-admin.sh`. Correct.

The orphaned `--set keycloak.adminUsername/adminPassword` removal was independently correct — those keys don't exist in `values.yaml`, so the flags were silent no-ops.

All static CI checks pass. Verdict: APPROVE.

Alan-Cha:
Superseded by #2211 which consolidates the install.md fixes along with a full symmetric authentication guide.

Alan-Cha:
All four comments are addressed in the superseding PR #2211.

The `docs/authentication.md` in #2211 now correctly describes the client-secrets flow:
- The `kagenti-agent-oauth-secret-job` creates `kagenti-keycloak-client-secret` in each **agent namespace** with per-workload OAuth2 client credentials — it does not create `keycloak-admin-secret` in `kagenti-system`
- The operator's `ClientRegistrationReconciler` reads `keycloak-initial-admin` directly from the `keycloak` namespace — there is no intermediate `keycloak-admin-secret` in the operator-managed path
- The break-glass section has been replaced with an operator restart + log verification, removing the incorrect `kubectl create secret keycloak-admin-secret` command that targeted the wrong secret

PR #2157, opened 2026-07-07
feat(skills): make in-cluster skillberry-store run the Claude Code CLI

### LINKED ISSUE #2156: In-cluster skillberry-store cannot generate skills: Claude Code CLI missing and blocked under root
## Problem

The in-cluster `skillberry-store` skill-generation feature shells out to the **Claude Code CLI**, but three things prevent it from working with the upstream `ghcr.io/skillberry-ai/skillberry-store:0.2.0` image:

1. **CLI not bundled** — skill generation fails with `Claude Code not found`.
2. **SSL hostname mismatch** — when `ANTHROPIC_BASE_URL` is set to a raw IP, `claude` fails with `Unable to connect to API: SSL certificate hostname mismatch` (the LiteLLM cert is issued for the hostname). This belongs in the env overlay, not the chart.
3. **Blocked under root** — the plugin invokes `claude --dangerously-skip-permissions`, which the CLI refuses under root/sudo: `--dangerously-skip-permissions cannot be used with root/sudo privileges for security reasons`.

## Proposed fix (chart)

- Install `@anthropic-ai/claude-code` at container startup via the existing entrypoint (`command:`) override so it lands on PATH before the app starts.
- Set `IS_SANDBOX=1` so the CLI accepts `--dangerously-skip-permissions` under root (the pod is the sandbox).
- (Overlay, not chart) point `ANTHROPIC_BASE_URL` at the LiteLLM **hostname**, not its IP.

## Notes / follow-ups

- The CLI wants Node >= 22; the image ships Node 20.19.2 — works today (engine warning only). Baking a derived image (Node 22 + preinstalled CLI) would remove the ~15s per-start npm fetch and the network dependency.

Verified end-to-end on a live Kind cluster: a headless `claude -p` call inside the pod using only pod-native env returns successfully after these changes.

eranra:
Closes #2156.

## Summary

The in-cluster `skillberry-store`'s skill-generation feature shells out to the **Claude Code CLI**, which is not bundled in the upstream `ghcr.io/skillberry-ai/skillberry-store:0.2.0` image. This wires the deployment so the CLI is present and usable at runtime.

## Changes

- **`values.yaml`** — extend the entrypoint (`command:`) override to `npm install -g @anthropic-ai/claude-code` at startup (best effort, `|| true`) so it lands on PATH (`/usr/local/bin`) before the app process starts, then `exec` the app.
- **`skillberry-store.yaml`** — set `IS_SANDBOX=1` so the CLI accepts `--dangerously-skip-permissions` under root (the pod is the sandbox); add `HF_HOME` on the persistent volume so the embedding model downloads once; loosen liveness/readiness/startup probes to absorb CPU-heavy embedding work during skill creation.

## Design decisions

- **Startup install vs. derived image:** chosen to avoid a new build pipeline and to fit the existing `command:` override pattern. Trade-off is a ~15s network fetch on each pod start; the values.yaml comment documents baking a derived image as the alternative.
- **`IS_SANDBOX=1` vs. running non-root:** keeping the container as root avoids PVC-ownership and `npm install -g` write-permission issues; the pod boundary is the sandbox.
- **`ANTHROPIC_BASE_URL`** is intentionally **not** hardcoded in the chart — it belongs in the env overlay and must use the LiteLLM **hostname** (not IP) to satisfy TLS cert validation.

## Testing

- `helm lint` and `helm template` render cleanly with the new `command`/env.
- `make lint` (pre-commit) passes on both files.
- Verified end-to-end on a live Kind cluster: after a rollout, the pod auto-installs the CLI, `IS_SANDBOX=1` is set, and a headless `claude -p` call using only pod-native env returns successfully.

### DIFF
--- a/charts/kagenti/values.yaml
+++ b/charts/kagenti/values.yaml
@@ -76,6 +76,22 @@ skillberryStore:
     repository: ghcr.io/skillberry-ai/skillberry-store
     tag: "0.2.0"             # override via --set or values overlay
     pullPolicy: IfNotPresent
+    # Override the container entrypoint. The default image command (`make run`)
+    # re-runs `uv install` at startup and requires network access to github.com.
+    # This bypasses that by running the pre-installed venv directly.
+    #
+    # The store's skill-generation feature shells out to the Claude Code CLI,
+    # which is NOT bundled in the upstream image. Install it at startup (best
+    # effort; `|| true` keeps the store up if the npm fetch fails) so it lands
+    # on PATH (/usr/local/bin) before the app process starts. This is a network
+    # fetch on every pod start (~15s); bake it into a derived image if that
+    # startup cost or the network dependency is unacceptable.
+    command:
+      - sh
+      - -c
+      - |
+        npm install -g @anthropic-ai/claude-code || true
+        exec .venv/bin/python -m skillberry_store.main
   imagePullSecrets: []        # public image; empty default
   apiPort: 8000
   uiPort: 8002
--- a/charts/kagenti/templates/skillberry-store.yaml
+++ b/charts/kagenti/templates/skillberry-store.yaml
@@ -28,6 +28,9 @@ spec:
       - name: skillberry-store
         image: "{{ .Values.skillberryStore.image.repository }}:{{ .Values.skillberryStore.image.tag }}"
         imagePullPolicy: {{ .Values.skillberryStore.image.pullPolicy }}
+        {{- with .Values.skillberryStore.image.command }}
+        command: {{- toYaml . | nindent 10 }}
+        {{- end }}
         env:
         - name: SBS_BASE_DIR
           value: "{{ .Values.skillberryStore.baseDir }}"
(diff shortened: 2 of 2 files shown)


eranra:
@PrielHaz FYI ^^^

(In the UI the execution of the plugin's when using Kagenti is direct and not inside a pod since the store instance is running in a pod/container anyways)

charts/kagenti/values.yaml
```diff
@@ -76,6 +76,22 @@ skillberryStore:
...
+    # which is NOT bundled in the upstream image. Install it at startup (best
+    # effort; `|| true` keeps the store up if the npm fetch fails) so it lands
+    # on PATH (/usr/local/bin) before the app process starts. This is a network
+    # fetch on every pod start (~15s); bake it into a derived image if that
+    # startup cost or the network dependency is unacceptable.
+    command:
+      - sh
+      - -c
+      - |
+        npm install -g @anthropic-ai/claude-code || true
```
pdettori, line 93: **suggestion**: `@anthropic-ai/claude-code` is unpinned, so it resolves to `@latest` on every pod start — different pods can get different CLI versions, and it's a first-party-but-unpinned fetch on each boot. This repo otherwise pins aggressively (Action SHAs, release image tags, the *Check Release Pins* CI). Consider pinning `@anthropic-ai/claude-code@<version>` for reproducibility.

Also, `|| true` masks a failed npm fetch and produces a *live-but-broken* pod where skill generation fails only later at runtime. Worth emitting an error log on failure so the broken state is diagnosable rather than silent.

charts/kagenti/values.yaml
```diff
@@ -76,6 +76,22 @@ skillberryStore:
...
+        exec .venv/bin/python -m skillberry_store.main
```
pdettori, line 94: **nit**: `exec .venv/bin/python ...` is a relative path that depends on the image `WORKDIR`. It works today (verified on Kind), but a future base-image `WORKDIR` change would silently break startup. An absolute path or an explicit `cd` to the app dir would be more robust.

pdettori (approved):
## Summary

Wires the `skillberry-store` Helm deployment so the Claude Code CLI is present at runtime: adds a `command:` entrypoint override that `npm install -g`s the CLI, sets `IS_SANDBOX=1` (to allow `--dangerously-skip-permissions` under root), points `HF_HOME` at the persistent volume, and adds/loosens probes to absorb CPU-heavy embedding work.

Well-documented, verified live on Kind, DCO-signed, and all CI green. The `IS_SANDBOX=1` + root tradeoff is sound and clearly justified inline (the pod boundary is the sandbox). Approving with two non-blocking inline notes on the startup entrypoint.

**Areas reviewed**: Helm/K8s template, YAML values, Security

_Assisted-By: Claude Code_

PR #2158, opened 2026-07-07
Fix: Don't forget about failed and progressing Tool builds

esnible:
## Summary

For #1906

See also https://github.com/kagenti/kagenti/pull/2111 , the same changes for agent builds (already merged)

Deploy Rossi
Bring in this PR
make build-load-ui-frontend and make build-load-ui-backend plus the usual rollout.

### DIFF
--- a/kagenti/backend/app/routers/tools.py
+++ b/kagenti/backend/app/routers/tools.py
@@ -498,9 +498,7 @@ def _get_tool_build_status(kube: KubernetesService, name: str, namespace: str) -
     return "Build Failed" if phase == "Failed" else "Building"
 
 
-def _get_tool_build_detail(
-    kube: KubernetesService, name: str, namespace: str
-) -> Optional[dict]:
+def _get_tool_build_detail(kube: KubernetesService, name: str, namespace: str) -> Optional[dict]:
     """Build a synthetic get_tool response for a source-built tool with no workload.
 
     Returns a response dict (same shape as get_tool's workload response) with a
--- a/kagenti/backend/app/routers/tools.py
+++ b/kagenti/backend/app/routers/tools.py
@@ -459,6 +459,100 @@ def _get_workload_status(workload: dict) -> str:
     return "Not Ready"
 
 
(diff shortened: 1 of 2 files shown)


kagenti/backend/app/routers/tools.py
```diff
@@ -459,6 +459,98 @@ def _get_workload_status(workload: dict) -> str:
...
+        build = kube.get_custom_resource(
+            group=SHIPWRIGHT_CRD_GROUP,
+            version=SHIPWRIGHT_CRD_VERSION,
+            namespace=namespace,
+            plural=SHIPWRIGHT_BUILDS_PLURAL,
+            name=name,
+        )
+    except ApiException as e:
+        if e.status != 404:
+            logger.warning("Failed to get Shipwright Build for a tool")
```
pdettori, line 520: nit (non-blocking): a **non-404** `ApiException` from this Build lookup is logged and swallowed (`return None`), so `get_tool` then reports `404 "Tool not found"` — masking a transient server error as not-found. Defensible for a best-effort fallback, but a short comment noting the intentional degradation would make the intent explicit.

pdettori (approved):
A focused, well-tested fix. `list_tools` and `get_tool` now surface source-built tools that have no workload yet — mapping the latest Shipwright BuildRun phase to `Building` / `Build Failed`, skipping `Succeeded` (the workload path handles it), and deduping against existing workloads. Mirrors the already-merged agent equivalent (#2111), extracted into two clean helpers, and additionally adds the fallback to the detail endpoint. Guards ensure a build-listing failure never breaks the core tool list.

**Areas reviewed**: Python (routers/tools.py), Tests
**CI**: green (E2E HyperShift pending `/run-e2e`, expected for PRs)

Verified against `upstream/main`: `existing_names` is the correct dedup set (not a copy-paste of the agent path's `agent_names`), `ToolSummary` has every field used, `get_tool` initializes `workload = None` so the new fallback is null-safe, and the CodeQL log-injection pattern (constant log messages) is preserved. Comprehensive tests (11 cases across both endpoints + the failure guard). One non-blocking nit inline.

Assisted-By: Claude Code

Issue #2100, opened 2026-06-26
MCP Inspector Connect fails on self-signed-cert OpenShift clusters — proxy Route cert not browser-trusted


cwiklik:
## Summary

On OpenShift clusters whose router uses a **self-signed / non-browser-trusted certificate**, the MCP Inspector **Connect** fails even after the Inspector Proxy Address is correctly populated (see #2085, which fixes the empty-address symptom). The browser console shows:

```
https://mcp-proxy-<ns>.apps.<cluster>/health  net::ERR_CERT_AUTHORITY_INVALID
checkProxyHealth → connect → "Couldn't connect to MCP Proxy Server: Failed to fetch"
```

## Root cause

The MCP Inspector and its proxy are exposed as **two separate Route hosts** (`mcp-inspector-<ns>…` and `mcp-proxy-<ns>…`). On Connect, the Inspector UI (served from the inspector host) makes a cross-origin `fetch()` to the **proxy host's** `/health`.

Browser cert trust is **per host**. A user reaches the Inspector UI (and Keycloak, and the Kagenti UI) by navigating to those hosts directly and accepting the self-signed cert. But the **proxy host is never visited directly** — it's only hit via a background `fetch()`, which **cannot prompt** the user to accept an untrusted cert. So the health check fails and Connect aborts.

This is **independent of #2085** (that fix passes the proxy address; this is a TLS-trust problem) and **independent of Kagenti code** — the proxy, CORS, auth, and the proxy→gateway MCP handshake all work (verified server-side). It only manifests on clusters without a browser-trusted (CA-signed) router cert.

## Impact

- Clusters with **CA-signed / trusted** router certs (typical managed OpenShift): unaffected — Connect works with the #2085 fix.
- **Self-signed** clusters (dev/ROSA-on-kubestellar etc.): Connect fails unless the user manually trusts the `mcp-proxy` host cert or the cluster CA.

## Proposed fixes

1. **Serve the Inspector proxy and UI on the same host (one Route, path-based)** so the connection is same-origin and uses the single already-trusted cert — eliminates both the cross-origin fetch and the separate cert acceptance. *(Recommended, robust.)*
2. Or **document** the prerequisite: on self-signed clusters, trust the `mcp-proxy` Route cert (or the cluster CA) before using the Inspector.

## Workaround (today)

Open `https://mcp-proxy-<ns>.apps.<cluster>/health` directly, accept the cert warning, then retry Connect.

[...]

Issue #2085, opened 2026-06-24
Bug: MCP Inspector 'Connect' button fails on OpenShift — Inspector Proxy Address not set


pdettori:
## Description

When installing Kagenti on OpenShift and navigating to **MCP Gateway → MCP Inspector**, the **"Connect"** button does not work because the **"Inspector Proxy Address"** field is not populated.

## Steps to Reproduce

1. Install Kagenti on an OpenShift cluster
2. Open the Kagenti UI
3. Navigate to **MCP Gateway** page
4. Click on **MCP Inspector**
5. Observe that the "Inspector Proxy Address" field is empty
6. Click the "Connect" button

## Expected Behavior

The Inspector Proxy Address should be automatically configured (or derived from the OpenShift route), and clicking "Connect" should establish a connection to the MCP Inspector proxy.

## Actual Behavior

The "Connect" button does not work because the "Inspector Proxy Address" is not set. The field is empty and no connection can be established.

## Environment

- Platform: OpenShift
- Component: MCP Gateway / MCP Inspector UI

PR #2159, opened 2026-07-07
Fix: serve MCP Inspector UI and proxy on one host (#2100)

cwiklik:
## Summary

On OpenShift clusters with a **self-signed router cert**, MCP Inspector **Connect** fails with `ERR_CERT_AUTHORITY_INVALID`. The Inspector UI (served from the `mcp-inspector-<ns>…` host, whose cert the user accepts when the page loads) makes a background `fetch()` to a **separate** `mcp-proxy-<ns>…` host. That host is never visited directly, so its self-signed cert is never accepted — and a background fetch can't prompt for it. Fixes the Connect failure described in #2100.

## Fix — serve UI + proxy on one host via an nginx sidecar

`charts/kagenti-deps/templates/mcp-inspector.yaml`:
- Add an **nginx reverse-proxy sidecar** in the mcp-inspector pod: `/` → `127.0.0.1:6274` (UI), `/mcp-proxy/` → `127.0.0.1:6277` (proxy, prefix-stripped), SSE/WebSocket-aware (`proxy_buffering off`, long read timeout).
- Collapse the Service to a single port (`8080` → nginx) and to **one** Route (OCP) / HTTPRoute (Kind); **delete** the separate `mcp-proxy` Route/HTTPRoute.
- OCP Route gets `haproxy.router.openshift.io/timeout: 1h` so the long-lived `streamable-http`/SSE stream survives HAProxy's ~30s default.
- nginx `pid`/temp paths under `/tmp` so it runs under OpenShift's restricted SCC (arbitrary non-root UID).

Plumbing so the browser is handed the same-origin proxy address:
- `charts/kagenti/templates/ui-routes-job.yaml` (OCP): `MCP_PROXY_FULL_ADDRESS` now **reuses the `mcp-inspector` route + a `/mcp-proxy` path suffix** instead of discovering a separate `mcp-proxy` route.
- `charts/kagenti/templates/ui.yaml` (Kind): `MCP_PROXY_FULL_ADDRESS = http://mcp-inspector.<domain>:8080/mcp-proxy`.
- `charts/kagenti-deps/values.yaml`: `mcpInspector.nginxImage` + `routeTimeout`.

Result: the proxy is same-origin with the UI, so the **one** cert the user already accepted covers the proxy calls too. **No frontend/backend code change** (the UI already forwards `MCP_PROXY_FULL_ADDRESS`; `ALLOWED_ORIGINS`, built from the inspector host, is now correctly same-origin).

## Why this approach (vs. the alternatives)

The Inspector can't natively serve UI + proxy on one host — its UI calls the proxy at the **absolute** `MCP_PROXY_FULL_ADDRESS`, its ports are hardcoded, and it has no base-path option. So same-origin has to be arranged around it. Three options were considered:

1. **Ingress-only path rewrite (no sidecar).** On Kind this is a clean single HTTPRoute with `URLRewrite`, but on OCP it needs **two Routes sharing the inspector's auto-generated host** — a host we don't know at template time, forcing a manual `appDomain`/`openshiftDomain` value (fragile: get it wrong and the proxy Route lands on an unreachable host, recreating this very bug), or pushing host-stitching into the imperative route-job. It also relies on the Inspector's **SSE** stream surviving an **edge-Route rewrite**, which is unproven and prone to HAProxy timeout/buffering quirks.
2. **Documentation only.** Tell users to open the `mcp-proxy` URL once and accept its cert. Zero code, but doesn't fix the UX and leaves a confusing manual step.
3. **nginx sidecar (this PR).** Collapses to **one auto-generated host** — no `appDomain` input, nothing to misconfigure — and handles SSE/WebSocket explicitly (`proxy_buffering off`, long timeouts) rather than hoping an edge rewrite does. Fully declarative in the chart, no frontend change, and the route-discovery job gets *simpler* (one host, not two).

Option 3 removes **two** sources of fragility (the OCP host-sharing config and the SSE-through-edge-rewrite risk) at the cost of one lightweight `nginx:1.27-alpine` sidecar (no build, kagenti already uses nginx in the UI). That trade — a few lines of chart YAML for a robust, self-contained, misconfiguration-proof fix — is why it's the optimal choice.

## Validation

- `helm lint` (kagenti-deps + kagenti, Kind & OCP) passes; renders confirm one host, the nginx sidecar, and the correct `MCP_PROXY_FULL_ADDRESS`.
- **Kind:** Connect works via the Kagenti UI (nginx path-routing + SSE proven).
- **OpenShift (self-signed router cert):** curl `/mcp-proxy/health`+`/config`+`/mcp` all `200` same-origin over the self-signed cert; then the **full real-UI flow** works (Kagenti UI → MCP Gateway → Open MCP Inspector → Connect) with no `ERR_CERT_AUTHORITY_INVALID`.

## Note

The UI passing `MCP_PROXY_FULL_ADDRESS` to the Inspector is a separate, already-merged fix (#2085/#2101); it ships in the next alpha frontend image (the released `v0.7.0-alpha.3` predates it). This PR is chart-only and depends on that FE fix being present to exercise the UI path.

Refs #2100

Assisted-By: Claude Code


charts/kagenti/templates/ui-routes-job.yaml
```diff
@@ -108,7 +108,10 @@ data:
     # Define the list of routes to process.
     ROUTES_DATA=(
       "kagenti-system,mcp-inspector,true,MCP_INSPECTOR_URL"
-      "kagenti-system,mcp-proxy,true,MCP_PROXY_FULL_ADDRESS"
+      # The proxy is served on the inspector host under /mcp-proxy by the nginx
+      # sidecar (single origin), so reuse the mcp-inspector route + path suffix
+      # instead of a separate mcp-proxy route. See kagenti#2100.
```
rubambiza: **Nit**: I see a lot of references to kagenti#2100 in the comments. These might go stale if the repo(s) get renamed. I think the comment are descriptive enough that we can forego these mentions.

rubambiza (commented):
One minor nit from human review. AI-assisted review coming next.

rubambiza (approved):
The PR does exactly what it claims and is cleanly scoped. Verified against source:

- The 5-field `ROUTES_DATA` entry (`...,MCP_PROXY_FULL_ADDRESS,/mcp-proxy`) is parsed correctly by `process_routes.sh` (`path_suffix` at line 128, appended at line 135) — the mechanism already exists (mlflow uses it), and empty suffixes on 4-field entries expand harmlessly.
- `ALLOWED_ORIGINS` is built from the inspector host, so serving the proxy same-origin keeps it valid — no backend/frontend change needed, as stated.
- nginx sidecar has resource requests/limits (matches the inspector container's convention), runs `/tmp` paths for the restricted SCC, and handles SSE/WebSocket explicitly.
- All `ui`/`proxy` Service port names and `targetPort` refs are consistently replaced by the single `http`/`8080` port; no dangling references remain in the chart.

**Areas reviewed**: Helm/K8s, Shell (route-job script), YAML
**CI**: all required checks passing; E2E HyperShift pending `/run-e2e`

### Suggestion (follow-up, not blocking — files outside this diff)

Since this PR **deletes** the separate `mcp-proxy` Route, two docs now reference a route that no longer exists:

- `docs/install.md:353-357` — the "For MCP Inspector, also accept the proxy certificate" step runs `kubectl get route mcp-proxy ...`, which will now return an empty host. This step is exactly what the PR makes obsolete (proxy is same-origin under `/mcp-proxy`, one cert covers both). Recommend removing/rewording it here or in a follow-up.
- `docs/identity-guide.md:205` — also references the `mcp-proxy` route, but this is a **pre-existing** bug unrelated to this PR: it's under "Keycloak Admin Access" and points at the wrong route entirely (should be `keycloak`). Worth a separate fix.

Assisted-By: Claude Code

cwiklik:
Thanks @rubambiza — addressed in `22760b7d`:

- ✅ **Nit (issue refs in comments):** removed the `kagenti#2100` mentions from all chart comments; kept the descriptive text so they won't go stale on a repo rename.
- ✅ **`docs/install.md` obsolete step:** removed the separate "accept the mcp-proxy certificate" step — the proxy is now same-origin under the Inspector host, so accepting the Inspector cert covers it.
- ↪️ **`docs/identity-guide.md:205`:** pre-existing, unrelated bug (it points "Keycloak Admin Access" at the wrong route) — leaving it out of this PR to stay focused; I'll file it as a separate fix.

Assisted-By: Claude Code

PR #2169, opened 2026-07-08
feat(simulation): provision simulated-tool workload (StatefulSet + PVC + Service)

bnayahu:
## Summary

Implements #2161 (part of epic #2151): a flag-gated backend create path that provisions a **simulated tool's workload** from an OpenAPI spec — a StatefulSet (`replicas: 1`, HPA-off marker) with a per-tool PVC mounted at the harness skills dir, plus a Service — discoverable by the MCP Gateway exactly like a real tool. Generation orchestration (posting the spec to the harness + status polling) is the next slice, #2162.

## What's included

- **`POST /api/v1/simulation/tools`** (`ROLE_OPERATOR`, flag-gated): validates the OpenAPI spec syntactically → **422 with no workload created** on invalid input; otherwise provisions the workload and returns **202** with `status: "Generating"`.
- **Pure-standalone manifest builders** in new `app/services/simulation_manifests.py` — **zero changes to `tools.py`** (minimal blast radius). A **parity test** asserts the simulated StatefulSet shares the tool StatefulSet's identity/mesh label subset + `securityContext`, guarding against drift.
- **StatefulSet**: `replicas: 1`, per-tool PVC (`volumeClaimTemplate` `data`) at `/app/skills-store`, `cache`/`tmp` emptyDirs, `/readyz`+`/healthz` probes, tool-parity `securityContext`.
- **Labels** on workload + service: `protocol.kagenti.io/mcp`, `kagenti.io/type=tool` (lists in the Tool Catalog), `kagenti.io/simulated="true"`; SPIRE + AuthBridge-inject honored when requested. HPA-off marker `kagenti.io/autoscaling="disabled"` (declarative).
- **No `AgentRuntime` CR** — the `type=tool` label is set directly (no new CRD, no operator changes).
- **`LLM_API_KEY`/`LLM_API_BASE`** via the generic `envVars` + `secretKeyRef` mechanism (no bespoke wiring).
- Additive `simulation_harness_image` setting and `kagenti.io/simulated` / HPA-off / skills-mount constants.

## Image pull secret (private-registry interim)

The simulation-harness image (`ghcr.io/kagenti/simulation-harness`) currently lives in a **private** GHCR package, so the provisioned pod needs an `imagePullSecrets` entry to start. Added, backend-only:

[...]

PR #2171, opened 2026-07-08
fix(k8s): sanitize user-controlled values in KubernetesService logs (CWE-117)

### LINKED ISSUE #2170: Log injection (CWE-117) in KubernetesService logging
## Summary

`kagenti/backend/app/services/kubernetes.py` logs user-controlled `namespace`, `name`, and `plural` values via f-strings in ~28 `logger.*` calls **without sanitization**, e.g.:

```python
logger.error(f"Error creating StatefulSet in {namespace}: {e}")
logger.info(f"Created ServiceAccount '{name}' in {namespace}")
logger.error(f"Error getting {plural}/{name} in {namespace}: {e}")
```

These values originate from HTTP request bodies (namespace/name in tool, agent, and simulation create paths). A value containing CR/LF/NUL can forge or split log entries (log injection, **CWE-117**). CodeQL (`py/log-injection`, `security-extended`) flags these as high/medium.

## Evidence

Surfaced by CodeQL on PR #2169 (simulated-tool workload), whose new create endpoint added another request→log data-flow path into these shared sinks. The sinks are **pre-existing** and already reachable from existing callers (e.g. `create_tool` in `tools.py` passes `request.namespace` into `ensure_service_account`/`create_statefulset`/`create_service`). This is shared-infrastructure tech debt, not specific to the simulation feature.

## Proposed fix

Apply the module's existing `_sanitize()` helper (strips `\n`/`\r`/`\x00`; CodeQL recognizes it as a sanitizer) to `namespace`/`name`/`plural` in every affected method, using the reassign-at-top idiom already established in `get_endpoint_slices` and `delete_service`:

```python
def create_statefulset(self, namespace: str, body: dict) -> dict:
    namespace = _sanitize(namespace)
    ...
```

Comprehensive across all affected methods (list/get/create/delete/patch for custom resources, deployments, services, statefulsets, jobs, service accounts, configmaps).

## Acceptance criteria

- No `logger.*` call in `kubernetes.py` interpolates an unsanitized `namespace`/`name`/`plural`.
- CodeQL `py/log-injection` alerts for `kubernetes.py` clear.
- No behavior change for valid inputs (sanitize is a no-op on values without control chars).
- Existing backend tests pass.

bnayahu:
## Summary

Fixes **CWE-117 log injection** (#2170) in `KubernetesService`. ~28 `logger.*` f-strings across the service interpolated user-controlled `namespace` / `name` / `plural` values (from tool/agent/simulation create paths) without sanitization, letting a value with CR/LF/NUL forge or split log entries. CodeQL (`py/log-injection`, `security-extended`) flags these as high/medium.

## Fix

Apply the module's existing `_sanitize()` helper (strips `\n`/`\r`/`\x00`; CodeQL-recognized sanitizer) to the affected parameters, using the **reassign-at-top idiom already established** in `get_endpoint_slices` and `delete_service`:

```python
def create_statefulset(self, namespace: str, body: dict) -> dict:
    namespace = _sanitize(namespace)
    ...
```

- **25 methods** sanitized (44 reassignments) — list/get/create/delete/patch across custom resources, deployments, services, statefulsets, jobs, plus `ensure_service_account`.
- Methods already sanitizing (`get_endpoint_slices`, `delete_service`, `create_secret`, `create_configmap`, `ensure_rolebinding`) left untouched.
- **Sink-side, not input-side, by design:** `KubernetesService` is shared by every router and background loop, so sanitizing at the log call (the single choke point) protects all callers at once — sanitizing in one endpoint would leave the shared sink injectable from the others.

## Behavior

No change for valid inputs — `_sanitize` is a no-op on values without control characters, and the sanitized value is safe to pass to the k8s API (a malformed name is rejected by the API regardless).

## Testing

- New proof test (`TestLogInjectionSanitization`): a newline-laced namespace passed to `create_statefulset` produces a single log record with no embedded newline and no forged `injected` line.
- Full backend suite: **487 passed** (excluding the pre-existing, unrelated `test_migration.py` local-venv collection quirk). `ruff check` + `format --check` clean.

## Notes

Surfaced by CodeQL on PR #2169 (simulated-tool workload); the sinks are pre-existing shared infra, so fixed here independently. Once this merges, #2169 rebases on top and its CodeQL gate clears.

Closes #2170

Assisted-By: Claude Code

### DIFF
--- a/kagenti/backend/app/services/kubernetes.py
+++ b/kagenti/backend/app/services/kubernetes.py
@@ -158,6 +158,8 @@ class KubernetesService:
         label_selector: Optional[str] = None,
     ) -> List[dict]:
         """List custom resources in a namespace."""
+        namespace = _sanitize(namespace)
+        plural = _sanitize(plural)
         try:
             response = self.custom_api.list_namespaced_custom_object(
                 group=group,
@@ -180,6 +182,7 @@ class KubernetesService:
         log_api_error: bool = True,
     ) -> dict:
         """List cluster-scoped custom resources (e.g., ClusterBuildStrategies)."""
+        plural = _sanitize(plural)
         try:
             return self.custom_api.list_cluster_custom_object(
                 group=group,
@@ -201,6 +204,9 @@ class KubernetesService:
         name: str,
     ) -> dict:
         """Get a specific custom resource."""
+        namespace = _sanitize(namespace)
+        plural = _sanitize(plural)
+        name = _sanitize(name)
         try:
             return self.custom_api.get_namespaced_custom_object(
                 group=group,
@@ -222,6 +228,9 @@ class KubernetesService:
         name: str,
     ) -> dict:
         """Delete a custom resource."""
+        namespace = _sanitize(namespace)
+        plural = _sanitize(plural)
+        name = _sanitize(name)
         try:
             return self.custom_api.delete_namespaced_custom_object(
                 group=group,
@@ -243,6 +252,8 @@ class KubernetesService:
         body: dict,
     ) -> dict:
         """Create a custom resource."""
+        namespace = _sanitize(namespace)
+        plural = _sanitize(plural)
         try:
             return self.custom_api.create_namespaced_custom_object(
                 group=group,
@@ -265,6 +276,9 @@ class KubernetesService:
         body: dict,
     ) -> dict:
         """Patch a custom resource."""
+        namespace = _sanitize(namespace)
+        plural = _sanitize(plural)
+        name = _sanitize(name)
         try:
             return self.custom_api.patch_namespaced_custom_object(
                 group=group,
@@ -289,6 +303,8 @@ class KubernetesService:
         the workload name (e.g. ``git-issue-agent``) rather than falling back
         to the ReplicaSet hash.
         """
+        namespace = _sanitize(namespace)
+        name = _sanitize(name)
         try:
             self.core_api.read_namespaced_service_account(name=name, namespace=namespace)
             logger.debug(f"ServiceAccount '{name}' already exists in {namespace}")
@@ -319,6 +335,8 @@ class KubernetesService:
         This is idempotent — if the ConfigMap already exists it is left unchanged
         so that user customizations are preserved.
         """
+        namespace = _sanitize(namespace)
+        name = _sanitize(name)
         try:
             self.core_api.read_namespaced_config_map(name=name, namespace=namespace)
             logger.debug(f"ConfigMap '{name}' already exists in {namespace}")
@@ -415,6 +433,7 @@ class KubernetesService:
 
     def create_deployment(self, namespace: str, body: dict) -> dict:
         """Create a Deployment in the specified namespace."""
+        namespace = _sanitize(namespace)
         try:
             result = self.apps_api.create_namespaced_deployment(
                 namespace=namespace,
@@ -427,6 +446,8 @@ class KubernetesService:
 
     def get_deployment(self, namespace: str, name: str) -> dict:
         """Get a Deployment by name."""
+        namespace = _sanitize(namespace)
+        name = _sanitize(name)
         try:
             result = self.apps_api.read_namespaced_deployment(
                 name=name,
@@ -439,6 +460,7 @@ class KubernetesService:
 
     def list_deployments(self, namespace: str, label_selector: Optional[str] = None) -> List[dict]:
         """List Deployments in a namespace with optional label selector."""
+        namespace = _sanitize(namespace)
         try:
             result = self.apps_api.list_namespaced_deployment(
                 namespace=namespace,
@@ -451,6 +473,8 @@ class KubernetesService:
 
     def delete_deployment(self, namespace: str, name: str) -> None:
         """Delete a Deployment by name."""
+        namespace = _sanitize(namespace)
+        name = _sanitize(name)
         try:
             self.apps_api.delete_namespaced_deployment(
                 name=name,
@@ -462,6 +486,8 @@ class KubernetesService:
 
     def patch_deployment(self, namespace: str, name: str, body: dict) -> dict:
         """Patch a Deployment with the provided body."""
+        namespace = _sanitize(namespace)
+        name = _sanitize(name)
         try:
             result = self.apps_api.patch_namespaced_deployment(
                 name=name,
@@ -479,6 +505,7 @@ class KubernetesService:
 
     def create_service(self, namespace: str, body: dict) -> dict:
         """Create a Service in the specified namespace."""
+        namespace = _sanitize(namespace)
         try:
             result = self.core_api.create_namespaced_service(
                 namespace=namespace,
@@ -491,6 +518,8 @@ class KubernetesService:
 
     def get_service(self, namespace: str, name: str) -> dict:
         """Get a Service by name."""
+        namespace = _sanitize(namespace)
+        name = _sanitize(name)
         try:
             result = self.core_api.read_namespaced_service(
                 name=name,
@@ -518,6 +547,7 @@ class KubernetesService:
 
     def list_services(self, namespace: str, label_selector: Optional[str] = None) -> List[dict]:
         """List Services in a namespace with optional label selector."""
+        namespace = _sanitize(namespace)
         try:
             result = self.core_api.list_namespaced_service(
                 namespace=namespace,
@@ -624,6 +654,7 @@ class KubernetesService:
 
     def create_statefulset(self, namespace: str, body: dict) -> dict:
         """Create a StatefulSet in the specified namespace."""
+        namespace = _sanitize(namespace)
         try:
             result = self.apps_api.create_namespaced_stateful_set(
                 namespace=namespace,
@@ -636,6 +667,8 @@ class KubernetesService:
 
     def get_statefulset(self, namespace: str, name: str) -> dict:
         """Get a StatefulSet by name."""
+        namespace = _sanitize(namespace)
+        name = _sanitize(name)
         try:
             result = self.apps_api.read_namespaced_stateful_set(
                 name=name,
@@ -648,6 +681,7 @@ class KubernetesService:
 
     def list_statefulsets(self, namespace: str, label_selector: Optional[str] = None) -> List[dict]:
         """List StatefulSets in a namespace with optional label selector."""
+        namespace = _sanitize(namespace)
         try:
             result = self.apps_api.list_namespaced_stateful_set(
                 namespace=namespace,
@@ -660,6 +694,8 @@ class KubernetesService:
 
     def delete_statefulset(self, namespace: str, name: str) -> None:
         """Delete a StatefulSet by name."""
+        namespace = _sanitize(namespace)
+        name = _sanitize(name)
         try:
             self.apps_api.delete_namespaced_stateful_set(
                 name=name,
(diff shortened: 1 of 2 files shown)


kagenti/backend/tests/test_kubernetes_service.py
```diff
@@ -810,3 +810,27 @@ def test_delete_custom_resource_success(self, kubernetes_service):
...
+            status=500, reason="Internal Server Error"
+        )
+
+        malicious_namespace = "team1\ninjected"
+
+        with caplog.at_level("ERROR"):
+            with pytest.raises(ApiException):
+                kubernetes_service.create_statefulset(malicious_namespace, {"metadata": {}})
+
+        assert len(caplog.records) == 1
```
clawgenti: **Nit:** `assert len(caplog.records) == 1` pins the test to exactly one log record. If the error path ever adds a second `logger.*` call (e.g. a future debug line), this assertion breaks the test unrelated to the sanitization guarantee. Consider `assert len(caplog.records) >= 1` and then asserting the *absence* of the newline in `caplog.records[0]`, which is the actual security invariant.

clawgenti (commented):
Applies the existing `_sanitize()` helper at the top of 25 `KubernetesService` methods that interpolate user-controlled `namespace`/`name`/`plural` into logger f-strings, closing the CWE-117 log injection surface flagged by CodeQL. Sink-side reassignment pattern is consistent with prior art in the same file, and the proof test correctly validates newline stripping.

---

kagenti/backend/tests/test_kubernetes_service.py
bnayahu: Good catch — relaxed to `>= 1` in 67b15438 so a future extra log call on the error path won't break the test, keeping the assertion focused on the actual invariant (no injected newline in `caplog.records[0]`).

clawgenti (commented):
Solid fix for CWE-117 log injection — consistent `_sanitize()` application across all 25 affected methods using the established idiom, with a targeted proof test.

**Findings:**

1. **Test fragility: `caplog.records[0]` index assumption** — the test asserts `len(caplog.records) >= 1` to allow future extra log calls, but then unconditionally reads `records[0]`. If any earlier log record (e.g., a DEBUG entry at a different log level captured by `caplog` propagation) is inserted before the ERROR record, the assertion `"team1injected" in message` could fail on the wrong record. Consider iterating over all records or searching for the record containing the sanitized namespace.

2. **`_sanitize` doesn't strip `\t`** — the helper removes `\n`, `\r`, `\x00` but not tab (`\t`). Tabs can be exploited in structured log formats (JSON, TSV, ELK). Not a blocker for this PR since the CodeQL rule is satisfied, but worth tracking.

---

kagenti/backend/tests/test_kubernetes_service.py
```diff
@@ -810,3 +810,30 @@ def test_delete_custom_resource_success(self, kubernetes_service):
...
+
+        with caplog.at_level("ERROR"):
+            with pytest.raises(ApiException):
+                kubernetes_service.create_statefulset(malicious_namespace, {"metadata": {}})
+
+        # We assert >= 1 (not == 1) so a future extra log call on the error path
+        # doesn't break this test; the security invariant is the absence of the
+        # injected newline in the record below, not the exact record count.
+        assert len(caplog.records) >= 1
+        message = caplog.records[0].getMessage()
```
clawgenti: `records[0]` is fragile given the `>= 1` assertion above. If an earlier log record (e.g., a DEBUG line from a logger not filtered to ERROR) lands first, this assertion will pass the wrong message. Consider: `message = next(r.getMessage() for r in caplog.records if 'team1' in r.getMessage())` or assert over all records to find the relevant one.

kagenti/backend/tests/test_kubernetes_service.py
```diff
@@ -810,3 +810,36 @@ def test_delete_custom_resource_success(self, kubernetes_service):
...
+        # and records[0] would then check the wrong message. Matching on "team1"
+        # finds the security-relevant record whether or not it was sanitized.
+        namespace_records = [r.getMessage() for r in caplog.records if "team1" in r.getMessage()]
+        assert namespace_records, "expected a log record referencing the namespace"
+        message = namespace_records[0]
+        # The sanitized namespace must be glued together with no embedded newline,
+        # so the attacker-supplied "injected" segment cannot appear as its own log line.
+        assert "team1injected" in message
+        assert "team1\ninjected" not in message
+        assert not any(line == "injected" for line in message.splitlines())
```
clawgenti, line 845: The test covers `create_statefulset` well. Consider a parametrized test covering at least one method per resource group (e.g. `get_custom_resource`, `get_deployment`, `get_service`, `get_job`) so that future accidental removal of a `_sanitize` call doesn't go undetected.

kagenti/backend/app/services/kubernetes.py
```diff
@@ -243,6 +252,8 @@ def create_custom_resource(
         body: dict,
     ) -> dict:
         """Create a custom resource."""
+        namespace = _sanitize(namespace)
+        plural = _sanitize(plural)
```
clawgenti, line 256: `create_custom_resource` sanitizes `namespace` and `plural` but not `name` (unlike `get_custom_resource`, `delete_custom_resource`, and `patch_custom_resource` which all include `name = _sanitize(name)`). If `name` is ever interpolated into a log message on this code path, log injection reappears. For consistency, add `name = _sanitize(name)` here.

clawgenti (commented):
Solid security fix applying the existing `_sanitize()` helper consistently across all 25 affected `KubernetesService` methods to close CWE-117 log injection. The sink-side sanitization strategy is well-justified and the proof test is thorough and robust.

- The test only exercises `create_statefulset` — while this covers the pattern, the other 24 sanitized methods have no analogous test coverage. Consider adding a parametrized test that iterates over representative methods across each resource type (or at minimum one per resource group: custom, deployment, service, statefulset, job) to prevent silent regressions if `_sanitize` calls are removed.
- `create_custom_resource` sanitizes `namespace` and `plural` but not `name` — unlike `get_custom_resource`, `delete_custom_resource`, and `patch_custom_resource` which all sanitize `name`. If `name` appears in any log call on the create path (now or in a future edit), that gap would reintroduce the vulnerability.

---

kagenti/backend/tests/test_kubernetes_service.py
bnayahu: Added in 00d9bd7f — a parametrized `test_read_methods_sanitize_namespace_in_log` now asserts the namespace-sanitization invariant across one read method per resource group (`get_custom_resource`, `get_deployment`, `get_service`, `get_job`), so an accidental removal of a `_sanitize` call on any group is caught rather than only `create_statefulset`.

kagenti/backend/app/services/kubernetes.py
bnayahu: `create_custom_resource` has no `name` parameter — unlike `get`/`delete`/`patch_custom_resource`, its signature is `(group, version, namespace, plural, body)` and the name lives inside `body`. The log line on this path is `f"Error creating {plural} in {namespace}: {e}"`, which references neither a `name` variable nor `body`, so there is nothing unsanitized reaching the log here. Adding `name = _sanitize(name)` would raise `NameError`. If a future change logs the name on this path, it should sanitize the value extracted from `body` at that point. Leaving as-is.

clawgenti (commented):
Targeted CWE-117 fix that applies `_sanitize()` at the method entry point for 25 `KubernetesService` methods, matching the idiom already established in `get_endpoint_slices` and `delete_service`. All CI checks pass, proof tests are thorough and well-reasoned.

---

kagenti/backend/tests/test_kubernetes_service.py
```diff
@@ -810,3 +810,102 @@ def test_delete_custom_resource_success(self, kubernetes_service):
...
+        # Select the record that actually logged the namespace rather than assuming
+        # it is records[0]: a future debug line on the error path could land first,
+        # and records[0] would then check the wrong message. Matching on "team1"
+        # finds the security-relevant record whether or not it was sanitized.
+        namespace_records = [r.getMessage() for r in caplog.records if "team1" in r.getMessage()]
+        assert namespace_records, "expected a log record referencing the namespace"
+        message = namespace_records[0]
+        # The sanitized namespace must be glued together with no embedded newline,
+        # so the attacker-supplied "injected" segment cannot appear as its own log line.
+        assert "team1injected" in message
```
clawgenti, line 843: `_sanitize` strips `\n`/`\r`/`\x00` — so `"team1\ninjected"` → `"team1injected"`. This assertion is correct as long as the sanitizer removes (not replaces) the newline. Worth a quick look at `_sanitize` to confirm there is no substitution character that would break this assertion.

clawgenti (commented):
Solid security fix — CWE-117 sanitization applied consistently across all 25 affected methods using the established sink-side reassign idiom. Tests are well-reasoned with good inline comments explaining the `>= 1` assertion and record-selection logic.

- The parametrized test covers custom resources, deployments, services, and jobs — good breadth across resource groups.
- `create_custom_resource` correctly omits a `name` sanitization since `name` isn't a separate parameter there (it lives in `body`); the asymmetry is intentional and consistent.
- Minor: the test at line 843 asserts `"team1injected" in message` but `_sanitize` strips `\n`/`\r`/`\x00` — double-check that the sanitized form actually concatenates to `team1injected` (no separator) rather than being replaced with a placeholder. If `_sanitize` only strips, that's fine; just worth a quick verification of the sanitizer output.

---

kagenti/backend/tests/test_kubernetes_service.py
bnayahu: _sanitize replaces "\n" with "".

cwiklik (approved):
Complete, well-tested CWE-117 log-injection fix. Applies the module's established reassign-at-top `_sanitize()` idiom to `namespace`/`name`/`plural` across the `KubernetesService` methods.

I audited all 42 logger calls in `kubernetes.py` for completeness (the thing that matters for a log-injection fix):
- Every logger interpolation of a user-controlled value (`namespace`/`name`/`plural`) is now inside a method that sanitizes that value — verified for both f-string and `%`-style calls (e.g. `create_configmap`'s `%s` line, `ensure_rolebinding`'s pre-existing `safe_name`/`safe_ns`, `get_endpoint_slices`, `delete_service`, `create_secret`).
- `label_selector`, `body`, `group`, and `version` never reach a log, so the create/list paths have no residual vector despite only sanitizing `namespace`.
- No logger uses concatenation to smuggle an unsanitized value.
- Since `_sanitize` only strips `\n`/`\r`/`\x00` (invalid in k8s names anyway), reassign-at-top doesn't change behavior for valid inputs.

Tests are solid: parametrized across resource groups, assert `>=1` record, match on `team1` rather than `records[0]`, and check no standalone `injected` line survives.

CI is green (CodeQL, Bandit, backend-tests, lint); E2E HyperShift is pending a maintainer `/run-e2e` (non-blocking). All 4 commits signed off.

One non-blocking observation for later (out-of-diff, so no inline comment): `_sanitize`'s docstring says it "strips other control characters," but the implementation only strips `\n`/`\r`/`\x00`. This PR makes that helper load-bearing across 25 methods, so it'd be worth either tightening the docstring or broadening the strip set in a follow-up. The current set is the CodeQL-recognized sanitizer, so it's a defensible scope.

LGTM.

Assisted-By: Claude Code

pdettori (approved):
LGTM

PR #2172, opened 2026-07-08
Fix: pre-apply kagenti-operator CRDs before helm upgrade (#1335)

### LINKED ISSUE #1335: bug: Helm upgrade does not install new CRDs from kagenti-operator subchart

cwiklik:
## Summary

Helm applies a chart's `crds/` directory **only on install, never on `helm upgrade`** — and the same holds for a **subchart's** `crds/`. The `kagenti` umbrella chart pulls `kagenti-operator-chart` as an OCI dependency, and that subchart ships its CRDs in `kagenti-operator-chart/crds/`. So when an existing cluster is re-run/upgraded to an operator version that introduces a **new** CRD, Helm silently skips it → the controller-manager crash-loops on the missing resource → the webhook loses healthy endpoints. Fixes #1335.

(The original report targeted the since-removed Ansible installer; the bug carries over unchanged to the script installers, which already pre-apply SPIRE and Gateway-API CRDs but had no step for the operator's own CRDs.)

## Fix

Add an explicit operator-CRD pre-apply to both installers, right before `helm upgrade --install kagenti`, mirroring the existing MCP Gateway CRD pre-apply (`scripts/ocp/setup-kagenti.sh`, whose comment already reads *"Helm does not update CRDs on upgrade"*).

- **`scripts/kind/setup-kagenti.sh`** — after `helm dependency update`, before the kagenti install.
- **`scripts/ocp/setup-kagenti.sh`** — after dependency resolution / the `--operator-repo` local-chart copy, before the kagenti install.

The CRDs are resolved from the operator subchart **already vendored** by `helm dependency update` (so the version always matches the `Chart.yaml` pin), handling both forms:
- unpacked dir `charts/kagenti/charts/kagenti-operator-chart/crds/` (OCP `--operator-repo` copy), else
- vendored `charts/kagenti/charts/kagenti-operator-chart-*.tgz` (extract `crds/`).

Applied with `kubectl apply --server-side --force-conflicts` (avoids the client-side last-applied-annotation size limit on large CRDs; idempotent across re-runs and prior Helm ownership), then waited on for `Established`. If no operator CRDs are found (agent operator disabled), it logs and skips — it does **not** fail. A genuine apply failure surfaces (no `|| true`), since a missing operator CRD is exactly what crashes the controller.

## Validation

- `bash -n` clean on both scripts; `shellcheck --severity=error` (CI's level) unaffected (the `$KUBECTL`/`ls` patterns are info/warning and already used in-repo).
- **Extraction logic exercised in bash** against a freshly `helm dependency update`-vendored `kagenti-operator-chart-0.3.0-alpha.6.tgz`: correctly resolved `crds/` and found all three CRDs — `agentcards`, `agentruntimes`, and **`authorizationpolicies`** (a CRD older operator versions lacked, i.e. the exact new-CRD-on-upgrade case this fixes).
- Fresh install is exercised by CI `Deploy & Test`. The **upgrade** path (install an operator version lacking a CRD, then re-run targeting one that adds it → new CRD created, controller does not crash-loop) is best verified manually on a cluster; steps in the issue.

No `charts/` changes. The vendored-tgz-vs-pin staleness noted in the issue is refreshed by the installer's own `helm dependency update` at run time and is out of scope.

Fixes #1335

Assisted-By: Claude Code


scripts/kind/setup-kagenti.sh
```diff
@@ -1084,6 +1084,30 @@ fi
...
+# Pre-apply the operator CRDs from the resolved operator subchart so an upgrade
+# that introduces a new CRD doesn't crash-loop the controller-manager.
+_op_crds=""
+_op_tmp=""
+if [ -d "$REPO_ROOT/charts/kagenti/charts/kagenti-operator-chart/crds" ]; then
+  _op_crds="$REPO_ROOT/charts/kagenti/charts/kagenti-operator-chart/crds"
+else
+  _op_tgz=$(ls "$REPO_ROOT"/charts/kagenti/charts/kagenti-operator-chart-*.tgz 2>/dev/null | head -1 || true)
+  if [ -n "$_op_tgz" ]; then
+    _op_tmp=$(mktemp -d)
```
clawgenti, line 1097: **Nit:** If `tar xzf` or a subsequent command fails with `set -e`, the script exits before the `if [ -n "$_op_tmp" ]; then rm -rf …` cleanup at line 1109. Consider a `trap 'rm -rf "$_op_tmp"' EXIT` guard (or add `|| true` to the tar/apply steps to keep existing error-propagation semantics).

scripts/ocp/setup-kagenti.sh
```diff
@@ -1268,6 +1268,31 @@ if [ -z "$OPERATOR_REPO" ]; then
...
+# crash-loop the controller-manager.
+_op_crds=""
+_op_tmp=""
+if [ -d "$KAGENTI_REPO/charts/kagenti/charts/kagenti-operator-chart/crds" ]; then
+  _op_crds="$KAGENTI_REPO/charts/kagenti/charts/kagenti-operator-chart/crds"
+else
+  _op_tgz=$(ls "$KAGENTI_REPO"/charts/kagenti/charts/kagenti-operator-chart-*.tgz 2>/dev/null | head -1 || true)
+  if [ -n "$_op_tgz" ]; then
+    _op_tmp=$(mktemp -d)
+    tar xzf "$_op_tgz" -C "$_op_tmp"
```
clawgenti: **Nit:** Same tmpdir-leak-on-early-exit as the `kind` script — `$_op_tmp` won't be cleaned if `tar xzf` or `kubectl apply` fires `set -e`. A `trap` guard or equivalent handles this.

clawgenti (commented):
Adds an explicit operator-CRD pre-apply step to both `kind` and `ocp` script installers, correctly mirroring the existing MCP Gateway CRD pattern and resolving the upgrade crash-loop (fixes #1335). Logic is sound — handles both extracted and vendored-tgz forms, is idempotent, and fails loudly on a genuine apply error.

- **Nit:** `_op_tmp` is only cleaned up by the inline `if [ -n "$_op_tmp" ]…` at the end of the block. If the script exits early (e.g. `set -e` fires on `tar xzf` or `kubectl apply`), `$_op_tmp` leaks. Low-severity since `/tmp` is ephemeral, but a `trap … EXIT` guard is the conventional fix.
- **Nit (SC2012):** `ls … | head -1` is used to find the vendored `.tgz`. The repo body notes this pattern is already present in-repo and CI `shellcheck --severity=error` passes; flagging for awareness only.

---

cwiklik:
Thanks @clawgenti — addressed in `e26e6f53`:

- ✅ **tmpdir leak on early exit:** fixed, but **not** via `trap … EXIT` — both scripts already install `EXIT` traps (`kind:497`, `ocp:1264` for the `--operator-repo` copy) and a second `trap … EXIT` would silently clobber them. Instead: the `tar` extraction is now guarded (`if tar xzf …`), and the apply/wait result is captured (`{ … } || _op_rc=$?`) so `set -e` can't skip the cleanup. The temp dir is then **always** removed, and a genuine apply failure is surfaced afterward (`exit $_op_rc`) — keeping the fail-loud behavior.
- ↪️ **SC2012 (`ls … | head`):** left as-is per your note — CI `shellcheck --severity=error` treats it as info/warning and passes, and the pattern is self-contained here.

Assisted-By: Claude Code

scripts/kind/setup-kagenti.sh
```diff
@@ -1084,6 +1084,36 @@ fi
...
 
+# Helm applies a subchart's crds/ only on install, never on upgrade (kagenti#1335).
+# Pre-apply the operator CRDs from the resolved operator subchart so an upgrade
+# that introduces a new CRD doesn't crash-loop the controller-manager.
+_op_crds=""
+_op_tmp=""
+if [ -d "$REPO_ROOT/charts/kagenti/charts/kagenti-operator-chart/crds" ]; then
+  _op_crds="$REPO_ROOT/charts/kagenti/charts/kagenti-operator-chart/crds"
+else
+  _op_tgz=$(ls "$REPO_ROOT"/charts/kagenti/charts/kagenti-operator-chart-*.tgz 2>/dev/null | head -1 || true)
```
clawgenti, line 1095: **nit**: `ls ... | head -1` for glob expansion triggers SC2012 (`ls` output may be mishandled on filenames with spaces/special chars). Since filenames here are controlled Helm artifact names this is low risk, but the POSIX idiom is:
```sh
for _f in "$REPO_ROOT"/charts/kagenti/charts/kagenti-operator-chart-*.tgz; do
  [ -e "$_f" ] && { _op_tgz="$_f"; break; }
done
```
The `ocp` variant has the same pattern. Both already pass CI shellcheck at `--severity=error`, so this is advisory only.

clawgenti (commented):
Pre-apply of operator CRDs before `helm upgrade` is a solid, idempotent fix for the upgrade-path crash-loop — both the unpack-from-tgz and dir-copy branches are handled cleanly, temp dir is reliably cleaned, and the failure mode surfaces correctly without a `|| true` swallow.

---

clawgenti (commented):
Fixes the Helm subchart CRD upgrade gap cleanly: temp-dir cleanup is leak-proof, error propagation is correct, and both `kubectl`/`$KUBECTL` usages match existing conventions in their respective scripts. All checks pass. Ready for human review.

---

mrsabath (approved):
## Summary

Focused, well-reasoned fix for the Helm subchart-CRD-on-upgrade limitation (#1335). Both installers pre-apply the operator CRDs from the already-resolved subchart (unpacked-dir first, else the vendored `.tgz`), using `kubectl apply --server-side --force-conflicts` + a `wait` for `Established`, with leak-proof temp-dir cleanup and a deliberate hard-fail (no `|| true`) since a missing operator CRD is exactly what crash-loops the controller-manager.

**What I verified**

- **Root cause match** — targets #1335 precisely (subchart `crds/` skipped on `helm upgrade`; missing CRD → controller crash-loop → webhook loses endpoints).
- **`set -euo pipefail` safety** — the `_op_tgz=$(ls … | head -1 || true)` assignment is protected, and `{ apply && wait; } || _op_rc=$?` keeps `errexit` from short-circuiting while still capturing the code. All new vars (`_op_crds`, `_op_tmp`, `_op_rc`, `_op_tgz`) are initialized, so `nounset` is safe.
- **Cleanup ordering** — temp dir is removed before the error is surfaced, so a failed apply doesn't leak the extraction dir. On OCP, `exit "$_op_rc"` correctly fires the existing `EXIT` trap that cleans the `--operator-repo` local chart copy.
- **CRD set** — the operator chart ships exactly three real CRDs (`agentcards`, `agentruntimes`, `authorizationpolicies`); `wait --for=condition=Established -f <dir>` waits per-object across all of them, and `oc wait` supports the same flags on OCP.
- **Dual-mode parity** — the kind (`$REPO_ROOT` / `kubectl`) and ocp (`$KAGENTI_REPO` / `$KUBECTL`) blocks are consistent: same CRD path, same flags, same fallback and cleanup. Unquoted `run_cmd $KUBECTL` matches the established in-repo idiom.

**Areas reviewed**: Shell, Security, Commit/PR conventions, dual-mode parity
**Commits**: 3 commits, all signed-off (DCO passes)
**CI note**: The `Shell Script Lint` CI job only scans `.github/scripts/`, so `scripts/kind/` and `scripts/ocp/` aren't CI-linted — the PR's manual `bash -n` / `shellcheck` is the effective gate here. Worth keeping in mind, not a blocker.

No must-fix issues found. LGTM — approving.

Issue #1899, opened 2026-06-11
Mesh-wide 503: expired SPIRE workload SVIDs break all gateway-routed services on long-running Kind clusters


yoavkatz:
# Mesh-wide outage: expired SPIRE workload SVIDs make all gateway-routed services return HTTP 503 (`localtest.me` URLs unreachable)

## Summary

On a long-running Kind cluster (~27h old), **every service exposed through the Istio
gateway became unreachable**, returning `HTTP 503 Service Unavailable`. This includes
`http://mlflow.localtest.me:8080/`, `http://kagenti-ui.localtest.me:8080/`,
`http://kagenti-api.localtest.me:8080/`, and others.

The trigger that surfaced it was MLflow, but **MLflow is not the cause** — the failure is
mesh-wide. The root cause is that **Istio Ambient workload mTLS certificates (SPIFFE SVIDs,
issued by SPIRE) expired and stopped rotating**. This happened because the **host machine was
suspended for ~16 hours** (laptop sleep), during which *every* time-bound SPIRE credential
expired at once — the agent's own SVID, its 2h Kubernetes service-account token, and even the
24h CA. On resume the `spire-agent` hit a **credential-expiry deadlock** (it needs a valid SA
token to re-attest and obtain a new SVID, but that token had also expired) and crashed. It
later recovered, but the Ambient data plane (ztunnel + waypoints) never re-fetched the fresh
SVIDs and kept serving expired certs.

This is a **reliability/lifecycle bug**: a freshly deployed cluster works, but a cluster that
is left idle/suspended longer than the SVID/CA + SA-token lifetime can silently lose all mesh
connectivity and not self-heal.

## Impact

- All HTTP routes behind the `http` gateway in `kagenti-system` return `503`.
- Inter-service mesh traffic fails (e.g. `otel-collector → mlflow`, agent → agent).
- The cluster looks healthy on the surface: all pods are `Running`, the gateway pod is
  `1/1`, the `HTTPRoute`/`Gateway` status is `Accepted` / `ResolvedRefs` / `ResolvedWaypoints`,
  and DNS resolves. Nothing in `kubectl get pods` indicates a problem — making this hard to
  diagnose without reading ztunnel logs.

## Environment

- Deployment: Kind, via `./.github/scripts/local-setup/kind-full-test.sh`
- Cluster age at time of failure: ~27h
- Mesh: Istio **Ambient** (ztunnel + waypoint proxies)
[...]

PR #2179, opened 2026-07-10
feat(k8s): mesh self-heal detect+recover for post-suspend 503 (Part of #1899)

cwiklik:
## What

Adds a detect-and-recover tool + skill for the Istio Ambient "everything returns 503" outage seen on long-running / suspended dev/Kind clusters. **Part A** of the agreed A+B mitigation for #1899 (B = feature-flagged CronJob remediator, targeted for v0.8).

- **`scripts/k8s/mesh-recover.sh`** — read-only by default: probes a *backend-dependent* gateway route (`kagenti-ui`/`kagenti-api`, which flip to 503 in an outage — not edge-redirect routes like kiali), checks ztunnel cert/logs (discovers the ztunnel namespace, since Kind uses `istio-system` and the chart uses `istio-ztunnel`), and optionally the SPIRE agent. `--fix` runs the proven recovery — `rollout restart` of ztunnel + waypoints + the gateway (and, with `--include-spire`, the SPIRE agent). `--json` for automation. Exit `0` healthy / `2` degraded.
- **`.claude/skills/istio:mesh-selfheal/`** — skill wrapping the above: when-to-use, the two-domain note (mesh mTLS vs. the separate SPIRE reattest issue), manual fallback commands, diagnostics, and testing.
- **`scripts/k8s/induce-mesh-outage.sh`** — Kind-guarded test harness that induces the symptom (CA down + ztunnel pods deleted → mesh-wide 503) and asserts the full `detect → --fix → re-probe` loop; `--break` / `--restore` / `--self-test`.
- Registration: `istio` parent skill table, a `k8s:health` cross-reference, and the skills `README.md` tree.

## Why

Root cause of #1899: the Ambient data plane serves **expired istiod-issued mTLS certs** and never re-fetches after a host suspend / clock jump, so every gateway route 503s while pods look healthy. (SPIRE is a *separate* identity plane; its reattest deadlock affects AuthBridge, not routing.) The durable data-plane fix is **upstream — istio/ztunnel#1679**; this is the kagenti-side dev-cluster mitigation.

## Scope / safety

- **dev/Kind only**; nothing auto-runs. `--fix` and the induce harness are opt-in and refuse to run outside a `kind-*` context (without `--force`).
- No production behavior change; no new always-on components.

## Testing

- `mesh-recover.sh` — healthy cluster → exit 0; degraded (bad target) → exit 2 + printed dry-run recovery; `--json` → clean JSON.
- `induce-mesh-outage.sh --self-test` on a live Kind cluster: induced 503 → detected (exit 2) → `--fix` → healthy (exit 0). This caught and fixed a real probe bug (edge-redirect routes gave a false-healthy).

Part of #1899. Durable fix tracked upstream: istio/ztunnel#1679.

Assisted-By: Claude Code


scripts/k8s/mesh-recover.sh
```diff
@@ -0,0 +1,304 @@
...
+    if grep -qi 'connection termination\|upstream connect error' <<<"$body"; then
+      record "mesh-reachability" "DEGRADED" "HTTP 503 (connection termination) for ${host} — Ambient upstream mTLS is failing"
+      log_error "503 with upstream connection termination for ${host} — classic expired-cert mesh outage"
+    else
+      record "mesh-reachability" "DEGRADED" "HTTP 503 for ${host}"
+      log_error "503 for ${host}"
+    fi
+  elif [[ "$code" == "000" ]]; then
+    record "mesh-reachability" "INCONCLUSIVE" "could not connect to ${GATEWAY_URL} — is the Kind port-map / port-forward up?"
+    log_warn "Gateway ${GATEWAY_URL} unreachable (curl 000). Not a 503; check the port-map. Treating as needs-attention."
```
clawgenti: **nit:** `INCONCLUSIVE` cases (curl 000, no ztunnel found) set `DEGRADED=true` inline rather than via `record()`, so they show up inconsistently in the JSON output — the `record()` call records `INCONCLUSIVE` status but the outer `DEGRADED` flag is set separately. Consider routing through `record "mesh-reachability" "INCONCLUSIVE" "..."` and then checking for both `DEGRADED` and `INCONCLUSIVE` statuses when deciding the exit code, for consistency.

scripts/k8s/induce-mesh-outage.sh
```diff
@@ -0,0 +1,156 @@
...
+MODE=""
+FORCE=false
+GATEWAY_URL="${GATEWAY_URL:-http://127.0.0.1:8080}"
+ISTIOD_NS="istio-system"
+STATE_DIR="/tmp/kagenti/mesh-test"
+STATE_FILE="${STATE_DIR}/istiod-replicas"
+HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+RECOVER="${HERE}/mesh-recover.sh"
+
+usage() { grep '^#' "$0" | sed 's/^# \{0,1\}//; 1d'; }
```
clawgenti: **nit:** `usage()` parses comment lines from the script header (`grep '^#' "$0" | sed ...`) — this is a common pattern but fragile if the shebang or leading comment format ever changes. Since the help text is short, an explicit `cat <<'EOF'` heredoc (like `mesh-recover.sh` uses) would be more robust.

clawgenti (commented):
Adds `mesh-recover.sh` + `induce-mesh-outage.sh` and the `istio:mesh-selfheal` skill — a clean, well-scoped detect/recover tool for the post-suspend Ambient 503 outage. All checks pass. Ready for human review.

---

clawgenti (commented):
Adds `mesh-recover.sh`, `induce-mesh-outage.sh`, and the `istio:mesh-selfheal` skill — a well-scoped detect/recover tool for the post-suspend Ambient 503 outage. Both scripts use `set -euo pipefail`, DCO sign-off is present on all commits, and CI is green.

- **Nit:** `SKILL.md` documents exit codes as `0` healthy · `2` degraded · `1` usage error, but `mesh-recover.sh` adds a third exit code `3` (INCONCLUSIVE, e.g. gateway unreachable or no ztunnel found). The script's own `--help` output and inline comments are correct; only the skill doc summary is missing this code.

---

.claude/skills/istio:mesh-selfheal/SKILL.md
```diff
@@ -0,0 +1,97 @@
...
+scripts/k8s/mesh-recover.sh
+
+# Apply the recovery (rollout restart ztunnel + waypoints + gateway)
+scripts/k8s/mesh-recover.sh --fix
+
+# Also check/restart the SPIRE agent (separate AuthBridge domain)
+scripts/k8s/mesh-recover.sh --fix --include-spire
+```
+
+Exit codes: `0` healthy · `2` degraded (or fix attempted) · `1` usage/precondition error.
```
clawgenti: **nit:** The exit code summary is missing exit `3` (INCONCLUSIVE — gateway unreachable or no ztunnel daemonset found). The script's own `usage()` output (and the inline comment at the `INCONCLUSIVE` variable declaration) are already correct; this line just needs `· 3 inconclusive (unreachable / not Ambient)` added. E.g.:
```
Exit codes: `0` healthy · `2` degraded (or fix attempted) · `3` inconclusive (unreachable / not Ambient) · `1` usage/precondition error.
```

cwiklik:
**Deploy & Test (e2e) failure is unrelated to this PR — safe to merge on the other green checks.**

Root cause (from the failed run): the **AuthBridge Weather (advanced) E2E** stage clones and runs `deploy_and_verify_advanced.sh` from `kagenti-extensions@main`, and its `pip install -r requirements.txt` fails with a dependency conflict:

```
ERROR: Cannot install -r requirements.txt (line 4), -r requirements.txt (line 6) and
langchain-core==1.4.8 because these package versions have conflicting dependencies.
ERROR: ResolutionImpossible
```

All Kubernetes steps before it (secret sync, `weather-tool-advanced` rollout) succeeded — the failure is purely a Python dependency-resolution conflict in the external `kagenti-extensions` repo.

This PR only adds two shell scripts under `scripts/k8s/` and Markdown skill files under `.claude/skills/` — it touches no Python requirements, no chart/deploy path, and not `kagenti-extensions`, so it cannot have caused this. The conflict is **deterministic (not flaky)**, so a re-run won't clear it; the fix belongs in `kagenti-extensions` requirements pins.

All other checks (lint, shellcheck, chart/helm, backend-tests, security scans, DCO) pass.

*Assisted-By: Claude Code*

scripts/k8s/mesh-recover.sh
```diff
@@ -0,0 +1,313 @@
...
+    record "spire-agent" "OK" "spire-agent healthy (restarts=${restarts}, ns ${ns})"
+    log_success "SPIRE agent healthy (ns ${ns})"
+  fi
+}
+
+# ---------------------------------------------------------------------------
+# Recovery
+# ---------------------------------------------------------------------------
+recover_mesh() {
+  local ns wp gw
```
esnible: **nit**: `local ns wp gw` declares `wp`, but it's never assigned or used in `recover_mesh()` (shellcheck SC2034). Safe to drop it from the declaration.

scripts/k8s/induce-mesh-outage.sh
```diff
@@ -0,0 +1,171 @@
...
+  --restore    Undo (scale istiod back, restart ztunnel).
+  --self-test  break → assert detect(exit 2) → restore CA → --fix → assert green.
+  --force      Allow running outside a kind-* context (dev only).
+  -h, --help   This help.
+EOF
+}
+
+while [[ $# -gt 0 ]]; do
+  case "$1" in
+    --break)     MODE=break; shift ;;
```
esnible: **nit**: `MODE=break` / `MODE=restore` trip shellcheck SC2209 because `break`/`restore` read like builtins. It works correctly (assigns the literal string), but quoting — `MODE='break'` / `MODE='restore'` — silences the warning and makes the string-assignment intent explicit.

.claude/skills/istio:mesh-selfheal/SKILL.md
esnible: **nit**: This lists exit codes `0` / `2` / `1`, but the script's own `--help` also documents `3 inconclusive` (gateway unreachable / not an Ambient cluster — a restart won't help). Worth adding `3` here so the skill doc matches the script.

esnible (approved):
## Summary

Well-scoped dev/Kind mesh self-heal tooling for the Ambient expired-cert 503 outage (#1899). `mesh-recover.sh` is read-only by default with `--fix` gated; `induce-mesh-outage.sh` is Kind-guarded and refuses to run outside a `kind-*` context without `--force`. No always-on components and no production behavior change. The backend-dependent probe-host choice (avoiding edge-redirect routes like kiali that would give a false-healthy) and the ztunnel-namespace discovery are nice touches. Reviewed the skill markdown and both scripts in full — no supply-chain concerns (the `.claude/` changes are documentation only, no hooks or executable config).

The failing `Deploy & Test` check is unrelated to this PR: a `langchain-core` dependency-resolution conflict in the externally-cloned kagenti-extensions AuthBridge E2E. These scripts are dev-only and not exercised by CI.

Approving. Three cosmetic nits inline — take or leave.

**Areas reviewed**: Shell, Docs, `.claude/` skills

Assisted-By: Claude Code

PR #2193, opened 2026-07-13
fix: :bug: Make --skip-cluster re-runs idempotent

### LINKED ISSUE #2192: :bug: Re-runs of setup with --skip-cluster failing on field-manager conflicts
### Describe the bug

Running `scripts/kind/setup-kagenti.sh --with-all --skip-cluster` against an existing cluster fails at multiple steps:

- Istio base/istiod upgrades fail with Apply failed with 1 conflict: conflict with "pilot-discovery" on webhook failurePolicy fields
- SPIRE install hangs for 5 minutes then times out (--wait watches wrong namespace)
- kagenti chart install hangs for 20 minutes (Helm can't confirm readiness of completed Jobs)
- SPIRE re-install fails with conflict with "kubectl-client-side-apply" on the OIDC configmap

### Steps To Reproduce

`scripts/kind/setup-kagenti.sh --with-all --skip-cluster` with already partially set-up cluster.

### Expected Behavior

 "Idempotent: safe to re-run" as claimed by the script header

### Kagenti Version

v0.6.1

### Additional Context

_No response_

evaline-ju:
## Summary
- Delete Istio ValidatingWebhookConfigurations before upgrade to avoid pilot-discovery field-manager conflicts
- Remove --wait from SPIRE install (pods deploy to a different namespace than the release, causing a 5-min timeout)
- Remove --wait --timeout 20m from kagenti install (Helm can't track readiness of completed Jobs)
- Add explicit kubectl wait with pod-existence checks for both SPIRE and kagenti
- Delete SPIRE OIDC configmap before re-install to avoid kubectl-client-side-apply conflict

Fixes #2192 

Assisted-By: Claude Code


scripts/kind/setup-kagenti.sh
```diff
@@ -668,6 +676,12 @@ if $WITH_SPIRE; then
     --set tornjak-frontend.service.type=ClusterIP \
     --set tornjak-frontend.service.port=3000
 
+  log_info "Waiting for SPIRE pods in zero-trust-workload-identity-manager..."
+  until kubectl get pod -l app.kubernetes.io/instance=spire \
+    -n zero-trust-workload-identity-manager 2>/dev/null | grep -q .; do sleep 2; done
```
clawgenti: `kubectl get pod` always prints a header line, so `grep -q .` matches even when no pods exist yet — the loop exits immediately rather than waiting. Use `--no-headers` to filter headers:
```bash
until kubectl get pod --no-headers -l app.kubernetes.io/instance=spire \
  -n zero-trust-workload-identity-manager 2>/dev/null | grep -q .; do sleep 2; done
```
Also consider adding a loop timeout (e.g. track `$SECONDS`) so a silent Helm failure doesn't spin indefinitely.

scripts/kind/setup-kagenti.sh
```diff
@@ -1263,10 +1277,16 @@ fi
...
 log_info "Installing kagenti..."
 run_cmd helm upgrade --install kagenti "$REPO_ROOT/charts/kagenti/" \
-  -n kagenti-system --wait --timeout 20m \
+  -n kagenti-system \
   "${SECRETS_FLAGS[@]+"${SECRETS_FLAGS[@]}"}" \
   "${KAGENTI_FLAGS[@]}"
 
+log_info "Waiting for kagenti pods..."
+until kubectl get pod -l app.kubernetes.io/name=kagenti-operator-chart \
+  -n kagenti-system 2>/dev/null | grep -q .; do sleep 2; done
```
clawgenti: Same `grep -q .` issue as the SPIRE wait above — header line causes immediate exit. Use `--no-headers`:
```bash
until kubectl get pod --no-headers -l app.kubernetes.io/name=kagenti-operator-chart \
  -n kagenti-system 2>/dev/null | grep -q .; do sleep 2; done
```
Same timeout-guard recommendation applies.

clawgenti (commented):
This PR correctly addresses the idempotency issues by removing `--wait` from Helm installs and replacing with explicit `kubectl wait` calls — the approach and comments are clear. Two issues in the `until` polling loops that affect both SPIRE and kagenti waits.

**Findings:**
1. `grep -q .` always matches: `kubectl get pod` prints a header line even when no pods exist, so the `until` loop exits immediately on every run. Use `--no-headers` to avoid false-positive matches.
2. No timeout on the `until` loop: if the Helm install fails silently, this spins indefinitely. Add a deadline, e.g. `SECONDS=0; until ... || (( SECONDS > 180 ))`.

---

scripts/kind/setup-kagenti.sh
```diff
@@ -668,6 +676,16 @@ if $WITH_SPIRE; then
     --set tornjak-frontend.service.type=ClusterIP \
     --set tornjak-frontend.service.port=3000
 
+  log_info "Waiting for SPIRE pods in zero-trust-workload-identity-manager..."
+  local deadline=$((SECONDS + 120))
```
clawgenti: **nit**: `local` is only valid inside a function — at top-level scope in bash it generates a `SC2168` warning (though it still works). The kagenti polling block below correctly omits `local` (line 1289). Consider using `deadline=` here to be consistent.

```bash
  deadline=$((SECONDS + 120))
```

clawgenti (commented):
Makes `--skip-cluster` re-runs idempotent by pre-deleting conflicting webhook configs and the SPIRE OIDC configmap, and replaces Helm `--wait` flags with explicit `kubectl wait` loops that correctly target the right namespaces.

One nit on `local` scoping — see inline. All other checks pass. Ready for human review.

---

scripts/kind/setup-kagenti.sh
```diff
@@ -668,6 +676,16 @@ if $WITH_SPIRE; then
     --set tornjak-frontend.service.type=ClusterIP \
     --set tornjak-frontend.service.port=3000
 
+  log_info "Waiting for SPIRE pods in zero-trust-workload-identity-manager..."
+  deadline=$((SECONDS + 120))
+  until kubectl get pod --no-headers -l app.kubernetes.io/instance=spire \
+    -n zero-trust-workload-identity-manager 2>/dev/null | grep -q .; do
+    (( SECONDS >= deadline )) && { log_error "Timed out waiting for SPIRE pods"; exit 1; }
```
clawgenti: **must-fix**: With `set -euo pipefail` (line 29), an arithmetic `(( expr ))` that evaluates to 0 (false — deadline not yet reached) exits with code 1, which kills the script immediately on every loop iteration where the timeout hasn't fired yet. This makes the entire wait loop non-functional and the script unreliable.

Safe replacement:
```bash
if (( SECONDS >= deadline )); then
  log_error "Timed out waiting for SPIRE pods"
  exit 1
fi
```
The `if` form exits non-zero only when the branch body runs.

scripts/kind/setup-kagenti.sh
```diff
@@ -1263,10 +1281,20 @@ fi
...
-  -n kagenti-system --wait --timeout 20m \
+  -n kagenti-system \
   "${SECRETS_FLAGS[@]+"${SECRETS_FLAGS[@]}"}" \
   "${KAGENTI_FLAGS[@]}"
 
+log_info "Waiting for kagenti pods..."
+deadline=$((SECONDS + 120))
+until kubectl get pod --no-headers -l app.kubernetes.io/name=kagenti-operator-chart \
+  -n kagenti-system 2>/dev/null | grep -q .; do
+  (( SECONDS >= deadline )) && { log_error "Timed out waiting for kagenti-operator pods"; exit 1; }
```
clawgenti: **must-fix**: Same `(( SECONDS >= deadline )) && { ...; exit 1; }` pattern under `set -euo pipefail` — the falsy arithmetic exits with code 1 and aborts the script before the timeout fires. Use `if (( SECONDS >= deadline )); then log_error ...; exit 1; fi` here as well.

clawgenti (commented):
Fix correctly addresses the idempotency issues with `--skip-cluster` re-runs — deleting stale webhooks/configmaps before Helm upgrades and replacing `--wait` with explicit `kubectl wait` loops is the right approach.

One must-fix: the arithmetic timeout checks will abort the script under `set -euo pipefail` before they even get a chance to trigger.

---

scripts/kind/setup-kagenti.sh
```diff
@@ -668,6 +676,16 @@ if $WITH_SPIRE; then
     --set tornjak-frontend.service.type=ClusterIP \
     --set tornjak-frontend.service.port=3000
 
+  log_info "Waiting for SPIRE pods in zero-trust-workload-identity-manager..."
+  deadline=$((SECONDS + 120))
+  until kubectl get pod --no-headers -l app.kubernetes.io/instance=spire \
+    -n zero-trust-workload-identity-manager 2>/dev/null | grep -q .; do
+    if (( SECONDS >= deadline )); then log_error "Timed out waiting for SPIRE pods"; exit 1; fi
```
clawgenti, line 683: **suggestion**: The 120-second deadline to detect pod existence may be tight on slow clusters (image pulls, resource pressure). If pods don't appear within 2 minutes, the script exits hard before the `kubectl wait --timeout=300s` ever runs. Consider bumping to 180s or making it configurable — the longer wait timeout below suggests this environment can be slow.

clawgenti (commented):
Shell script changes make re-runs idempotent by pre-deleting conflicting webhook configs and configmaps, then replacing Helm's broken `--wait` with explicit `kubectl wait` loops — the approach is solid and well-commented.

---

Alan-Cha (approved):
This PR fixes real idempotency bugs preventing `--skip-cluster` re-runs of the Kind setup script (webhook field-manager conflicts, SPIRE/kagenti `--wait` timeouts, stale OIDC configmap).

Verified against a live Kind cluster:
- Both webhook configs (`istiod-default-validator`, `istio-validator-istio-system`) exist and match the delete calls.
- Both new `kubectl wait` label selectors (`app.kubernetes.io/name=kagenti-operator-chart` in `kagenti-system`, `app.kubernetes.io/instance=spire` in `zero-trust-workload-identity-manager`) match real running pods.
- The `until ... | grep -q .` loops correctly use `--no-headers` and have an explicit deadline — the earlier bot-review findings (false-positive header match, missing timeout) are resolved.
- Confirmed the `(( SECONDS >= deadline ))` check is safe under `set -euo pipefail`: it's inside an `if` condition, which bash exempts from `errexit`, so it won't abort the script (verified with a standalone repro).
- The unwrapped `kubectl delete --ignore-not-found` calls follow the file's existing convention (other cleanup deletes in the script are unwrapped too), so this isn't a `--dry-run` regression.

**Author**: evaline-ju (MEMBER — maintainer)
**Areas reviewed**: Shell
**Commits**: 8 commits (6 requiring sign-off, all signed-off; 2 merge commits, exempt)
**CI status**: passing (Deploy & Test still pending at review time)

No must-fix issues found.

pdettori (approved):
LGTM

PR #2197, opened 2026-07-13
fix(installer): --with-skills must enable the backend

### LINKED ISSUE #2194: `--with-skills` leaves the backend disabled (skill auto-sync never runs)
### Summary
`scripts/kind/setup-kagenti.sh --with-skills` enables the UI flag but leaves the
backend **off**, so the subsequent `helm upgrade` removes a running
`kagenti-backend`. Because the skill **auto-sync loop runs inside the backend**
(`app/main.py` starts `run_skill_autosync_loop` when
`kagenti_feature_flag_external_skills` is on), auto-sync silently never runs.

### Root cause
The flag-dependency block resolves in the wrong order:

```bash
# UI requires backend API
if $WITH_UI && ! $WITH_BACKEND; then WITH_BACKEND=true; fi   # runs first
# Skills requires UI (and transitively backend)
if $WITH_SKILLS && ! $WITH_UI; then WITH_UI=true; fi         # sets UI *after*
```

With only `--with-skills`, the UI→backend rule evaluates `WITH_UI` while it is
still false, so `WITH_BACKEND` stays false; the skills→UI rule then sets
`WITH_UI=true` but nothing re-derives the backend. Installer reports
`Backend API: false`.

### Impact
Anyone enabling skills via `--with-skills` gets a broken skill auto-sync with no
error — the backend Deployment is removed.

### Fix
Reorder so dependencies cascade top-down (skills → UI → backend). One-line move.

### Repro
`./scripts/kind/setup-kagenti.sh --with-skills --skip-cluster` → summary shows
`Backend API: false`; `kubectl get deploy -n kagenti-system` has no
`kagenti-backend`.

PrielHaz:
# fix(installer): `--with-skills` must enable the backend

Closes #2194

## Problem
`setup-kagenti.sh --with-skills` enabled the UI flag but left `WITH_BACKEND=false`,
so `helm upgrade` removed a running `kagenti-backend`. Since the skill auto-sync
loop runs inside the backend, auto-sync silently never ran.

## Cause
The flag-dependency block resolved `UI → backend` **before** `skills → UI`, so the
backend rule saw `WITH_UI` still false.

## Fix
Reorder so dependencies cascade top-down: **skills → UI → backend**. Now
`--with-skills` brings the backend up as intended.

## Verification
```
$ ./scripts/kind/setup-kagenti.sh --with-skills --skip-cluster
    Backend API:   true      # was: false
```
`kubectl get deploy -n kagenti-system` shows `kagenti-backend` running; the
backend logs `Skill auto-sync started`.


pdettori (approved):
## Summary
Correct, minimal fix for the flag-dependency ordering bug. Verified against `main`: with the original order the `UI → backend` rule ran before `skills → UI`, so `--with-skills` alone left `WITH_BACKEND=false` and the subsequent `helm upgrade` (which reads `components.ui.enabled=${WITH_BACKEND}` and `metricsServer.enabled=${WITH_BACKEND}`) removed a running backend — disabling the in-backend skill auto-sync loop. Reordering to `skills → UI → backend` makes the cascade fire fully; the other dependency rules (Kiali/MLflow/Kuadrant) key off unrelated flags, so there are no side effects. The added comment is accurate.

**Author**: PrielHaz (FIRST_TIME_CONTRIBUTOR — first-time) — diff manually inspected and confirmed safe.
**Areas reviewed**: Shell
**Commits**: 1 change commit, DCO sign-off: passing
**CI status**: passing (DCO)

**nit (non-blocking):** the title/commit prefix `fix(installer):` uses scope parens; the org PR-title convention prefers a capitalized prefix without parens (`Fix: ...`). Not blocking — no title check is failing.

Thanks for the clear problem/cause/fix writeup and the verification steps. LGTM.

Assisted-By: Claude Code

Issue #2151, opened 2026-07-07
epic: Simulated MCP Tools from OpenAPI Specs in Kagenti (Runtime Simulation MVP)


bnayahu:
## Problem

Teams building agents against enterprise B2B, CRUD-oriented APIs (CRM, ITSM/ticketing,
HR/ERP, reservations, order/inventory) are routinely blocked because the real tool is
missing, unstable, rate-limited, costly, or gated behind credentials that aren't available
yet. The choices today are to hand-write and maintain an MCP server per tool, or to stand
up static mocks that can't hold state across a call sequence. Both are slow, throwaway work
that stalls agent development, demos, and evaluation before a backend exists.

Kagenti has robust support for skill and tool management, but has no framework-neutral
way to turn an API contract into a running, stateful tool that agents and the MCP Gateway
treat exactly like the real thing - and no way for a user to create and manage such a tool
from within the platform.

This epic is the first concrete step (or MVP) towards the runtime simulation north star. See
the *Additional context* section below.

## Design

The [Simulation Harness](https://github.com/kagenti/lab-runtime-simulation) turns an OpenAPI
spec into a running, LLM-driven MCP tool that makes two deliberately separate promises:

- **Interface-identical.** The simulated MCP server exposes exactly the operations,
  parameters, and response schemas of the source spec. Agents and the MCP Gateway interact
  with it precisely as they would with the real tool; the contract is guaranteed the same.
- **Behaviorally sufficient.** Backed by session-scoped state, responses are close enough to
  the real service to support the target use cases below - coherent across a CRUD lifecycle
  (create → read → update → list/delete) - without claiming byte-for-byte production fidelity.

In this epic, these promises are scoped to three target use cases:

- **Development stand-in.** Build and iterate an agent against a tool whose real backend, tenant,
  or credentials aren't available yet.
- **Demos & onboarding.** Seed demo and onboarding namespaces with working tools, with no real
  external backends.
- **Testing & evaluation.** Exercise and assert agent behavior against a controlled, reproducible
  seed dataset - known records, edge cases, scenario fixtures.

A simulation starts from mock data the harness generates automatically from the spec, but users
can supply their own test data to seed it instead, replacing the generated dataset. This is a
key capability for testing and evaluation: it lets a simulated tool be primed with known
records, edge cases, or scenario-specific fixtures so that agent behavior can be exercised and
asserted against a controlled, reproducible starting state rather than arbitrary generated data.

### User-facing flow

[...]

PR #2201, opened 2026-07-14
fix(ui): omit unset optional args on tool invoke (#2200)

### LINKED ISSUE #2200: bug(ui): tool invoke modal sends sentinel values for optional args, failing schema validation
## Description

The MCP tool **invoke modal** (`ToolDetailPage`) pre-seeds *every* input-schema
property with a sentinel value and sends the entire args object on invoke — even
for **optional** fields the user never touched. When an optional field carries a
JSON-schema constraint that the sentinel violates (e.g. `minimum: 1`), the call
is rejected by the tool's own request-body validator.

This is **not simulation-specific** — it lives in the general tool-invocation UI
and affects any MCP tool with an optional numeric parameter constrained
`minimum ≥ 1` (or a backend that rejects empty optional strings). It surfaced now
because simulated tools from OpenAPI specs (epic #2151) are the first tools with
such optional params being invoked from the UI.

## Steps to reproduce

1. Deploy a simulated tool whose spec has an optional numeric param with a
   `minimum ≥ 1` constraint. Example: the `restaurant-reservation-api`
   simulated tool, operation `searchRestaurants` — only `city` is required;
   `party_size` and `price_tier` are optional with `minimum: 1`.
2. Open the tool's detail page and click **Invoke** on `searchRestaurants`.
3. Fill in only the required `city` (e.g. `Boston`), leave the optional fields
   untouched, and submit.

## Expected

Optional fields the user did not set are omitted from the request; the call
succeeds with just `{ "city": "Boston" }`.

## Actual

The UI transmits sentinel values for the untouched optional fields
(`party_size: 0`, `price_tier: 0`, empty strings for optional strings), and the
call fails:

```
Input validation error: 0 is less than the minimum of 1
```

### Verified against the live tool (MCP call, bypassing the UI)

| Payload | Result |
|---------|--------|
| `{ "city": "Boston" }` (optionals omitted) | ✅ success → `[]` |
| `{ "city":"Boston", "cuisine":"", "date_time":"", "party_size":0, "price_tier":0, "distance_km":0 }` (UI default) | ❌ `Input validation error: 0 is less than the minimum of 1` |

The generated MCP `inputSchema` is **correct** (`required: ["city"]`, all else
optional, `party_size`/`price_tier` have `minimum: 1`). The defect is purely in
the UI's arg initialization/submission.

## Root cause

`kagenti/ui-v2/src/pages/ToolDetailPage.tsx`

- **`openInvokeModal` (~L249-267)** seeds *all* properties — required or not —
  with a sentinel: `0` for `number`/`integer`, `''` for `string`, `false` for
  `boolean`.
- **`handleInvoke` (~L280-284)** sends the full `toolArgs` object verbatim;
  untouched optional fields are never stripped.

So an optional `integer` with `minimum: 1` is sent as `0`, violating the schema.

## Suggested fix

Do not send optional fields the user hasn't set. Preferred: stop seeding
sentinels — initialize optional fields as absent and, on submit, drop keys that
are empty/untouched and not in `required`. (Dropping the `0` sentinel also avoids
the "did the user really type 0?" ambiguity that a naive strip would introduce.)

## Environment

- Component: `kagenti/ui-v2` (Tool detail / invoke modal)
- Related epic: #2151 (Simulated MCP Tools) — blocks UI testing of that feature
- Related (different defect, closed): #1715 (invoke dialog Enter-key)

bnayahu:
## Summary

The MCP tool **invoke modal** (`ToolDetailPage`) pre-seeded *every* input-schema property with a sentinel value (`0` for number/integer, `''` for string, `false` for boolean) and submitted the whole args object — including **optional** fields the user never touched. When an optional field carried a JSON-schema constraint the sentinel violates (e.g. `minimum: 1`), the tool's own request-body validator rejected the call.

This surfaced on simulated tools from OpenAPI specs (epic #2151): invoking `searchRestaurants` with only the required `city` failed with `Input validation error: 0 is less than the minimum of 1`, because untouched `party_size`/`price_tier` (both `minimum: 1`) were sent as `0`. The display logic (`value={String(toolArgs[key] || '')}`) even rendered the seeded `0` as blank, so the field *looked* empty while sending `0`.

It is not simulation-specific: any MCP tool with an optional numeric param constrained `minimum ≥ 1` (or a backend rejecting empty optional strings) hits it.

## Changes

- Add `kagenti/ui-v2/src/utils/toolInvoke.ts` with two pure, unit-tested helpers:
  - `initialInvokeArgs` — seeds only keys with an explicit schema `default` (no sentinel fabrication).
  - `buildInvokeArgs` — keeps a key iff it is `required` **or** its value is neither `undefined` nor `''`, so unset optionals are omitted while explicit `0`/`false` are preserved.
- Wire both into `ToolDetailPage`: `openInvokeModal` seeds via `initialInvokeArgs`; `handleInvoke` submits `buildInvokeArgs(toolArgs, required)`.
- Number input clears to `undefined` (not `0`); number/enum/string inputs display via `?? ''` (not `|| ''`) so a real `0` no longer renders blank.
- Keep the boolean `Switch` controlled (`isChecked={toolArgs[key] === true}`).

## Testing

- 6 new unit tests in `toolInvoke.test.ts` (covering `0`/`false`/empty-string/`undefined`/explicit-default); full `ui-v2` unit suite green (54/54); `npm run build` clean.
- Verified end-to-end on a Kind cluster against a live simulated tool: `searchRestaurants` with `city` only → success (previously failed); `city` + `cuisine` → success.

## Out of scope

Client-side required-field validation (blocking submit on an empty required field). A blank required field now yields a clear backend "missing required" error rather than a confusing `minimum` error — an improvement; stricter client-side validation is a separate follow-up.

## Related issue(s)

Fixes #2200

Assisted-By: Claude Code


pdettori (approved):
A focused, correct fix for #2200. The prior modal fabricated sentinel values (`0`/`''`/`false`) for every schema property and submitted untouched optionals, breaking tools with `minimum: 1` constraints. This PR extracts two pure, unit-tested helpers — `initialInvokeArgs` (seeds only explicit schema `default`s) and `buildInvokeArgs` (drops unset optionals while preserving explicit `0`/`false`) — wires them into `ToolDetailPage`, and fixes the `|| ''` → `?? ''` display bug that rendered a real `0` as blank.

Logic is sound and the edge cases (`0` / `false` / `''` / `undefined` / explicit default) are directly covered by the 6 new tests. Type compatibility checks out: `JSONSchemaProperty` carries `default?: unknown`, so `Record<string, JSONSchemaProperty>` is structurally assignable to the helper signature. Co-locating `toolInvoke.test.ts` matches the existing `validation.test.ts` convention. All 3 commits are DCO-signed and completed CI checks are green.

LGTM. Assisted-By: Claude Code

PR #2205, opened 2026-07-14
Fix: Rollback partial failure on agent/tool create

### LINKED ISSUE #1906: 🐛 Bug: Disappearing Agents and Tools on Dashboard

esnible:
## Summary

Resolves #1906

If Kagenti can create some but not all of the K8s resources for an agent or tool, the tool call fails, leaving the resources partially created.  These partial resources were usually invisible in the Kagenti UI and API.  If the underlying problem (resource exhaustion, etc) is fixed, the remains of the partially created agent or tool prevents the creation retry from being successful.

### DIFF
--- a/kagenti/backend/app/routers/agents.py
+++ b/kagenti/backend/app/routers/agents.py
@@ -12,7 +12,7 @@ import re
 import socket
 import ipaddress
 from datetime import datetime, timezone
-from typing import Any, Dict, List, Literal, Optional
+from typing import Any, Dict, List, Literal, Optional, Tuple
 from urllib.parse import urlparse
 
 import httpx
@@ -107,6 +107,7 @@ from app.routers.skills import _sanitize_k8s_name
 from app.utils.routes import (
     create_route_for_agent_or_tool,
     detect_platform,
+    rollback_workload_resources,
     route_exists,
     sanitize_log,
     select_route_port,
@@ -3757,6 +3758,10 @@ async def create_agent(
 
     request = apply_agent_import_defaults(request, kube)
 
+    # Persistent resources created during this call, tracked so we can roll them
+    # back if a later creation step fails (avoids leaking a workload, Service,
+    # AgentRuntime, or route). Only used by the image-deployment path below.
+    created: List[Tuple[str, str]] = []
     try:
         if request.deploymentMethod == "image":
             # Deploy from existing container image
@@ -3829,6 +3834,7 @@ async def create_agent(
                     namespace=request.namespace,
                     body=workload_manifest,
                 )
+                created.append(("Deployment", request.name))
                 logger.info(
                     f"Created Deployment '{request.name}' in namespace '{request.namespace}'"
                 )
@@ -3846,6 +3852,7 @@ async def create_agent(
                     namespace=request.namespace,
                     body=workload_manifest,
                 )
+                created.append(("StatefulSet", request.name))
                 logger.info(
                     f"Created StatefulSet '{request.name}' in namespace '{request.namespace}'"
                 )
@@ -3863,6 +3870,7 @@ async def create_agent(
                     namespace=request.namespace,
                     body=workload_manifest,
                 )
+                created.append(("Job", request.name))
                 logger.info(f"Created Job '{request.name}' in namespace '{request.namespace}'")
             elif request.workloadType == WORKLOAD_TYPE_SANDBOX:
                 sandbox_manifest = _build_sandbox_manifest(
@@ -3878,6 +3886,7 @@ async def create_agent(
                     namespace=request.namespace,
                     body=sandbox_manifest,
                 )
+                created.append(("Sandbox", request.name))
                 logger.info(f"Created Sandbox '{request.name}' in namespace '{request.namespace}'")
 
             # Create Service (not needed for Jobs).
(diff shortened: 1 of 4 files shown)


kagenti/backend/app/routers/agents.py
```diff
@@ -3924,6 +3935,11 @@ async def create_agent(
                     service_name=request.name,
                     service_port=service_port,
                 )
+                # create_route_for_agent_or_tool makes an HTTPRoute or an OpenShift
+                # Route depending on platform; track both so rollback deletes the
+                # right one (the other 404s and is swallowed).
+                created.append(("HTTPRoute", request.name))
```
clawgenti, line 3941: **nit**: Both `HTTPRoute` and `Route` are appended unconditionally before the platform is known, relying on rollback swallowing the 404 for whichever doesn't exist. The comment above (lines 3938-3940) explains this clearly — just noting it for reviewers who might flag it as a bug at first glance. Consider adding a short `# one of these will 404 and be swallowed` inline note on line 3942 to make the intent immediately obvious without reading the block comment above.

clawgenti (commented):
Adds a well-scoped `rollback_workload_resources` helper that cleans up partially-created K8s resources (Deployment/StatefulSet/Job/Sandbox, Service, AgentRuntime, HTTPRoute/Route) when an agent or tool creation fails mid-flight. Implementation is solid — LIFO delete order, best-effort with error swallowing, shared across `create_agent` and `create_tool`, with good test coverage.

- Commit subjects don't follow the repo's emoji-prefix convention (e.g. `Rollback partial failure on agent/tool create` → `🐛 Rollback partial failure on agent/tool create`)

---

pdettori (approved):
Reviewed the rollback logic end-to-end and verified every cross-reference against `upstream/main` and the PR head.

**Correctness — all verified:**
- Every `KubernetesService` delete method exists with matching signatures (`delete_deployment/statefulset/job/sandbox/service/custom_resource`, `list/delete_persistent_volume_claim`).
- Tracked names match created names: workload/route/AgentRuntime = `request.name`; agent Service = `request.name`; tool Service correctly tracks the `{name}-mcp` suffix.
- AgentRuntime rollback uses the same `CRD_GROUP/CRD_VERSION/AGENTRUNTIMES_PLURAL` as both `_ensure_agentruntime` and `_ensure_tool_agentruntime`; HTTPRoute/Route groups/versions/plurals match `create_httproute`/`create_openshift_route`.
- Sandbox PVC selector `app.kubernetes.io/name=<name>` matches both the creation label and `delete_agent`'s own cleanup.
- Handler ordering is correct: `ApiException` → `HTTPException` (re-raises 400 validations untouched) → generic `Exception`. `.append` runs only after a successful create, so a 409 on the first create leaves `created` empty and rollback is a no-op — a pre-existing agent/tool is never deleted. Rollback swallows all errors so it never masks the original exception.

**Design decisions look right:** shared/idempotent resources (SA, ConfigMaps, RoleBinding) are intentionally left in place, and the source-build path is scoped out of rollback since it already self-heals via its existing pre-import Build/BuildRun cleanup. Test coverage is thorough (reverse order, per-kind, PVCs, both route kinds, empty no-op, error swallowing, untracked resources).

**Author**: esnible (MEMBER — maintainer)
**Areas reviewed**: Python (backend routers + utils + tests)
**Commits**: 3, all signed-off: yes
**CI status**: passing (E2E HyperShift pending `/run-e2e`)

No blocking issues.

Assisted-By: Claude Code

PR #2214, opened 2026-07-15
Fix: Retry transient transport timeouts in agent conversation e2e tests

pdettori:
## Problem

The Kind e2e workflow (`E2E K8s 1.35.0`) intermittently fails on `test_agent_conversation.py::test_agent_simple_query` with a bare `httpx.ReadTimeout`. The same commit passes on re-run (observed on both `main` push and PR branches), so it is a flake, not a regression — the test code is unchanged.

## Root cause

Both agent-conversation tests wrap the streaming call in a retry loop, but the transient-error classification only ran when the agent **returned** a FAILED task with a matching error string in its body. A transport-level `httpx.ReadTimeout` — raised when the agent hangs on its downstream LLM/MCP call past the 120s client read timeout — is thrown out of `client.send_message()` **before any task result**, so it hit the outer `except` and called `pytest.fail` immediately with no retry:

```python
try:
    last_result = await _send_and_collect_response(client, user_message)
except Exception as e:
    pytest.fail(f"Error during A2A conversation: {e}")   # no retry
if last_result["task_failed"]:
    ... transient retry list (never reached for a transport timeout) ...
```

`ReadTimeout` also carries no message string, which is why the failure surfaced as an empty `Error during A2A conversation:`.

## Fix

- Add a shared `_is_transient_conversation_error()` classifier: transient httpx transport exceptions by type (`TimeoutException` — covers `ReadTimeout`/`ConnectTimeout`/etc. — plus `ConnectError`, `RemoteProtocolError`), and the existing message fragments.
- Route transient transport exceptions through the same retry + backoff path already used for FAILED tasks, in both `test_agent_simple_query` and `test_agent_multiturn_conversation`.
- Failure messages now include `type(e).__name__`, so a genuine timeout is diagnosable instead of showing an empty error.
- The multi-turn body check now reuses the full transient-fragment list instead of its previous two-item subset.

No production code changes; test resilience only. This complements the existing TODO to migrate to the sandbox agent (which has built-in SSE reconnection).

## Testing

- `python -m py_compile` + `ruff check` clean; pre-commit hooks pass.
- Logic-only change to retry classification; behavior on a genuine, persistent timeout is unchanged (fails after `_LLM_QUERY_MAX_ATTEMPTS`).

Assisted-By: Claude Code

### DIFF
--- a/kagenti/tests/e2e/common/test_agent_conversation.py
+++ b/kagenti/tests/e2e/common/test_agent_conversation.py
@@ -44,6 +44,37 @@ from kagenti.tests.e2e.conftest import (
 
 logger = logging.getLogger(__name__)
 
+# Error signatures that indicate a transient failure worth retrying. These match
+# either the text of a FAILED task's status message or the string form of an
+# exception raised while streaming the response.
+_TRANSIENT_ERROR_FRAGMENTS = (
+    "Cannot connect",
+    "Connection error",
+    "Expecting value",
+    "Error calling tool",
+    "timed out",
+    "Read timed out",
+    "ConnectionPool",
+)
+
+# httpx transport exceptions that are transient: the agent hung on its downstream
+# LLM/MCP call past the client read timeout, or the connection dropped. These are
+# raised by client.send_message() before any task result is returned, so they must
+# be classified by type — httpx.ReadTimeout in particular carries no message text.
+_TRANSIENT_TRANSPORT_EXCEPTIONS = (
+    httpx.TimeoutException,  # base of ReadTimeout, ConnectTimeout, WriteTimeout, PoolTimeout
+    httpx.ConnectError,
+    httpx.RemoteProtocolError,
+)
+
+
+def _is_transient_conversation_error(exc: Exception) -> bool:
+    """Return True if an exception raised while streaming the agent response is
+    transient and the query should be retried rather than failing the test."""
+    if isinstance(exc, _TRANSIENT_TRANSPORT_EXCEPTIONS):
+        return True
+    return any(frag in str(exc) for frag in _TRANSIENT_ERROR_FRAGMENTS)
+
 
 def _is_openshift_from_config():
     """Detect if running on OpenShift from KAGENTI_CONFIG_FILE."""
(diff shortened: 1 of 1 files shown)


mrsabath (approved):
Clean, well-scoped fix for a real e2e flake. The transport-level `httpx.ReadTimeout` was thrown out of `client.send_message()` before any task result, hitting the outer `except` → `pytest.fail` with no retry (and an empty error string). The shared `_is_transient_conversation_error()` classifier (type-based for httpx transport exceptions + the existing message fragments) routes those through the same retry/backoff path already used for FAILED tasks, in both test functions.

Verified:
- `httpx` / `asyncio` are imported.
- Retry-loop bounds (`for attempt in range(1, MAX+1)` with the `attempt < MAX` guard) match the existing `task_failed` path exactly — retries fire on attempts 1–4, attempt 5 falls through to `pytest.fail`. The `continue` re-enters the loop without double-sleeping.
- The httpx exception types are valid and the inline comment describing the hierarchy (`TimeoutException` as base of `ReadTimeout`/`ConnectTimeout`/`WriteTimeout`/`PoolTimeout`) is accurate.
- Test-only change; behavior on a genuine persistent timeout is unchanged (fails after `_LLM_QUERY_MAX_ATTEMPTS`).

Nice consolidation too: both tests now share `_TRANSIENT_ERROR_FRAGMENTS` instead of the multiturn path's previous two-item subset, and failures surface `type(e).__name__` so an empty-message timeout is finally diagnosable.

Optional nit (non-blocking): the commit subject `fix(test): ...` uses conventional-commits style rather than the emoji-prefix `git:commit` convention — the PR title (`Fix:`) passed `verify-pr-title`, so no action needed.

**Areas reviewed**: Python (retry / exception classification), commit & PR conventions
**Commits**: 3 (1 change + 2 main merges); DCO passed
**CI**: all green except the two e2e jobs (`Deploy & Test` pending, `E2E HyperShift` awaiting maintainer `/run-e2e`) — the very jobs this PR stabilizes.

Good catch on classifying the timeout by type — grazie, Paolo. LGTM. 👍

PR #2221, opened 2026-07-17
Chore: Remove orphaned in-cluster LiteLLM proxy deployment and scripts

### LINKED ISSUE #1126: fix: remove anyuid SCC requirement for LiteLLM proxy
LiteLLM's Prisma cache is stored under /root/.cache and requires root access. The current deployment uses anyuid SCC on OpenShift.

Options:
1. Custom Dockerfile: COPY --chown=1001 /root/.cache /home/litellm/.cache
2. Init container to copy Prisma binaries to emptyDir with world-read
3. Upstream PR to litellm

Reference: deployments/litellm/deployment.yaml

pdettori:
## Summary

The in-cluster LiteLLM proxy is **not deployed by any active path**, so its deployment manifests and scripts are dead code. This PR removes them, which also resolves the `anyuid` SCC concern in #1126 by deletion rather than by building a custom LiteLLM image.

### Why these are orphaned
- Nothing invokes `38-deploy-litellm.sh`. The main deploy flows (`kind/deploy-platform.sh`, `hypershift/create-cluster.sh`, `kind/run-e2e-tests.sh`) do not reference it.
- The Kind deployment was reverted when the sandbox infra landed (commit `095f85a8`, "revert Kind deployment").
- `91-test-litellm.sh` only tests a proxy that is never deployed.

### Removed
- `deployments/litellm/deployment.yaml` — the manifest #1126 references (still carried a live `TODO(#1126)` and the `anyuid` workaround)
- `deployments/litellm/service.yaml`
- `.github/scripts/kagenti-operator/38-deploy-litellm.sh` — the script that ran `oc adm policy add-scc-to-user anyuid`
- `.github/scripts/kagenti-operator/91-test-litellm.sh`

### Follow-up (not in this PR)
`.github/scripts/kagenti-operator/39-setup-llm-teams.sh` lists `38-deploy-litellm.sh` as a prerequisite (comment only) and creates LiteLLM teams/virtual-keys via the backend `llm_keys` router. It is functionally dead without the proxy and is also not invoked anywhere, but it touches live backend surface, so it's left out of this scope. Worth a separate decision on whether to retire the in-cluster LLM-team flow (and the backend router) entirely.

Closes #1126

Assisted-By: Claude Code


esnible (approved):
Clean, well-scoped deletion-only PR removing genuinely orphaned in-cluster LiteLLM proxy manifests and scripts.

**Verification (against `upstream/main`, freshly fetched):**
- All 4 deleted files exist on `main` — deletion is valid.
- The active deploy flows (`kind/deploy-platform.sh`, `kind/run-e2e-tests.sh`, `hypershift/create-cluster.sh`, `90-run-e2e-tests.sh`) are clean of any `litellm` / `38-deploy` / `91-test` references.
- The only lingering reference to the deleted scripts is the acknowledged comment in `39-setup-llm-teams.sh:8`, which the PR body already calls out as follow-up.
- Deletion legitimately closes #1126 — the deleted `deployment.yaml` and `38-deploy-litellm.sh` carried the `TODO(#1126)` and the `oc adm policy add-scc-to-user anyuid` workaround.

**One consumer the body doesn't mention (non-blocking):** `deployments/sandbox/llm-budget-proxy.yaml` sets `LITELLM_URL="http://litellm-proxy.kagenti-system.svc.cluster.local:4000"` and `kagenti/llm-budget-proxy/app/main.py` actively forwards LLM traffic to it — i.e. it points at exactly the `litellm-proxy` Service this PR deletes. This is consistent with the orphaned-code thesis, since the sandbox deploy path (`76-deploy-sandbox-agents.sh`) is itself not invoked by any active flow. Worth listing alongside `39-setup-llm-teams.sh` in the follow-up so the dangling `LITELLM_URL` default gets revisited whenever the in-cluster LLM-team / sandbox-budget flow is retired.

**Author:** pdettori (MEMBER — maintainer) · **Commits:** 1, signed-off · **Config files (.claude/.vscode):** none · **CI:** passing (E2E HyperShift pending `/run-e2e`, expected).

Approve.

Assisted-By: Claude Code

Issue #1972, opened 2026-06-15
renaming Kagenti to rossoctl


Ibrahim2595:
## Summary

This issue tracks the rename of **Kagenti** → **rossoctl** across our GitHub organization, repositories, documentation, website, release artifacts, and developer experience.

The goal is to execute the rename with minimal disruption to users while keeping the migration coordinated across all repositories.

**Please use this issue to raise concerns or dependencies before work begins.**

---

# Rollout Plan

### Repository cleanup

* [x] Archive repositories that are no longer maintained or relevant.
* [x] Confirm with Red Hat which repositories, APIs, or package names must remain stable and for how long.

### PR freeze

* [x] Merge or close all ready pull requests before the rename.
* [x] Notify maintainers to land pending work before the migration begins.

### Release checkpoint

* [x] Cut a final **Kagenti** alpha release to preserve a pre-rename baseline.

### Website

Landing page + docs-hide items moved to the cleanup tracker #2291.

### Rename automation

Dropped as obsolete — the rename completed without a formalized playbook.

---

### GitHub organization

* [x] Rename the GitHub organization to **rossoctl**.
* [x] Verify GitHub redirects are working correctly (`kagenti` org redirects to `rossoctl`).

### Repository rename

Rename repositories using the following convention:

* `rossoctl`
* `operator`
* `examples`
* ...

Experimental repositories:

* `experimental-<name>`

### Repository migration

Rename repositories one at a time in dependency order to minimize breakage.

For each repository:

* [ ] Rename the repository.
* [ ] Apply the rename playbook.
* [ ] Run tests and CI.
* [ ] Update release artifacts.
* [ ] Verify links and documentation.
* [ ] Merge.

---

## Phase 3 — Launch

* [x] Announce the rename across community channels.
* [x] Update external references where we have control.

  * [x] Rename the Kagenti YouTube channel to Rossoctl (#2282). Done — https://www.youtube.com/@Rossoctl
  * [x] Rename the Kagenti Medium blog to Rossoctl (#2283). Done — https://medium.com/rossoctl-the-agentic-platform (site refs: rossoctl/.github#105)
  * [ ] Rename the Kagenti Slack workspace and Google Group to Rossoctl (#2285). Moved to cleanup tracker #2291.

> Landing-page publish + docs-hide (target Jul 31) and the Slack/Google Group work moved to the cleanup tracker **#2291**.

Ibrahim2595:
[...]

Issue #485, opened 2025-12-15
Update documentation to new installer and deprecate old installer


rubambiza:

pdettori:
this has been completed - we deprecated the old installer, removed it and updated all docs.

PR #2230, opened 2026-07-19
⚠️ Rename Kagenti → Rossoctl across the flagship repo (→ rossoctl)

Ibrahim2595:
Full content + path rename Kagenti → Rossoctl. All 7928 case-insensitive `kagenti` references updated and 29 kagenti-named paths renamed; grep is clean.

Paths renamed (git mv):
- Python package kagenti/ → rossoctl/ (backend, ui-v2, tui, keycloak, auth, tests)
- charts/kagenti → charts/rossoctl; charts/kagenti-deps → charts/rossoctl-deps
- Java pkg io/kagenti → io/rossoctl; setup/cleanup-kagenti.sh → -rossoctl.sh; .claude/skills/kagenti* → rossoctl*; .github/scripts/kagenti-operator → operator; docs/diagrams

Content:
- Brand / prose / comments: Kagenti → Rossoctl
- Python: dist kagenti-project → rossoctl-project; packages=["rossoctl"]; every import kagenti / from kagenti → rossoctl (0 remaining)
- Go TUI module: github.com/kagenti/kagenti/kagenti/tui → github.com/rossoctl/rossoctl/rossoctl/tui
- npm: @kagenti/ui → @rossoctl/ui; kagenti-app-demo → rossoctl-app-demo
- Java: io.kagenti.keycloak(.authenticator) → io.rossoctl.keycloak(.authenticator)
- Images ghcr.io/kagenti/* → ghcr.io/rossoctl/*; chart kagenti-operator-chart → operator-chart; cross-repo kagenti-extensions → rossocortex, kagenti-operator → operator (repo) / rossoctl-operator (deployed identity)
- CODEOWNERS @kagenti/* → @rossoctl/*; CI reusable-workflow refs kagenti/.github → rossoctl/.github
- Installer env interface KAGENTI_* → ROSSOCTL_* (~90 vars: KAGENTI_NS, KAGENTI_REPO, KAGENTI_CONFIG_FILE, ...). No KAGENTI_* backward-compat fallback (per "no kagenti").
- Runtime contracts: CRD groups agent.kagenti.dev → agent.rossoctl.dev, mcp.kagenti.com → mcp.rossoctl.com; label/annotation domains kagenti.io/*, protocol/inject/integrations/openshell.kagenti.io → *.rossoctl.io; namespaces kagenti-system/-webhook-system → rossoctl-*; Keycloak realm kagenti → rossoctl

Verification:
- grep sweep: 0 `kagenti` (content + paths); no malformed names (two rossoctl-rossoctl-* are faithful renames of pre-existing doubled source strings)
- Python: compileall clean; pytest rossoctl/backend/tests → 673 passed; no broken internal rossoctl.* imports; uv.lock consistent (uv lock --check)
- ruff check ✅ and ruff format ✅ (7 files reformatted — renamed identifiers grew 1 char)
- Go TUI: go build ✅, go vet ✅, go test ✅ (all packages)
- NOT run locally: root e2e suite (needs kubectl + kind/hypershift cluster), Java/Maven build of keycloak SPI (no maven), Helm chart rendering (no helm)

Related to #1972

Assisted-By: Claude Opus 4.8 <noreply@anthropic.com>

Fixes #


charts/rossoctl/Chart.lock
```diff
@@ -1,6 +1,6 @@
 dependencies:
-- name: kagenti-operator-chart
-  repository: oci://ghcr.io/kagenti/kagenti-operator
+- name: operator-chart
+  repository: oci://ghcr.io/rossoctl/operator
   version: 0.3.0-alpha.8
 digest: sha256:6a51019293663bc08a44202438a00936c3c4b378045f344fd92061ec7b197d19
```
mrsabath: **must-fix — this is what fails `chart-tests`.**

```
Error: the lock file (Chart.lock) is out of sync with the dependencies file (Chart.yaml). Please update the dependencies
```

The dependency was renamed in both files — `Chart.yaml` now has `name: operator-chart` / `repository: oci://ghcr.io/rossoctl/operator`, and this lock was string-substituted to match — but `digest: sha256:6a51019293663bc08a44202438a00936c3c4b378045f344fd92061ec7b197d19` was carried over verbatim from the old `kagenti-operator-chart` entry. Helm computes the lock digest from the resolved dependency set (name + repo + version), so renaming the dep invalidates it and `helm dependency build` refuses to proceed.

Regenerate instead of hand-editing:

```bash
helm dependency update charts/rossoctl
helm dependency update charts/rossoctl-deps   # if its Chart.lock exists too
```

then commit the updated `Chart.lock`. (Requires the renamed `ghcr.io/rossoctl/operator` chart to be pullable, which ties into the registry-publish sequencing noted in the summary — you may need that published first to regenerate.)

mrsabath (changes requested):
## Summary

Structurally this is a faithful, thorough rename — I spot-checked the sensitive substitutions (image path doubling `rossoctl/rossoctl/*`, `kagenti-extensions`→`rossocortex`, CRD groups `*.rossoctl.dev`/`*.rossoctl.com`, `kagenti.io/*`→`rossoctl.io/*` labels, `KAGENTI_*`→`ROSSOCTL_*` env) and they match the pre-rename source and the #1972 rollout plan. All residual `kagenti`-named files are old paths being removed, each with an added `rossoctl` counterpart. Sign-off/DCO ✅.

Requesting changes on **one genuine, self-inflicted CI blocker** (Chart.lock). The other failing checks are rename-transition artifacts, explained below so they aren't mistaken for defects.

### Blocking

- **`chart-tests` — Chart.lock out of sync (inline).** The dependency was renamed in both `Chart.yaml` and `Chart.lock`, but the stored `digest` was carried over unchanged. Helm derives the digest from name+repo+version, so `helm dependency build` fails. Regenerate the lock (see inline).

### Non-blocking (transition sequencing — surfaced so reviewers don't misread them)

- **`Deploy & Test` — 403 on `oci://ghcr.io/rossoctl/operator/operator-chart`.** The refs are rewritten correctly, but the `rossoctl` org/registry isn't published yet. Per #1972 this is a dependency-ordering step (publish `ghcr.io/rossoctl/*` images+charts first). E2E can't go green until then — not a code defect in this PR.
- **`CodeQL` — "82 new alerts incl. 4 critical".** False-positive artifact of a 555-file rename: findings like `Incomplete URL substring sanitization` fire on `rossoctl.io/` string literals whose logic is identical to `main`. CodeQL's own note: *"Alerts not introduced by this PR might have been detected because the changes were too large."* Recommend confirming zero net-new alerts vs `main` before merge.
- **`Dependency Review` — `cryptography@46.0.5` (high).** Pre-existing vuln re-surfaced only because the lockfile path moved (`kagenti/…`→`rossoctl/…`). Worth a separate bump; shouldn't block the rename.

### Nit

- Title is 74 chars (guideline is <72) and uses a bare `⚠️` prefix. The org title-rule Ruleset expects an allowed word prefix (e.g. `Refactor:`); the emoji alone may trip that required check on a future re-run even though it's currently green. Consider e.g. `Refactor: Rename Kagenti → Rossoctl across the flagship repo`.

**Areas reviewed**: Helm/charts, CI (5 failing checks investigated end-to-end), path renames (974 files), Python packaging, security scanners, commit conventions.
**CI status**: 5 failing (1 fixable now, 4 transition artifacts), rest passing; E2E HyperShift awaiting `/run-e2e`.

charts/rossoctl/Chart.lock
Ibrahim2595: Fixed in b36471d. Didn't hand-edit — regenerated the lock digest for the renamed `operator-chart` dependency via Helm's `HashReq`, and validated the recompute by reproducing the pre-rename digest (`sha256:6a51019…`) byte-for-byte, so the new value is trustworthy.

`helm dependency build charts/rossoctl` now passes the digest check (`Saving 1 charts`) and fails only on the expected 403 pulling the not-yet-published `oci://ghcr.io/rossoctl/operator` — which folds into the `ghcr.io/rossoctl/*` publish sequencing from #1972 (same class as Deploy & Test), not a defect in this PR.

Ibrahim2595:
Thanks for the thorough review, @mrsabath 🙏

**Blocking — fixed (b36471d):** regenerated `charts/rossoctl/Chart.lock`'s digest for the renamed `operator-chart` dep (validated by reproducing the pre-rename digest exactly). `helm dependency build` now clears the digest check and only 403s on the not-yet-published `oci://ghcr.io/rossoctl/operator`, collapsing into the same `ghcr.io/rossoctl/*` publish ordering as Deploy & Test. Details in the inline thread.

**Non-blocking — agree with your read:**
- **Deploy & Test 403** → publish `ghcr.io/rossoctl/*` images + charts first (dependency ordering per #1972).
- **CodeQL "82 new alerts"** → large-diff false positives (findings fire on `rossoctl.io/` string literals whose logic is identical to `main`). Will confirm zero net-new vs `main` before merge.
- **`cryptography@46.0.5`** → pre-existing, surfaced only because the lockfile path moved. Tracking as a separate bump.

**Title nit:** happy to switch to a word prefix. One question — the emoji `pr-verifier` check is currently green on the `⚠️` prefix, so which form satisfies *both* it and the title Ruleset? (e.g. does `⚠️ Refactor: …` pass, or should it be a bare `Refactor:`?) I'll apply the same across all four rename PRs (#733, #678, #485, and this one).

mrsabath:
**Blocked on Phase 3** of the rename rollout ordering (see kagenti/kagenti#1972). This PR unblocks once `oci://ghcr.io/rossoctl/operator/operator-chart` is published — then `helm dependency update charts/rossoctl` + `charts/rossoctl-deps` regenerates the correct `Chart.lock` digest. Note: string-replacing the digest won't pass `chart-tests` — Helm computes it from the resolved dependency, so the OCI chart must be pullable first.

PR #2250, opened 2026-07-23
Feat: Add --with-simulated-tools convenience flag to Kind setup

bnayahu:
## Summary

Adds a `--with-simulated-tools` convenience flag to `scripts/kind/setup-rossoctl.sh` so the simulated MCP tools feature (`featureFlags.simulatedTools`) can be enabled at install time without hand-writing a Helm values override file.

Key changes:
- New `--with-simulated-tools` flag wired through the script: variable declaration, arg parser, `--help` text, config summary echo, and `--set featureFlags.simulatedTools`.
- Auto-enables `--with-backend` and `--with-ui` (the feature-flag env var lands on the backend and the feature surfaces in the UI), resolved before the UI→backend cascade — mirrors the `--with-skills` dependency rule.
- Opt-in only: **not** included in `--with-all` (consistent with `--with-skills`).
- Feature remains off by default; nothing changes unless the flag is passed.

Tested via an incremental re-run against a local Kind cluster: the installer exits 0, the summary prints `Simulated Tools: true`, and `ROSSOCTL_FEATURE_FLAG_SIMULATED_TOOLS` flips to `true` on the backend (with the `simulatedTools`-gated PVC RBAC applied) while co-installed components stay intact.

## Related issue(s)

_None._

Assisted-By: Claude Code

### DIFF
--- a/scripts/kind/setup-rossoctl.sh
+++ b/scripts/kind/setup-rossoctl.sh
@@ -10,7 +10,8 @@
 # Optional:        --with-istio (ambient mesh), --with-spire, --with-backend,
 #                  --with-ui, --with-mcp-gateway, --with-kuadrant, --with-otel,
 #                  --with-mlflow, --with-builds, --with-kiali,
-#                  --with-agent-sandbox, --with-skills, --with-all
+#                  --with-agent-sandbox, --with-skills, --with-simulated-tools,
+#                  --with-all
 #
 # Idempotent: safe to re-run. Uses helm upgrade --install and kubectl apply.
 # Re-running with additional --with-* flags adds components incrementally.
@@ -49,6 +50,7 @@ WITH_KIALI=false
 WITH_KUADRANT=false
 WITH_AGENT_SANDBOX=false
 WITH_SKILLS=false
+WITH_SIMULATED_TOOLS=false
 SKILL_REGISTRY_ALLOWED_HOSTS=""
 WITH_ALL=false
 SKIP_CLUSTER=false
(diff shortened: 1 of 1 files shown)


mrsabath (approved):
Clean, correctly-structured installer flag addition — adds `--with-simulated-tools` to `setup-rossoctl.sh`, modeled precisely on the existing `--with-skills` pattern. All six wiring points are present and consistent: variable init (off by default), arg parser, `--help` text, dependency cascade, config-summary echo, and the Helm `--set`.

Verified against the repo:
- `featureFlags.simulatedTools` is a **real** chart value (`charts/rossoctl/values.yaml:40`, epic #2151) — not invented.
- `--with-all` correctly **excludes** it (opt-in only, matching `--with-skills`).
- The dependency-cascade ordering is right: the `simulatedTools → UI` rule is placed *before* the `UI → backend` rule, so enabling the flag cascades to both UI and backend. The comment explains why.

Feature-gated, off by default, backward compatible. **Verdict: APPROVE** — no issues.

**Areas reviewed**: Shell (flag wiring + dependency cascade), Helm value existence. **Commits**: 1, signed-off (DCO). **CI**: passing.

cwiklik (approved):
Adds a `--with-simulated-tools` convenience flag to `scripts/kind/setup-rossoctl.sh`, flipping the existing `featureFlags.simulatedTools` chart value. Textbook mirror of the `--with-skills` pattern: off-by-default variable, arg parser, `--help` text, config-summary echo, and the `--set`, plus a correctly-ordered UI→backend auto-enable cascade (the simulated-tools→UI rule is placed before the UI→backend rule, so backend gets pulled in).

Verified beyond the diff:
- `featureFlags.simulatedTools` is a real chart key (values.yaml:40) and the feature is wired end-to-end — `backend/app/core/config.py`, `routers/simulation.py`, the `/config/features` endpoint, UI hooks/pages, and `tests/test_simulation_flag.py` — so the `--set` genuinely activates the feature rather than being a no-op.
- Off by default and correctly excluded from `--with-all` (not in the `WITH_ALL` block), consistent with the repo's opt-in feature-flag convention.

All CI green (shellcheck, Deploy & Test, verify-pr-title, Trivy, CodeQL, DCO); single signed commit; no `.claude`/`.vscode`. LGTM.

Assisted-By: Claude Code

Issue #2167, opened 2026-07-08
Simulated tools: worked example, E2E test, and docs


bnayahu:
Part of epic #2151.

## User story

**As an Agent Developer**, I want a worked example, an automated E2E test, and documentation, so that the simulated-tool path is proven end to end and easy to follow.

## Scope — in

- A **named sample OpenAPI spec** → simulated tool → an agent calling it through a coherent **create → read → update → list/delete** flow (this exercises and **asserts the session-scoped state coherence** that is epic story #4).
- An **E2E test in `kagenti/tests/e2e/`** driven through the UI/API flow, covering create → Generating → Ready → agent CRUD flow, plus at least one failure path.
- **Docs:** update `docs/new-tool.md` (or a sibling page) to cover the simulated-tool path end to end, including **egress allow-listing** of the inference endpoint (needed at generation and serving time) and the pre-provisioned `LLM_API_KEY` Secret.
- **Demo/onboarding seeding** (folded in from epic story #7): document/script standing up a simulated tool in a demo/onboarding namespace with no real external backend, using the worked example as the seed.

## Scope — out

- CI integration of simulated tools as pipeline fixtures (out of epic scope).
- Evaluation/admission gating (out of epic scope).

## Key files

- `kagenti/examples/` (sample spec + agent)
- `kagenti/tests/e2e/` (new E2E)
- `docs/new-tool.md` (or sibling)
- Demo/onboarding seeding assets (installer/demo docs)

## Acceptance criteria

- The worked example passes as an **automated E2E test**: with the flag on, a simulated tool is created from a spec, moves Generating → Ready, shows the `SIMULATED` badge, and an agent completes a coherent CRUD flow against it.
- A failure path (e.g. missing LLM Secret or forced generation failure) is asserted to surface Failed/Error with a reason.
- `docs/new-tool.md` documents the path end to end, including egress allow-listing and the LLM Secret prerequisite.
- The demo/onboarding namespace can be seeded with a working simulated tool with no real external backend.

## Dependencies

- All other simulated-tools issues in this epic.

PR #2251, opened 2026-07-23
fix(simulated-tools): adopt workloads via AgentRuntime instead of self-labeling

bnayahu:
## Summary

Makes the simulated-tools feature actually functional when the `agent-label-protection` ValidatingAdmissionPolicy is installed. The backend was baking `rossoctl.io/type=tool` directly onto the StatefulSet (metadata + pod template), which the VAP reserves for the operator SA — so every create from the `rossoctl-backend` SA was denied (`403 -> API 502`) and the feature was non-functional. This routes simulated tools through the same operator-adoption path regular tools already use, so no VAP exemption is needed.

> **Depends on #2250** (the `--with-simulated-tools` flag). This branch is stacked on it; until #2250 merges, the diff here also includes that flag commit. Please merge after #2250.

## Key changes

- **Bare workload + AgentRuntime adoption.** The StatefulSet is created without `rossoctl.io/type`; a new `AgentRuntime{type: tool, targetRef -> StatefulSet}` CR lets the operator (the only VAP-exempt SA) stamp the label during reconciliation. Mirrors the regular-tool path.
- **Generation gating.** Adoption rolls the pod exactly once (the operator writes a `config-hash` annotation onto the pod template that the backend cannot precompute). Generation is therefore deferred until the AgentRuntime reports `Ready/Configured` with `configuredPods >= 1`, then the spec is posted to the settled pod — so it can't land on a pod about to be recreated. The gate short-circuits if the AgentRuntime 404s (tool deleted mid-provisioning).
- **Delete path.** Removes the AgentRuntime CR (after the workload, so the operator finalizer no-ops).
- **Standalone.** `build_simulation_agentruntime` lives in `simulation_manifests` and uses the generic `create`/`delete_custom_resource` helpers, so `app/routers/tools.py` is untouched.

## Testing

- Backend unit tests: **674 passed** (new manifest / AgentRuntime-builder / provision-wait / lifecycle coverage added); pylint 9.97/10.
- E2E on Kind (`tests/e2e/common/test_simulated_tool.py`) with **no VAP exemption** — both happy and failure paths pass. Verified: create returns 202 (no 403/502), the operator adopts the workload and stamps `rossoctl.io/type=tool`, the gate defers generation ~35s until `Configured` with no mid-generation pod restart, then generation -> Ready -> MCP CRUD -> clean delete (StatefulSet, AgentRuntime, Service, PVC).

## Follow-ups (out of scope here)

- **Orphan per-tool ServiceAccount.** The backend has create-but-not-delete RBAC on serviceaccounts in agent namespaces, so the per-tool SA is not removed on delete. This is a pre-existing, platform-wide leak (agents and regular tools do the same). Preferred fix: ownerReference-based GC in `ensure_service_account` so all workload types are covered uniformly.
- **CI coverage.** No workflow runs `--with-simulated-tools` (and `--with-all` excludes it), so this path is not yet exercised in CI. Worth adding a job that deploys with the flag and runs the simulated-tool E2E.

Related: epic #2151, issue #2167.

Assisted-By: Claude Code

### DIFF
--- a/rossoctl/backend/app/core/config.py
+++ b/rossoctl/backend/app/core/config.py
@@ -102,6 +102,11 @@ class Settings(BaseSettings):
     simulation_harness_request_timeout: float = 10.0
     # Seconds between trigger-task POST attempts while the harness is still starting.
     simulation_trigger_poll_interval: int = 5
+    # Max seconds to wait for the operator to adopt the workload (AgentRuntime
+    # targetRef) and roll the pod onto the configured revision before the trigger
+    # task posts the spec. Bounds the "provisioning" wait so a stuck operator does
+    # not hang generation forever; on timeout the trigger proceeds best-effort.
+    simulation_provision_timeout: int = 180
     # Auto-inject MCP_URL / LLM env vars on agent import (TUI parity; weather demo defaults)
     rossoctl_feature_flag_agent_import_defaults: bool = False
     skill_autosync_interval: int = (
--- a/rossoctl/backend/app/routers/simulation.py
+++ b/rossoctl/backend/app/routers/simulation.py
@@ -28,7 +28,10 @@ from pydantic import BaseModel, field_validator
 from app.core.auth import ROLE_OPERATOR, ROLE_VIEWER, require_roles
 from app.core.config import settings
 from app.core.constants import (
+    AGENTRUNTIMES_PLURAL,
     APP_KUBERNETES_IO_NAME,
+    CRD_GROUP,
+    CRD_VERSION,
     DEFAULT_IN_CLUSTER_PORT,
     ROSSOCTL_SIMULATED_LABEL,
     TOOL_SERVICE_SUFFIX,
@@ -45,6 +48,7 @@ from app.services.simulation_harness_client import (
 from app.services.simulation_manifests import (
     MAX_SIMULATION_NAME_LEN,
     EnvVar,
+    build_simulation_agentruntime,
     build_simulation_env_vars,
     build_simulation_service,
     build_simulation_statefulset,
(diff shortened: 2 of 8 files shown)


mrsabath (approved):
High-quality fix that makes simulated-tools work under the `agent-label-protection` VAP. Root cause is correct: the backend SA was setting `rossoctl.io/type=tool` directly on the StatefulSet, which the VAP reserves for the operator SA → every create 403'd. The fix creates a **bare workload + `AgentRuntime{targetRef→StatefulSet}`** and lets the operator (the VAP-exempt SA) stamp the label during reconciliation.

**Verified against the repo:**
- `build_simulation_agentruntime` mirrors the regular-tool path exactly — same CR shape as `tools._ensure_tool_agentruntime` (`spec.type` + `spec.targetRef` + optional `authBridgeMode`). The "no VAP exemption needed" thesis holds because the label-stamping is done by the operator SA, identical to regular tools.
- Constants (`CRD_GROUP` / `CRD_VERSION` / `AGENTRUNTIMES_PLURAL`) are real (resolve from settings), not invented.

**The adopt-and-roll race is handled carefully:** `_wait_for_runtime_configured` defers spec delivery until `Ready=True` AND `configuredPods >= 1`, so the spec can't land on a pod about to be recreated by the one-time adoption roll. A 404 short-circuits (tool deleted mid-provision), other API errors retry to a bounded deadline, and timeout proceeds best-effort (later caught by the status watchdog). Create/delete ordering is correct — CR after workload on create; workload before CR on delete so the operator finalizer no-ops.

**Tests are genuinely assertive:** the 6 `_wait_for_runtime_configured` tests cover every branch (ready+configured, poll-until-configured with exact call/sleep counts, timeout, transient-retry, 404 short-circuit asserting `sleep.assert_not_awaited()`, non-404 retry), plus manifest/lifecycle/trigger coverage. Reported 674 passing, pylint 9.97.

**Verdict: APPROVE** — correct root-cause fix, verified against the real operator path, careful about the reconciliation race, thoroughly tested. The two documented follow-ups (orphan per-tool SA; missing `--with-simulated-tools` CI coverage) are pre-existing / out-of-scope and correctly flagged.

Notes: stacked on #2250 (the `setup-rossoctl.sh` change is that PR's flag commit `56363f0` — already approved; only `447d4b2` is new here), so merge after #2250. E2E HyperShift is pending a maintainer `/run-e2e`, not a failure.

**Areas reviewed**: Python (adoption + generation-gating logic), K8s CR shape (verified vs regular-tool path), tests, config. **Commits**: 2 (1 new), signed-off (DCO). **CI**: passing (E2E manual-trigger pending).

PR #2286, opened 2026-07-27
fix(kind): enforce safe SPIFFE CSI driver minimum

### LINKED ISSUE #2288: SPIFFE CSI mount explosion can degrade nodes and disrupt storage
## Summary

The Rossoctl SPIRE installation path currently uses SPIRE chart `0.27.0`, which embeds SPIFFE CSI driver `0.2.7`. That driver version can create duplicate bind mounts during kubelet reconciliation because `NodePublishVolume` is not idempotent.

With the CSI driver's bidirectional mount propagation, repeated publish calls can cause mount counts to grow rapidly enough to degrade the entire node.

## Observed incident

This occurred on a three-node vanilla Kubernetes development cluster using the Rossoctl SPIRE installation path.

- One workload using the SPIRE CSI volume accumulated 16,384 duplicate mounts at its CSI target path.
- The affected node reached approximately 33,000 mounts in total.
- SSH login took tens of seconds.
- Processes that inspect the mount table became extremely slow.
- systemd-managed network, DNS, and time services repeatedly missed their startup windows.
- Distributed-filesystem disks served by the affected node were marked down even though the node remained reachable.

The mount stack was visible under a path of this form:

```text
/var/lib/kubelet/pods/<pod-uid>/volumes/kubernetes.io~csi/spire-agent-socket/mount
```

## Root cause

Affected SPIFFE CSI driver versions unconditionally perform another bind mount when kubelet calls `NodePublishVolume` for a target that is already published. Kubelet reconciliation or restart can therefore add duplicate mount entries rather than receiving the idempotent success required by the CSI specification.

The upstream defect is documented in:

- https://github.com/spiffe/spiffe-csi/issues/331
- https://github.com/spiffe/spiffe-csi/pull/332

The fix was released in SPIFFE CSI driver `0.2.12`. The current patch release is `0.2.13`.

## Recovery and verification

1. Stopped the workload responsible for the active CSI target.
2. Detached the duplicate mount stack.
3. The affected node immediately dropped from approximately 33,000 mounts to a few hundred.
4. SSH and systemd-managed services returned to normal behavior.
5. Upgraded only `spiffe-csi-driver` from `0.2.7` to `0.2.13` through the SPIRE Helm value override.
6. The CSI DaemonSet became healthy on all nodes and the mount leak did not recur.

No SPIRE chart or workload configuration changes were otherwise required.

## Installation-path coverage

- **Kind:** covered directly by the proposed installer change.
- **Vanilla Kubernetes:** covered because `scripts/k8s/setup-rossoctl.sh` delegates to `scripts/kind/setup-rossoctl.sh` with `ROSSOCTL_SETUP_FLAVOR=k8s --skip-cluster`. This is the path on which the incident was observed.
- **OpenShift with ZTWIM:** not changed by this fix. ZTWIM manages its own CSI driver image through OLM, so the deployed operator bundle must be checked separately to determine whether it contains the upstream idempotency fix.
- **OpenShift with `useSpireHelmChart=true`:** not directly covered because that mode expects SPIRE Helm installation outside this installer. The external installation must explicitly select SPIFFE CSI driver `0.2.12` or newer.

## Expected behavior

Rossoctl should not install a SPIFFE CSI driver older than `0.2.12`. The installer should use a concrete safe default while allowing newer versions and rejecting overrides below the fixed minimum.

## Proposed fix

See #2286, which:

- defaults to SPIFFE CSI driver `0.2.13`;
- allows a newer version through `SPIFFE_CSI_DRIVER_VERSION`;
- rejects versions below the safe minimum, `0.2.12`; and
- rejects tags that do not use the exact numeric `X.Y.Z` image-tag format; and
- passes the selected tag explicitly to the SPIRE Helm release.

moonlight16:
## What changed

- Default the SPIFFE CSI driver used by the Kind installer to `0.2.13`.
- Allow newer versions through the `SPIFFE_CSI_DRIVER_VERSION` environment variable while rejecting versions below the safe minimum, `0.2.12`, or tags outside the numeric `X.Y.Z` image-tag format.
- Pass the selected image tag explicitly to the SPIRE Helm release instead of inheriting the SPIRE chart's current `0.2.7` default.

## Why

SPIFFE CSI driver versions through `0.2.11` do not make `NodePublishVolume` idempotent. Kubelet reconciliation or restart can therefore create duplicate bind mounts. With bidirectional mount propagation, those mounts can accumulate rapidly and exhaust node resources.

The defect is documented in [spiffe/spiffe-csi#331](https://github.com/spiffe/spiffe-csi/issues/331) and fixed by [spiffe/spiffe-csi#332](https://github.com/spiffe/spiffe-csi/pull/332). The fix was released in `0.2.12`; this change defaults to the current patch release, `0.2.13`, while permitting later safe versions.

The SPIRE `0.27.0` chart currently embeds SPIFFE CSI driver `0.2.7`, so an explicit override is needed until a fixed chart version is available and adopted.

The observed mount explosion, node-level impact, root cause, and recovery are documented in #2288.

The change covers both Kind and vanilla Kubernetes: `scripts/k8s/setup-rossoctl.sh` delegates installation to the patched script with the Kubernetes setup flavor. It does not change the OpenShift/ZTWIM path, where OLM controls the CSI driver image and the operator bundle must be verified separately.

Fixes #2288.

## Validation

- `bash -n scripts/kind/setup-rossoctl.sh`
- `git diff --check`
- Verified that `0.2.11`, `v0.2.13`, `0.2.13-rc1`, and malformed values are rejected, while `0.2.12`, `0.2.13`, `0.3.0`, and `1.0.0` are accepted.
- Confirmed that `spiffe-csi-driver.image.tag=0.2.13` renders and deploys successfully with SPIRE chart `0.27.0` in a live Kubernetes cluster.

### DIFF
--- a/scripts/kind/setup-rossoctl.sh
+++ b/scripts/kind/setup-rossoctl.sh
@@ -97,7 +97,9 @@ log_warn()    { echo -e "${YELLOW}⚠${NC} $1"; }
 log_error()   { echo -e "${RED}✗${NC} $1"; }
 
 version_at_least() {
-  local version="${1#v}" minimum="${2#v}"
+  # Image tags use numeric X.Y.Z releases. Reject prefixes and pre-release
+  # suffixes so a version can pass validation only if that exact tag is usable.
+  local version="$1" minimum="$2"
   local version_major version_minor version_patch
   local minimum_major minimum_minor minimum_patch
 
@@ -264,7 +266,8 @@ while [[ $# -gt 0 ]]; do
       echo "  --with-istio        Enable full Istio ambient mesh (mTLS, waypoints)"
       echo "                      Gateway API controller is always installed as core"
       echo "  --with-spire        Install SPIRE + SPIFFE IdP setup"
-      echo "                      Override CSI with SPIFFE_CSI_DRIVER_VERSION (minimum 0.2.12)"
+      echo "                      Override CSI with SPIFFE_CSI_DRIVER_VERSION"
+      echo "                      (numeric X.Y.Z tag; minimum 0.2.12)"
       echo "  --with-backend      Install Rossoctl backend API"
       echo "  --with-ui           Install Rossoctl UI (auto-enables backend)"
       echo "  --with-mcp-gateway  Install MCP Gateway"
@@ -374,7 +377,7 @@ if $WITH_KUADRANT && ! $WITH_MCP_GATEWAY; then
 fi
 
 if $WITH_SPIRE && ! version_at_least "$SPIFFE_CSI_DRIVER_VERSION" "$SPIFFE_CSI_DRIVER_MIN_VERSION"; then
-  log_error "SPIFFE_CSI_DRIVER_VERSION must be a semantic version at least"
+  log_error "SPIFFE_CSI_DRIVER_VERSION must be a numeric X.Y.Z image tag at least"
   log_error "${SPIFFE_CSI_DRIVER_MIN_VERSION}; got ${SPIFFE_CSI_DRIVER_VERSION}"
   exit 1
 fi
--- a/scripts/kind/setup-rossoctl.sh
+++ b/scripts/kind/setup-rossoctl.sh
@@ -70,6 +70,10 @@ CERT_MANAGER_VERSION="v1.17.2"
 ISTIO_VERSION="1.28.0"
 SPIRE_CRD_VERSION="0.5.0"
 SPIRE_VERSION="0.27.0"
+# v0.2.12 fixed a mount leak that can exhaust the host mount table and disrupt
+# node services. Allow newer versions while enforcing the safe minimum.
+SPIFFE_CSI_DRIVER_MIN_VERSION="0.2.12"
+SPIFFE_CSI_DRIVER_VERSION="${SPIFFE_CSI_DRIVER_VERSION:-0.2.13}"
 GATEWAY_API_VERSION="v1.4.0"
 TEKTON_VERSION="v0.66.0"
 SHIPWRIGHT_VERSION="v0.14.0"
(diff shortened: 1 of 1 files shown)


evaline-ju:
@moonlight16 the kind CI test failures showing up here should be remediated in the merged #2313, if you wanted to rebase

scripts/kind/setup-rossoctl.sh
```diff
@@ -244,6 +266,8 @@ while [[ $# -gt 0 ]]; do
       echo "  --with-istio        Enable full Istio ambient mesh (mTLS, waypoints)"
       echo "                      Gateway API controller is always installed as core"
       echo "  --with-spire        Install SPIRE + SPIFFE IdP setup"
+      echo "                      Override CSI with SPIFFE_CSI_DRIVER_VERSION"
```
Alan-Cha, line 269: nit: `docs/developer/kind.md` and `docs/getting-started/install.md` both have a flag table listing `--with-spire` that isn't updated to mention this new `SPIFFE_CSI_DRIVER_VERSION` override. It's discoverable via `--help` here, so not blocking, but worth a one-line addition to those tables for consistency.

Alan-Cha (approved):
### Summary
Defaults the SPIFFE CSI driver to 0.2.13 and validates any override via `SPIFFE_CSI_DRIVER_VERSION` against a safe minimum (0.2.12), rejecting malformed or pre-release tags. Passes the resolved tag explicitly to the SPIRE Helm release via `--set-string`, overriding the SPIRE chart's stale 0.2.7 default. Fixes the mount-leak/exhaustion issue described in spiffe/spiffe-csi#331, tracked in #2288.

Verified `version_at_least()` by hand: correctly accepts 0.2.12/0.2.13/0.3.0/1.0.0 and rejects 0.2.11, `v0.2.13`, `0.2.13-rc1`, short (`1.2`, `5`) and malformed (`0.2.12.1`, empty) inputs — matches the PR's stated validation exactly, including edge cases not explicitly listed. The check is correctly gated behind `$WITH_SPIRE` so it doesn't fire for installs that skip `--with-spire`. `--set-string` is the right choice for the Helm flag (avoids numeric-tag coercion issues that plain `--set` could hit).

**Areas reviewed**: Shell
**Commits**: 2 commits, all signed-off: yes
**CI status**: passing (all checks green)

mrsabath (approved):
Independently verified this fix end to end.

- Reproduced `version_at_least`: accepts `0.2.12 / 0.2.13 / 0.3.0 / 1.0.0`, rejects `0.2.11 / v0.2.13 / 0.2.13-rc1 / 1.2 / 5 / 0.2.12.1 / ""`, and behaves correctly under the script's `set -euo pipefail` (the trailing `(( ))` chain is safely consumed by the `if` context).
- Confirmed the Helm key path against the pulled SPIRE `0.27.0` chart: the `spiffe-csi-driver` subchart ships `image.tag: ""`, which defaults to its `appVersion: 0.2.7`, so the explicit `--set-string "spiffe-csi-driver.image.tag=..."` override is genuinely required. `--set-string` correctly avoids numeric-tag coercion and matches sibling flags in the same release.
- The gate is correctly behind `$WITH_SPIRE`, and the fix reaches the vanilla-Kubernetes path via `scripts/k8s/setup-rossoctl.sh` `exec`-ing this script; the OLM-managed OpenShift path is untouched, as stated.

Areas reviewed: Shell, Helm value wiring. Commits: 2, all signed-off. CI: all green. Corroborates @Alan-Cha's approval.

Issue #2284, opened 2026-07-27
release-validation (Kind E2E): MLflow SDK trace read-back returns 0 despite traces persisted — test_mlflow_traces fails on v0.7.0-rc.1


cwiklik:
## Summary
`release-validation.yaml` → **Kind E2E → "Run backend E2E tests"** fails on `tests/e2e/common/test_mlflow_traces.py`: the MLflow Python SDK reports **0 traces** even though the trace pipeline is healthy and traces **are** persisted in MLflow's database. Reproduced on **two consecutive `v0.7.0-rc.1` runs** — not flaky.

Failing tests:
- `TestWeatherAgentTracesInMLflow::test_weather_agent_traces_exist`
- `TestWeatherAgentTracesInMLflow::test_weather_trace_has_spans`
- `TestGenAITracesInMLflow::test_trace_has_tree_structure`

Runs:
- https://github.com/rossoctl/rossoctl/actions/runs/30291969076
- https://github.com/rossoctl/rossoctl/actions/runs/30295693206

## Impact / severity
- **Non-functional — observability read-path only.** The trace *pipeline* works; traces are captured and persisted (see Evidence). This is a **read-back** failure in the e2e, not a platform defect.
- **GA blocker, not an RC blocker.** Per the release SOP, RC validation must be green on the *final RC before GA*. `v0.7.0-rc.1` can ship with this as a documented known issue; it **must be resolved and re-validated before `v0.7.0` GA**.
- **Regression window:** these same tests **passed on 2026-06-30** (run 28478512307, pre-rename) and fail now — introduced somewhere in the kagenti→rossoctl rename window.

## Evidence (the pipeline is healthy)
- ✅ Weather-agent conversation ran and returned 200s.
- ✅ Agent → OTEL collector: weather-service POSTs OTLP to `otel-collector.rossoctl-system:8335/v1/traces` → **HTTP 200**.
- ✅ Collector → MLflow → DB: **`test_mlflow_otel_metrics_received` PASSES**, querying MLflow's Postgres (`trace_info`) directly and finding traces present.
- ❌ SDK read-back: the 3 failing tests use `MlflowClient.search_experiments()` → `search_traces(locations=[exp.experiment_id])` over the `mlflow.<domain>:8080` ingress and get **`[]`** despite the DB rows.

## Root-cause analysis (narrowed)
**Ruled out (with evidence):**
- **Not the trace pipeline** — traces flow + persist (above).
[...]

PR #2309, opened 2026-07-28
Test(e2e): diagnose MLflow trace read-back (#2284)

cwiklik:
## What
Adds fail-path diagnostics to `get_all_traces()` in the MLflow-traces E2E to root-cause #2284 — the MLflow SDK returns 0 traces even though traces are persisted in the DB (a read-back regression that surfaced in the kagenti→rossoctl rename window).

## Why a diagnostic, not a fix
Static analysis is exhausted: the test and the MLflow SDK are unchanged since these tests last passed (2026-06-30), and `mlflow.auth.enabled=false` — so the cause needs one live signal. On the **empty-read path only**, this logs:
- resolved tracking URI + `search_experiments()` result (names / ids / count)
- per-experiment `search_traces(locations=...)` errors (previously swallowed)
- **PROBE A** — deprecated `search_traces(experiment_ids=...)`
- **PROBE B** — explicit experiment `"0"`/`"1"` (the OTLP exporter writes `x-mlflow-experiment-id: "0"`)
- **PROBE C** — raw REST `experiments/search` (bypasses the SDK)

## How the re-run pins the fix
- empty over REST / `search_experiments` → read-path/ingress issue → **chart fix**
- experiments present but `locations=` misses them while `experiment_ids=` / explicit-id finds them → **test/SDK fix**

## Testing
- `ruff check` clean; mock-exercised `_diagnose_empty_traces` with a fake client → runs to completion, no raise, all probes execute and errors are caught.
- **No behavior change on healthy runs** — diagnostics fire only when 0 traces are found.

Refs #2284

Assisted-By: Claude Code

### DIFF
--- a/rossoctl/tests/e2e/common/test_mlflow_traces.py
+++ b/rossoctl/tests/e2e/common/test_mlflow_traces.py
@@ -402,6 +402,68 @@ def get_mlflow_client():
     return _mlflow_client
 
 
+def _diagnose_empty_traces(client, experiments) -> None:
+    """Diagnostics for issue #2284: the MLflow SDK returns 0 traces even though
+    traces are persisted in the DB (a read-back regression that surfaced in the
+    kagenti->rossoctl rename window). This distinguishes a read-path/ingress issue
+    (the SDK sees no experiments over the ingress) from an experiment-linkage/API
+    issue (experiments present, but ``search_traces(locations=...)`` misses the
+    OTLP-ingested traces). Runs ONLY when 0 traces were found, so it adds no noise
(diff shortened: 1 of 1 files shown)


cwiklik:
Superseded by #2313.

Root cause of #2284 turned out to be MLflow **auth** on the read path (`MLFLOW_TRACKING_TOKEN` missing/expired → the SDK read-back returned empty, swallowed by the test) — test/backend-side, not a chart/pipeline bug. #2313 both surfaces the auth error and makes the e2e tests pass (on top of the mcp pin #2311), so the diagnostic probes added here are no longer needed.

Closing in favor of #2313. Refs #2284.

PR #2311, opened 2026-07-28
fix(backend): pin mcp SDK <2.0.0

evaline-ju:
mcp 2.0.0 removed/renamed streamablehttp_client from mcp.client.streamable_http, breaking the backend on startup. Pin to <2 until we migrate to the new API.

## Summary

[MCP 2.0.0](https://pypi.org/project/mcp/2.0.0/) released today removed `streamablehttp_client` from `mcp.client.streamable_http` and is breaking the rossoctl backend at this time. For now, we pin the upper bound of `mcp` to avoid the breaking API, but follow-ups will be needed to address this (follow-up issue #2312 created)


evaline-ju:
backend e2e tests for mlflow traces already failing and tracked in #2284

cwiklik:
The pin here (`mcp>=1.0.0,<2`) is correct — but it's now byte-identical to the pin included in #2313 (which also carries the MLflow auth fix and closes #2284). To avoid double-merging the same pin, recommend **closing this in favor of #2313** once that's out of draft and its Deploy & Test is green.

(If the 0.7 pin is needed *sooner* than #2313 clears e2e, the alternative is to merge this first and let #2313 rebase — your call.)

_Assisted-By: Claude Code_

evaline-ju:
Will just close in favor of/consolidate on #2313

Issue #1178, opened 2026-04-09
epic: OpenClaw Link Health Agent for kagenti org


rubambiza:
## Problem

Documentation link rot is a silent problem. CI-based link checking (lychee in PR workflows) only catches broken links when files are modified -- it does not detect:

- External links that rot over time (sites go down, pages move) between PRs
- Cross-repo broken references across the kagenti org
- Cumulative link health trends across the organization

There is no proactive, org-wide mechanism to surface broken links, triage them, or fix them automatically.

## Design

A two-agent system built on OpenClaw using the Standing Order + isolated Cron Job pattern:

**Scanner Agent** (cheap model: gemini-flash, runs Mon/Wed/Fri 11:00 UTC)
- A bash script does the mechanical work: scanning, diffing, issue creation, reporting. The model's role is to invoke the script and summarize results.
- Repos are maintained as full clones by an existing nightly cron job that auto-discovers new repos via `gh repo list`.
- Runs lychee on each repo, reusing existing `.lychee.toml` configs where present
- Diffs results against previous scan to identify new, recurring, and fixed broken links
- Creates GitHub Issues following the [bug report template](https://github.com/kagenti/.github/blob/main/.github/ISSUE_TEMPLATE/3-bug-report.yml) structure (`:bug:` prefix, `kind/bug` + `broken-link/internal` or `broken-link/external` labels when permissions allow)
- Auto-closes issues for links fixed since the last scan
- Writes structured reports and updates `docs/link-health.md` in kagenti/kagenti via a fork-based PR

**Fixer Agent** (model: sonnet, runs Tue/Thu 14:00 UTC)
- A bash script handles deterministic work: parsing issues, re-verifying links, searching for moved files via `git log --follow` and `find`. The model handles ambiguous cases (multiple candidate matches, deleted files).
- Queries GitHub for open issues matching the scanner's issue format (structured body with repo, file, URL, category)
- Re-verifies each broken link before attempting a fix; closes issues where links have become valid
- For internal links: resolves moved files, fixes relative paths, anchors, cross-repo references
- For external links: comments on the issue with context for manual resolution (no automated PRs)
- Opens fork-based PRs with fixes for human review (internal links only)
- Comments on the original issue with the PR link or explanation of why automated fix is not possible

[...]

PR #2315, opened 2026-07-29
docs: Link health report (auto-updated)

clawgenti:
## Summary

Auto-updated by Kagenti Link Health Scanner. This PR is continuously updated with each scan. Merge when convenient.

| Metric | Value |
|--------|-------|
| Repos scanned | 8 |
| Broken (internal) | 14 |
| Broken (external) | 12 |
| New since last scan | +2 |
| Fixed since last scan | -12 |

## Related issue(s)

- kagenti/kagenti#1178

### DIFF
--- a/docs/link-health.md
+++ b/docs/link-health.md
@@ -1,22 +1,23 @@
 # Link Health Report
 
-> Last scan: 2026-07-29 07:06 ET | Scan ID: 2026-07-29-001
+> Last scan: 2026-07-31 07:01 ET | Scan ID: 2026-07-31-001
 
 ## Summary
 
 | Metric | Value |
 |--------|-------|
 | Repos scanned | 8 |
-| Total links checked | 1802 |
-| Broken (internal) | 24 |
+| Total links checked | 1836 |
+| Broken (internal) | 14 |
 | Broken (external) | 12 |
-| New since last scan | +36 |
-| Fixed since last scan | -735 |
+| New since last scan | +2 |
+| Fixed since last scan | -12 |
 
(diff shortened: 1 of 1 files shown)


docs/link-health.md
```diff
@@ -1,26 +1,24 @@
----
-draft: true       # excluded from https://www.rossoctl.dev/
----
-
```
evaline-ju, line 4: we should probably keep clawgenti from deleting headers cc: @rubambiza

rubambiza: I partly agree. I think this is part of a broader rethinking I need to do now that the `docs` folder has been re-established as the source for the docs page under `rossoctl.dev`. I am curious if the machinery can ignore subfolders of docs where I can move these reports, or perhaps these reports _do_ belong somewhere on the docs page now that `automation` and `agent-skills` are core repos. We can discuss further offline.

Ultimately, the original file said clearly that this file should not be edited "manually", and the header was _rightly_ inserted at some point to avoid rendering it. So, I see both sides and why clawgenti would opt to remove it.

Btw, the remote host of clawgenti is currently down, and I cannot do much at the moment in terms of edits to the standing order instructions that could temporarily address this.

rubambiza: FYI: https://github.com/rossoctl/automation/issues/44

rubambiza:
Closing the PR because the reports now have a new home per https://github.com/rossoctl/automation/issues/44. Moreover, it cannot merge because removing the headers would mean it starts showing up on rossoctl website under [documentation](https://www.rossoctl.dev/docs/), which is unintended.

PR #2322, opened 2026-07-29
fix(backend): resolve OpenShift Route targetPort by service port name

### LINKED ISSUE #2191: bug: OpenShift Route targetPort set to port number instead of port name, causing 503 for agent traffic
## Summary

When Kagenti deploys an agent on OpenShift, it creates a Route with `targetPort` set to the numeric service port (`8080`) instead of the service port name (`http`). OpenShift's router resolves `targetPort` by name when the Service port has a name set — a numeric `targetPort` that doesn't match any named port causes the router to fail to find a healthy backend and return HTTP 503 to all external traffic, even though the pod is running and healthy.

## Steps to Reproduce

1. Deploy an agent to an OpenShift cluster via Kagenti (e.g. `POST /api/v1/agents` with `servicePorts: [{name: "http", port: 8080, targetPort: 8000}]` and `createHttpRoute: true`)
2. Wait for the pod to reach Running state
3. Curl the external Route URL

```
HTTP 503
```

Internal in-cluster traffic (e.g. `curl http://<agent>.<ns>.svc.cluster.local:8080`) works fine.

## Expected Behavior

The Route is accessible externally with HTTP 200. The router correctly forwards traffic through the Service to the pod.

## Root Cause

Kagenti creates the OpenShift Route with:

```yaml
spec:
  port:
    targetPort: 8080   # numeric — does not match any named port
```

But the Service it points to has:

```yaml
spec:
  ports:
  - name: http
    port: 8080
    targetPort: 8000
```

On OpenShift, when a Service port carries a name, the Route's `targetPort` must use that **name** (e.g. `http`), not the port number. From the [OpenShift Route docs](https://docs.openshift.com/container-platform/4.14/networking/routes/route-configuration.html):

> `targetPort` — The target port on pods selected by the service this route points to. If this is a string, it will be looked up as a named port in the target endpoints port list.

Because no port in the Service is *named* `8080`, the router cannot resolve the backend and returns 503.

## Fix

Set `targetPort` in the generated Route to the service port **name** instead of the port number:

```yaml
spec:
  port:
    targetPort: http   # matches the service port name
```

## Workaround

Patch the Route after deployment:

```bash
kubectl patch route <agent-name> -n <namespace> \
  --type=json \
  -p='[{"op":"replace","path":"/spec/port/targetPort","value":"http"}]'
```

## Environment

- OpenShift 4.x (tested on `apps.ykt3.hcp.res.ibm.com`)
- Kagenti agent deployment via `POST /api/v1/agents` with `protocol: a2a`, `deploymentMethod: image`

webchang:
## Related issue(s)

Fixes #2191

## Summary
- OpenShift's router matches a Route's `spec.port.targetPort` against the **name** of the backing Service port whenever that port is named. The backend currently emits the numeric Service `port`, so the router fails to resolve it and returns **HTTP 503** for agent/tool routes on OpenShift.
- Adds `resolve_openshift_target_port()`, which returns the Service port **name** when the matching port is named, and falls back to the numeric port otherwise. `create_openshift_route()` now uses it.
- The Gateway API `HTTPRoute` path is intentionally unchanged, since `backendRefs.port` requires an integer.

This reproduces on OpenShift clusters (observed on internal `ykt2`/`ykt3` deployments): agent/tool Services expose a named `http` port, and the generated Route's numeric `targetPort` returns 503 until patched to the port name.

## Test plan
- [x] `pytest tests/test_routes.py` — 18 passed (3 new in `TestCreateOpenShiftRoute`)
  - named port → `targetPort` == `"http"`
  - unnamed port → falls back to numeric
  - Service lookup failure (ApiException) → falls back to numeric

### DIFF
--- a/rossoctl/backend/app/utils/routes.py
+++ b/rossoctl/backend/app/utils/routes.py
@@ -165,6 +165,34 @@ def create_httproute(
             raise
 
 
+def resolve_openshift_target_port(
+    kube: KubernetesService,
+    service_name: str,
+    namespace: str,
+    service_port: int,
+):
+    """Resolve the value an OpenShift Route should use for spec.port.targetPort.
+
+    OpenShift's router matches a Route's targetPort against the *name* of the
+    backing Service port whenever that port is named; a numeric value equal to
+    the Service's `port` (rather than its container `targetPort`) fails to
+    resolve and the router returns 503. Return the port name when the matching
+    Service port is named, otherwise fall back to the numeric port.
+    """
+    try:
+        service = kube.get_service(namespace=namespace, name=service_name)
+        for sp in service.get("spec", {}).get("ports", []) or []:
+            if sp.get("port") == service_port and sp.get("name"):
+                return sp["name"]
+    except ApiException:
+        logger.warning(
+            "Could not look up Service %s in %s to resolve route targetPort; using numeric port",
+            sanitize_log(service_name),
+            sanitize_log(namespace),
+        )
+    return service_port
+
+
 def create_openshift_route(
     kube: KubernetesService,
     name: str,
@@ -185,6 +213,7 @@ def create_openshift_route(
     name = sanitize_log(name)
     namespace = sanitize_log(namespace)
     service_name = sanitize_log(service_name)
+    target_port = resolve_openshift_target_port(kube, service_name, namespace, service_port)
     route_manifest = {
         "apiVersion": "route.openshift.io/v1",
         "kind": "Route",
@@ -198,7 +227,7 @@ def create_openshift_route(
         "spec": {
             "path": "/",
             "port": {
-                "targetPort": service_port,
+                "targetPort": target_port,
             },
             "to": {
                 "kind": "Service",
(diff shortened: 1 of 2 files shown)


cwiklik (approved):
Correct, well-scoped fix for the OpenShift 503s (#2191). `resolve_openshift_target_port()` returns the Service port **name** when the matching port is named and falls back to the numeric port otherwise, and `create_openshift_route()` uses it — matching OpenShift router semantics (named service ports must be referenced by name in `spec.port.targetPort`).

Verified:
- The resolver reads only `port` and `name`, which are identical across the raw-API JSON and the Python-client dict shapes, so it's robust to whichever `get_service` returns.
- Safe fallbacks: unnamed port → numeric; `ApiException` → numeric (with a sanitized warning log).
- The lookup key is the `sanitize_log`'d `service_name`/`namespace`, and `sanitize_log` only strips `\n`/`\r`/`\x00` (invalid in k8s names anyway), so for valid names the lookup resolves the real Service and stays consistent with the route's `to.name`.
- Gateway API `HTTPRoute` correctly left numeric (backendRefs.port requires an int).
- 3 new tests assert the named / unnamed / lookup-failure branches.

22 checks green; E2E HyperShift pending a maintainer `/run-e2e` (non-blocking). Signed commit, no `.claude`/`.vscode`. LGTM.

Assisted-By: Claude Code

PR #2325, opened 2026-07-30
Docs: Make titles smaller, reword to provide context to reader

esnible:
## Summary

- Rename https://www.rossoctl.dev/docs/category/concepts to "Kubernetes deployment pattern"
- Remove "RossoCortex delivers apabilities and abstractions required for secure deployments"
- Remove https://www.rossoctl.dev/docs/overview/why-choose
- Make many headers smaller visually


Alan-Cha (approved):

docs/concepts/ibac-plugin.md
```diff
@@ -2,7 +2,7 @@
 description: IBAC (Intent-Based Access Control).
 ---
 
-# IBAC (Intent-Based Access Control)
+### IBAC (Intent-Based Access Control)
```
mrsabath, line 5: **Suggestion (non-blocking):** the H1 is demoted to `###` here, but this page's frontmatter has only `description:` — no `title:` or `sidebar_label:`. In the standard Docusaurus setup the page title is derived from the first H1, so demoting it may leave the page without a proper `<h1>`/title. The sibling `sparc-plugin.md` (unchanged in this PR) keeps its `#` H1 with the same frontmatter shape, so this also diverges from that convention. Separately, the page now runs `###` (this heading) straight to `####` (Threat Model / Try the Demo!) with no H1/H2 above it — a heading-level skip. Suggest either keeping `#` here or adding a `title:`/`sidebar_label:` to the frontmatter. Could you confirm how it renders on the built site?

docs/concepts/contextguru.md
```diff
@@ -2,7 +2,7 @@
 description: Shrink your LLM context.
 ---
 
-# Context-guru
+### Context-guru
```
mrsabath, line 5: **Suggestion (non-blocking):** same as `ibac-plugin.md` — H1 demoted to `###` on a page whose frontmatter is `description:`-only (no `title:`/`sidebar_label:`), which risks a title-less page since Docusaurus derives the title from the first H1. The headings also go `###` → `####` with nothing above. Suggest keeping `#` or adding a frontmatter title.

mrsabath (approved):
Docs-only PR that normalizes heading levels site-wide (demotes in-body H1s and shifts subordinate headings down a level), renames the components.md title to "Kubernetes deployment pattern", merges the new-tool.md H1 into its Overview heading, trims a sentence from 1-what-is.md, and deletes the why-choose overview page. No broken links or stale anchors — all same-file TOCs key off heading text and still resolve, and nothing links to the deleted why-choose page.

**Areas reviewed:** Docs (Markdown), internal links/anchors, heading structure
**Commits:** 1, signed-off (DCO passes)
**CI:** all checks green

Approving. The two inline suggestions are non-blocking — they ask you to confirm how a couple of pages render on the built site, since their in-body H1 was the page-title source.

**Nit on `docs/overview/3-why-choose.md` (deleted):** removing this page drops the value-proposition list ("Rossoctl gives you: Zero-Trust Security Architecture, Authentication and Authorization, Trusted workload identity (SPIRE), …") from the overview. Nothing links to `why-choose`, so no broken links, and this looks intentional per the PR summary. Worth confirming the value-prop content is captured elsewhere — the "Supported AI Use-Case Types" part overlaps `docs/use-case-types.md`, but the gives-you list appears unique to this page. (Left in the body rather than inline since the file is fully deleted.)

PR #2335, opened 2026-08-01
fix(backend): fall back to legacy agent card endpoint

### LINKED ISSUE #858: feature: Use new URL for fetching Agent Cards
### Feature Description

Many of our agent examples, such as the Image Agent, show warnings when Kagenti talks to them:

```
WARNING:a2a.server.apps.jsonrpc.jsonrpc_app:Deprecated agent card endpoint '/.well-known/agent.json' accessed. Please use '/.well-known/agent-card.
```

This issue is to modify Kagenti so that it doesn't use the deprecated endpoint triggering the warnings.  We should also verify our agent examples work with the new endpoint.

### Proposed Solution

Fetch /.well-known/agent-card.json and fall back to /.well-known/agent.json if the former is not found.

### Want to contribute?

### Additional Context

_No response_

w3lld1:
## Summary

- I added a shared agent-card fetch path that tries `/.well-known/agent-card.json` first.
- I fall back to the deprecated `/.well-known/agent.json` endpoint only when the new endpoint returns 404, preserving other HTTP error behavior.
- I added focused regression coverage for both invoke URL resolution and the agent-card API.

## Related issue(s)

Fixes #858

## Testing Instructions

- `uv run --frozen --extra dev pytest tests/test_chat_invoke_url.py -q` (21 passed)
- `uv run --frozen --extra dev ruff check app/routers/chat.py tests/test_chat_invoke_url.py`
- `uv run --frozen --extra dev ruff format --check app/routers/chat.py tests/test_chat_invoke_url.py`
- `python3 -m py_compile app/routers/chat.py tests/test_chat_invoke_url.py`
- `git diff --check origin/main...HEAD`

I did not run the cluster-backed end-to-end suite locally; I left that validation to CI because this change is covered by focused backend unit tests.

### DIFF
--- a/rossoctl/backend/app/routers/chat.py
+++ b/rossoctl/backend/app/routers/chat.py
@@ -50,11 +50,11 @@ _invoke_url_cache: dict[tuple[str, str], tuple[str, float]] = {}
 _CACHE_TTL_SECONDS = 30.0
 
 
-async def _fetch_agent_card(client: httpx.AsyncClient, base_url: str) -> httpx.Response:
+async def _fetch_agent_card(client: httpx.AsyncClient) -> httpx.Response:
     """Fetch an agent card, falling back to the deprecated endpoint on 404."""
-    response = await client.get(f"{base_url}{A2A_AGENT_CARD_PATH}")
+    response = await client.get(A2A_AGENT_CARD_PATH)
     if response.status_code == 404:
-        response = await client.get(f"{base_url}{A2A_LEGACY_AGENT_CARD_PATH}")
+        response = await client.get(A2A_LEGACY_AGENT_CARD_PATH)
     response.raise_for_status()
     return response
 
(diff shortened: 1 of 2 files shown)


w3lld1:
Addressed both CodeQL findings in `1ca592ba`: agent-card requests now use constant relative paths on an `httpx` client configured with the validated internal service base URL, and the agent-card route validates both Kubernetes labels before URL construction. I added coverage for invalid identities and reran the focused module (22 passed), Ruff, and `git diff --check`.

cwiklik (approved):
Clean, well-scoped fix (Fixes #858) — and it improves the security posture along the way.

- `_fetch_agent_card` tries `/.well-known/agent-card.json`, falls back to legacy `/.well-known/agent.json` **only on 404**, then `raise_for_status()`. So non-404 errors are preserved rather than masked, and a both-404 case raises → the caller's existing `except` returns the base URL (the old "404 → base URL" behavior is preserved via the fetch-failure path, not lost).
- The fetch now uses a constant path against a `base_url` client, so there's no caller-controlled request path.
- `get_agent_card` now validates `name`/`namespace` against the RFC1123 regex and returns 400 before `resolve_agent_url` — input hardening the endpoint previously lacked.

Tests are assertive: both fallback tests assert the exact call order (`agent-card.json` → `agent.json`), and the invalid-identity test asserts `resolve` is not called (validation-before-resolve). No regression — the old 404 test is subsumed by the fetch-failure + legacy-fallback tests.

No `.claude`/`.vscode`, no secrets, no dependency changes; both commits signed; 21 checks green (E2E HyperShift pending a maintainer trigger). Nicely done — thanks for the contribution. LGTM.

Assisted-By: Claude Code

Issue #1426, opened 2026-05-02
epic: Eliminate Keycloak Admin Credentials from Agent Pods (Phases 1–3)


mrsabath:
> **✅ CLOSED / SPLIT 2026-07-16.** Phases 1–3 (drive Keycloak **admin** credentials out of agent namespaces and out of the operator) are complete. The remaining Phase 4 (replace agent **client secrets** with federated SPIFFE client auth — externally blocked on RHBK/Keycloak 26.6.x) and Phase 5 (sunset the client-registration sidecar) have moved to the follow-up epic **#2220**. This epic is closed on its shipped milestone: **admin-credential elimination complete.**

**Executive summary.** AuthBridge has required Keycloak admin credentials and client secrets to be reachable from agent pods — a productization blocker on OpenShift, since a single compromised namespace could mean full realm compromise and the long-lived secrets demand manual rotation. This epic drives those credentials out in phases: centralizing client registration in the operator, then replacing admin credentials with SPIFFE-authenticated Admin API access, and finally replacing client secrets with federated SPIFFE (JWT-SVID) client authentication. The outcome is zero long-lived Keycloak credentials in any agent namespace — with the final phase gated on RHBK shipping Keycloak 26.6.x.

## Problem

AuthBridge currently requires Keycloak admin credentials and client secrets to be present in or accessible from agent pods. This is a **blocker for productizing AuthBridge on OpenShift**:

1. **Admin credentials in agent namespaces** — The client-registration sidecar (or the operator) uses Keycloak admin credentials to register OAuth clients. These credentials grant full realm admin access (create/delete users, realms, clients, roles). A compromised agent namespace means full Keycloak compromise.

2. **Client secrets in pods** — Agents use client secrets to authenticate with Keycloak for token exchange. These are long-lived credentials that require manual rotation.

Both issues violate the principle of least privilege and create a large blast radius for any namespace-level compromise.

## Owner

Alan drives this work.

---

### Phase 1: Remove admin credentials from agent namespaces ✅

Move client registration from per-agent sidecar to a centralized operator that runs in `kagenti-system`.

| Item | Status | Reference |
|------|--------|-----------|
| Operator-managed client registration (webhook side) | **MERGED** 2026-03-31 | [kagenti-extensions#262](https://github.com/kagenti/kagenti-extensions/pull/262) |
[...]

PR #2338, opened 2026-08-03
fix: 🐛 Publish operator-spiffe-bootstrap image via CI

### LINKED ISSUE #2333: Operator SPIFFE auth (--enable-operator-spiffe-auth) fails: operator-spiffe-bootstrap image unpublished
## Summary

Following `docs/users-guides/authentication.md` exactly (`scripts/kind/setup-rossoctl.sh --with-spire --enable-operator-spiffe-auth`) on tag `v0.7.0-alpha.6` causes the entire `helm install rossoctl` release to fail, not just the SPIFFE feature.

## Root cause

`charts/rossoctl/templates/operator-client-bootstrap-job.yaml` defaults the bootstrap Job's image to:

```
ghcr.io/rossoctl/rossoctl/operator-spiffe-bootstrap:0.0.0-unreleased
```

The template's own comment says:

> NOTE: operator-spiffe-bootstrap is not published yet and this Job is off by default ... The default below is a placeholder version to satisfy the release pin-check; override via operator-chart.spiffe.operatorAuth.bootstrapImage once the image is built and pushed.

Confirmed via the ghcr.io API that `rossoctl/rossoctl/operator-spiffe-bootstrap` has no published tags (`403 Forbidden` / `UNAUTHORIZED` on anonymous pull).

`scripts/kind/setup-rossoctl.sh`'s `--enable-operator-spiffe-auth` flag only sets `operator-chart.spiffe.operatorAuth.enabled=true` — it never overrides `bootstrapImage`, so any user following the docs literally hits the broken placeholder.

## Impact

Because this is a Helm post-install hook, the Job's `ImagePullBackOff` causes `helm --wait` to time out, and the **entire `rossoctl` release is marked `failed`** — the operator, backend, and UI never get installed either, well beyond just the SPIFFE auth feature.

## Reproduction

```bash
git worktree add .worktrees/repro v0.7.0-alpha.6
cd .worktrees/repro
scripts/kind/setup-rossoctl.sh --with-spire --enable-operator-spiffe-auth
# ...
# Error: failed post-install: resource Job/keycloak/rossoctl-operator-client-bootstrap not ready. status: InProgress, message: Job in progress
```

```bash
kubectl describe pod -n keycloak -l app.kubernetes.io/name=rossoctl-operator-client-bootstrap
# Failed to pull image "ghcr.io/rossoctl/rossoctl/operator-spiffe-bootstrap:0.0.0-unreleased":
# ... 403 Forbidden
```

## Verified fix works once the image exists

I built the image locally from `rossoctl/auth/operator-spiffe-bootstrap/` (source is present in-repo, just never published/pushed), loaded it into the Kind cluster, and overrode `operator-chart.spiffe.operatorAuth.bootstrapImage` to point at it via `--rossoctl-values`. With a real image present, the bootstrap Job completes successfully and the operator authenticates to Keycloak using its JWT-SVID as documented (confirmed via the operator log line `"SPIFFE ID authentication enabled: using JWT-SVID for client registration"` and successful registration of test workloads without touching `keycloak-admin-secret`).

## Suggested fix

Add CI to build and publish `ghcr.io/rossoctl/rossoctl/operator-spiffe-bootstrap` on release (same as other pinned images), and update the release-pin process to set a real tag instead of leaving the `0.0.0-unreleased` placeholder when `operator-chart.spiffe.operatorAuth.enabled` is expected to be usable.

## Environment

- Tag: `v0.7.0-alpha.6`
- `operator-chart` version: `0.3.0-alpha.10`
- Local Kind cluster, macOS/Colima

Alan-Cha:
## Summary

`--enable-operator-spiffe-auth` breaks the entire `helm install rossoctl` release because the bootstrap Job's image was never built or published by CI, even though the Dockerfile and source (`rossoctl/auth/operator-spiffe-bootstrap/`) have been in-repo since the feature was added. The chart fell back to a nonsensical placeholder tag (`ghcr.io/rossoctl/rossoctl/operator-spiffe-bootstrap:0.0.0-unreleased`), which 403s on pull and times out the post-install hook, failing the whole release.

Key changes:
- Add `operator-spiffe-bootstrap` to the `build.yaml` image matrix, matching the sibling `spiffe-idp-setup` entry (same Dockerfile shape, no smoke test).
- Add a dedicated `operatorSpiffeBootstrap: { image, tag }` block to `charts/rossoctl/values.yaml`, matching the plain `image`/`tag` shape every sibling `*OAuthSecret` block already uses (kept separate from `operator-chart.spiffe.operatorAuth`, which only needs to hold the `enabled` flag shared with the operator subchart).
- Register `operatorSpiffeBootstrap.tag` in `scripts/pin-release-tags.sh`'s `PINNABLE_IMAGES` list so future releases pin it automatically, same as every other image (verified via `--dry-run`).
- Remove the stale "not published yet" comment and placeholder fallback from `charts/rossoctl/templates/operator-client-bootstrap-job.yaml`.

The resulting `tag: latest` is expected and non-blocking — confirmed via `scripts/check-release-pins.sh` and `.github/workflows/ci-release-pins.yaml` that the hard release-pin gate only runs on PRs targeting `release/*` branches or manual dispatch, precisely so `tag: latest` is allowed during normal development (same as every other image before its first release pin).

Verified the rendered template produces the correct `image:` string via an isolated `helm template` run.

## Related issue(s)

Fixes #2333

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -54,6 +54,9 @@ jobs:
           - name: spiffe-idp-setup
             context: ./rossoctl
             dockerfile: auth/spiffe-idp-setup/Dockerfile
+          - name: operator-spiffe-bootstrap
+            context: ./rossoctl
+            dockerfile: auth/operator-spiffe-bootstrap/Dockerfile
 
     steps:
       # 1. Checkout the repository code
--- a/charts/rossoctl/templates/operator-client-bootstrap-job.yaml
+++ b/charts/rossoctl/templates/operator-client-bootstrap-job.yaml
@@ -75,12 +75,7 @@ spec:
       restartPolicy: OnFailure
       containers:
       - name: bootstrap
-        # NOTE: operator-spiffe-bootstrap is not published yet and this Job is off
-        # by default (operator-chart.spiffe.operatorAuth.enabled=false). The
-        # default below is a placeholder version to satisfy the release pin-check;
-        # override via operator-chart.spiffe.operatorAuth.bootstrapImage once
-        # the image is built and pushed.
-        image: {{ (index .Values "operator-chart" "spiffe" "operatorAuth" "bootstrapImage") | default "ghcr.io/rossoctl/rossoctl/operator-spiffe-bootstrap:0.0.0-unreleased" }}
+        image: "{{ index .Values "operator-chart" "spiffe" "operatorAuth" "bootstrapImage" }}:{{ index .Values "operator-chart" "spiffe" "operatorAuth" "bootstrapImageTag" }}"
         imagePullPolicy: IfNotPresent
         env:
         - name: KEYCLOAK_URL
--- a/charts/rossoctl/templates/operator-client-bootstrap-job.yaml
+++ b/charts/rossoctl/templates/operator-client-bootstrap-job.yaml
@@ -75,7 +75,7 @@ spec:
       restartPolicy: OnFailure
       containers:
       - name: bootstrap
-        image: "{{ index .Values "operator-chart" "spiffe" "operatorAuth" "bootstrapImage" }}:{{ index .Values "operator-chart" "spiffe" "operatorAuth" "bootstrapImageTag" }}"
+        image: "{{ .Values.operatorSpiffeBootstrap.image }}:{{ .Values.operatorSpiffeBootstrap.tag }}"
         imagePullPolicy: IfNotPresent
         env:
         - name: KEYCLOAK_URL
--- a/charts/rossoctl/values.yaml
+++ b/charts/rossoctl/values.yaml
@@ -274,6 +274,16 @@ apiOAuthSecret:
   useServiceAccountCA: false
 
 # ------------------------------------------------------------------
+#  Operator SPIFFE Authentication Bootstrap Job Configuration
+#  Used by charts/rossoctl/templates/operator-client-bootstrap-job.yaml when
+#  operator-chart.spiffe.operatorAuth.enabled=true. Built by this repo's
+#  build.yaml CI workflow — kept in sync via scripts/pin-release-tags.sh.
+# ------------------------------------------------------------------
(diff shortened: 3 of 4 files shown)


cwiklik (approved):
Correct root-cause fix for the `operator-spiffe-bootstrap` gap (epic #1426): the image is added to `build.yaml`'s matrix (matching `spiffe-idp-setup`), the `0.0.0-unreleased` placeholder in the bootstrap Job is replaced by a dedicated `operatorSpiffeBootstrap` values block, and — importantly — it's registered in `pin-release-tags.sh`'s `PINNABLE_IMAGES`. That last part is what makes the `tag: latest` default fine: per this repo's own `ci-release-pins.yaml`, `tag: latest` on `main` is intentional (dev), and the hard pin-gate + `pin-release-tags.sh` stamp a concrete version on `release/*` branches, so a release can never ship `:latest`. Functionally, `build.yaml` also publishes `:latest` on main pushes, so default installs resolve. CI is green (22 checks incl. Check Release Pins). No `.claude`/`.vscode`; both commits signed.

Two notes for the maintainers (not blocking this review):

1. **Release positioning:** this PR is milestoned v0.7.0 but targets `main`, and `release-0.7` (already at rc.2) still carries the old `0.0.0-unreleased` placeholder. To actually ship in 0.7 it needs to be cherry-picked into `release-0.7` for rc.3/GA after merge. Since the bootstrap Job is off by default, rc.2 is unaffected on the default path — only the opt-in `--enable-operator-spiffe-auth` path was broken.

2. **Pin-gate gap (pre-existing, not this PR):** the placeholder `0.0.0-unreleased` slipped through the release pin-gate because `check-release-pins.sh` only rejects `:latest`/`tag: latest`, not a fake-pinned `0.0.0-unreleased`. Worth teaching the check to also reject non-existent/placeholder version tags so this class can't recur.

LGTM.

Assisted-By: Claude Code

PR #2340, opened 2026-08-03
🐛 Pin operatorSpiffeBootstrap.tag in package-charts CI safety net

### LINKED ISSUE #2339: operatorSpiffeBootstrap.tag has no release-time safety net in package-charts
## Summary

`.github/workflows/build.yaml`'s `package-charts` job force-corrects six image tag fields directly in the CI runner's ephemeral checkout, right before `helm package`/`helm push`, regardless of what's actually committed in the git tag:

```bash
yq e ".ui.frontend.tag = \"${REF_NAME}\"" -i values.yaml
yq e ".ui.backend.tag = \"${REF_NAME}\"" -i values.yaml
yq e ".uiOAuthSecret.tag = \"${REF_NAME}\"" -i values.yaml
yq e ".agentOAuthSecret.tag = \"${REF_NAME}\"" -i values.yaml
yq e ".apiOAuthSecret.tag = \"${REF_NAME}\"" -i values.yaml
yq e ".mlflowOAuthSecret.tag = \"${REF_NAME}\"" -i values.yaml
```

This is a safety net independent of `scripts/pin-release-tags.sh`: even if that manual release-prep step was skipped or done incorrectly before tagging, the **packaged OCI chart artifact** pushed to `oci://ghcr.io/rossoctl/rossoctl` still ends up with the correct tag for these six fields (the git tag itself would still be wrong, but the published chart self-corrects).

`operatorSpiffeBootstrap.tag` (added in #2338, alongside the fix that makes CI actually build `operator-spiffe-bootstrap`) was added to `scripts/pin-release-tags.sh`'s `PINNABLE_IMAGES` list, but **not** to this `package-charts` `yq` list. So it currently has no automatic protection at all: if `pin-release-tags.sh` is skipped before a tag is pushed, `operatorSpiffeBootstrap.tag` ships wrong in *both* the git tag and the published OCI chart, unlike its six siblings which at least self-correct in the published chart.

## Impact

Low likelihood (requires skipping a documented release-prep step), but if it happens, the failure mode is exactly the one #2333/#2334 already demonstrated: a plausible-looking tag pointing at a nonexistent or wrong image, breaking operator SPIFFE auth silently until someone notices.

## Suggested fix

Add to both `yq` blocks in the `package-charts` job of `.github/workflows/build.yaml`:

```bash
yq e ".operatorSpiffeBootstrap.tag = \"${REF_NAME}\"" -i values.yaml
```

(Only needs adding to the `rossoctl` chart's packaging step, not `rossoctl-deps` — same as its siblings.)

## Related

- #2333, #2334 (root cause bugs this field was added to fix)
- #2338 (added `operatorSpiffeBootstrap.tag` to `pin-release-tags.sh` but not this CI safety net)

Alan-Cha:
## Summary

**Stacked on #2338** — branched from that PR's branch since it depends on the `operatorSpiffeBootstrap` values field #2338 introduces. The diff below will include #2338's changes until that merges; only the last commit is new here. Please merge #2338 first.

`package-charts` in `.github/workflows/build.yaml` force-corrects six image tag fields directly in the CI runner's ephemeral checkout, right before `helm package`/`helm push` — independent of whether `scripts/pin-release-tags.sh` was actually run before tagging. This guarantees the **published OCI chart** artifact is correct for those six fields even if that manual release-prep step was skipped (the git tag itself would still be wrong, but the published chart self-corrects).

`operatorSpiffeBootstrap.tag` (added in #2338) was registered in `pin-release-tags.sh`'s `PINNABLE_IMAGES` list, but not in this `package-charts` safety net — leaving it with zero automatic protection. If the release-prep step were ever skipped, this field would ship wrong in *both* the git tag and the published OCI chart, unlike its six siblings.

Key changes (last commit only — rest is #2338):
- Add `yq e ".operatorSpiffeBootstrap.tag = \"${REF_NAME}\"" -i values.yaml` to the `rossoctl` chart's packaging step in `package-charts`, matching its five siblings exactly.

Verified the exact `yq` expression against a copy of the current `values.yaml` — correctly sets `operatorSpiffeBootstrap.tag` without disturbing the sibling `image:` field.

## Related issue(s)

Fixes #2339

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -180,6 +180,7 @@ jobs:
           yq e ".agentOAuthSecret.tag = \"${REF_NAME}\"" -i values.yaml
           yq e ".apiOAuthSecret.tag = \"${REF_NAME}\"" -i values.yaml
           yq e ".mlflowOAuthSecret.tag = \"${REF_NAME}\"" -i values.yaml
+          yq e ".operatorSpiffeBootstrap.tag = \"${REF_NAME}\"" -i values.yaml
           helm package . --destination . --version ${chartVersion} --app-version ${chartVersion}
           helm push "./${chartPackageName}" oci://${{ env.REGISTRY }}/${{ env.REPO }}
 
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -54,6 +54,9 @@ jobs:
           - name: spiffe-idp-setup
             context: ./rossoctl
             dockerfile: auth/spiffe-idp-setup/Dockerfile
+          - name: operator-spiffe-bootstrap
+            context: ./rossoctl
+            dockerfile: auth/operator-spiffe-bootstrap/Dockerfile
 
     steps:
       # 1. Checkout the repository code
--- a/charts/rossoctl/templates/operator-client-bootstrap-job.yaml
+++ b/charts/rossoctl/templates/operator-client-bootstrap-job.yaml
@@ -75,12 +75,7 @@ spec:
       restartPolicy: OnFailure
       containers:
       - name: bootstrap
-        # NOTE: operator-spiffe-bootstrap is not published yet and this Job is off
-        # by default (operator-chart.spiffe.operatorAuth.enabled=false). The
-        # default below is a placeholder version to satisfy the release pin-check;
-        # override via operator-chart.spiffe.operatorAuth.bootstrapImage once
-        # the image is built and pushed.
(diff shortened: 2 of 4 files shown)


Alan-Cha:
Closing — this was a narrow addition (mirroring the existing package-charts CI safety net for the five sibling *OAuthSecret fields onto operatorSpiffeBootstrap.tag), not a broader versioning change. Not needed: the normal release process (scripts/pin-release-tags.sh) will move this off `latest` to a tagged version when the next release is cut, same as every other image before its first pin.

PR #2343, opened 2026-08-05
feat(ui): add "Open Data Governance" observability card

too-far-away:
## Summary

Adds an **"Open Data Governance"** card to the Observability Dashboard, mirroring the existing "Open Kiali" card: always shown, opens the data-governance UI in a new browser tab.

The card links to the data-governance UI (a separate, optional extension), and its description notes it is an **early alpha version**.

## Changes

- **Backend** — `GET /config/dashboards` now returns a `dataGovernance` URL, sourced from the `DATA_GOVERNANCE_DASHBOARD_URL` env var (ConfigMap-driven), falling back to `http://dg.<domain>:8080` when unset — the same pattern as Kiali's `network` URL. (`config.py`, `responses.py`, `routers/config.py`)
- **Frontend** — new `<DashboardCard>` in `ObservabilityPage.tsx` ("Open Data Governance", `SecurityIcon`), `dataGovernance` added to the `DashboardConfig` type. (`ObservabilityPage.tsx`, `services/api.ts`)
- **Helm** — `DATA_GOVERNANCE_DASHBOARD_URL` added to the UI ConfigMap and backend env wiring in `charts/rossoctl/templates/ui.yaml`.
- **Dockerfile fix** — `chmod 0644` the nginx `default.conf`. `COPY` preserves the builder's umask; under podman that yields root-only `0640`, and since the container runs as non-root UID 1001, nginx crashloops with `[emerg] Permission denied`. BuildKit/docker happens to copy `0644`, which masked this. The explicit chmod makes the image builder-independent.

## Tests

- `backend/tests/test_dashboard_config.py`: added `dataGovernance` coverage — fallback (`http://dg.<domain>:8080`) and explicit-URL cases; existing tests updated. All pass (`6 passed`).
- Frontend type-checks via the image build (`tsc -b` in the Dockerfile builder stage).

## Verification

Built + deployed to a local Kind cluster and confirmed:
- `/config/dashboards` → `dataGovernance = http://dg.localtest.me:8080`
- The served UI bundle contains both `Open Data Governance` and `Early alpha version`.

## Notes

- Data governance is deployed **independently** of the platform chart; this PR only adds the UI link + config plumbing, not the component itself. The card is always shown (matching Kiali).

---
Assisted-By: Claude Code


rossoctl/backend/app/models/responses.py
```diff
@@ -73,6 +73,7 @@ class DashboardConfigResponse(BaseModel):
     traces: str
     network: str
     mlflow: str
+    dataGovernance: str
```
abigailgold, line 76: Is this purposefully non-optional?

clawgenti (commented):
Adds a Data Governance card to the Observability Dashboard with full-stack plumbing (Helm ConfigMap, backend config/router/model, frontend card + TypeScript type) and the nginx `chmod` fix for podman/non-root builders. All three commits are signed off, DCO passes, and the new fallback URL pattern mirrors the existing Kiali pattern consistently.

---

abigailgold:
Additional findings:
1. Suggestion: The OpenShift route-discovery job (ui-routes-job.yaml) wasn't updated with a data-governance entry, so on OpenShift the feature always falls back to the synthesized URL rather than discovering a real Route — an asymmetry from the Kiali pattern it claims to mirror, worth confirming with the author.
2. Nit: The new card was omitted from the page's "Note" alert tool-list array, despite being always shown like Kiali/Phoenix/MLflow which are all included there.

clawgenti (commented):
Adds the Data Governance card to the Observability Dashboard with consistent full-stack plumbing (Helm ConfigMap, backend config/router/model, frontend card + TypeScript type) and the nginx `chmod` fix for podman/non-root builders — all four commits signed off, DCO passes.

---

too-far-away:
Thanks for the review, @abigailgold. Went through all three against the current head (`052e576a`):

**Inline (`responses.py` — "Is `dataGovernance` purposefully non-optional?")** — Yes, intentional. The bare-`str` fields are the ones with a guaranteed value (`traces`, `network`, `mlflow`, `keycloakConsole`, `domainName`); `Optional[str] = None` is reserved for genuinely-absent ones (`traceAnalysis`, `mcpInspector`, `mcpProxy`). `get_dashboard_config` always yields a value for data-governance via `settings.data_governance_dashboard_url or f"http://dg.{domain}:8080"` — exactly mirroring how `network` handles Kiali — so non-optional is the consistent choice.

**Nit (card missing from the "Note" tool-list array)** — Already present on the current head: `ObservabilityPage.tsx` includes `dataGovernanceUrl && 'Data Governance for interaction traces'` in that array, alongside Phoenix/Kiali/MLflow. Likely read from an earlier commit.

**Suggestion (OpenShift `ui-routes-job.yaml` has no data-governance entry)** — Accurate observation, but correct as-is. The route-discovery job only knows about routes for components this chart installs (phoenix/mlflow/kiali/keycloak/mcp-inspector, each gated on `.Values.components.*.enabled`). Data-governance is a **separate, out-of-chart sibling extension** — the rossoctl chart neither creates nor can gate on a data-governance `Route`, so the job correctly omits it, and OpenShift uses the synthesized `http://dg.{domain}:8080` fallback. Auto-discovering a data-governance route would be a deliberate follow-up (assumes such a Route exists + a values gate), not part of this PR.

Net: findings 2 and 3 need no change; finding 1 is a valid design observation whose answer is "data-governance is out-of-chart."

abigailgold (approved):

PR #2348, opened 2026-08-05
docs: Link health report (auto-updated)

clawgenti:
## Summary

Auto-updated by Kagenti Link Health Scanner. This PR is continuously updated with each scan. Merge when convenient.

| Metric | Value |
|--------|-------|
| Repos scanned | 8 |
| Broken (internal) | 7 |
| Broken (external) | 12 |
| New since last scan | +19 |
| Fixed since last scan | -0 |

## Related issue(s)

- kagenti/kagenti#1178

### DIFF
--- a/docs/link-health.md
+++ b/docs/link-health.md
@@ -1,22 +1,23 @@
 # Link Health Report
 
-> Last scan: 2026-07-29 07:06 ET | Scan ID: 2026-07-29-001
+> Last scan: 2026-07-31 07:01 ET | Scan ID: 2026-07-31-001
 
 ## Summary
 
 | Metric | Value |
 |--------|-------|
 | Repos scanned | 8 |
-| Total links checked | 1802 |
-| Broken (internal) | 24 |
+| Total links checked | 1836 |
+| Broken (internal) | 14 |
 | Broken (external) | 12 |
-| New since last scan | +36 |
-| Fixed since last scan | -735 |
+| New since last scan | +2 |
+| Fixed since last scan | -12 |
 
 ## Trend (last 10 scans)
 
 | Date | Internal | External | Delta |
 |------|----------|----------|-------|
+| 07-31 | 14 | 12 | -10 |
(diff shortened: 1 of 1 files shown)


docs/link-health.md
```diff
@@ -1,58 +1,42 @@
----
-draft: true       # excluded from https://www.rossoctl.dev/
----
-
 # Link Health Report
```
cwiklik, line 1: must-fix: `main`'s version of this file starts with frontmatter —

```yaml
---
draft: true       # excluded from https://www.rossoctl.dev/
---
```

This scan dropped it. Per its own comment, that `draft: true` is what excludes this internal broken-links report from the public docs site; removing it publishes the dashboard. Almost certainly a scanner-template regression — the generator should keep emitting the frontmatter. Please confirm intent before merge.

cwiklik (changes requested):
### Summary

The data churn in this auto-generated Link Health Report is fine (table/trend updates only). **But this scan also silently removed the YAML frontmatter that `main` carries:**

```yaml
---
draft: true       # excluded from https://www.rossoctl.dev/
---
```

The frontmatter's own comment documents its purpose — it keeps this internal broken-links dashboard **off** the public docs site (`https://www.rossoctl.dev/`). Dropping it would publish the report (repo names + hundreds of broken URLs) publicly, contrary to that documented intent. This looks like a regression in the scanner's template rather than a deliberate decision — the PR body is the generic auto-update template and doesn't mention it.

**Ask (for the scanner's maintainer — the bot force-updates this PR each scan, so this won't self-resolve):** restore `draft: true` frontmatter emission in the generator template. If publishing the report publicly is now intentional, disregard and say so in the PR.

Everything else is clean: CI green (22), DCO signed on all 3 commits, no `.claude/`/`.vscode/` changes.

**Author**: clawgenti (CONTRIBUTOR — returning external / bot)
**CI**: passing.

Issue #2342, opened 2026-08-04
token-exchange E2E (community KC): invalid_client + SPIFFE port-forward timeout after injection is fixed


cwiklik:
## Context

The `release-validation` **community-KC token-exchange** leg (the `… — fatal` step) has been red since ~2026-07-16 (last green 2026-06-30), spanning the org rename. It accumulated multiple independent bugs.

**#2341 fixes the first one** (client-registration was pointed at the public Keycloak URL, unreachable from the operator pod → no credential secret → wedged rollout → sidecar-injection tests saw a 1-container pod). With that fix validated (run 30916392962), injection now works — pods reach `2/2` Ready with the `authbridge-proxy` sidecar, and the 3 sidecar/credential tests pass. This unmasked two remaining problems tracked here.

### Bug A — `invalid_client` on the client-credentials grant (primary)
Keycloak rejects the workload's client-credentials grant:
```
{"error":"invalid_client","error_description":"Invalid client or Invalid client credentials"}
```
Failing directly: `TestClientCredentials::test_agent_client_credentials`, `test_tool_client_credentials`, `TestTokenExchange::test_user_to_agent_exchange`, `TestAuthbridgeExtProc::test_outbound_token_is_different`, `test_user_identity_preserved_through_exchange`. Everything else that needs an exchanged token then cascades to `assert 401 == 200` (7 more tests).

**Lead:** the credential the pod presents no longer matches the Keycloak client record — a secret-lifecycle/ordering race. The operator (re-)registers/rotates the `rossoctl-keycloak-client-credentials-*` secret asynchronously (registrations observed *after* the test run began), so the value mounted in the already-running pod is stale relative to Keycloak. Needs a decision on rotation semantics (restart pods on secret change) and/or the test setup waiting for a stable registration before exercising the grant. Likely rename-era regression.

### Bug B — SPIFFE port-forward read timeout (infra/test-harness)
```
requests.exceptions.ReadTimeout: HTTPConnectionPool(host='localhost', port=8081): Read timed out. (read timeout=10)
```
Failing: `TestSpiffeTokenExchange::test_spiffe_client_credentials`, `test_spiffe_token_exchange`. Separate from Bug A (kubectl port-forward to `svc/keycloak-service` in the harness).

## Refs
- #2341 — the injection/KEYCLOAK_URL fix (bug #1, merged separately)
- #2129 — earlier report of the same leg's symptom (pod not exec-able); root cause there is superseded by #2341 + this issue.

[...]

PR #2351, opened 2026-08-07
fix(e2e): stabilize token-exchange E2E (community KC) — #2342 (test-only)

cwiklik:
## Summary

Test-only stabilization of the community-KC token-exchange E2E leg (the \`… — fatal\` step, red since ~2026-07-16). Per RCA, the root cause is a **credential-rotation race**: the operator re-registers client credentials asynchronously after \`56-enable-spiffe.sh\` rotates them, so a session-scoped fixture that read the secret once cached a stale value → \`invalid_client\` cascading to 12 tests. Plus an independent SPIFFE port-forward read-timeout (2 tests).

This is **Track 1** of a two-track fix, scoped for the RC: **test-only, no product/operator code**.

## Changes (test-only)

- \`rossoctl/tests/e2e/token_exchange/conftest.py\` — \`agent_credentials\` validates creds with a real \`client_credentials\` grant and re-reads/retries on \`invalid_client\`, riding out the async re-registration.
- \`.github/scripts/token-exchange/80-run-tests.sh\` — Keycloak/agent port-forward now waits-for-ready and self-heals (fixes Bug B read-timeout).
- \`.github/scripts/token-exchange/70-deploy-test-workloads.sh\` — credential-convergence gate (present + resourceVersion-stable) replaces the count-only wait (belt-and-suspenders).
- Design + plan under \`docs/superpowers/\`.

## Not in this PR

The operator idempotent client-registration hardening (the root of the re-registration churn) is **Track 2**, deferred to 0.8 as its own PR on \`rossoctl/operator\` — not required to make the leg green.

## Verification

- **Static:** \`bash -n\` + \`shellcheck\` + \`ruff\`/\`ast.parse\` clean; each change peer-reviewed (caught + fixed a keepalive PID leak and a \`pipefail\` abort).
- **Behavioral:** pending this PR's \`release-validation\` Kind run — the real gate.

Refs #2342. Supersedes #2129.

Assisted-By: Claude Code

### DIFF
--- a/.github/scripts/token-exchange/70-deploy-test-workloads.sh
+++ b/.github/scripts/token-exchange/70-deploy-test-workloads.sh
@@ -255,7 +255,7 @@ PREV_RV=""; STABLE=0
 for i in $(seq 1 90); do
   RV=$(kubectl get secrets -n "$TX_NAMESPACE" \
         -o jsonpath='{range .items[*]}{.metadata.name}={.metadata.resourceVersion}{"\n"}{end}' 2>/dev/null \
-        | grep rossoctl-keycloak-client-credentials | sort)
+        | grep rossoctl-keycloak-client-credentials | sort || true)
   CNT=$(printf '%s\n' "$RV" | grep -c . || echo 0)
   if [[ "$CNT" -ge 2 && "$RV" == "$PREV_RV" ]]; then
     STABLE=$((STABLE + 1))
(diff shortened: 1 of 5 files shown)


cwiklik:
Closing — this PR was built on an incorrect root cause.

Corrected RCA: the community-KC token-exchange leg fails because Keycloak rejects the workloads' **SPIFFE / federated-jwt** client authentication (`CLIENT_LOGIN_ERROR ... did not complete for federated-jwt`), not the client-secret rotation race this PR targeted. The operator registered the clients cleanly (no thrashing), Keycloak was healthy, and the SVID was issued — the failure is in the SPIFFE→Keycloak JWT trust path. Notably, the `agent_credentials` fixture change here (validate-by-client-secret-grant + 180s retry) is *counterproductive* under SPIFFE mode: the clients no longer accept secret auth, so the retry can never succeed and burns a ~3-minute stall that then cascades admin-token 401s.

Confirmed gap (real, but not the whole fix): the Keycloak SPIRE-OIDC truststore setup in `55-setup-spiffe-idp.sh` is OCP-only, so on Kind Keycloak has no CA for the OIDC JWKS endpoint. Open questions (need a live-cluster / DEBUG-logged run to answer) captured for follow-up.

This is a deeper SPIFFE↔Keycloak integration issue, not a quick test-only fix — deferring as non-RC-blocking and handing the diagnosis back to #2342 rather than shipping a speculative change.

PR #2353, opened 2026-08-08
fix(ocp): add --preserve-cert-manager flag to cleanup-rossoctl.sh

### LINKED ISSUE #2352: cleanup-rossoctl.sh deletes the cert-manager namespace, silently breaking ingress/API cert renewal on shared clusters
## Summary

`scripts/ocp/cleanup-rossoctl.sh` **Step 7** unconditionally deletes the `cert-manager` and `cert-manager-operator` namespaces:

```bash
_delete_ns cert-manager-operator &
_delete_ns cert-manager          &
wait
```

On clusters where cert-manager manages resources **outside** rossoctl, this deletes cluster-specific secrets that `setup-rossoctl.sh` does **not** recreate — producing a silent, latent failure.

## Impact (observed on our OpenShift/HyperShift `ykt*` clusters)

The cluster ingress wildcard cert (`*.apps.<cluster>`) and the `api.<cluster>` cert are issued by cert-manager via a `letsencrypt-prod` ClusterIssuer using a **Route53 DNS-01** solver. The solver reads AWS creds from a **cluster-specific `route53-credentials` secret** in the `cert-manager` namespace.

Running cleanup deletes that namespace → the solver secret goes with it. `setup-rossoctl.sh --with-all` then reinstalls cert-manager and recreates the `letsencrypt-prod` ClusterIssuer, so everything **looks** healthy (issuer `Ready=True`, certs still valid). But the cluster-specific `route53-credentials` secret is **not** part of the rossoctl install, so it is never restored.

The failure is invisible until the next ~90-day renewal cycle, when:
- the DNS-01 Challenge stalls (`secrets "route53-credentials" not found`),
- the wildcard cert expires,
- the `console` / `authentication` ClusterOperators go `Available=False`,
- **browser Console login breaks** while `oc login` still works (CLI hits the kube-apiserver serving cert, a separate valid cert).

We hit exactly this on `ykt3` and had to recover by copying the secret from a sibling cluster.

## Root cause

Step 7 assumes rossoctl **owns** cert-manager. On any cluster where cert-manager predates rossoctl or serves other consumers (cluster ingress/API certs being the common case), that assumption is wrong and the teardown is destructive beyond rossoctl's own footprint.

## Proposed fix

Add a `--preserve-cert-manager` flag that skips Step 7 (keeps the operator + namespace and any issuer/solver secrets intact). Off by default, so existing behavior is unchanged. PR attached.

Longer term, cleanup should arguably only remove cert-manager if rossoctl installed it (e.g. detect an ownership label/annotation) rather than assuming ownership.

webchang:
## What

Adds a `--preserve-cert-manager` flag to `scripts/ocp/cleanup-rossoctl.sh` that **skips Step 7** — the step that deletes the `cert-manager` and `cert-manager-operator` namespaces. When the flag is passed, the cert-manager operator, namespace, issuers, and any cluster-specific solver secrets are left intact.

Off by default → existing behavior is unchanged.

Fixes #2352.

## Why

Step 7 assumes rossoctl owns cert-manager. On clusters where cert-manager also manages resources **outside** rossoctl — most commonly the ingress/API wildcard certs via a `letsencrypt-prod` ClusterIssuer with a **cluster-specific DNS-01 solver secret** (`route53-credentials`) — deleting the namespace drops that solver secret.

`setup-rossoctl.sh --with-all` reinstalls cert-manager and recreates the ClusterIssuer, so the platform *looks* healthy (issuer `Ready=True`, certs still valid). But the cluster-specific secret is not part of the rossoctl install, so it is never restored. The failure is invisible until the next ~90-day renewal, when the wildcard cert expires, the `console`/`authentication` ClusterOperators degrade, and browser Console login breaks (while `oc login` still works).

We hit this on `ykt3` and recovered by copying the secret from a sibling cluster. This flag lets operators tear down rossoctl on such shared clusters without collateral damage.

## Changes

- New `--preserve-cert-manager` flag (parsed alongside `--yes`/`-y`).
- Step 7 wrapped in `if $PRESERVE_CERT_MANAGER; then ... skip ... else ... existing teardown ... fi`.
- Confirmation summary and completion banner note when cert-manager is preserved.
- Header/usage docs + `--help` updated to explain the flag and its rationale.

## Testing

- `bash -n scripts/ocp/cleanup-rossoctl.sh` → syntax OK.
- `--help` renders the new option.
- Default path (no flag) is byte-identical in behavior to before (Step 7 still runs).

## Notes

A more thorough long-term fix would be to only remove cert-manager when rossoctl installed it (detect an ownership label/annotation) rather than assuming ownership. This flag is the minimal, backward-compatible mitigation.

Assisted-By: Claude Code

### DIFF
--- a/scripts/ocp/cleanup-rossoctl.sh
+++ b/scripts/ocp/cleanup-rossoctl.sh
@@ -8,8 +8,9 @@
 # Uses parallel deletion where possible for faster cleanup.
 #
 # Usage:
-#   ./scripts/ocp/cleanup-rossoctl.sh            # Interactive (prompts for confirmation)
-#   ./scripts/ocp/cleanup-rossoctl.sh --yes      # Skip confirmation prompt
+#   ./scripts/ocp/cleanup-rossoctl.sh                        # Interactive (prompts for confirmation)
+#   ./scripts/ocp/cleanup-rossoctl.sh --yes                  # Skip confirmation prompt
+#   ./scripts/ocp/cleanup-rossoctl.sh --preserve-cert-manager # Keep cert-manager (Step 7)
 #
 # This script:
 #   1. Uninstalls Helm releases (parallel): rossoctl, mcp-gateway, kuadrant-operator, rossoctl-deps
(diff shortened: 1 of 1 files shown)


scripts/ocp/cleanup-rossoctl.sh
```diff
@@ -306,30 +328,41 @@ echo ""
...
+fi
 echo ""
 
 ELAPSED=$(( SECONDS - START_SECONDS ))
 MINS=$(( ELAPSED / 60 ))
 SECS=$(( ELAPSED % 60 ))
 
 echo "============================================"
-echo "  Rossoctl Cleanup Complete  (${MINS}m ${SECS}s)"
+if $PRESERVE_CERT_MANAGER; then
```
cwiklik, line 360: nit: both branches of this `if/else` print the identical `Rossoctl Cleanup Complete (${MINS}m ${SECS}s)` line. Could simplify to print the base line once, then conditionally add the preserved note — e.g. `echo "...Complete..."` followed by `if $PRESERVE_CERT_MANAGER; then echo "  cert-manager PRESERVED (--preserve-cert-manager)"; fi`. Trivial / non-blocking.

cwiklik (approved):
**APPROVE** — reviewed via github:pr-review.

Small, correct, well-scoped OCP-safety fix: `--preserve-cert-manager` gates the Step 7 cert-manager teardown so cleanup on shared clusters no longer drops cert-managers DNS-01 solver secret and silently breaks ingress/API cert renewal (#2352). Verified the flag actually skips the subscription/CSV/namespace deletes and that the pre-cleanup summary + completion banner reflect it. Shell hygiene fine (`set -euo pipefail`, no unquoted-var regressions); DCO signed; all CI green; no agent/IDE config touched. One trivial nit inline.

Assisted-By: Claude Code

PR #2369, opened 2026-08-12
feat: Add PUT /{namespace}/{name}/identity-config endpoint

esnible:
## Summary

- Adds `PUT /{namespace}/{name}/identity-config` as a sibling to the existing `GET` for the same path, under the `rossoctl_feature_flag_authbridge_api` feature flag and gated by the same `ROLE_OPERATOR` permission.
- Accepts the raw request body as the AuthBridge runtime YAML (`Body(media_type="text/plain")` — no JSON envelope required).
- Patches the `config.yaml` key of the `authbridge-config-{name}` ConfigMap in the target namespace via the existing `KubernetesService.upsert_configmap` helper, which creates the ConfigMap if missing or merges the key if it exists.

## Test plan

- [x] Manual: `curl -X PUT -H "Authorization: Bearer <token>" http://rossoctl-ui.localtest.me:8080/api/v1/agents/<namespace>/<name>/identity-config --data-binary "@config.yaml"` and confirm the `authbridge-config-<name>` ConfigMap's `config.yaml` key is updated
- [x] Confirm `GET .../identity-config` still returns the AuthBridge-reported config for the same agent

Assisted-By: Claude Code

### DIFF
--- a/rossoctl/backend/app/routers/agents.py
+++ b/rossoctl/backend/app/routers/agents.py
@@ -4884,6 +4884,9 @@ async def fetch_env_from_url(request: FetchEnvUrlRequest) -> FetchEnvUrlResponse
         raise HTTPException(status_code=500, detail=f"Unexpected error: {str(e)}")
 
 
+_K8S_NAME_RE = re.compile(r"^[a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?$")
+
+
 if settings.rossoctl_feature_flag_authbridge_api:
 
     @router.get(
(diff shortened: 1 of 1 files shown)


rossoctl/backend/app/routers/agents.py
```diff
@@ -4926,6 +4926,30 @@ async def get_agent_identity_config(
...
+        Set the AuthBridge configuration for an Agent.
+        """
+
+        namespace = sanitize_log(namespace)
+        name = sanitize_log(name)
+
+        kube.upsert_configmap(
+            namespace=namespace,
+            name=f"authbridge-config-{name}",
+            data={"config.yaml": new_authbridge_config_yaml},
```
evaline-ju, line 4958: suggestion: The YAML body is stored verbatim without parsing. A `yaml.safe_load()` round-trip before `upsert_configmap` would catch malformed config at write time rather than silently storing it for AuthBridge to fail on later

rossoctl/backend/app/routers/agents.py
```diff
@@ -4926,6 +4926,30 @@ async def get_agent_identity_config(
...
+        name: str,
+        new_authbridge_config_yaml: str = Body(media_type="text/plain"),
+        kube: KubernetesService = Depends(get_kubernetes_service),
+    ) -> dict:
+        """
+        Set the AuthBridge configuration for an Agent.
+        """
+
+        namespace = sanitize_log(namespace)
+        name = sanitize_log(name)
```
evaline-ju: suggestion: `sanitize_log` only prevents log injection and doesn't validate namespace/name as legal K8s names. Bad values will propagate and fail as 500 instead of 400. Consider `_K8S_NAME_RE.fullmatch` on both params

evaline-ju (approved):
lgtm with a couple suggestions

PR #2370, opened 2026-08-13
feat: expose cpu/memory resource requests and limits in agent/tool deployment API

### LINKED ISSUE #2203: feat: expose cpu/memory resource requests and limits in agent/tool deployment API
## Summary

CPU and memory resources for deployed agents and tools are currently hardcoded constants in the backend. There is no way for API callers to specify per-deployment resource requirements.

## Current behavior

`CreateAgentRequest` and `CreateToolRequest` do not expose resource fields. All generated Kubernetes manifests (Deployment, StatefulSet, Job, Sandbox CR) use hardcoded values from `rossoctl/backend/app/core/constants.py`:

```python
DEFAULT_RESOURCE_LIMITS   = {"cpu": "500m",  "memory": "1Gi"}
DEFAULT_RESOURCE_REQUESTS = {"cpu": "100m",  "memory": "256Mi"}
```

The only user-controllable resource field today is `persistentStorage.size` (PVC capacity).

## Desired behavior

API callers should be able to specify resource requests and limits when deploying an agent and tool. 
```json
{
  "name": "my-agent",
  "resources": {
    "requests": { "cpu": "250m", "memory": "512Mi" },
    "limits":   { "cpu": "2",    "memory": "4Gi"   }
  }
}
```

When not specified, the existing constants serve as defaults (no breaking change).

## Proposed implementation

1. Add an optional `ResourceConfig` Pydantic model with `requests` and `limits` fields
2. Add `resources: Optional[ResourceConfig]` to `CreateAgentRequest` (`agents.py:233`) and `CreateToolRequest` (`tools.py`)
3. Wire the value (with fallback to constants) through the four manifest builders:
   - `_build_deployment_manifest`
   - `_build_statefulset_manifest`
   - `_build_job_manifest`
   - `_build_sandbox_manifest`
4. Mirror the field in the Go TUI types (`tui/internal/api/types.go`)

## Affected files

- `rossoctl/backend/app/core/constants.py` — defaults remain, used as fallback
- `rossoctl/backend/app/routers/agents.py` — request model + manifest builders
- `rossoctl/backend/app/routers/tools.py` — request model + manifest builders
- `rossoctl/tui/internal/api/types.go` — Go mirror of request types

esnible:
## Summary
- Add `resources` (requests/limits) to `CreateAgentRequest`/`CreateToolRequest`, overriding platform defaults
- Wire through all agent workload builders (Deployment, StatefulSet, Job, Sandbox) and tool builders (Deployment, StatefulSet)
- Mirror `ResourceConfig` in the Go TUI types

Fixes #2203

## Test plan
- [ ] `uv run pytest rossoctl/tests/e2e/ -v`
- [x] `go build ./...` in `rossoctl/tui`
- [x] Backend routers import cleanly; `resources` field confirmed on both Pydantic models

Assisted-By: Claude Code

### DIFF
--- a/rossoctl/backend/app/models/shipwright.py
+++ b/rossoctl/backend/app/models/shipwright.py
@@ -37,7 +37,7 @@ class ResourceConfig(BaseModel):
     limits: Optional[Dict[str, str]] = None
 
 
-def resolve_resources(resources: Optional["ResourceConfig"]) -> Dict[str, Dict[str, str]]:
+def resolve_resources(resources: Optional[ResourceConfig]) -> Dict[str, Dict[str, str]]:
     """Merge a ResourceConfig override over the platform defaults, per requests/limits."""
     return {
         "limits": (resources.limits if resources and resources.limits else DEFAULT_RESOURCE_LIMITS),
--- a/rossoctl/backend/app/models/shipwright.py
+++ b/rossoctl/backend/app/models/shipwright.py
@@ -13,6 +13,8 @@ from typing import Any, Dict, List, Optional
 from pydantic import BaseModel
 
 from app.core.constants import (
+    DEFAULT_RESOURCE_LIMITS,
+    DEFAULT_RESOURCE_REQUESTS,
     SHIPWRIGHT_DEFAULT_DOCKERFILE,
     SHIPWRIGHT_DEFAULT_TIMEOUT,
 )
@@ -23,6 +25,26 @@ class ResourceType(str, Enum):
 
     AGENT = "agent"
     TOOL = "tool"
+
+
+class ResourceConfig(BaseModel):
+    """Container CPU/memory requests and limits, overriding the platform defaults.
+
+    Shared by agents and tools (source-build and direct-image paths alike).
+    """
+
+    requests: Optional[Dict[str, str]] = None
+    limits: Optional[Dict[str, str]] = None
+
+
+def resolve_resources(resources: Optional["ResourceConfig"]) -> Dict[str, Dict[str, str]]:
+    """Merge a ResourceConfig override over the platform defaults, per requests/limits."""
+    return {
+        "limits": (resources.limits if resources and resources.limits else DEFAULT_RESOURCE_LIMITS),
+        "requests": (
+            resources.requests if resources and resources.requests else DEFAULT_RESOURCE_REQUESTS
+        ),
+    }
 
 
 class ShipwrightBuildConfig(BaseModel):
--- a/rossoctl/backend/app/routers/agents.py
+++ b/rossoctl/backend/app/routers/agents.py
@@ -2612,6 +2612,8 @@ def _build_agent_shipwright_build_manifest(
         resource_config["tlsBridgeEnabled"] = True
     if request.persistentStorage:
         resource_config["persistentStorage"] = request.persistentStorage.model_dump()
+    if request.resources:
+        resource_config["resources"] = request.resources.model_dump(exclude_none=True)
     # Add env vars if present
     if request.envVars:
         resource_config["envVars"] = [ev.model_dump(exclude_none=True) for ev in request.envVars]
@@ -4083,6 +4085,7 @@ class FinalizeShipwrightBuildRequest(BaseModel):
     outboundPortsExclude: Optional[str] = None
     inboundPortsExclude: Optional[str] = None
     defaultOutboundPolicy: Optional[Literal["passthrough", "exchange"]] = None
+    resources: Optional[ResourceConfig] = None
     persistentStorage: Optional[PersistentStorageConfig] = None
     mcpToolName: Optional[str] = None
     llmPreset: Optional[str] = None
@@ -4417,6 +4420,10 @@ async def finalize_shipwright_build(
         if final_persistent_storage is None and stored_config.get("persistentStorage"):
             final_persistent_storage = PersistentStorageConfig(**stored_config["persistentStorage"])
 
+        final_resources = request.resources
+        if final_resources is None and stored_config.get("resources"):
+            final_resources = ResourceConfig(**stored_config["resources"])
+
         final_mcp_tool_name = (
             request.mcpToolName
             if request.mcpToolName is not None
@@ -4454,6 +4461,7 @@ async def finalize_shipwright_build(
             inboundPortsExclude=final_inbound_ports_exclude,
             defaultOutboundPolicy=final_default_outbound_policy,
             persistentStorage=final_persistent_storage,
+            resources=final_resources,
             gitPath=stored_config.get("gitPath")
             or build.get("spec", {}).get("source", {}).get("contextDir", ""),
             mcpToolName=final_mcp_tool_name,
--- a/rossoctl/backend/app/routers/agents.py
+++ b/rossoctl/backend/app/routers/agents.py
@@ -114,6 +114,8 @@ from app.utils.routes import (
 )
 from app.models.shipwright import (
     ResourceType,
+    ResourceConfig,
+    resolve_resources,
     ShipwrightBuildConfig,
     BuildSourceConfig,
     BuildOutputConfig,
@@ -230,13 +232,6 @@ class PersistentStorageConfig(BaseModel):
 
     enabled: bool = False
     size: str = "1Gi"
-
-
-class ResourceConfig(BaseModel):
-    """Container CPU/memory requests and limits, overriding the platform defaults."""
-
-    requests: Optional[Dict[str, str]] = None
-    limits: Optional[Dict[str, str]] = None
 
 
 class CreateAgentRequest(BaseModel):
@@ -3125,17 +3120,6 @@ def _ensure_agentruntime(
             logger.warning("Failed to create AgentRuntime '%s': %s", name, e.reason)
 
 
-def _resolve_resources(request: "CreateAgentRequest") -> Dict[str, Dict[str, str]]:
-    """Merge request.resources over the platform defaults, per requests/limits."""
-    resources = request.resources
-    return {
-        "limits": (resources.limits if resources and resources.limits else DEFAULT_RESOURCE_LIMITS),
-        "requests": (
-            resources.requests if resources and resources.requests else DEFAULT_RESOURCE_REQUESTS
-        ),
-    }
-
-
 def _build_deployment_manifest(
     request: "CreateAgentRequest",
     image: str,
@@ -3211,7 +3195,7 @@ def _build_deployment_manifest(
                             "name": "agent",
                             "image": image,
                             "imagePullPolicy": DEFAULT_IMAGE_POLICY,
-                            "resources": _resolve_resources(request),
+                            "resources": resolve_resources(request.resources),
                             "env": env_vars,
                             "ports": [
                                 {
@@ -3411,7 +3395,7 @@ def _build_statefulset_manifest(
                             "name": "agent",
                             "image": image,
                             "imagePullPolicy": DEFAULT_IMAGE_POLICY,
-                            "resources": _resolve_resources(request),
+                            "resources": resolve_resources(request.resources),
                             "env": env_vars,
                             "ports": [
                                 {
@@ -3552,7 +3536,7 @@ def _build_job_manifest(
                             "name": "agent",
                             "image": image,
                             "imagePullPolicy": DEFAULT_IMAGE_POLICY,
-                            "resources": _resolve_resources(request),
+                            "resources": resolve_resources(request.resources),
                             "env": env_vars,
                             "ports": [
                                 {
@@ -3658,7 +3642,7 @@ def _build_sandbox_manifest(
                             "name": "agent",
                             "image": image,
                             "imagePullPolicy": DEFAULT_IMAGE_POLICY,
-                            "resources": _resolve_resources(request),
+                            "resources": resolve_resources(request.resources),
                             "env": env_vars,
                             "ports": [
                                 {
--- a/rossoctl/backend/app/routers/agents.py
+++ b/rossoctl/backend/app/routers/agents.py
@@ -232,6 +232,13 @@ class PersistentStorageConfig(BaseModel):
     size: str = "1Gi"
 
 
+class ResourceConfig(BaseModel):
+    """Container CPU/memory requests and limits, overriding the platform defaults."""
+
+    requests: Optional[Dict[str, str]] = None
+    limits: Optional[Dict[str, str]] = None
+
+
 class CreateAgentRequest(BaseModel):
     """Request to create a new agent."""
 
@@ -318,6 +325,9 @@ class CreateAgentRequest(BaseModel):
 
     # Persistent storage (for Sandbox and StatefulSet workloads)
     persistentStorage: Optional[PersistentStorageConfig] = None
+
+    # CPU/memory requests and limits, overriding DEFAULT_RESOURCE_REQUESTS/LIMITS
+    resources: Optional[ResourceConfig] = None
 
     # Shipwright build configuration
     shipwrightConfig: Optional[ShipwrightBuildConfig] = None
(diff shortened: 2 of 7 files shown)


clawgenti (commented):
Solid implementation of per-agent/tool resource override — the plumbing is consistent and the fallback-to-defaults logic is correct. Two findings worth addressing before merge.

**Findings:**
- `FinalizeToolBuildRequest` does not include a `resources` field; if the Shipwright finalize path creates the final Deployment/StatefulSet, resource overrides will silently revert to platform defaults at finalize time.
- No new test cases exercise the resource override path — the `_make_request` helper is updated in the two test files but no test verifies that a custom `ResourceConfig` actually propagates into the generated manifest.

---

rossoctl/backend/app/routers/tools.py
```diff
@@ -253,6 +260,9 @@ class CreateToolRequest(BaseModel):
     # Outbound routing rules (authproxy-routes ConfigMap)
     outboundRoutes: Optional[List["OutboundRoute"]] = None
 
+    # CPU/memory requests and limits, overriding DEFAULT_RESOURCE_REQUESTS/LIMITS
+    resources: Optional[ResourceConfig] = None
+
 
 class FinalizeToolBuildRequest(BaseModel):
```
clawgenti, line 260: `FinalizeToolBuildRequest` (line ~267) doesn't include a `resources` field. If the finalize endpoint creates the Deployment or StatefulSet (rather than `create_tool`), resource overrides from the original `CreateToolRequest` won't be available and will silently fall back to platform defaults. Either propagate `resources` here or confirm the finalize path always defers to the already-created workload.

rossoctl/backend/tests/test_sandbox_manifest.py
```diff
@@ -32,6 +32,7 @@ def _make_request(**overrides):
     req.defaultOutboundPolicy = overrides.get("defaultOutboundPolicy", None)
     req.persistentStorage = overrides.get("persistentStorage", None)
     req.skills = overrides.get("skills", None)
+    req.resources = overrides.get("resources", None)
```
clawgenti, line 35: `req.resources` is wired into `_make_request` but no test case exercises it. Consider adding a test like `test_sandbox_manifest_custom_resources` that passes a `ResourceConfig` override and asserts the generated manifest contains the expected `requests`/`limits` values.

rossoctl/backend/app/routers/agents.py
```diff
@@ -232,6 +232,13 @@ class PersistentStorageConfig(BaseModel):
     size: str = "1Gi"
 
 
+class ResourceConfig(BaseModel):
```
clawgenti: `ResourceConfig` is defined again in `tools.py` (line 191) with identical fields. Extract to a shared module to avoid schema drift.

rossoctl/backend/app/routers/agents.py
```diff
@@ -3113,6 +3125,17 @@ def _ensure_agentruntime(
             logger.warning("Failed to create AgentRuntime '%s': %s", name, e.reason)
 
 
+def _resolve_resources(request: "CreateAgentRequest") -> Dict[str, Dict[str, str]]:
```
clawgenti: `_resolve_resources` and `_resolve_tool_resources` in `tools.py` are near-identical. A shared helper would remove the duplication.

clawgenti (commented):
This PR cleanly adds per-agent/tool resource overrides via a new `ResourceConfig` model, wiring it through all relevant workload builders with a simple merge-over-defaults pattern.

- `ResourceConfig` is defined identically in both `agents.py` and `tools.py` — consider extracting it to a shared `models.py` or `common.py` to avoid drift if the schema evolves.
- `_resolve_resources` / `_resolve_tool_resources` are also near-identical — could be unified as a single free function in shared code.
- The `UpdateAgentRequest` path correctly falls back to stored config when `final_resources is None`, consistent with how `persistentStorage` is handled.

---

clawgenti (commented):
Adds `ResourceConfig` + `resolve_resources` with clean fallback-to-defaults semantics, wired consistently through agents and tools routers with round-trip persistence via Build annotations. One import gap in the new test class and a minor forward-ref style nit.

- **Missing imports in `test_shipwright.py`**: `TestResourceOverridePropagation` calls `_build_deployment_manifest`, `_build_statefulset_manifest`, and `_build_agent_shipwright_build_manifest` (lines 1076, 1091, 1108, 1125), but none of these are present in the `from app.routers.agents import (...)` block — the tests will fail with `NameError` at runtime.
- **Unnecessary forward reference in `tools.py`**: `resources: Optional["ResourceConfig"]` at lines 1224 and 1378 uses a string annotation, but `ResourceConfig` is already imported at the top of the file (line 68), so the quotes are redundant.

---

rossoctl/backend/tests/test_shipwright.py
```diff
@@ -14,6 +14,7 @@
 import json
 import pytest
 
+from app.models.shipwright import ResourceConfig
 from app.routers.agents import (
```
clawgenti, line 18: **Missing imports**: `TestResourceOverridePropagation` calls `_build_deployment_manifest`, `_build_statefulset_manifest`, and `_build_agent_shipwright_build_manifest` (used at lines 1076, 1091, 1108, 1125), but they are not imported here. All three must be added to this import block, e.g.:
```python
from app.routers.agents import (
    CreateAgentRequest,
    ShipwrightBuildConfig,
    _build_agent_shipwright_build_manifest,
    _build_deployment_manifest,
    _build_statefulset_manifest,
)
```
Without this the entire new test class will raise `NameError` on collection.

rossoctl/backend/app/routers/tools.py
```diff
@@ -1215,6 +1221,7 @@ def _build_tool_deployment_manifest(
     outbound_ports_exclude: Optional[str] = None,
     inbound_ports_exclude: Optional[str] = None,
     auth_bridge_mode: Optional[str] = None,
+    resources: Optional["ResourceConfig"] = None,
```
clawgenti: `ResourceConfig` is already imported at the top of the file (line 68), so the forward-reference string `"ResourceConfig"` is unnecessary here. Use `Optional[ResourceConfig]` directly.

rossoctl/backend/app/routers/tools.py
```diff
@@ -1371,6 +1375,7 @@ def _build_tool_statefulset_manifest(
...
```
clawgenti: Same as line 1224 — `ResourceConfig` is already in scope; drop the quotes: `Optional[ResourceConfig]`.

clawgenti (commented):
All three prior rounds of feedback have been addressed: `ResourceConfig`/`resolve_resources` unified into `shipwright.py`, the Shipwright finalize path now propagates resources via Build annotations, and `TestResourceOverridePropagation` covers custom/default resources for all four workload types (agent deployment, agent statefulset, agent shipwright, tool deployment, tool statefulset, tool shipwright). All CI checks pass.

Two minor nits below (neither blocking).

---

rossoctl/backend/app/models/shipwright.py
```diff
@@ -25,6 +27,26 @@ class ResourceType(str, Enum):
...
+    """Container CPU/memory requests and limits, overriding the platform defaults.
+
+    Shared by agents and tools (source-build and direct-image paths alike).
+    """
+
+    requests: Optional[Dict[str, str]] = None
+    limits: Optional[Dict[str, str]] = None
+
+
+def resolve_resources(resources: Optional["ResourceConfig"]) -> Dict[str, Dict[str, str]]:
```
clawgenti: **nit**: `ResourceConfig` is defined just above this function in the same file, so `Optional["ResourceConfig"]` doesn't need the quotes — a direct annotation works fine here (unlike a forward reference where the class is defined later). Not a bug, just a minor inconsistency.

rossoctl/tui/internal/api/types.go
```diff
@@ -84,12 +84,12 @@ type DeleteResponse struct {
 
 // DashboardConfigResponse holds dashboard URLs.
 type DashboardConfigResponse struct {
-	Traces         string `json:"traces"`
-	Network        string `json:"network"`
-	MCPInspector   string `json:"mcpInspector"`
-	MCPProxy       string `json:"mcpProxy"`
+	Traces          string `json:"traces"`
```
clawgenti, line 87: **nit**: The `DashboardConfigResponse` field alignment has been changed from single-space to multi-space padding here and through line 92. This is cosmetic-only (no functional change) but will dirty `git blame` for those lines with an unrelated formatting commit. If this was intentional to match a project-wide style pass, that's fine — otherwise consider reverting the whitespace changes to keep blame clean.

esnible: Not something to revert. Adding the code elsewhere triggered a gofmt pass, and gofmt -l/gofmt -d confirm the file is currently canonical. DashboardConfigResponse's realignment is gofmt's own required output, not a stray manual edit; reverting it would make the file fail formatting.

clawgenti (commented):
Adds `ResourceConfig` + `resolve_resources` to centralize container CPU/memory overrides across agents and tools, replacing duplicated inline default dicts at every manifest-build site. The two-phase build annotation persistence (store on create, restore on finalize) is correctly handled for both agent and tool paths.

---

rossoctl/backend/tests/test_shipwright.py
```diff
@@ -1050,3 +1054,132 @@ def test_tool_statefulset_has_spire_label_when_enabled(self):
...
+            protocol="a2a",
+            framework="LangGraph",
+            deploymentMethod="image",
+            containerImage="registry.example.com/test-agent:v1",
+            resources=ResourceConfig(
+                requests={"cpu": "250m", "memory": "512Mi"},
+                limits={"cpu": "1", "memory": "2Gi"},
+            ),
+        )
+        manifest = _build_deployment_manifest(request, image="registry.example.com/test-agent:v1")
```
clawgenti, line 1076: `_build_deployment_manifest` is not imported — only `_build_tool_deployment_manifest` etc. are imported from `app.routers.tools`. The agent-side helpers (`_build_deployment_manifest`, `_build_statefulset_manifest`, `_build_agent_shipwright_build_manifest`) need to be added to the `from app.routers.agents import (...)` block, otherwise these tests will fail with `NameError` at collection time.

rossoctl/tui/internal/api/types.go
```diff
@@ -191,26 +191,32 @@ type PersistentStorageConfig struct {
...
+	GitPath           string        `json:"gitPath,omitempty"`
+	GitBranch         string        `json:"gitBranch,omitempty"`
+	ImageTag          string        `json:"imageTag,omitempty"`
+	ContainerImage    string        `json:"containerImage,omitempty"`
+	ImagePullSecret   string        `json:"imagePullSecret,omitempty"`
+	ServicePorts      []ServicePort `json:"servicePorts,omitempty"`
+	CreateHTTPRoute   bool          `json:"createHttpRoute"`
+	AuthBridgeEnabled bool          `json:"authBridgeEnabled"`
+	SpireEnabled      bool          `json:"spireEnabled"`
+	AuthBridgeMode    string        `json:"authBridgeMode,omitempty"`
```
clawgenti, line 219: `CreateAgentRequest` (and presumably `CreateToolRequest`) does not include a `Resources *ResourceConfig` field here, so the TUI has no way to pass resource overrides to the API. If this is intentional (API-first rollout), a comment like `// TODO: wire ResourceConfig once TUI resource-picker form is ready` or a follow-up issue would make that clear.

clawgenti (commented):
Adds `ResourceConfig` + `resolve_resources()` and wires per-workload CPU/memory overrides through all agent and tool manifest builders cleanly. Two items worth addressing before merge.

**Findings:**

- Missing imports in `test_shipwright.py`: `_build_deployment_manifest`, `_build_statefulset_manifest`, and `_build_agent_shipwright_build_manifest` are called in the new `TestResourceOverridePropagation` tests but are not imported from `app.routers.agents`. The test file currently only imports those names from `app.routers.tools`.
- `types.go` does not add `Resources *ResourceConfig` to `CreateAgentRequest` / `CreateToolRequest`, so the TUI cannot pass resource overrides yet. If this is intentional (API-first rollout), a comment or follow-up issue would clarify intent.

---

rossoctl/backend/app/models/shipwright.py
```diff
@@ -13,6 +13,8 @@
 from pydantic import BaseModel
 
 from app.core.constants import (
+    DEFAULT_RESOURCE_LIMITS,
```
rubambiza, line 16: I presume these constants were changed in a previous PR/commit because I see changes only to `app.models` and `app.routers.`

rossoctl/backend/app/models/shipwright.py
```diff
@@ -25,6 +27,26 @@ class ResourceType(str, Enum):
     TOOL = "tool"
 
 
+class ResourceConfig(BaseModel):
```
rubambiza, line 30: I am assuming the parent `BaseModel` class is a [Python data class](https://docs.python.org/3/library/dataclasses.html). Please correct me if I'm wrong.

rossoctl/backend/app/models/shipwright.py
```diff
@@ -25,6 +27,26 @@ class ResourceType(str, Enum):
...
+    """Container CPU/memory requests and limits, overriding the platform defaults.
+
+    Shared by agents and tools (source-build and direct-image paths alike).
+    """
+
+    requests: Optional[Dict[str, str]] = None
+    limits: Optional[Dict[str, str]] = None
+
+
+def resolve_resources(resources: Optional[ResourceConfig]) -> Dict[str, Dict[str, str]]:
```
rubambiza, line 40: This method definition is telling me `ResourceConfig` might not be a data class as [asked above](https://github.com/rossoctl/rossoctl/pull/2370/changes#r3785253842). If so, then the argument becomes stronger for making it one because it seems like this can be a method that gets resolved internally in the class either by passing in custom limits or letting the class's `__init__` method call `resolve_resources` internally and set the defaults during object creation.

I have not read the rest of the files yet, so my comment above may be off base depending on how/where the method is invoked.

rossoctl/backend/app/routers/agents.py
```diff
@@ -4074,6 +4069,7 @@ class FinalizeShipwrightBuildRequest(BaseModel):
     outboundPortsExclude: Optional[str] = None
     inboundPortsExclude: Optional[str] = None
     defaultOutboundPolicy: Optional[Literal["passthrough", "exchange"]] = None
+    resources: Optional[ResourceConfig] = None
```
rubambiza, line 4072: Random thought 🤔 Is there a reason we have a file with 4000+ LOCs? 😱 Surely there must be some components to all of this that can be pulled out into their own modules.

If AI was involved in some of the maintenance, I can see why because the agent is totally okay with continuing to add to the same file. That is why I recently had to [breakdown an automation file](https://github.com/rossoctl/automation/pull/50) before it got too overwhelming. Just a suggestion here.

rossoctl/backend/tests/test_sandbox_manifest.py
rubambiza: @esnible was this already resolved?

rossoctl/backend/tests/test_shipwright.py
rubambiza: @esnible Same here.

rossoctl/backend/tests/test_shipwright.py
rubambiza: Has this been resolved?

rossoctl/tui/internal/api/types.go
rubambiza: Ignore hallucinated comment.

rubambiza (changes requested):
Some clarifying questions with a recommended suggestion for the `ResourceConfig` class structure in Python.

The comment on refactoring can be addressed in a separate issue/PR.

rossoctl/backend/app/models/shipwright.py
esnible: `BaseModel` is from Pydantic, not a dataclasses.  Thy are similar.  _shipright.py_ already had `from pydantic import BaseModel` and I am afraid to touch it as part of this small PR.

esnible:
Closing in favor of https://github.com/rossoctl/rossoctl/pull/2383

PR #2388, opened 2026-08-17
Refactor: Split agents.py router into modules under 1000 lines

esnible:
## Summary

Splits the 5091-line `app/routers/agents.py` by functional area. No behavior change.

Key changes:

- `agents.py` (899) keeps the router, the read/delete endpoints, and re-exports every moved name, so existing importers and tests are untouched
- New modules: `agents_manifests` (891), `agents_finalize` (637), `agents_migration` (583), `agents_skills` (490), `agents_authbridge` (480), `agents_models` (454), `agents_shipwright` (437), `agents_create` (407), `agents_env` (240), `agents_status` (197)
- Removes the dead `GET /agents/migration/migratable` route — it was registered after `GET /{namespace}/{name}`, which matches the same two path segments, so requests already resolved to `get_agent`. Only the decorator is gone; `migrate_all_agents` still calls the function

Verification:

- 728 tests pass, matching the pre-refactor baseline, with zero existing test files modified
- All 83 moved definitions exist exactly once and are textually identical to the original (decorators excluded)
- Route table matches baseline apart from the one intended removal, in both `authbridge_api` flag states
- `make lint` at 9.94/10, unchanged

Notes for reviewers:

- `list_agents` deliberately stays in `agents.py`: `test_shipwright_builds_api.py` patches `app.routers.agents.settings`, which only reaches code whose module globals live there. Moving it would leave the test green while silently testing nothing.
- `create_agent` is registered via `router.post("")(create_agent)` rather than a sub-router, because FastAPI rejects an empty path on a prefix-less sub-router.

## Related issue(s)

Fixes #


rossoctl/backend/app/routers/agents_skills.py
```diff
@@ -0,0 +1,490 @@
...
+    SKILL_TYPE_LABEL,
+    SKILL_TYPE_VALUE,
+)
+from app.models.shipwright import (
+    BuildOutputConfig,
+    BuildSourceConfig,
+    ResourceType,
+)
+from app.routers.agents_models import CreateAgentRequest
+from app.routers.skills import _sanitize_k8s_name
```
clawgenti: **suggestion** — `_sanitize_k8s_name` is a private helper from `app.routers.skills`. Importing private symbols across router modules creates hidden coupling; consider moving it to `app.utils` or `app.core` so both modules can consume it without depending on each other's internals.

clawgenti (commented):
Clean mechanical split of `agents.py` (5091 lines → 10 focused modules, each ≤ 1000 lines) with zero behavior change and 728 passing tests — well-structured refactor that will pay dividends in maintainability.

**Findings:**
- **PR title fails `verify-pr-title` CI check** — the emoji prefix `🌱` is not in the allowed list; rename to e.g. `Refactor: Split agents.py router into focused modules under 1000 lines`
- All other CI failures are infrastructure 429/network timeouts unrelated to this PR's code
- `_sanitize_k8s_name` is imported from `app.routers.skills` (a private symbol) in `agents_skills.py`; consider extracting it to a shared utility module to avoid cross-module private coupling

---

esnible:
Good catch — done in 89476ade.

`sanitize_k8s_name` now lives in `app/utils/naming.py` as a public helper, next to the existing `sanitize_log` in `app/utils`. `skills.py`, `agents_skills.py` and `skill_autosync.py` all import it, so neither router depends on the other's internals.

One extra thing surfaced while doing it: `app/services/skill_autosync.py` had its own copy of the function, byte-identical apart from a missing docstring. Since the shared home makes it redundant, that duplicate is gone too — the two can no longer drift apart.

`tests/test_skills.py` now imports from the new location; its assertions are unchanged. 728 tests pass, lint unchanged.

rossoctl/backend/app/routers/agents_manifests.py
```diff
@@ -0,0 +1,891 @@
...
+
+"""
+Kubernetes manifest builders for agent workloads.
+
+Builds the env vars, labels/annotations, resource blocks and the four supported
+workload manifests (Deployment, StatefulSet, Job, Sandbox) plus the Service and
+AgentRuntime CR that accompany them.
+
+All four workload builders share the same ``_build_env_vars`` +
+``_get_linked_skill_mounts`` pair, which is why they live together here.
```
clawgenti: `_get_linked_skill_mounts` is actually defined in `agents_skills.py` and imported from there (`from app.routers.agents_skills import _get_linked_skill_mounts`). The docstring claiming it lives here is inaccurate — suggest updating to reflect that the pair is shared across modules, not co-located.

rossoctl/backend/app/routers/agents_skills.py
```diff
@@ -0,0 +1,490 @@
...
+
+
+def _build_fetcher_scripts_data() -> Dict[str, str]:
+    """Return ConfigMap data dict containing per-registry-type fetch scripts."""
+    return {"skillberry.sh": _SKILLBERRY_SH, "generic.sh": _GENERIC_SH}
+
+
+def _ensure_fetcher_scripts_cm(kube: "KubernetesService", namespace: str) -> None:
+    """Create or replace the rossoctl-skill-fetcher-scripts ConfigMap in namespace."""
+    import kubernetes.client as k8s_client
```
clawgenti: Unnecessary function-scoped import: `kubernetes.client` is already accessible at module level via `from kubernetes.client import ApiException` (L18). Consider adding `import kubernetes.client as k8s_client` to the module-level imports instead.

clawgenti (commented):
Large module split — clean refactor with good per-module docstrings, thorough re-exports, and a well-explained dead-route removal.

- `agents_manifests.py` docstring says `_get_linked_skill_mounts` "lives together here", but it is defined in `agents_skills.py` and imported from there (L58).
- `_ensure_fetcher_scripts_cm` in `agents_skills.py` uses a function-scoped `import kubernetes.client as k8s_client` (L371) while `kubernetes.client` is already reachable at module level via the existing `from kubernetes.client import ApiException`.

---

rossoctl/backend/app/utils/naming.py
```diff
@@ -0,0 +1,13 @@
...
+"""
+Helpers for deriving valid Kubernetes resource names from user-supplied text.
+"""
+
+
+def sanitize_k8s_name(name: str) -> str:
+    """Sanitize a name to be valid for Kubernetes resource names."""
+    out = "".join(c.lower() if c.isalnum() or c in ("-", ".") else "-" for c in name)
+    out = out.strip("-.")
+    return out or "skill"
```
clawgenti: The fallback value `"skill"` is domain-specific (it made sense in `skills.py` where the function originated), but `utils/naming.py` is now a general-purpose utility used by agent modules too. Consider `"resource"` as a more neutral default, or document the constraint that callers must validate the fallback is appropriate for their context.

rossoctl/backend/app/routers/agents.py
```diff
@@ -1269,3823 +781,119 @@ async def delete_agent(
...
+router.post(
+    "", response_model=CreateAgentResponse, dependencies=[Depends(require_roles(ROLE_OPERATOR))]
+)(create_agent)
+router.include_router(finalize_router)
+router.include_router(env_router)
+router.include_router(authbridge_router)
+
+# Names re-exported for backwards compatibility (imported by app.routers.tools,
+# app.services.reconciliation, app.services.agent_env_defaults and the tests).
+__all__ = [
```
clawgenti, line 816: Several private helpers (`_build_agent_shipwright_build_manifest`, `_get_linked_skill_mounts`, etc.) appear in `__all__`. The PR description explains this is required for test-patch compatibility — a brief inline comment here noting *why* private names appear in `__all__` would help future readers avoid accidentally removing them.

clawgenti (commented):
Clean, well-structured refactor — the 5000-line `agents.py` is now broken into focused, sub-1000-line modules with no behavior change, and the `__all__` re-export strategy preserves backward compatibility for all existing importers and test patches.

Two minor findings below.

---

esnible:
Both fixed in 7a88b1d4.

**Fallback value** — agreed, `"skill"` was a leftover from the function's origin in `skills.py`. Now `"resource"`, with a docstring noting when it applies.

Worth flagging that this *is* a (narrow) behavior change, not just a rename: the fallback is reachable only when the input sanitizes to nothing — `""`, `"---"`, `"@#$"`. Any input with at least one alphanumeric character is unaffected, so no real skill or agent name changes. If some caller was relying on degenerate input silently becoming a ConfigMap named `skill`, it now becomes `resource` instead.

One subtlety in the tests: `test_leading_trailing_stripped` also asserts `== "skill"` for `"..skill.."` and `"-.skill.-"`, but there `"skill"` is the *stripped result*, not the fallback — those assertions correctly stay unchanged. Only `test_empty_and_invalid_fallback` was updated.

**`__all__` comment** — added above the list:

> The underscore-prefixed entries below are deliberate, not an oversight. Before this module was split up they were module-level privates of `agents.py`, and roughly twenty test modules still import them from `app.routers.agents` (e.g. `from app.routers.agents import _build_sandbox_manifest`); the tests also patch some of them by that dotted path. Listing them keeps those imports working and stops linters from stripping the "unused" re-export imports above. Removing an entry here will break the importing test rather than fail anything in this file, so prefer updating the importers first if you want one gone.

For scale: 42 of the 88 entries are private, which is why the warning felt worth spelling out.

728 tests pass; lint steady at 9.94/10.

esnible:
Both fixed in a0861b66.

**Docstring** — you're right, the pair is split across modules. Reworded to name the location of each:

> All four workload builders go through the same pair of helpers -- `_build_env_vars` (defined here) and `_get_linked_skill_mounts` (imported from `agents_skills`) -- which is why the four builders live together here.

**Nested import** — hoisted to module level as you suggested. One detail worth noting for the record: `from kubernetes.client import ApiException` does *not* bind the name `kubernetes`, so the explicit `import kubernetes.client as k8s_client` is genuinely required rather than merely tidier — dropping the function-scoped import without adding it would have been a `NameError`.

I also applied the same fix to the identical case one module over: `_ensure_authproxy_routes` in `agents_authbridge.py` did `import yaml as _yaml` while `yaml` was already imported at module level and used in four other places. Same defect, same fix. That clears both remaining pylint `W0404` reimport warnings.

728 tests pass; lint steady at 9.94/10.

clawgenti (commented):
Clean completion of the `agents.py` split — the final commit addresses the `sanitize_k8s_name` fallback (now `"resource"` for a general-purpose utility) and adds the `__all__` comment explaining why private names are re-exported for test patching.

---

rossoctl/backend/app/routers/agents_env.py
```diff
@@ -0,0 +1,240 @@
...
+        logger.info(f"Using CA bundle: {ca_bundle_path}")
+
+        # Create SSL context with system certificates
+        ssl_context = ssl.create_default_context(cafile=ca_bundle_path)
+
+        async with httpx.AsyncClient(
+            timeout=10.0, follow_redirects=True, verify=ssl_context
+        ) as client:
+            logger.debug(f"Making HTTP request to {request.url}")
+            response = await client.get(request.url)
```
cwiklik, line 215: Disposition note for CodeQL: this is the critical `py/full-ssrf` alert, but it's `fetch_env_from_url` moved verbatim from `agents.py` (~L4178) — the SSRF guards (private/loopback/link-local IP blocking via `is_ip_blocked`, scheme validation, hostname→IP resolution) moved with it. Pre-existing debt re-attributed by the large move, not introduced by this refactor. Dismiss as pre-existing / accept via baseline rather than changing behavior in a pure split. Same applies to the 39 log-injection / 8 url-substring-sanitization / stack-trace alerts.

cwiklik (approved):
Behavior-preserving split of the 5091-line `agents.py` — nicely done. `agents.py` re-exports every moved name via an explicit `__all__` with good comments (preserving the dotted paths ~20 test modules patch, and the `finalize_shipwright_build` late-import for reconciliation). backend-tests pass at the 728 baseline, and the moved code is verbatim on the paths I spot-checked.

The two failing checks are **gate artifacts, not code defects**:

1. **CodeQL — "49 new alerts (1 critical)"**: these are pre-existing findings re-attributed because the move is large (CodeQL says as much: "alerts not introduced by this PR… because the code changes were too large"). I confirmed the lone **critical `py/full-ssrf`** at `agents_env.py:215` is `fetch_env_from_url` **moved verbatim from `agents.py` (~L4178)** — its SSRF guards (private/loopback/link-local IP blocking, scheme + hostname checks) came along. The 39× log-injection / 8× url-substring / stack-trace are likewise relocated debt from the old router. These should be dispositioned as pre-existing (baseline/dismiss), not "fixed" in a pure move.
2. **verify-pr-title**: fails on the `🌱` prefix — please retitle with a conventional prefix, e.g. `Refactor: Split agents.py router into modules`.

Follow-up worth filing (out of scope here): actually pay down the 39× log-injection in these modules, à la the KubernetesService `_sanitize` work.

Minor: the description says "zero existing test files modified," but `test_skills.py` (+15/-15) and `skill_autosync.py` were touched for import-path updates — accurate to say "no agent test files modified."

Code LGTM; clear the title + CodeQL gate for green.

Assisted-By: Claude Code

PR #2392, opened 2026-08-18
feat: integrate optional Context Service

moonlight16:
## Summary

- add an optional proxy for Context Service named-resource APIs
- resolve Context Service PVC attachments when creating StatefulSet or Sandbox agents
- configure the integration solely through `CONTEXT_SERVICE_URL`
- expose Helm configuration and document the user workflow
- preserve normal Rosso behavior when Context Service is not configured

## Why

This gives Rosso agents access to independently managed, shared or dedicated context storage without making Context Service a platform requirement. The integration validates the returned attachment contract rather than coupling agent creation to a specific context type.

## Dependency

This integrates with the named context-resource API merged through https://github.com/rossoctl/context-service/pull/1. Context Service remains an optional, separately installed component.

## Validation

- Ruff on changed backend files
- 747 backend tests passed
- Helm rendered with Context Service disabled and enabled
- `git diff --check`

Assisted-By: Codex

### DIFF
--- a/rossoctl/backend/app/routers/agents.py
+++ b/rossoctl/backend/app/routers/agents.py
@@ -327,7 +327,7 @@ class CreateAgentRequest(BaseModel):
     # Persistent storage (for Sandbox and StatefulSet workloads)
     persistentStorage: Optional[PersistentStorageConfig] = None
 
-    # Named Context Service resources. Requires the context-service feature flag.
+    # Named Context Service resources. Requires CONTEXT_SERVICE_URL.
     contexts: Optional[List[ContextAttachment]] = None
 
     # Shipwright build configuration
--- a/rossoctl/backend/app/routers/agents.py
+++ b/rossoctl/backend/app/routers/agents.py
@@ -232,6 +232,14 @@ class PersistentStorageConfig(BaseModel):
     size: str = "1Gi"
 
 
+class ContextAttachment(BaseModel):
+    """A named Context Service resource mounted into the agent."""
+
+    name: str
+    mountPath: str = "/workspace"
+    readOnly: bool = False
+
+
 class CreateAgentRequest(BaseModel):
     """Request to create a new agent."""
 
@@ -318,6 +326,9 @@ class CreateAgentRequest(BaseModel):
 
     # Persistent storage (for Sandbox and StatefulSet workloads)
     persistentStorage: Optional[PersistentStorageConfig] = None
+
+    # Named Context Service resources. Requires the context-service feature flag.
+    contexts: Optional[List[ContextAttachment]] = None
 
     # Shipwright build configuration
     shipwrightConfig: Optional[ShipwrightBuildConfig] = None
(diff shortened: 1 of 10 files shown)


rossoctl/backend/app/routers/contexts.py
```diff
@@ -0,0 +1,97 @@
...
+    """Validate and encode a Kubernetes-name URL path segment."""
+    if len(value) > max_length or not _KUBERNETES_NAME.fullmatch(value):
+        raise HTTPException(
+            status_code=400,
+            detail=f"{field} must be a lowercase Kubernetes name",
+        )
+    return quote(value, safe="")
+
+
+async def _request(method: str, path: str, body: dict | None = None) -> httpx.Response:
```
cwiklik, line 47: suggestion (non-blocking): `agents.py` guards on empty `CONTEXT_SERVICE_URL` and returns a clean 400 "integration disabled", but these proxy endpoints don't — with the router registered and `CONTEXT_SERVICE_URL` unset, `_url()` produces a relative path and `_request` fails with a confusing 503 "Context Service unavailable". A URL-presence guard here (clean 501/400 "disabled") would give a misconfigured cluster a clear signal instead of a transport error, matching the agents.py behavior.

rossoctl/backend/app/routers/agents.py
```diff
@@ -319,6 +327,9 @@ class CreateAgentRequest(BaseModel):
     # Persistent storage (for Sandbox and StatefulSet workloads)
     persistentStorage: Optional[PersistentStorageConfig] = None
 
+    # Named Context Service resources. Requires the context-service feature flag.
```
cwiklik: nit: this says "Requires the context-service feature flag," but there's no canonical `rossoctl_feature_flag_context_service` — the gate is `CONTEXT_SERVICE_URL` presence plus the module-load guard (surfaced as `contextService` in the features endpoint). Either add the boolean flag per CLAUDE.md's canonical mechanism, or reword to "requires CONTEXT_SERVICE_URL" so the comment matches the actual gate.

cwiklik (approved):
Backend side of the optional Context Service integration — well-hardened. The proxy (`contexts.py`) enforces roles on every endpoint (operator for write, viewer+operator for read), validates+encodes path segments (`_KUBERNETES_NAME` + `quote(safe="")`, i.e. the path-validation fix — blocks path-injection/SSRF), sets an httpx timeout, uses Pydantic `Literal` enums, and validates the returned attachment contract (kind==pvc, claimName present).

Notably, `agents.py` **confines context resolution to the agent's own namespace**: `ContextAttachment` has only `name`/`mountPath`/`readOnly` (no namespace), and `_resolve_context_mounts` resolves against `request.namespace`. So a caller can't attach another tenant's PVC by name — this closes the exact cross-tenant gap I flagged on the harness-side counterpart. Plus RBAC (operator) on agent create, feature-gating, and absolute+unique mountPath checks.

Two minor consistency notes inline (non-blocking):
1. Proxy endpoints don't guard on empty `CONTEXT_SERVICE_URL` — they'd 503 "unavailable" rather than a clean "disabled" like agents.py does.
2. The "context-service feature flag" comment doesn't match the actual gate (CONTEXT_SERVICE_URL + module-load); either add the canonical boolean flag or reword.

24 checks green; E2E HyperShift pending trigger. Both commits signed, no `.claude`/`.vscode`. LGTM.

Assisted-By: Claude Code

PR #2393, opened 2026-08-18
test: Add transparent inbound interception E2E suite

huang195:
## Summary

E2E suite gating transparent inbound interception (rossoctl/cortex#330). Companions: rossoctl/cortex#776 (listener + iptables, **merged**), rossoctl/operator#511 (the switch).

The load-bearing test is **`test_direct_pod_dial_is_validated`**. The reason the feature exists is that inbound JWT validation must not be sidestepped by another pod dialing the agent's real port, and nothing in the existing suite covers that.

## Design notes

**Two agents, two namespaces.** One `transparent`, one default `reverse-proxy`. A single-shape run cannot distinguish "validation works" from "nothing reached the agent at all". The reverse-proxy namespace is a control that deliberately documents the bypass as it exists today — if it starts failing because the bypass is gone, the default changed and the test should be retired with it.

**The probe always runs from a separate pod.** Dialing from inside the agent pod would traverse loopback, which shares the network namespace, is inside the trust boundary by construction, and is deliberately not intercepted — so it would prove nothing.

**The stand-in agent honors `PORT`.** Without that the reverse-proxy control could not be exercised at all: an agent that ignores `PORT` collides with AuthBridge on the stolen port and never starts, which is itself one of the failure modes transparent interception removes.

## Coverage beyond the boundary property

- **Shape** — agent keeps its own port and gets no `PORT` env var; sidecar declares `transparent-in` and does not claim the agent's port; proxy-init has both `INBOUND_TRANSPARENT_PORT` and `POD_IP` (from the downward API, without which the ambient inbound path is uncaptured); `AB_INBOUND` is live in the pod's nat table, not merely configured.
- **Don't break the neighbours** — `:9091` is not JWT-gated (it would fail kubelet probes and crash-loop the pod), `:9094` stays reachable for `abctl`, and `AB_REDIRECT` is still present so inbound capture did not disturb the egress guard.

## Three platform constraints the fixtures satisfy

Each of these otherwise produces a silently *meaningless* run rather than a failure — found by running against a live cluster:

1. The namespace needs `rossoctl-enabled=true`, or the injection webhook's `namespaceSelector` skips it and the pod comes up with **no sidecar at all**.
2. `rossoctl.io/type` cannot be applied by hand — the `agent-label-protection` ValidatingAdmissionPolicy rejects it. Injection is driven through an AgentRuntime CR, which is how agents are deployed for real.
3. Readiness alone is not a usable gate: the AgentRuntime controller labels the pod template *after* the Deployment exists, so the first ready pod predates injection. The fixtures wait for the `authbridge-proxy` container.

Images are ones already cached on Kind nodes (`nginx-unprivileged:alpine`, whose busybox `wget` doubles as the probe), so the suite needs no registry pull.

## Result: 10 passed, 0 skipped on Kind

The A/B that matters, verified live and attributed to the sidecar rather than to a waypoint (`Transparent inbound server listening addr=[::]:8083`, then `jwt-validation … status=401 … reason="missing Authorization header"`):

| Mechanism | Port dialed pod-to-pod | Result |
|---|---|---|
| `reverse-proxy` (default) | 8000 (sidecar) | 401 |
| `reverse-proxy` (default) | **8001 (relocated agent)** | **200 — the bypass** |
| `transparent` | 8000 (agent's real port) | **401** |

Also confirmed: an **undeclared port `:80` returns 401** (multi-port coverage is real), and health/stats/session return 200/200/404 — not gated.

## Three things the first live run changed

Running this for real surfaced two bugs in the suite and one in the platform:

- **`test_agent_keeps_its_own_port` could not pass even on correct behavior.** It asserted `PORT` was absent, but the stand-in agent declares `PORT` itself — it must, or the reverse-proxy control cannot be exercised at all. Now asserts the operator did not *rewrite* it, which is the real discriminator.
- **Two tests could only ever skip.** The chain and egress-guard tests exec'd `iptables-save` in the sidecar, whose image has no iptables — quietly dropping both assertions from a green run. They now assert on proxy-init's own report, which is meaningful because that container runs `require_jump` guards that abort on a rule that did not land.
- **Retargeted at pre-onboarded namespaces** (`team1`/`team2`, overridable). A fresh namespace lacks the per-namespace plumbing — `authbridge-config`, plus the Keycloak client registration that produces each agent's credentials Secret — so an injected pod there never starts, *including default reverse-proxy ones*. Fixtures flip `inboundInterception` in place and **restore the original body on teardown**, since these namespaces are shared.

## `TI_STUB_CREDENTIALS`

On clusters where Keycloak client registration is broken (no `k8s.keycloak.org` CRD, as on community-provider installs), every injected pod stays `Pending` on `FailedMount`. Without the flag the suite **skips with that cause named**, rather than failing and pointing suspicion at the interception rules.

`TI_STUB_CREDENTIALS=1` creates a placeholder so the boundary can still be tested. Opt-in on purpose: stubbing by default would let this suite go green on a broken cluster. Sound for these assertions — the Secret is for *outbound* token-exchange, and inbound rejects a request with no Authorization header before any IdP contact — but explicitly **not** sound for any test needing a valid token.

## Not covered here

The **ambient (HBONE) path**. The cluster used runs only `istiod` — no ztunnel — so there is no ambient data plane to exercise; counters confirm all traffic took plain PREROUTING. Adding an ambient case needs a cluster with ztunnel deployed.

`E2E HyperShift` is `pending` awaiting a maintainer `/run-e2e`.

### DIFF
--- a/.github/scripts/operator/90-run-e2e-tests.sh
+++ b/.github/scripts/operator/90-run-e2e-tests.sh
@@ -44,7 +44,11 @@ fi
 # Support filtering tests via PYTEST_FILTER or PYTEST_ARGS
 # PYTEST_FILTER: pytest -k filter expression (e.g., "test_mlflow" or "TestGenAI")
 # PYTEST_ARGS: additional pytest arguments (e.g., "-x" for stop on first failure)
-PYTEST_TARGETS="${PYTEST_TARGETS:-tests/e2e/common tests/e2e/rossoctl_operator}"
+# transparent_inbound self-gates: it skips with a named reason when the deployed
+# platform predates transparent inbound interception, or when the per-agent
+# Keycloak credentials Secret cannot be created. So including it here is safe on
+# older images and becomes a real gate as soon as they carry the feature.
+PYTEST_TARGETS="${PYTEST_TARGETS:-tests/e2e/common tests/e2e/rossoctl_operator tests/e2e/transparent_inbound}"
 PYTEST_OPTS="-v --timeout=300 --tb=short"
 
 if [ -n "${PYTEST_FILTER:-}" ]; then
--- a/rossoctl/tests/e2e/transparent_inbound/README.md
+++ b/rossoctl/tests/e2e/transparent_inbound/README.md
@@ -20,6 +20,27 @@ be retired with it.
 - A cluster with the operator + authbridge images built from this branch.
 - `proxy.allowedInboundInterception` must permit `transparent` (the default).
 - Linux nodes: the feature is iptables-based. Skipped elsewhere.
+- **A namespace the platform can onboard.** The injected sidecar mounts a
+  per-agent Keycloak client-credentials Secret, created by the operator's
+  client-registration path. If that path is broken — e.g. the Keycloak CRD
+  (`k8s.keycloak.org/v2alpha1`) is absent, as on a cluster installed with the
+  community Keycloak provider — every injected pod stays `Pending` on
+  `FailedMount` and the suite errors in fixture setup regardless of this
+  feature. Check `kubectl describe pod` for a missing
+  `rossoctl-keycloak-client-credentials-*` Secret before debugging the
+  interception rules.
+
+Three platform constraints the fixtures satisfy, each of which otherwise
+produces a silently meaningless run:
+
+- The namespace needs `rossoctl-enabled=true`, or the injection webhook's
+  `namespaceSelector` skips it and the pod comes up with no sidecar at all.
+- `rossoctl.io/type` cannot be applied by hand — the `agent-label-protection`
+  ValidatingAdmissionPolicy rejects it. Injection is driven through an
+  AgentRuntime CR, which is how agents are deployed for real.
+- Readiness alone is not a usable gate: the AgentRuntime controller labels the
+  pod template *after* the Deployment exists, so the first ready pod predates
+  injection. The fixtures wait for the `authbridge-proxy` container.
 
 ```sh
 uv run pytest rossoctl/tests/e2e/transparent_inbound/ -v
--- a/rossoctl/tests/e2e/transparent_inbound/README.md
+++ b/rossoctl/tests/e2e/transparent_inbound/README.md
@@ -15,20 +15,38 @@ port answers without validation. If that test ever fails because the bypass is
 gone, the default changed — which is a decision, not a bug, and the test should
 be retired with it.
 
+## Namespaces
+
+Runs in **pre-onboarded** namespaces (`team1`/`team2` by default, override with
+`TI_NS_TRANSPARENT` / `TI_NS_CONTROL`) rather than creating its own. A fresh
+namespace lacks the platform's per-namespace plumbing — `authbridge-config`, and
+the Keycloak client registration that produces each agent's credentials Secret —
+so an injected pod there never starts, for reasons unrelated to this feature.
+
+Each fixture flips the namespace's `inboundInterception` in place and **restores
+the original body on teardown**. These are shared namespaces: leaving one flipped
+would silently change every other agent in it on its next pod recreation.
+
 ## Requirements
 
 - A cluster with the operator + authbridge images built from this branch.
 - `proxy.allowedInboundInterception` must permit `transparent` (the default).
 - Linux nodes: the feature is iptables-based. Skipped elsewhere.
-- **A namespace the platform can onboard.** The injected sidecar mounts a
-  per-agent Keycloak client-credentials Secret, created by the operator's
-  client-registration path. If that path is broken — e.g. the Keycloak CRD
-  (`k8s.keycloak.org/v2alpha1`) is absent, as on a cluster installed with the
-  community Keycloak provider — every injected pod stays `Pending` on
-  `FailedMount` and the suite errors in fixture setup regardless of this
-  feature. Check `kubectl describe pod` for a missing
-  `rossoctl-keycloak-client-credentials-*` Secret before debugging the
-  interception rules.
+- **Working Keycloak client registration.** The injected sidecar mounts a
+  per-agent credentials Secret created by the operator. Where that path is broken
+  — e.g. the `k8s.keycloak.org/v2alpha1` CRD is absent, as on a cluster installed
+  with the community Keycloak provider — every injected pod stays `Pending` on
+  `FailedMount`, including default reverse-proxy ones. The suite **skips with that
+  cause named** rather than failing and pointing suspicion at the interception
+  rules.
+
+  To test the boundary anyway on such a cluster, set `TI_STUB_CREDENTIALS=1` to
+  create a placeholder Secret. This is opt-in on purpose: stubbing by default
+  would let the suite go green on a cluster whose Keycloak registration is
+  broken. The stub is sound for these assertions — the Secret is for *outbound*
+  token-exchange, and inbound validation rejects a request with no Authorization
+  header before any IdP contact — but it would **not** be sound for any test
+  needing a valid token.
 
 Three platform constraints the fixtures satisfy, each of which otherwise
 produces a silently meaningless run:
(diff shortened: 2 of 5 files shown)


huang195:
## Green on a real cluster: 10 passed, 0 skipped

First live run surfaced three problems — two in this suite, one in the platform.

**Retargeted at pre-onboarded namespaces** (`team1`/`team2`, overridable). A fresh namespace lacks the per-namespace plumbing (`authbridge-config`, plus the Keycloak client registration that produces each agent's credentials Secret), so an injected pod there never starts — including default reverse-proxy ones. Fixtures now flip `inboundInterception` in place and **restore the original body on teardown**, since these namespaces are shared.

**Fixed an assertion that could not pass even on correct behavior.** `test_agent_keeps_its_own_port` asserted `PORT` was absent, but the stand-in agent declares `PORT` itself — it must, or the reverse-proxy control can't be exercised at all. Now asserts the operator didn't *rewrite* it, which is the real discriminator.

**Removed two silent skips.** The chain and egress-guard tests exec'd `iptables-save` in the sidecar, whose image has no iptables — so they could only ever skip, quietly dropping two assertions from a green run. They now assert on proxy-init's own report, which is meaningful because that container runs `require_jump` guards that abort on a rule that didn't land. The egress test also checks egress setup completed *before* inbound setup, since the ambient DNAT is installed at the head of the egress chain.

**`TI_STUB_CREDENTIALS=1`** for clusters where Keycloak registration is broken (this one: no `k8s.keycloak.org` CRD). Opt-in on purpose — stubbing by default would let the suite go green on a broken cluster. Sound for these assertions (the Secret is for *outbound* token-exchange; inbound rejects a header-less request before any IdP contact), explicitly **not** sound for a test needing a valid token. Without the flag the suite skips with that cause named.

## The A/B that matters

| Mechanism | Port | Result |
|---|---|---|
| `reverse-proxy` (default) | 8000 (sidecar) | 401 |
| `reverse-proxy` (default) | **8001 (relocated agent)** | **200 — the bypass** |
| `transparent` | 8000 (agent's real port) | **401** |

Attributed to the sidecar, not a waypoint: `Transparent inbound server listening addr=[::]:8083` followed by `jwt-validation … status=401 … reason="missing Authorization header"`.

Also verified an **undeclared port :80 returns 401**, so multi-port coverage is real, and health/stats/session (9091/9093/9094) return 200/200/404 — not gated.

Assisted-By: Claude Code

huang195:
All seven addressed (`3a828db6`) — every one was valid.

## #1 The suite gated nothing — the important one

Confirmed exactly. `openshell/` and `token_exchange/` each have a runner; `transparent_inbound/` was referenced by no workflow or script, so it ran only when a human typed the command. Not what a regression gate is for.

Added to the shared target list in `90-run-e2e-tests.sh`, so it runs wherever that runner does (Kind and HyperShift e2e, release validation).

That is only safe because the suite now **self-gates**. It runs against whatever images a cluster happens to have, and an operator predating transparent inbound silently ignores the namespace flip and port-steals instead — which would fail the transparent assertions for a reason unrelated to the code under test. Detected from the injected pod (the sidecar declares `transparent-in` only when the feature is present) and skipped with the cause named. It becomes a real gate as soon as a cluster carries the feature, which matters here since cortex#776 is merged but the operator switch (rossoctl/operator#511) is not.

## #2 Stub Secret — two bugs, both real

**It leaked.** Teardown restored the namespace config and deleted the workload but never the placeholder Secret, so a fake credential outlived the run in a shared namespace. Now recorded on creation and removed on teardown — verified: today's run cleaned its own, and the three left in `team1`/`team2` turned out to be debris from my earlier manual debugging, which I've also removed.

**And the stdout/stderr point is exactly right.** `"already exists"` goes to stderr, which `kubectl()` drops, so from the second poll onward it always read as failure and re-attempted the create every 3s. Keyed on the exit code now, via a `kubectl_rc` helper.

## #3 `_set_mechanism` not idempotent

Correct. Prepending blind meant `TI_KEEP=1` or a killed run left keys behind, and the next run emitted duplicate top-level YAML keys. Strips any it finds first.

## #4 `_namespace_config` dead — and worse than dead

Right on both counts: dead after the retarget to pre-onboarded namespaces, *and* it built exactly the synthetic config that `_set_mechanism`'s own docstring argues against. Leaving it there invited someone to use it. Deleted.

## #5–#7 Three assertions that could pass while broken

These are the ones I'm most glad you caught — all three were false confidence rather than cosmetic:

- `"9091" in logs` matched the **whole log**, where that number appears in plenty of other contexts, so it never actually checked the exempt list. Now parses the `exempt sidecar ports=` field.
- `logs.index()` raises `ValueError` when a marker is missing, so that assertion's message could never print — the failure would have surfaced as an unrelated traceback. `find()` plus an explicit presence check.
- The control asserting `status != 401` would stay green if the bypass closed with a 403, or if the port stopped answering at all. The control would silently stop being a control while still reporting success. Now asserts 200, with the message explaining that a failure here may mean the default changed and the test should be retired.

## Verification

10 passed on Kind after the changes; `ruff check` and `ruff format --check` clean; stub Secret confirmed removed on teardown.

Assisted-By: Claude Code

cwiklik (approved):
### Summary

Excellent, strict E2E suite for the transparent-inbound boundary (companion to operator#511 / cortex#776). The load-bearing test and its control are exactly right, and the design decisions in the body (two agents/two namespaces, probe from a *separate* pod not loopback, PORT-honoring stand-in) are all faithfully implemented. Approving — the tests are strong and this is test-only.

**Verified strict:**
- `test_direct_pod_dial_is_validated` — dials the agent's real `podIP:port` from a separate probe pod with no auth header and asserts `status != 200` **and** `status in (401, 403)` (specific codes, not a loose "not 5xx").
- `test_reverse_proxy_control_bypass_is_reachable` — a genuine contrasting control: asserts the relocated port `!= AGENT_PORT` (it really is port-stealing) and `status == 200`, with a comment on why a degrading control (`!= 401`) would be wrong.
- `test_service_routed_request_is_validated` — `(401, 403)`. Fixtures clean up (restore shared-namespace config, delete stub Secret, `--rm` probes), waits are deadline-bounded, and the only "secret" is an opt-in placeholder — good hygiene.

**Main concern — gate robustness, not assertion quality (fast-follow):** this suite is a security gate that can **silently disarm itself** on a cluster that is supposed to have the feature:
- `_require_platform_support` reads a missing `transparent-in` sidecar port as "platform predates the feature" and **skips** — but on a branch-built CI cluster that is a *regression*, and it would skip rather than fail. It can't distinguish feature-absent from feature-broken.
- `_wait_ready` skips when the Keycloak-credentials Secret can't be created; if the CI cluster lacks working Keycloak registration, every injected pod stays Pending and the **entire suite skips green** — the boundary test never runs. Nothing in `90-run-e2e-tests.sh` sets `TI_STUB_CREDENTIALS=1`.

Neither can make the security test *pass falsely* — it becomes a named SKIP — but a SKIP reads as "not a failure," and nothing positively asserts the transparent path actually executed. The shell comment's "becomes a real gate as soon as they carry the feature" isn't enforced anywhere. Recommend a fast-follow: a CI flag (e.g. `TI_REQUIRE_FEATURE=1`) that converts these two skips into hard failures for branch-built images, and/or set `TI_STUB_CREDENTIALS=1` in the runner if the CI cluster has no Keycloak CRD. (You reported 10 passed / 0 skipped on Kind, so the gate does execute in your env — this is about keeping it armed in shared CI over time.)

**Minor:**
- `test_health_port_not_gated` / `test_session_api_not_gated` assert only `not in (401, 403)`, so a crash-looping 5xx passes — worth also asserting `< 500`.
- `_set_mechanism`'s empty-config skip can fire on a transient `kubectl` error (`check=False`).
- Stale module docstring says "scratch namespace," but the suite mutates the shared team1/team2 namespaces (contradicts the README and the teardown). Worth fixing so a reader doesn't think teardown nukes a scratch ns.

**Author**: huang195 (MEMBER — maintainer)
**Areas reviewed**: Python E2E tests, Shell (CI runner), test quality / skip-safety, fixtures & secrets
**CI**: 23 green; E2E HyperShift pending maintainer `/run-e2e`.

PR #2397, opened 2026-08-18
feat: Report backend version from GET /auth/config

esnible:
## Summary

Adds a `version` key to `GET /api/v1/auth/config`, reporting the same string the UI shows in its `rossoctl-brand-version` badge.

```json
{"enabled": false, "version": "v0.7.0-alpha.6", "keycloak_url": null, ...}
```

Key changes:

- `app/routers/auth.py` — `version: str` on `AuthConfigResponse`, resolved by `_backend_version()`. Set on **both** return paths, including the early return when auth is disabled.
- `app/core/config.py` — new `rossoctl_backend_version` setting (`ROSSOCTL_BACKEND_VERSION`).
- `backend/Dockerfile` — `ARG RELEASE_TAG` baked in as `ROSSOCTL_BACKEND_VERSION`, mirroring what `ui-v2/Dockerfile` does to `package.json`. This is what makes both images report the same version from the same source.
- `Makefile` — the backend build now passes `--build-arg RELEASE_TAG`, matching the frontend build.
- `tests/test_auth.py` — 5 tests.

Resolution order: `ROSSOCTL_BACKEND_VERSION` → `rossoctl-backend` package metadata → `"unknown"`.

## Notes for reviewers

**The Makefile change is a fix, not cosmetic.** CI already passes `RELEASE_TAG` to every image in the build matrix, but the Makefile passed it only to the frontend (line 76) and not the backend (line 93). Without it, `make build-load-ui` produced a UI showing the release tag next to a backend reporting `unknown`.

**Why the env var rather than package metadata alone**: `backend/Dockerfile` deliberately installs dependencies only, not the app package, so `importlib.metadata` raises `PackageNotFoundError` in the runtime image. Metadata is the local-dev/test fallback (`backend/pyproject.toml`); the env var is the only source inside a built image.

**Verified**: 752 tests pass; `ROSSOCTL_BACKEND_VERSION=v0.7.0-alpha.6` returns that string over HTTP; the `ARG` sits in the runtime stage (an `ARG` in the builder stage would silently yield `unknown`). The Dockerfile line itself is not exercised by the unit tests — worth a `docker build --build-arg RELEASE_TAG=test` on review.

Note `RELEASE_TAG` carries the `v` prefix (`v0.7.0-alpha.6`) since it comes from `git describe`/`github.ref_name`, unlike `Chart.yaml`'s `appVersion`.

## Related issue(s)

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,7 +90,7 @@ build-load-ui-backend:
 		echo "Info: Podman backend detected. Using --load flag for build."; \
 	fi
 	@echo ""
-	docker build -t $(UI_BACKEND_REPO):$(UI_BACKEND_TAG) -f $(UI_BACKEND_DIR)/Dockerfile $(DOCKER_BUILD_CONTEXT) $(DOCKER_BUILD_FLAGS)
+	docker build -t $(UI_BACKEND_REPO):$(UI_BACKEND_TAG) -f $(UI_BACKEND_DIR)/Dockerfile $(DOCKER_BUILD_CONTEXT) $(DOCKER_BUILD_FLAGS) --build-arg RELEASE_TAG=$(RELEASE_TAG)
 	@echo ""
 	@echo "Loading backend image into kind cluster $(KIND_CLUSTER_NAME)..."
 	kind load docker-image $(UI_BACKEND_REPO):$(UI_BACKEND_TAG) --name $(KIND_CLUSTER_NAME)
--- a/rossoctl/backend/Dockerfile
+++ b/rossoctl/backend/Dockerfile
@@ -41,6 +41,13 @@ COPY --from=builder /app/.venv /app/.venv
 # Copy application code
 COPY backend/app/ ./app/
 
+# Bake the release tag in as the reported backend version. Mirrors what
+# ui-v2/Dockerfile does to package.json, so the backend's GET /auth/config and the
+# UI's version badge report the same string from the same source. CI passes
+# RELEASE_TAG for every image in the build matrix (.github/workflows/build.yaml).
+ARG RELEASE_TAG=unknown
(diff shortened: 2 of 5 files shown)


cwiklik (approved):
### Summary

Clean feature — adds a `version` field to `GET /auth/config`, resolved via `ROSSOCTL_BACKEND_VERSION` (baked from `RELEASE_TAG` in `backend/Dockerfile`) → `rossoctl-backend` package metadata → `"unknown"`. The resolution order matches runtime reality (the image installs dependencies only, so the env var is the real source and metadata is the dev/test fallback), and both response paths set it, including the auth-disabled early return.

I checked the one real risk of adding a **required** Pydantic field: it breaks no caller. `AuthConfigResponse(...)` is only constructed in `auth.py` (both sites updated); the other `AuthConfigResponse` references are the Go TUI client and the TS UI, which consume the JSON and tolerate an additive key. The Makefile change is a genuine parity fix (backend build now gets `--build-arg RELEASE_TAG`, so it no longer reports `unknown` next to a versioned UI). Tests are solid — 5 cases covering both paths, metadata fallback, env-override precedence, and the `unknown` fallback; `backend-tests` and `lint` are green.

Approving.

**Non-blocking note:** the `version` field isn't behind a feature flag. CLAUDE.md states all new features ship gated/off-by-default, but a passive version string on a config endpoint isn't the functional-subsystem kind of feature that policy targets — leaving it un-gated seems right; just flagging it as a conscious call.

**Author**: esnible (MEMBER — maintainer)
**Areas reviewed**: Python (router/config), Dockerfile, Makefile, Tests, API-contract consumers (Go TUI + TS UI)
**CI**: 21 green; Deploy & Test running; E2E HyperShift pending maintainer `/run-e2e`.

PR #2399, opened 2026-08-19
chore(release): pin image tags for v0.8.0-alpha.1

cwiklik:
## Summary

Cut **`v0.8.0-alpha.1`** — the first alpha off `main` since the `0.7.0` GA — so people can install pinned, reproducible artifacts to test features in `main`.

Pins all rossoctl-built images and aligns core dependency versions:

| Component | Version |
|-----------|---------|
| platform (this repo) | `0.8.0-alpha.1` |
| operator-chart | `0.3.1-alpha.1` |
| cortex / AuthBridge injection images | `v0.7.0-alpha.3` |
| examples | `v0.2.0` (unchanged — no cut warranted) |

### Changes
- `Chart.yaml` version/appVersion `0.7.0-alpha.6 → 0.8.0-alpha.1`; operator-chart dep `0.3.0-alpha.10 → 0.3.1-alpha.1` (`Chart.lock` regenerated).
- rossoctl images (ui-v2, backend, ui/agent/api/phoenix/mlflow-oauth-secret, operator-spiffe-bootstrap, spiffe-idp-setup) `→ v0.8.0-alpha.1`.
- **`operator-spiffe-bootstrap: latest → v0.8.0-alpha.1`** — no floating tags, even for alphas.
- operator image tag `→ 0.3.1-alpha.1`.
- AuthBridge injection overrides `v0.6.0-alpha.12 → v0.7.0-alpha.3` (match cortex alpha.3, and the operator chart's own pinned defaults).
- rossoctl-deps spiffe-idp-setup `→ v0.8.0-alpha.1`.

### Verification
- `scripts/check-release-pins.sh` → **PASS**
- Full-render `:latest` gate (`helm template` incl. operator subchart) → **PASS**, all injected images pinned.

### Notes
Two `pin-release-tags.sh` gaps worked around here (filed separately): yq write path broken on yq v4.45.1, and the managed-image list misses `phoenix-oauth-secret` + the rossoctl-chart `spiffe-idp-setup` copy.

Assisted-By: Claude Code

### DIFF
--- a/charts/rossoctl/values.yaml
+++ b/charts/rossoctl/values.yaml
@@ -400,25 +400,11 @@ operator-chart:
   # the explicit value is required because neither ZTWIM nor spire-bundle exist at startup.
   signatureVerification:
     spireTrustDomain: localtest.me
-  # Pin sidecar image tags injected by the AuthBridge webhook.
-  #
-  # After cortex#411 there are three combined images:
-  #   * authbridge      — proxy-sidecar mode (default), full plugin set
-  #                       including the a2a/mcp/inference parsers.
-  #   * authbridge-envoy — envoy-sidecar mode, full plugin set, expects
-  #                       proxyInit's iptables setup.
-  #   * authbridge-lite  — proxy-sidecar shape, auth-only plugins
-  #                       (parsers dropped). For size-constrained deployments;
-  #                       not selected by default.
-  # Spiffe-helper is bundled inside each combined image and gated per
-  # workload by the SPIRE_ENABLED env var; it is no longer a separate
-  # sidecar. Client registration is operator-managed.
-  defaults:
-    images:
-      authbridge: ghcr.io/rossoctl/cortex/authbridge:v0.7.0-alpha.3
-      envoyProxy: ghcr.io/rossoctl/cortex/authbridge-envoy:v0.7.0-alpha.3
-      authbridgeLite: ghcr.io/rossoctl/cortex/authbridge-lite:v0.7.0-alpha.3
-      proxyInit: ghcr.io/rossoctl/cortex/proxy-init:v0.7.0-alpha.3
+  # AuthBridge sidecar image tags (authbridge / -envoy / -lite / proxy-init) are
+  # NOT overridden here — they come from the operator subchart's own defaults,
+  # pinned via the operator-chart dependency version in Chart.yaml. Keeping a
+  # single source of truth avoids the silent drift we hit when this override
+  # block sat stale behind the subchart default (rossoctl/rossoctl#2399).
 
 # ------------------------------------------------------------------
 #  SPIFFE Configuration
--- a/charts/rossoctl/values.yaml
+++ b/charts/rossoctl/values.yaml
@@ -194,7 +194,7 @@ ui:
   frontend:
     enabled: true
     image: ghcr.io/rossoctl/rossoctl/ui-v2
-    tag: v0.7.0-alpha.6
+    tag: v0.8.0-alpha.1
     resources:
       limits:
         cpu: 100m
@@ -211,7 +211,7 @@ ui:
   # Backend configuration
   backend:
     image: ghcr.io/rossoctl/rossoctl/backend
-    tag: v0.7.0-alpha.6
+    tag: v0.8.0-alpha.1
     # Optional Context Service API URL. Empty disables context resource APIs and
     # agent context attachments. Example:
     # http://context-service.serverless-harness.svc.cluster.local:8080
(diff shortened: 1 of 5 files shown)


charts/rossoctl/values.yaml
```diff
@@ -415,10 +415,10 @@ operator-chart:
   # sidecar. Client registration is operator-managed.
   defaults:
     images:
-      authbridge: ghcr.io/rossoctl/cortex/authbridge:v0.6.0-alpha.12
-      envoyProxy: ghcr.io/rossoctl/cortex/authbridge-envoy:v0.6.0-alpha.12
-      authbridgeLite: ghcr.io/rossoctl/cortex/authbridge-lite:v0.6.0-alpha.12
-      proxyInit: ghcr.io/rossoctl/cortex/proxy-init:v0.6.0-alpha.12
+      authbridge: ghcr.io/rossoctl/cortex/authbridge:v0.7.0-alpha.3
```
huang195: As of `v0.3.1-alpha.1` the operator chart pins these same four images to `v0.7.0-alpha.3` in its own `values.yaml` (kagenti/kagenti-operator#512), so this override block is now byte-identical to the subchart default. That means the next cortex bump has to land in two places, and if one is missed the drift is invisible — the override silently wins.

Two defensible answers: drop the block and let the subchart carry it, or keep it deliberately as this chart's own statement of what it ships. If it's the latter, worth a one-line comment saying so — otherwise someone will "clean up the duplication" later and quietly change behavior.

huang195 (approved):
Dobra robota — pins are correct and I checked them against the registries rather than the PR body. operator-chart `0.3.1-alpha.1` exists (cut 52 min before this PR), and its own `values.yaml` at tag `v0.3.1-alpha.1` really does carry `v0.7.0-alpha.3` for all four injection images, so the "matches the operator chart's own pinned defaults" claim holds exactly — that pin (kagenti/kagenti-operator#512) merged 11 minutes before the operator tag, so the ordering worked. `git grep` for the three outgoing versions finds no straggler anywhere in the charts. `operator-spiffe-bootstrap: latest → v0.8.0-alpha.1` is the good catch here.

Two things outside the diff:

**`docs/overview/5-quickstart.md:37`** still reads `git checkout v0.7.0-alpha.6`, directly under the sentence "Check out the latest stable release (recommended)". That was already stale before this PR — the `v0.7.0` GA cut on 08-17 missed it — and it is wrong twice over, since an alpha is not a stable release. The right value is **`v0.7.0`**, not this cut's alpha. The reason it keeps drifting: `check-release-pins.sh` has 7 checks and none of them look at docs, so nothing catches it. Cheap to fix here, since this is the release PR.

**The pin check does not prove the images exist.** Check 7 (`--verify-images`) is opt-in, and `ci-release-pins.yaml` invokes the script without it (lines 25 and 63), so both the green check and the PASS in the body attest to pin *syntax* only. The `v0.8.0-alpha.1` images do not exist until the tag is pushed — which is the established convention here (prior cut: pin commit `78091d26` on 07-21, tag published 07-22), no objection — but it does mean `main` HEAD is briefly uninstallable after merge. Running `scripts/check-release-pins.sh --verify-images` right after the tag lands would close that loop.

On the examples row: nothing in this repo pins an examples version — no `EXAMPLES_VERSION`, no `v0.2.0` under `charts/`, `deployments/`, or `.github/` — so there was nothing here to leave unchanged.

---

**Areas reviewed**: Helm values, Chart.yaml/Chart.lock, Docs · **Agent/IDE config (.claude/.vscode)**: none · **Commits**: 1, signed-off: yes · **CI**: 23 pass (incl. Check Release Pins), `E2E HyperShift` pending `/run-e2e`

Assisted-By: Claude Code

cwiklik:
Thanks for the thorough review — addressed both in-diff items in `9fa493d8`:

- **AuthBridge override block (values.yaml:418)** → dropped it. It duplicated the operator subchart's own default (both `v0.7.0-alpha.3` as of operator-chart `0.3.1-alpha.1`) and had been silently winning while stale at `v0.6.0-alpha.12` — exactly the drift you flagged. Single source of truth is now the subchart, pinned via the operator-chart dependency version. Rendered cortex image tags are unchanged (`v0.7.0-alpha.3`), verified with `helm template`. Left a comment where the block was so nobody re-adds it.
- **quickstart.md:37** → now `git checkout v0.7.0` (the GA), not the stale alpha. Good catch that "stable" + an alpha tag was wrong twice over.

On the two out-of-diff notes:
- **`--verify-images`**: agreed, the green check is pin-*syntax* only; `v0.8.0-alpha.1` images don't exist until the tag is pushed (the established convention). I'll run `scripts/check-release-pins.sh --verify-images` right after the tag lands to close the loop.
- **examples**: confirmed — nothing in this repo pins an examples version, so there was nothing to leave unchanged.

Assisted-By: Claude Code

PR #2403, opened 2026-08-19
Fix: derive pinnable images from values + fix yq write path in pin-release-tags.sh

### LINKED ISSUE #2400: pin-release-tags.sh: yq write path broken on yq v4.45.1 (dumps help, applies nothing)
## Problem
`scripts/pin-release-tags.sh <version>` exits 0 but **applies no changes** on yq `v4.45.1` (mikefarah). Instead of writing, the yq invocation dumps its full `--help` text to stdout and the tag/version edits never land. `--dry-run` prints the correct intended changes, masking the bug.

## Repro
```
$ yq --version   # v4.45.1
$ bash scripts/pin-release-tags.sh v0.8.0-alpha.1
...dumps yq help text...
--- exit:0 ---
$ git diff --stat   # EMPTY — nothing changed
```

## Impact
During the `v0.8.0-alpha.1` cut the script silently no-op'd; all pins had to be applied by hand. A release manager trusting the exit code would tag a release with **stale/unpinned** image tags.

## Suggested fix
- Pin/detect the yq invocation syntax to the installed major (the write form changed across yq v4 minors), or
- Assert post-conditions: after writing, re-read each key and fail loudly if it doesn't equal the requested version (turn the silent no-op into a hard error).

Found during the v0.8.0-alpha.1 cut (rossoctl/rossoctl#2399).

### LINKED ISSUE #2401: pin-release-tags.sh: managed-image list misses phoenix-oauth-secret + rossoctl-chart spiffe-idp-setup copy
## Problem
`scripts/pin-release-tags.sh` pins only a subset of the `ghcr.io/rossoctl/rossoctl/*` images. Two rossoctl-built image tags in `charts/rossoctl/values.yaml` are **not** in its managed list and are left stale:

1. `phoenixOAuthSecret.tag` — `ghcr.io/rossoctl/rossoctl/phoenix-oauth-secret`
2. `spiffeIdp.image.tag` (the copy **inside `charts/rossoctl/values.yaml`**) — `ghcr.io/rossoctl/rossoctl/spiffe-idp-setup`. The script only updates the `charts/rossoctl-deps/values.yaml` copy.

## Impact
A `0.8.x` release would ship a `0.7.x` phoenix-oauth-secret / spiffe-idp-setup — version-skewed, though not a `:latest` gate failure. Had to be pinned by hand during the v0.8.0-alpha.1 cut.

## Suggested fix
Derive the managed set from **all** `image: ghcr.io/rossoctl/rossoctl/*` occurrences across both chart values files (grep + pin the following `tag:`), rather than a hardcoded key list — so new rossoctl images are covered automatically.

Found during the v0.8.0-alpha.1 cut (rossoctl/rossoctl#2399).

cwiklik:
## Summary

Fixes two `scripts/pin-release-tags.sh` bugs surfaced during the `v0.8.0-alpha.1` cut, and removes a dead config the fix exposes.

- **#2401 — hardcoded image list drifted.** The pinner skipped `phoenix-oauth-secret` and the `charts/rossoctl/values.yaml` copy of `spiffe-idp-setup` (only the `rossoctl-deps` copy was pinned), so a 0.8.x cut shipped version-skewed images that had to be pinned by hand. Now the managed set is **derived** from every `ghcr.io/rossoctl/rossoctl/*` image in both values files — a uniform rule (replace the image string's last path segment with `tag`) handles both the flat (`image:` + sibling `tag:`) and nested (`image.repository` + `image.tag`) layouts — so new rossoctl images are covered automatically. Errors out if the derivation finds nothing.
- **#2400 — yq write applied nothing.** The write used jq-style `--arg`, which mikefarah `yq` rejects (`unknown flag: --arg`); it printed `--help` and made **no change while exiting 0**. Switched image-tag and `Chart.yaml` version/appVersion writes to `strenv()`, and added a **post-write assertion** so a future syntax drift fails loudly instead of silently no-op'ing.
- **Dead-config removal.** `phoenix-oauth-secret` has no source dir, isn't in `build.yaml`, and no template renders it — it can never be published and would fail the now-correct deriver / `--verify-images`. Dropped its `values.yaml` block and the `docs/releasing.md` row. (The `rossoctl-deps` `secretName: phoenix-oauth-secret` is a Kubernetes Secret name, not an image — left intact.)

## Test plan

- [x] `--dry-run` lists all 9 tags, **no phoenix line**
- [x] a real pin applies a **non-empty diff** (the old code produced an empty diff — that was the #2400 bug), then reverted
- [x] `scripts/check-release-pins.sh` passes
- [x] `scripts/check-release-pins.sh --verify-images` passes against the live registry (phoenix no longer checked)

Closes #2401. Fixes #2400.

_Assisted-By: Claude Code_

### DIFF
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -149,7 +149,6 @@ these include:
 | `ghcr.io/rossoctl/rossoctl/ui-oauth-secret` | `uiOAuthSecret.tag` | UI Keycloak client registration |
 | `ghcr.io/rossoctl/rossoctl/agent-oauth-secret` | `agentOAuthSecret.tag` | Agent Keycloak client registration |
 | `ghcr.io/rossoctl/rossoctl/api-oauth-secret` | `apiOAuthSecret.tag` | API Keycloak client registration |
-| `ghcr.io/rossoctl/rossoctl/phoenix-oauth-secret` | `phoenixOAuthSecret.tag` | Phoenix observability auth |
 | `quay.io/ladas/mlflow-oauth-secret` | `mlflowOAuthSecret.tag` | MLflow auth (move to `ghcr.io/rossoctl/rossoctl/mlflow-oauth-secret` once published) |
 
 Additionally, some Helm templates hardcode `:latest` for utility images
--- a/scripts/pin-release-tags.sh
+++ b/scripts/pin-release-tags.sh
@@ -162,6 +162,53 @@ else
 fi
 
 # ---------------------------------------------------------------------------
+# Surgical in-place value replacement
+#
+# `yq -i` rewrites (and reformats) the whole document — it drops blank lines and
+# re-indents sequences, which turned a 9-tag pin into ~100 lines of churn. To keep
+# the diff to only the version lines, we locate each target node's line with
+# `yq '... | line'` (read-only) and rewrite just that line's value with sed. The
+# line count never changes (same-line value swap), so line numbers stay valid
+# across successive edits within a file.
+# ---------------------------------------------------------------------------
+
+# Detect GNU vs BSD/macOS sed once: GNU accepts `--version`; BSD needs `-i ''`.
+if sed --version >/dev/null 2>&1; then
+    sed_i() { sed -i "$@"; }
+else
+    sed_i() { sed -i '' "$@"; }
+fi
+
+# pin_scalar FILE YQ_PATH VALUE — rewrite only the value of the scalar at YQ_PATH
+# to VALUE, in place, touching just that one line (no document reflow).
+#
+# yq's `line` operator is used as an anchor, but its offset is not uniform (0 for
+# top-level keys like Chart.yaml's .version, but -1 for nested keys in
+# values.yaml), so we resolve the exact line by matching YQ_PATH's trailing key
+# within a one-line window of the anchor. The post-write assertion at the call
+# site re-reads via yq and fails loudly if the substitution missed.
+pin_scalar() {
+    local file="$1" yq_path="$2" value="$3"
+    local key="${yq_path##*.}"
+    local anchor cand target=""
+    anchor=$(yq eval "$yq_path | line" "$file")
+    if [[ -z "$anchor" || "$anchor" == "0" ]]; then
+        echo "error: could not locate $yq_path in ${file#$REPO_ROOT/}" >&2
+        exit 1
+    fi
+    for cand in "$anchor" $((anchor + 1)); do
+        if sed -n "${cand}p" "$file" | grep -qE "^[[:space:]]*${key}:"; then
+            target="$cand"; break
+        fi
+    done
+    if [[ -z "$target" ]]; then
+        echo "error: could not resolve '$key:' line near $anchor in ${file#$REPO_ROOT/}" >&2
+        exit 1
+    fi
+    sed_i "${target}s#: .*#: ${value}#" "$file"
+}
+
+# ---------------------------------------------------------------------------
 # Verify images exist (optional)
 # ---------------------------------------------------------------------------
 
@@ -222,7 +269,7 @@ for entry in "${PINNABLE_IMAGES[@]}"; do
     if [[ "$DRY_RUN" == "true" ]]; then
         printf '%s[DRY]%s  %s (%s): %s → %s\n' "$YELLOW" "$NC" "$image_name" "$rel_path" "$current" "$VERSION"
     else
-        VERSION="$VERSION" yq eval "$yq_path = strenv(VERSION)" -i "$target_file"
+        pin_scalar "$target_file" "$yq_path" "$VERSION"
         got=$(yq eval "$yq_path" "$target_file")
         [[ "$got" == "$VERSION" ]] || { echo "error: pin verify failed for $yq_path in $rel_path (got '$got')" >&2; exit 1; }
         printf '%s[PIN]%s  %s (%s): %s → %s\n' "$GREEN" "$NC" "$image_name" "$rel_path" "$current" "$VERSION"
@@ -254,8 +301,11 @@ if [[ "$SKIP_CHART_VERSION" != "true" ]]; then
         printf '%s[DRY]%s  Chart.yaml version: %s → %s\n' "$YELLOW" "$NC" "$current_chart_ver" "$chart_ver"
         printf '%s[DRY]%s  Chart.yaml appVersion: → %s\n' "$YELLOW" "$NC" "$chart_ver"
     else
-        chart_ver="$chart_ver" yq eval '.version = strenv(chart_ver)' -i "$ROSSOCTL_CHART"
-        chart_ver="$chart_ver" yq eval '.appVersion = strenv(chart_ver)' -i "$ROSSOCTL_CHART"
+        pin_scalar "$ROSSOCTL_CHART" '.version' "$chart_ver"
+        pin_scalar "$ROSSOCTL_CHART" '.appVersion' "$chart_ver"
+        got_v=$(yq eval '.version' "$ROSSOCTL_CHART")
+        got_av=$(yq eval '.appVersion' "$ROSSOCTL_CHART")
+        [[ "$got_v" == "$chart_ver" && "$got_av" == "$chart_ver" ]] || { echo "error: Chart.yaml version pin verify failed (version='$got_v' appVersion='$got_av')" >&2; exit 1; }
         printf '%s[PIN]%s  Chart.yaml version + appVersion → %s\n' "$GREEN" "$NC" "$chart_ver"
     fi
 fi
--- a/scripts/pin-release-tags.sh
+++ b/scripts/pin-release-tags.sh
@@ -134,6 +134,13 @@ REGISTRY="ghcr.io/rossoctl/rossoctl"
 #   flat:    image: <ref>            + tag: <ver>   -> <parent>.tag
 #   nested:  image: { repository: <ref>, tag: <ver> } -> <parent>.image.tag
 # Format of each derived entry: "values-file|yq-path|image-name"
+#
+# By design, any layout OTHER than these two aborts the run (fail-loud, never a
+# silent skip): an image carrying an inline tag with no sibling `tag:` key yields
+# a `.tag` path that doesn't exist and the pin loop exits with an error; likewise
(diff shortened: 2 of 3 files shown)


cwiklik:
Verified this end-to-end in a throwaway worktree — the core fixes are correct and complete:

- **#2400** write path fixed (`--arg` → `strenv()` at all 3 sites; post-write assertion added) — a real `pin` now actually applies (the old code produced an empty diff).
- **#2401** image set is derived from the values files — dry-run lists the right 9 tags and correctly **excludes** the now-removed `phoenix-oauth-secret`.
- Dead `phoenixOAuthSecret` block + `docs/releasing.md` row removed; no other content stripped.

**One tweak requested before merge — keep the write surgical (don't reformat the files).**

`yq -i` reflows the whole file when it writes: it drops blank lines between sections and re-indents YAML sequences (e.g. `parentRefs:` list `2-space → 4-space`). It does *not* strip comments, and the YAML stays valid — but a single real pin came out as **~110 deletions / 50 insertions across the values files + a 13-line Chart.yaml diff**, for what is semantically **9 tag changes**. Every future release-pin commit would carry that cosmetic churn, burying the actual version bumps and making review (and spotting an accidental change) much harder. For comparison, the hand-done v0.8.0-alpha.1 pin was a clean 21/21 touching only tag lines.

Ask: **only the version lines should change** — roughly one line per image, plus `version:`/`appVersion:` in Chart.yaml — with no blank-line or indentation churn.

Suggested approach (keeps everything you've built): keep `yq` for the read-only **derivation**, but do the **substitution** surgically instead of `yq -i`. E.g. have the deriver also emit each tag's line number (`yq`'s `line` operator), then edit exactly that line:
```bash
# per derived entry: known line number $ln in $target_file
sed -i '' "${ln}s#: .*#: ${VERSION}#" "$target_file"   # BSD/macOS sed; drop the '' on GNU/CI
```
Same idea for `version:` / `appVersion:` in Chart.yaml. The post-write assertion you added still guards correctness.

**Acceptance check:** after `bash scripts/pin-release-tags.sh <ver>`, `git diff charts/` should show only the version-number lines changed — no reflowed blank lines or re-indented lists.

Assisted-By: Claude Code

scripts/pin-release-tags.sh
```diff
@@ -154,6 +161,53 @@ else
...
+    for cand in "$anchor" $((anchor + 1)); do
+        if sed -n "${cand}p" "$file" | grep -qE "^[[:space:]]*${key}:"; then
+            target="$cand"; break
+        fi
+    done
+    if [[ -z "$target" ]]; then
+        echo "error: could not resolve '$key:' line near $anchor in ${file#$REPO_ROOT/}" >&2
+        exit 1
+    fi
+    sed_i "${target}s#: .*#: ${value}#" "$file"
```
huang195: `s#: .*#: ${value}#` replaces everything after the first `: `, so it also eats a trailing comment and any existing quoting.

Concrete case in the tree today: `charts/rossoctl/Chart.yaml` has `appVersion: "0.8.0-alpha.1"` (quoted). The next pin writes `appVersion: 0.9.0` — unquoted. Valid YAML, Helm accepts it, but it is an unrelated line change in a release commit and drifts from the usual quoted-appVersion convention.

Same mechanism drops comments: `tag: v0.8.0  # keep in sync with build.yaml` → `tag: v0.9.0`. I ran the script to confirm both.

The new post-write assertion can't catch either — it compares the *parsed* value, which is correct in both cases. So this is the one silent-write path the PR leaves open.

Keeping the tail fixes it, roughly:

```bash
sed_i -E "${target}s#^([[:space:]]*[^:]+:)[[:space:]]*[^#]*(#.*)?\$#\1 ${value}\2#" "$file"
```

Not blocking — no `tag:` line has a trailing comment right now, and the quote loss is cosmetic.

scripts/pin-release-tags.sh
```diff
@@ -129,20 +123,33 @@ fi
...
-)
+# For every string value matching the registry, we pin its sibling `tag`. This
+# handles both layouts in the values files by replacing the last path segment
+# with "tag":
+#   flat:    image: <ref>            + tag: <ver>   -> <parent>.tag
+#   nested:  image: { repository: <ref>, tag: <ver> } -> <parent>.image.tag
+# Format of each derived entry: "values-file|yq-path|image-name"
+derive_expr='.. | select((tag == "!!str") and (. == "'"$REGISTRY"'/*"))
+    | (path | .[-1] = "tag" | join("."))
+    + "|" + (. | sub("^'"$REGISTRY"'/"; "") | sub(":.*$"; ""))'
```
huang195, line 146: Two brittle spots here. Both fail loud, which already beats the old silent skip — I'd just make the intent explicit in the comment.

1. An image with an inline tag and no sibling `tag:` key aborts the run:

   ```yaml
   newThing:
     image: ghcr.io/rossoctl/rossoctl/new-thing:v0.8.0
   ```
   → `error: yq path .newThing.tag not found`, exit 1. Because the pin loop writes as it goes, an image added late in `values.yaml` leaves the tree half-pinned. The `sub(":.*$"; "")` on this line says inline refs were anticipated, but the derived pin path can't handle them.

2. `join(".")` isn't yq-path-safe. A values key containing a dot yields `.my.thing.tag` and aborts. Unlikely in these charts, but it's the same shape of failure.

A one-liner saying "layouts other than these two abort by design" would make it clear that's intended rather than an oversight.

Separately: `spiffe-idp-setup` is derived from both values files, so `--verify-images` runs `docker manifest inspect` on it twice and prints a duplicate `[OK]`. Cosmetic — dedup on image name if you care.

docs/releasing.md
```diff
@@ -149,7 +149,6 @@ these include:
 | `ghcr.io/rossoctl/rossoctl/ui-oauth-secret` | `uiOAuthSecret.tag` | UI Keycloak client registration |
 | `ghcr.io/rossoctl/rossoctl/agent-oauth-secret` | `agentOAuthSecret.tag` | Agent Keycloak client registration |
 | `ghcr.io/rossoctl/rossoctl/api-oauth-secret` | `apiOAuthSecret.tag` | API Keycloak client registration |
-| `ghcr.io/rossoctl/rossoctl/phoenix-oauth-secret` | `phoenixOAuthSecret.tag` | Phoenix observability auth |
 | `quay.io/ladas/mlflow-oauth-secret` | `mlflowOAuthSecret.tag` | MLflow auth (move to `ghcr.io/rossoctl/rossoctl/mlflow-oauth-secret` once published) |
```
huang195, line 152: While you're in this file — the docs still carry **two** hardcoded image tables, which is the exact drift you just removed from the script.

This table is prefaced "As of v0.5.0" and lists 4 of the 9 images the deriver now finds. Its mlflow row says `quay.io/ladas/mlflow-oauth-secret`, but `charts/rossoctl/values.yaml:468` has `ghcr.io/rossoctl/rossoctl/mlflow-oauth-secret`. Worth noting a `quay.io/*` image would not be pinned by the new deriver at all, since it only matches `$REGISTRY/*`.

The second table, "Images pinned by the script" (~line 753), lists 7 rows and is missing `operatorSpiffeBootstrap.tag` and the `charts/rossoctl` copy of `spiffeIdp.image.tag` — that second one is the other half of #2401.

Suggest replacing both with a pointer to `--dry-run` for the current set. Same argument as the script change: a hand-maintained list will drift again.

huang195 (approved):
Dobra robota — deriving the set from the values files is the right fix, and the surgical write is a real improvement over `yq -i`.

I reconstructed the tree from the PR head and actually ran the script rather than only reading the diff:

- `--dry-run` lists **9** tags (8 in `charts/rossoctl/values.yaml`, 1 in `rossoctl-deps`) — no phoenix line.
- A real pin produces exactly **9 changed lines**, zero yq reflow. That confirms both #2400 (the old `--arg` path wrote nothing) and the reflow motivation for `pin_scalar`.
- Re-running the same version → all `[SKIP]`, no writes. Idempotent.
- The `yq ... | line` offset is `-1` for nested keys and `0` for top-level, exactly as your comment says, so the 2-line window holds. A missing path returns `0`, which the guard catches correctly.

Dead-config removal checks out: `phoenixOAuthSecret` appears only in `charts/rossoctl/values.yaml` and `docs/releasing.md` — no template, no `build.yaml` entry. The `rossoctl-deps` `secretName: phoenix-oauth-secret` is correctly left alone.

Three non-blocking things in the comments. The one I would fix before the next cut is `pin_scalar`'s sed — `Chart.yaml` `appVersion` is quoted today, so the next pin silently unquotes it.

**Areas reviewed**: Shell, Helm values, Docs
**Commits**: 2 commits, all signed-off: yes (DCO green)
**CI status**: passing — `Deploy & Test` and `E2E HyperShift` still pending (E2E needs `/run-e2e`)

Nit: title is ~82 chars, over the 72 guideline. `verify-pr-title` passes anyway.

_Assisted-By: Claude Code_

cwiklik:
Thanks @huang195 — pushed `89b846a4` addressing the two write-path comments:

**Comment #1 (sed eats quoting + trailing comments):** fixed, and a bit beyond the suggested regex. `pin_scalar` now rewrites *only the scalar value*, preserving:
- **existing quoting** — verified `Chart.yaml` `appVersion: "0.8.0-alpha.1"` → `appVersion: "0.9.9-test"` (stays quoted); your regex kept the comment but would still have unquoted it.
- **trailing comments** — a `tag: v0.8.0  # keep in sync` line keeps its comment.

The post-write assertion is retained. Re-verified: a real pin is still 11/11 churn, only version-number lines change.

**Comment #2 (derivation brittleness):** added a comment at the deriver stating that layouts other than the two handled — an inline tag with no sibling `tag:`, or a values key containing a literal `.` — **abort by design** (fail-loud, never a silent half-pin), and to extend the deriver if one is introduced. Left the `--verify-images` double-`[OK]` on `spiffe-idp-setup` as-is (cosmetic).

**Comment #3 (docs/releasing.md's two drifting tables + the mlflow `quay.io`/`ghcr` mismatch):** good catch — deferring to a follow-up rather than widen this PR, since it's a docs-content cleanup (and the `quay.io` mlflow row is a separate correctness issue: the deriver only matches `$REGISTRY/*`, so a `quay.io` image wouldn't be pinned at all).

Nit (title length): noted; `verify-pr-title` passes so leaving it.

_Assisted-By: Claude Code_

PR #330, opened 2025-10-23
📖 update OpenShift install doc

pdettori:
## Summary

This PR updates the OpenShift installation documentation to improve accuracy and usability. The changes remove outdated limitations, enhance installation instructions with better automation, and provide clearer guidance for accessing the UI and configuring certificate trust.

Key changes:

- Removed several outdated limitation notices that are no longer applicable
- Enhanced helm installation instructions with automated tag retrieval command
- Expanded UI access section with detailed certificate handling and MCP Inspector proxy configuration

PR #2404, opened 2026-08-20
test(e2e): cover inbound mTLS and the ambient/HBONE path in transparent inbound

huang195:
## What

Adds two modules to the transparent-inbound e2e suite, closing **both open halves
of rossoctl/cortex#780**: `test_transparent_inbound_mtls.py` for inbound mTLS, and
`test_transparent_inbound_ambient.py` for the Istio ambient / HBONE delivery path.
Suite goes from 13 tests to 15, 0 skipped locally.

## Why

#330 shipped the transparent inbound listener reusing the reverse proxy's mTLS
posture via `WrapListener`, and the sidecar logged `mtls=true` — but no TLS
client had ever been driven at `:8083`. The path was exercised only by its own
startup message.

The untested part is specifically the **combination**: `SO_ORIGINAL_DST` recovery
happens on the raw connection, *before* `tlssniff` peeks the first byte. Had
those two interfered, the logs would still have looked correct.

## The three cases

| # | Case | Expected | What it proves |
|---|---|---|---|
| A | `strict` + plaintext | connection closed, `http_code=000` | `tlssniff` is genuinely in the transparent path — the same request is a **401** under permissive, so a closed connection can only come from the sniffer |
| B | `strict` + TLS, no client cert | TLS alert `certificate required` | server-side TLS really terminates on the recovered connection (to demand a client cert it had to send its own first) |
| C | `strict` + TLS, **valid SVID** | **`http_code=401`** | the whole chain in one request: REDIRECT → `SO_ORIGINAL_DST` → `tlssniff` → `tls.Server` (client cert verified) → `jwt-validation` |

**C is the one that matters.** It lands on 401 rather than 200 deliberately:
authentication of the *transport* must not be mistaken for authorization of the
*request*. **B doubles as C's control** — the same endpoint refuses a caller with
no certificate, so C cannot be explained by verification being lax.

## Notes on approach

- **Borrows the agent sidecar's own SVID** rather than minting a second SPIRE
  registration. Any identity in the trust domain satisfies
  `RequireAndVerifyClientCert`, and the assertion is that the handshake completes
  at all, not which identity completes it. Skips when SPIRE supplies no SVIDs.
- `--resolve` so SNI/Host match the SVID's DNS SANs while still dialing the **pod
  IP** directly — via the Service the request would be indistinguishable from the
  Service-routed case this module is not testing.
- **`conftest` gains a `_baseline` registry.** Fixtures now restore the namespace
  to what it looked like *before the suite touched it*, not to whatever they
  happened to observe. Without this, a second module flipping the same shared
  namespace makes restoration order-dependent.
- Module-scoped, not session-scoped: it flips the same namespace the permissive
  fixture uses, so it must set up and tear down within its own module.

## Verification

Run on kind `rossoctl` against operator + authbridge + proxy-init images built
from **merged main** (so the shipped code was under test):

```
13 passed in 89.56s   # 10 existing + 3 new, 0 skipped
```

Both namespaces confirmed restored afterwards (no `inboundInterception`/`mtls`
keys left behind, no leftover test objects).

### The suite was inert in CI — root-caused and fixed

Worth stating plainly, because the green checkmark on the first two commits should
not be over-read: **all 13 tests were skipping on the shared Kind e2e run**,
including the 10 that merged in #2393. Signature was three 3-minute fixture stalls
(exactly `READY_TIMEOUT`), each followed by an instant batch of skips, because the
per-agent Keycloak client-credentials Secret was never created.

My first read of that was wrong. I called it the "intermittent Kind client
registration" that `e2e-kind.yaml` works around, and the earlier skip message
blamed a missing `k8s.keycloak.org` CRD. **Neither is the cause.** It is
deterministic, and the CRD is not in this code path at all — the controller uses
Keycloak's admin REST API directly.

The actual cause: with SPIRE enabled, the operator's ClientRegistration controller
derives the Keycloak client ID from the **Deployment pod template's**
`serviceAccountName`, and refuses on `default`:

```
"msg":"cannot resolve Keycloak client id yet"
"reason":"SPIRE enabled: set spec.template.spec.serviceAccountName to a dedicated
          ServiceAccount (not default) on the workload for a stable SPIFFE client ID"
```

It then requeues every 30s forever at `Info` level. The fixtures' Deployment set no
`serviceAccountName`, so they hit this every single run. The platform's own
`74-deploy-weather-agent.sh` creates a dedicated SA for exactly this reason, which
is why weather-service works in CI and these agents did not.

Proven both directions on Kind: fixture shape → `cannot resolve` + zero Secrets +
pod `Pending`; add `serviceAccountName` → Secret appears in 20s **with the exact
hash the webhook had already pre-populated**, pod Ready 10s later.

The third commit adds the SA to the manifest and rewrites the skip message and
README to point at the controller log, which prints the real reason verbatim.

**This resolves the gating question** — better than either option I floated
earlier. No stubbing, no waiting on a platform fix:

```
13 passed, 0 skipped, no TI_STUB_CREDENTIALS   # real controller-issued credentials
```

### A latent platform inconsistency this surfaced (not a defect, not fixed here)

The AuthBridge webhook creates a per-agent ServiceAccount, sets it on the **pod**,
and pre-populates the annotation naming the credentials Secret — but never touches
the **Deployment template** the controller reads, so the webhook's fixup is
invisible to it.

That is latent rather than broken. Every path that creates a workload for real
already sets `serviceAccountName` — the backend's agent manifests
(`agents_manifests.py:379,581,723,832`), tools (`tools.py:1330,1492`),
`simulation_manifests.py:253`, `agents_migration.py:487` (comment: *"Ensure
serviceAccountName is set so the webhook's SPIFFE identity…"*), and the
`7x-deploy-*` scripts. Only a hand-written Deployment reaches the refusal, which
is why this fixture did and nothing else does. Nor is anything regressed: the
refusal has been present since the commit introducing that controller
(`9637c03`, 2026-03-26) and `spire.enabled` has defaulted true since 2025-11-06.

I initially filed this as operator#513 and generalized it to "any workload";
that was wrong and the issue is closed with a retraction.

One thing there does still seem worth fixing upstream, unrelated to the SA: those
failures surface only as `logger.Info` plus `RequeueAfter: 30s` forever — no Event,
no status condition — while the pod sits on `FailedMount`. That applies to every
reason client-ID resolution can fail, not just this one.

Second, minor: webhook-created SAs carry no `ownerReference`, so nothing
garbage-collects them. Eight had accumulated across `team1`/`team2` during
development. Teardown now deletes ours.

## The ambient / HBONE half

The suite could only ever exercise one of the two capture rules. Without a
ztunnel a pod-to-pod dial arrives as ordinary TCP and is caught in
`nat PREROUTING` by `AB_INBOUND`. Under ambient it never reaches `PREROUTING`:
ztunnel terminates HBONE on `:15008` and re-originates a local connection, caught
instead by a mark-based DNAT at the head of `AB_REDIRECT`, ahead of that chain's
ztunnel-mark `RETURN`. Two paths, two rules, **and the same 401 either way** —
which is the hazard #780 names, since a PREROUTING-only implementation looks
correct while passing every mesh-delivered request through unvalidated.

So the new module asserts the **delivery path** as well as the status, read from
ztunnel's own access log. Without that it silently becomes a duplicate of
`test_direct_pod_dial_is_validated` the moment ambient stops being engaged.

That the DNAT counter previously read 0 was a config gate, not missing
infrastructure: `charts/rossoctl-deps/templates/istio-operand.yaml` wraps the
`Istio`, `IstioCNI` and `ZTunnel` operands in
`{{- if and .Values.components.istio.enabled .Values.openshift }}`, and
`deployments/envs/dev_values*.yaml` set `openshift: false`. Kind has no ambient
data plane by construction.

### Verified

Ambient installed by hand on Kind (istiod 1.28.0 `profile=ambient` with
`PILOT_ENABLE_AMBIENT=true`, istio-cni, ztunnel), agent in `team1`, probe in
`team2`, both `istio.io/dataplane-mode=ambient`:

| Request from `team2` | Result | `AB_INBOUND` REDIRECT | ambient DNAT | `:9091` RETURN |
|---|---|---|---|---|
| baseline | — | `[2:120]` | `[0:0]` | `[1:60]` |
| Service `:8000` | **401** | unchanged | **+1** | — |
| direct pod dial `:8000` | **401** | unchanged | **+1** | — |
| `:9091/healthz` | **200** | unchanged | unchanged | **+1** |

The ambient DNAT carried every mesh-delivered request; `AB_INBOUND` never fired
for one. The exemption is confirmed *positively* — the dport-9091 rule's own
counter incremented — not by absence. The plain `--mark 0x539 -j RETURN` stayed
flat having taken 6 *outbound* HBONE packets earlier, live confirmation that
`--dst-type LOCAL` on the DNAT is load-bearing. Corroborated at three layers in
the same millisecond: ztunnel logged `dst.hbone_addr=<podIP>:8000` with SPIFFE
identities both ends, and authbridge logged
`plugin=jwt-validation status=401 reason="missing Authorization header"`.

**Negative control**, since a test that cannot fail proves nothing: deleting that
one DNAT rule from the live pod's netns makes the mesh-delivered request return
**200** — the app answering unvalidated — and the module fails naming that cause,
while the health-port test keeps passing. Recreating the pod reprograms the rule
from proxy-init and both pass again.

Rule ordering and counters are not asserted in the test: not reachable from a test
pod, and unnecessary — every way the rule can regress (deleted, ordered after the
mark `RETURN`, matching too narrowly) ends in that same 200. The counter evidence
is recorded on #780.

### One thing to flag

The ambient module is deliberately **not** `kind_only`, unlike the rest of this
suite. Ambient ships on OpenShift only, so a Kind-only ambient test could never
run where ambient is the shipped configuration; it self-gates instead on a ready
ztunnel (found by label across namespaces — the chart uses `istio-ztunnel`, a
community install defaults to `istio-system`) plus `dataplane-mode=ambient` on
both namespaces, naming whichever is missing.

Consequence: **this becomes the first module in the suite to run on OpenShift**,
so its first run there may surface pre-existing suite or platform issues rather
than ambient ones. Narrowing it to `kind_only` would make it skip everywhere the
feature actually ships, which is why it is not — but that is a call worth making
explicitly rather than by default.

Also worth noting the data plane used here is community Helm charts, not the
sail-operator operands the platform ships. The ztunnel namespace does not enter
the iptables tuple under test, but confirming on OpenShift is what the unmarked
test now does automatically.

## Follow-up, not in this PR

The mTLS coverage lives in this repo while the code under test lives in cortex, so
a cortex regression would not fail cortex CI. The narrow gap there is a Go test
for the assembled `NewInboundListener` → `WrapListener` with mTLS **on** — the two
weaker cases here already duplicate existing `tlssniff` and `origdst` unit tests.

Assisted-By: Claude Code

### DIFF
--- a/rossoctl/tests/e2e/transparent_inbound/README.md
+++ b/rossoctl/tests/e2e/transparent_inbound/README.md
@@ -69,12 +69,28 @@ feature.
 - `proxy.allowedInboundInterception` must permit `transparent` (the default).
 - Linux nodes: the feature is iptables-based. Skipped elsewhere.
 - **Working Keycloak client registration.** The injected sidecar mounts a
-  per-agent credentials Secret created by the operator. Where that path is broken
-  — e.g. the `k8s.keycloak.org/v2alpha1` CRD is absent, as on a cluster installed
-  with the community Keycloak provider — every injected pod stays `Pending` on
-  `FailedMount`, including default reverse-proxy ones. The suite **skips with that
-  cause named** rather than failing and pointing suspicion at the interception
-  rules.
+  per-agent credentials Secret the operator registers *asynchronously*. Where that
+  path does not complete, every injected pod stays `Pending` on `FailedMount` —
+  including default reverse-proxy ones — so it says nothing about interception. The
+  suite **skips with the cause named** rather than failing and pointing suspicion
+  at the interception rules. Two causes are known and they need distinguishing:
+
+  1. **On Kind, registration is intermittent.** `.github/workflows/e2e-kind.yaml`
+     carries an explicit wait loop for these Secrets and marks the token-exchange
+     suite `continue-on-error` for exactly this reason.
+  2. **The `k8s.keycloak.org/v2alpha1` CRD is absent**, as on community-provider
+     installs. Then *no* agent in the cluster ever gets a Secret.
+
+  Check `kubectl get crd | grep k8s.keycloak.org`, and whether other agents have
+  credentials Secrets, before suspecting this feature.
+
+  > **This suite is currently inert in CI.** All of its tests skip on the shared
+  > Kind e2e run via cause (1) above — verified on the run for #2404. That has
+  > been true since the suite merged, so CI passing is **not** evidence these
+  > tests ran. Until it is resolved, the meaningful signal is a local run with
+  > `TI_STUB_CREDENTIALS=1`. Making it gate in CI means either fixing registration
+  > or setting `TI_STUB_CREDENTIALS=1` in the runner — the latter reverses the
+  > opt-in reasoning below, so it is a deliberate call, not a default.
 
   To test the boundary anyway on such a cluster, set `TI_STUB_CREDENTIALS=1` to
   create a placeholder Secret. This is opt-in on purpose: stubbing by default
--- a/rossoctl/tests/e2e/transparent_inbound/README.md
+++ b/rossoctl/tests/e2e/transparent_inbound/README.md
@@ -69,42 +69,47 @@ feature.
 - `proxy.allowedInboundInterception` must permit `transparent` (the default).
 - Linux nodes: the feature is iptables-based. Skipped elsewhere.
 - **Working Keycloak client registration.** The injected sidecar mounts a
-  per-agent credentials Secret the operator registers *asynchronously*. Where that
-  path does not complete, every injected pod stays `Pending` on `FailedMount` —
-  including default reverse-proxy ones — so it says nothing about interception. The
-  suite **skips with the cause named** rather than failing and pointing suspicion
-  at the interception rules. Two causes are known and they need distinguishing:
+  per-agent credentials Secret the operator's ClientRegistration controller creates
+  *asynchronously*. Where that path does not complete, every injected pod stays
+  `Pending` on `FailedMount` — including default reverse-proxy ones — so it says
+  nothing about interception. The suite **skips with the cause named** rather than
+  failing and pointing suspicion at the interception rules.
 
-  1. **On Kind, registration is intermittent.** `.github/workflows/e2e-kind.yaml`
-     carries an explicit wait loop for these Secrets and marks the token-exchange
-     suite `continue-on-error` for exactly this reason.
-  2. **The `k8s.keycloak.org/v2alpha1` CRD is absent**, as on community-provider
-     installs. Then *no* agent in the cluster ever gets a Secret.
+  The controller logs its exact reason and then requeues every 30s, so read the
+  log rather than guessing:
 
-  Check `kubectl get crd | grep k8s.keycloak.org`, and whether other agents have
-  credentials Secrets, before suspecting this feature.
+  ```sh
+  kubectl logs -n rossoctl-system deployment/rossoctl-controller-manager \
+    | grep -E 'cannot resolve|waiting for|registration failed|skipping'
+  ```
 
-  > **This suite is currently inert in CI.** All of its tests skip on the shared
-  > Kind e2e run via cause (1) above — verified on the run for #2404. That has
-  > been true since the suite merged, so CI passing is **not** evidence these
-  > tests ran. Until it is resolved, the meaningful signal is a local run with
-  > `TI_STUB_CREDENTIALS=1`. Making it gate in CI means either fixing registration
-  > or setting `TI_STUB_CREDENTIALS=1` in the runner — the latter reverses the
-  > opt-in reasoning below, so it is a deliberate call, not a default.
+  The reason that bit this suite, and the reason the fixtures create a dedicated
+  ServiceAccount, is documented at the manifest in `conftest.py`: with SPIRE
+  enabled the controller derives the Keycloak client ID from the **Deployment pod
+  template's** `serviceAccountName` and refuses on `default`. The AuthBridge
+  webhook does create a per-agent SA and set it on the **pod**, but never on the
+  template — so the webhook's fixup is invisible to the controller, and the pod
+  ends up mounting a Secret that the controller has already declined to create.
+  That split is a platform defect, tracked separately; setting the SA ourselves is
+  the half we control. Note the `k8s.keycloak.org` CRD is **not** in this path —
+  the controller uses Keycloak's admin REST API directly.
 
-  To test the boundary anyway on such a cluster, set `TI_STUB_CREDENTIALS=1` to
-  create a placeholder Secret. This is opt-in on purpose: stubbing by default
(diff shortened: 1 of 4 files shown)


cwiklik (approved):
### Summary

Strong extension of the transparent-inbound e2e suite covering the mTLS chain (SO_ORIGINAL_DST → tlssniff → tls.Server → jwt-validation) — the mTLS half of cortex#780. Approving; test-only, and the load-bearing assertions are precise.

**Verified strict:**
- **A** (`test_strict_rejects_plaintext`) triangulates well — asserts `code=200 not in`, `code=401 not in`, **and** `code=000 in`, so a closed connection is distinguished from both a served response and a pipeline-reached 401. Proves `tlssniff` is genuinely in the path.
- **C** (`test_strict_completes_mtls_and_still_enforces`) asserts **exactly** `code=401` — the right call (transport auth ≠ request authz), not a loose "not 200".
- No hardcoded credentials: the client SVID (cert/key/bundle) is fetched **at runtime** from the live sidecar and staged as a Secret. TLS verification is correctly *enabled* where C's claim depends on it (`-k` appears only in B, where skipping server-cert verification doesn't undermine what B tests). The new dedicated ServiceAccount is added to the manifest and torn down — and it actually reduces false skips by removing the `default`-SA refusal cause the prior suite hit.

**Fast-follow (should-fix, not blocking a test-only PR):**
1. The new `svid_secret` fixture (`conftest.py`) `cat`s `/opt/svid*.pem` with `check=False` and skips on empty output — but a kubectl-exec failure (renamed container, moved path, transient error) is indistinguishable from "SPIRE genuinely supplies no SVID." That gates **test C, the one that matters**, so an infra regression would silently green the security-critical case. Same skip-safety class as #2393 — please distinguish exec-failure (check the exit/stderr) from legitimate absence before skipping. A `kind_only` marker + the CI cluster having SPIRE is the only thing keeping C armed today; nothing in the test enforces it.

**Minor:**
2. Test B's assertion `"certificate required" in out.lower() or "alert" in out.lower()` accepts *any* TLS alert, which is looser than the "certificate required" behavior it claims (and than its role as C's control). Consider dropping the `or "alert"`.
3. `svid_secret` writes the SVID **private key** to a tempfile that's never unlinked (only the in-cluster Secret is torn down). Unlink after `--from-file` — SVID keys are short-lived, but leaving key material on the runner is avoidable.

**Author**: huang195 (MEMBER — maintainer)
**Areas reviewed**: Python e2e tests, conftest fixtures, mTLS/SVID handling, skip-safety, secrets
**CI**: 22 green; Deploy & Test pending; E2E HyperShift gated on maintainer `/run-e2e`.

PR #2447, opened 2026-08-20
Feat: report attached contexts in agent detail responses

moonlight16:
## Summary

Expose context attachments through `GET /api/v1/agents/{namespace}/{name}`.

When Rosso creates a StatefulSet or Sandbox agent with context attachments, it now preserves the resolved attachment metadata on the workload and returns it as part of the agent detail response.

```json
"contexts": [
  {
    "name": "research",
    "type": "workspace",
    "mountPath": "/workspace",
    "readOnly": false,
    "claimName": "context-research"
  }
]
```

## Why

Agent creation already accepts context attachments, but callers could not subsequently determine which named contexts were attached to an agent or where they were mounted.

## Behavior

- Supports StatefulSet and Sandbox workloads.
- Covers direct-image and source-build/finalization paths.
- Omits `contexts` for agents created without context attachments.
- Safely ignores missing or malformed legacy attachment metadata.
- Rejects incomplete Context Service responses that omit the PVC claim or context type.
- Returns the attachment declaration recorded when Rosso created the workload; it is not continuously reconciled with manual Kubernetes changes.

Existing agents are not automatically backfilled.

## Validation

- 21 focused backend tests pass.
- Live-tested StatefulSet and Sandbox creation in `team2`.
- Verified context name, type, claim, mount path, and access in agent GET responses.
- Verified agents without contexts remain compatible.
- Verified nonexistent contexts fail creation cleanly.
- Verified temporary test resources were removed.

### DIFF
--- a/rossoctl/backend/README.md
+++ b/rossoctl/backend/README.md
@@ -77,7 +77,9 @@ Once running, access the OpenAPI documentation:
 
 ### Agents
 - `GET /api/v1/agents` - List all agents across namespaces
-- `GET /api/v1/agents/{namespace}/{name}` - Get specific agent details
+- `GET /api/v1/agents/{namespace}/{name}` - Get specific agent details. When present,
+  `contexts` reports the attachments declared when Rosso created the workload; it is a
+  metadata snapshot and is not continuously reconciled with manual Kubernetes changes.
 - `GET /api/v1/agents/{namespace}/{name}/route-status` - Check HTTPRoute/Route status for agent
 - `DELETE /api/v1/agents/{namespace}/{name}` - Delete agent
 - `POST /api/v1/agents` - Create new agent (supports deployment from image or source code via Shipwright)
--- a/rossoctl/backend/app/routers/agents.py
+++ b/rossoctl/backend/app/routers/agents.py
@@ -22,6 +22,7 @@ Names those submodules define are re-exported here (see ``__all__``) because
 other modules and the test suite import them from ``app.routers.agents``.
 """
 
+import json
 import logging
 from typing import Any
 
@@ -90,6 +91,7 @@ from app.routers.agents_env import (
 # the patch effective -- do not promote it to a module-level import there.
 from app.routers.agents_finalize import finalize_router, finalize_shipwright_build
 from app.routers.agents_manifests import (
+    CONTEXTS_ANNOTATION,
     build_container_resources,
     _agentruntime_supported_workload,
     _build_agentruntime_manifest,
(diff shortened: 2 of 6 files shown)


cwiklik (approved):
### Summary

Clean feature — adds a `contexts` field to `GET /agents/{ns}/{name}`, snapshotting resolved attachment metadata on the workload via a single `rossoctl.io/contexts` annotation and reading it back. Approving.

**Verified:**
- **Multi-path parity (§3.10) holds.** One key (`CONTEXTS_ANNOTATION = "rossoctl.io/contexts"`), one shape `{name,type,mountPath,readOnly,claimName}`; `_record_contexts` is called at all create+finalize sites (deployment/statefulset/job/sandbox), and the read side is workload-type-agnostic at `.metadata.annotations` — so both StatefulSet and Sandbox surface through the same path. No resolve-but-forget-to-record path.
- **Robustness matches the body's claims.** The read-side `json.loads` is guarded (`except (TypeError, json.JSONDecodeError)` + warning), so a malformed/hand-edited annotation can't 500 the agent-detail endpoint; missing annotation → field omitted. Incomplete Context Service responses are rejected strictly (502 on missing PVC claim / missing type, *before* any partial record is persisted).
- **Tenancy safe.** `claimName` comes from `resolve_context(request.namespace, name)` scoped to the tenant's own namespace — no caller-supplied namespace override, no cross-tenant claim leaked into the response.

**Non-blocking:**
1. Two behaviors the PR body advertises aren't tested: (a) **malformed-annotation tolerance** — add a test that a bad JSON annotation returns GET 200 with `contexts` omitted (exercises the `except`); (b) the **Sandbox** GET read path (only StatefulSet GET is asserted), plus a write-side assertion that `_record_contexts` actually populates `metadata.annotations[CONTEXTS_ANNOTATION]` — right now write→read is bound only through a shared string literal.
2. Nits: `CONTEXTS_ANNOTATION`/`_record_contexts` are inserted into the middle of the import block in `agents_manifests.py` (move below the imports); the GET-reporting test hardcodes `"rossoctl.io/contexts"` rather than importing the constant, so a future rename wouldn't be caught.

The `contexts` field is additive and omitted when there are no attachments — consistent with the already-shipped context-attachments feature.

**Areas reviewed**: Python (routers/manifests), multi-path parity, deserialization robustness, tenancy, tests
**CI**: 23 green; E2E HyperShift gated on maintainer `/run-e2e`.

moonlight16:
Addressed all review suggestions in b0e88073:

- added malformed-annotation GET coverage
- added Sandbox GET coverage and write-side annotation assertions
- moved the annotation constant/helper below the import block
- imported the annotation constant in tests instead of hardcoding it

Validation: `23 passed` in `tests/test_contexts.py`, plus Ruff check/format.

Issue #2398, opened 2026-08-19
Broken link in README.md: https://github.com/rossoctl/rossoctl/blob/main/docs/components.md


abigailgold:
### Describe the bug

In the Documentation section of the readme there is a table with links. The link for Component Details leads to a non-existing md file (https://github.com/rossoctl/rossoctl/blob/main/docs/components.md)

### Steps To Reproduce

1. Go to rossoctl main readme
2. Scroll down to Documentation
3. Press Component Details link
4. Link is broken (no such md file).

### Expected Behavior

Link opens to md file

PR #2402, opened 2026-08-19
Fix: Correct broken doc links in README Documentation table (#2398)

cwiklik:
## Summary

The docs were reorganized into `getting-started/`, `concepts/`, and `users-guides/` subdirectories, but five links in the README **Documentation** table still pointed at the old flat `docs/` paths and 404:

| Table row | Old (broken) | Corrected |
|---|---|---|
| Components | `docs/components.md` | `docs/concepts/components.md` |
| Import Your Own Agent | `docs/new-agent.md` | `docs/getting-started/new-agent.md` |
| Import Your Own Tool | `docs/new-tool.md` | `docs/getting-started/new-tool.md` |
| Skills Configuration & Usage | `docs/skills.md` | `docs/concepts/skills.md` |
| Troubleshooting | `docs/troubleshooting.md` | `docs/users-guides/troubleshooting.md` |

#2398 reported the `components.md` link; a sweep of the whole table found four more with the same root cause. All corrected targets verified to exist on `main`.

## Test plan

- [x] Swept every relative link in the Documentation table; the five above were the only broken ones (the rest resolve).
- [x] Confirmed all five corrected paths exist on `main`.

Fixes #2398.

_Assisted-By: Claude Code_

PR #2301, opened 2026-07-28
docs: Documentation improvements for 0.7 launch

esnible:
## Summary

More documentation cleanup for 0.7 launch.

PR #2448, opened 2026-08-20
CI: gate broken relative links in changed docs (#2412)

### LINKED ISSUE #2412: docs-ci: lychee link-check passes on broken relative links (fail: false)
## Problem

`.github/workflows/docs-ci.yaml` runs lychee with `fail: false`, so the **Link Check** job stays green even when relative links 404. This is why #2398 (broken README links) and the additional dead links found while fixing it (in `CLAUDE.md`) all shipped to `main` with a green Link Check.

## Evidence

At the time #2398 was filed, `README.md` had **5** broken relative links and `CLAUDE.md` had **5** more — none caught by CI.

## Suggested fix

Make lychee fail the job on broken **local/relative** links, while optionally keeping **external** URLs advisory (external checks are flake-prone — rate limits, transient 5xx):

- Flip `fail: true` for the relative-link pass, or
- split into two lychee invocations — relative links gating (`fail: true`), external links advisory (`fail: false`).

## Related

- #2398 (README links) — fixed in #2402
- Dead `CLAUDE.md` links (`docs/ai-ops/`, `docs/auth/keycloak-patterns.md`, `docs/skills/`) — also fixed in #2402; the `keycloak-patterns` doc was removed with no successor and may warrant restoring as separate content work.

Found via review of #2402.

_Assisted-By: Claude Code_

cwiklik:
## Summary

`docs-ci.yaml` ran lychee with `fail: false`, so broken **relative** links passed the Link Check job — how #2398 reached `main` with a green check.

Splits the single advisory pass into two:

- **Gating pass** (`fail: true`) — `lychee --offline` over only the `.md` files the PR changes. `--offline` checks local/relative links only, so it is deterministic and free of external-URL flakiness. Verified locally (lychee 0.24.2 + repo `.lychee.toml`): a broken relative link exits non-zero (job red), a clean one exits 0.
- **Advisory pass** (`fail: false`) — the existing full check (external URLs + all docs) preserved unchanged; external links are flake-prone, so they stay non-gating.

`fetch-depth: 0` makes the PR base merge-base reachable for the changed-files diff; changed paths are `./`-prefixed so a filename can't be parsed as a lychee flag (and lychee fails **closed** on a bad input path, so a mangled path reds the job rather than passing silently).

## Coverage (and its limit)

This catches a broken relative link in an **edited** `.md`. It does **not** catch a rename that moves a link target while leaving its *unedited* referrers stale — e.g. #2398 broke `README.md` (edited in the reorg PR → **would be caught**) *and* `CLAUDE.md` (not edited → **would slip through**). That move-the-target class stays with the advisory full pass and the cleanup issue.

## Why changed-files-scoped, not repo-wide

`main` still has **~74 broken relative links** (lychee's count under `.lychee.toml`'s `exclude_path`): in-repo doc breakage, cross-repo `../cortex/…` refs that can't resolve locally, and a few placeholders. A repo-wide `fail: true` would block on all of them. This PR fixes **recurrence** (the issue's intent) without that cleanup. Pre-existing debt is tracked in #2449; note the OpenClaw link-health agent (#1178) only auto-files **external** broken-link issues, so relative-link debt is otherwise untracked.

Closes #2412.

_Assisted-By: Claude Code_

### DIFF
--- a/.github/workflows/docs-ci.yaml
+++ b/.github/workflows/docs-ci.yaml
@@ -36,8 +36,39 @@ jobs:
     timeout-minutes: 10
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+        with:
+          fetch-depth: 0  # need the merge-base to diff changed files against the PR base
 
-      - name: Run lychee link checker
+      # Gate only on links in the files THIS PR changes, so a docs reorg can't
+      # introduce a broken relative link (rossoctl/rossoctl#2412) without failing
+      # review — without forcing an up-front cleanup of the repo's pre-existing
+      # relative-link debt (tracked separately).
+      - name: Determine changed markdown files
+        id: changed
+        run: |
+          base='${{ github.event.pull_request.base.sha }}'
+          files="$(git diff --name-only --diff-filter=d "${base}...HEAD" -- '*.md' | sort -u | tr '\n' ' ')"
+          echo "files=${files}" >> "$GITHUB_OUTPUT"
+          [ -n "${files}" ] && echo "any=true" >> "$GITHUB_OUTPUT" || echo "any=false" >> "$GITHUB_OUTPUT"
+          echo "Changed markdown: ${files:-<none>}"
+
+      # Gating pass: --offline checks only local/relative links (no network), so it
+      # is deterministic and free of external-URL flakiness. Fails the job on a
+      # broken relative link in a changed doc.
+      - name: Link check — changed files, relative links (gating)
+        if: steps.changed.outputs.any == 'true'
+        uses: lycheeverse/lychee-action@e7477775783ea5526144ba13e8db5eec57747ce8  # v2.9.0
+        with:
+          args: >-
+            --offline
+            --config .lychee.toml
+            ${{ steps.changed.outputs.files }}
+          fail: true
+
+      # Advisory pass: full check (external URLs + all docs). External links are
+      # flake-prone (rate limits, transient 5xx), so this stays non-gating —
+      # preserving today's behavior for everything the gating pass doesn't cover.
+      - name: Link check — full, external + all docs (advisory)
         uses: lycheeverse/lychee-action@e7477775783ea5526144ba13e8db5eec57747ce8  # v2.9.0
         with:
           args: >-
--- a/.github/workflows/docs-ci.yaml
+++ b/.github/workflows/docs-ci.yaml
@@ -39,15 +39,24 @@ jobs:
         with:
           fetch-depth: 0  # need the merge-base to diff changed files against the PR base
 
-      # Gate only on links in the files THIS PR changes, so a docs reorg can't
-      # introduce a broken relative link (rossoctl/rossoctl#2412) without failing
-      # review — without forcing an up-front cleanup of the repo's pre-existing
-      # relative-link debt (tracked separately).
+      # Gate on broken relative links in the .md files THIS PR changes
+      # (rossoctl/rossoctl#2412), without forcing an up-front cleanup of the repo's
+      # pre-existing relative-link debt (~74 links; tracked in #2449).
+      #
+      # Scope note: this catches a broken relative link in an *edited* file. It does
+      # NOT catch a rename that moves a link target while leaving its (unedited)
+      # referrers stale — e.g. #2398 broke README.md (edited in the reorg PR, would
+      # be caught here) *and* CLAUDE.md (not edited, would slip through). That
+      # move-the-target class is only surfaced by the advisory full pass today.
       - name: Determine changed markdown files
         id: changed
         run: |
           base='${{ github.event.pull_request.base.sha }}'
-          files="$(git diff --name-only --diff-filter=d "${base}...HEAD" -- '*.md' | sort -u | tr '\n' ' ')"
+          # Prefix each path with ./ so a name can never be parsed as a lychee flag.
+          # (A pathological name with a space still can't inject: lychee errors on a
+          # nonexistent input path and fails the job, rather than passing silently.)
+          files="$(git diff --name-only --diff-filter=d "${base}...HEAD" -- '*.md' \
+                   | sort -u | sed 's,^,./,' | tr '\n' ' ')"
           echo "files=${files}" >> "$GITHUB_OUTPUT"
           [ -n "${files}" ] && echo "any=true" >> "$GITHUB_OUTPUT" || echo "any=false" >> "$GITHUB_OUTPUT"
           echo "Changed markdown: ${files:-<none>}"


.github/workflows/docs-ci.yaml
```diff
@@ -36,8 +36,39 @@ jobs:
...
+          files="$(git diff --name-only --diff-filter=d "${base}...HEAD" -- '*.md' | sort -u | tr '\n' ' ')"
+          echo "files=${files}" >> "$GITHUB_OUTPUT"
+          [ -n "${files}" ] && echo "any=true" >> "$GITHUB_OUTPUT" || echo "any=false" >> "$GITHUB_OUTPUT"
+          echo "Changed markdown: ${files:-<none>}"
+
+      # Gating pass: --offline checks only local/relative links (no network), so it
+      # is deterministic and free of external-URL flakiness. Fails the job on a
+      # broken relative link in a changed doc.
+      - name: Link check — changed files, relative links (gating)
+        if: steps.changed.outputs.any == 'true'
```
huang195, line 68: **suggestion** — the gate has never actually executed, and its failure mode is skip → green.

You call this out already: lychee couldn't be run locally, and this PR touches no `.md` so the gating pass is skipped by design. Combined with `if: steps.changed.outputs.any == 'true'`, any bug in the `Determine changed markdown files` step degrades to *step skipped, job green* — the same fail-open-and-green shape as the `fail: false` bug this PR fixes.

Worth one throwaway commit before merge: add a `.md` with a deliberately broken relative link, confirm the job goes red, revert. Cheap, and it's the difference between a gate and the appearance of one.

.github/workflows/docs-ci.yaml
```diff
@@ -36,8 +36,39 @@ jobs:
     timeout-minutes: 10
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+        with:
+          fetch-depth: 0  # need the merge-base to diff changed files against the PR base
+
+      # Gate only on links in the files THIS PR changes, so a docs reorg can't
```
huang195: **suggestion** — this comment (and the matching line in the PR description) overstates coverage for the rename case.

I verified the README half holds: README.md *was* in reorg PR #2301's changed files and still carried the stale `./docs/components.md` link, so the gate would have failed it.

But `CLAUDE.md` was **not** in that PR's changed files, and its `docs/components.md` link broke in exactly the same way — invisible to a changed-files-scoped gate. So "a docs reorg can't introduce a broken relative link without failing review" is not quite true: a rename whose inbound linkers aren't also edited still slips through. Same for the description's "and the dead `CLAUDE.md` links fixed in #2402" — README yes, CLAUDE.md no.

Either soften the claim, or add a cheap targeted check: on any `.md` rename/delete in the PR, grep the repo for references to the old path and fail if any remain. That covers the incident class directly without touching the pre-existing debt.

.github/workflows/docs-ci.yaml
```diff
@@ -36,8 +36,39 @@ jobs:
...
+      # is deterministic and free of external-URL flakiness. Fails the job on a
+      # broken relative link in a changed doc.
+      - name: Link check — changed files, relative links (gating)
+        if: steps.changed.outputs.any == 'true'
+        uses: lycheeverse/lychee-action@e7477775783ea5526144ba13e8db5eec57747ce8  # v2.9.0
+        with:
+          args: >-
+            --offline
+            --config .lychee.toml
+            ${{ steps.changed.outputs.files }}
```
huang195, line 74: **nit** — this splices a contributor-controlled path list straight into `args:`.

A filename containing a space splits into two args; one starting with `-` gets parsed as a lychee flag. With `permissions: contents: read` and no secrets in this workflow the blast radius is gate fidelity, not compromise, so this is low priority.

The part worth confirming: does lychee **error** on a nonexistent input path, or warn and continue? If it warns, a split path means the gating step checks nothing while the job stays green. If you want it airtight, write the list to a file and feed it via stdin, or pass through an env var with `xargs -0`.

huang195 (approved):
Dobra robota — this closes the hole properly, and the two-pass split (deterministic offline gating, flaky external advisory) is the right shape.

### What I verified

The interesting check: **would this gate have caught #2398?** Yes. README.md *was* in reorg PR #2301's changed files, and at commit `5872043` it still pointed at `./docs/components.md`, which that same commit deleted. The offline gating pass would have failed it. I expected changed-files scoping to miss the motivating incident; it doesn't.

Also confirmed:

- **`${{ }}` in `run:`** — trigger is `pull_request` only, so `base.sha` is always populated. No push-event breakage.
- **`base...HEAD`** is the right form for PRs, and `fetch-depth: 0` makes the merge-base reachable.
- **`--diff-filter=d`** correctly drops deletions.
- **`tr '\n' ' '`** prevents `$GITHUB_OUTPUT` newline injection.
- Both action refs SHA-pinned; the advisory pass is byte-identical to before.
- **"~45 broken relative links on main"** — my independent count is **74** applying `.lychee.toml`'s `exclude_path`. The debt is understated, so "not landable repo-wide" is the right call.

Three non-blocking comments inline. Nothing holding this up.

### Summary

**Areas reviewed**: CI / GitHub Actions
**Commits**: 1, signed off
**CI status**: 22 passing; `Deploy & Test` pending, `E2E HyperShift` awaiting `/run-e2e`

_Assisted-By: Claude Code_

cwiklik:
Thanks @huang195 — pushed `1a0eb6ac`. All three addressed:

**#1 (gate never executed / skip→green):** executed it. Installed lychee 0.24.2 and ran the gating command with the repo's `.lychee.toml`:
- clean relative link → **exit 0**
- broken relative link → **exit 2** (`[ERROR] … File not found`) → job red

So the gate genuinely fails on a broken relative link. Happy to also do the throwaway-`.md` CI demonstration if you'd like it on the record, but the local run against the real config covers the concern without churn.

**#2 (overstates coverage — CLAUDE.md class):** correct, and softened. The workflow comment (and I'll fix the PR description) now says explicitly: the gate catches a broken relative link in an *edited* file (the README half of #2398), but **not** a rename that moves a target while leaving unedited referrers stale (the CLAUDE.md half) — that class stays with the advisory pass and the #2449 cleanup. I looked at your rename-grep suggestion and decided against it for now: referrers use varied relative forms (`../components.md`, `./docs/…`, bare), so a substring grep misses the `../`-up cases and a basename grep is false-positive-heavy — a robust version reimplements lychee's resolution. Not worth the fragility here; noted as a possible enhancement in #2449.

**#3 (args splicing):** hardened + verified it fails **closed**, not open. I checked lychee's behavior on a bad input path: it **errors** (`Input '…' not found`, exit non-zero), so a split/space-mangled path fails the job rather than silently checking nothing. Added a `./` prefix so a filename can't be parsed as a flag either. Airtight enough for a read-only, no-secrets workflow.

Re your count: **74** (yours, proper lychee parse) vs my crude-regex **~45** — yours is authoritative; I'll bump #2449's figure. Understated either way, so changed-files scoping stands.

_Assisted-By: Claude Code_

PR #2452, opened 2026-08-21
feat: feature-flagged mesh self-heal CronJob

cwiklik:
## Summary

**Part B** of the mesh-wide-503 self-heal (rossoctl/rossoctl#1899). Part A — `scripts/k8s/mesh-recover.sh` + the `k8s:health` / `istio:mesh-selfheal` skills — shipped in 0.7 (#2179). This adds the hands-off remediator: a **feature-flagged CronJob** (in `rossoctl-deps`) that periodically probes the gateway and, on a recoverable mesh outage (expired ztunnel/waypoint mTLS certs after a host suspend → HTTP 503 on all `*.localtest.me` routes), runs `mesh-recover.sh --fix` with cooldown / max-restart safety and loud Events.

**Part of #1899** (not Closes — #1899 stays open as the umbrella; the durable fix is upstream, istio/ztunnel#1679).

## Design

- **OFF by default.** Helm value `meshSelfHeal.enabled` (chart-level flag). Renders only on **non-OpenShift** — the ambient cert-expiry failure mode is Kind/dev-specific.
- **Reuses `mesh-recover.sh` unchanged** (auto-discovers the ztunnel namespace + probe host; exit `0` healthy / `2` degraded / `3` inconclusive). Helm `.Files.Get` can't read outside the chart, so a **verbatim copy** lives in `charts/rossoctl-deps/files/`, kept identical by `scripts/check-mesh-recover-sync.sh`.
- **Wrapper** (`mesh-selfheal-wrapper.sh`) adds only scheduling + safety: detect → cooldown / max-restart(rolling-window) gating (persisted in a small state ConfigMap) → `--fix` → k8s Events. Probes the in-cluster gateway Service `http-istio.rossoctl-system.svc`.
- **Safe by construction:** CronJob pod is kept **out of ambient** (`istio.io/dataplane-mode: none`) so its own `kubectl` calls don't traverse the broken ztunnel it's repairing; `concurrencyPolicy: Forbid`; nonroot / read-only rootfs / drop-ALL; least-privilege RBAC (get/list + **patch** deployments+daemonsets for `rollout restart`; state ConfigMap get/patch; events create).

## Files

- `charts/rossoctl-deps/values.yaml` — `meshSelfHeal` block (flag, schedule, image, gatewayUrl, cooldown/max/window, resources).
- `charts/rossoctl-deps/files/mesh-recover.sh` — verbatim copy of the Part A engine.
- `charts/rossoctl-deps/files/mesh-selfheal-wrapper.sh` — the scheduling/gating/Events wrapper.
- `charts/rossoctl-deps/templates/mesh-selfheal.yaml` — scripts CM + state CM + SA + ClusterRole/Binding + Role/Binding + CronJob, gated `enabled AND not openshift`.
- `scripts/check-mesh-recover-sync.sh` — `diff -q` guard for the chart copy.

## Verification

- `helm template`: **OFF → no objects**, **ON+Kind → full set**, **ON+OpenShift → no objects** (Kind-only gate).
- `helm lint` clean · wrapper **shellcheck-clean** · sync guard passing.

## Follow-ups (not in this PR)

- Wire `scripts/check-mesh-recover-sync.sh` into pre-commit.
- Live Kind test via `scripts/k8s/induce-mesh-outage.sh`.
- File/link the suspend/clock-jump repro to istio/ztunnel#1679 (durable upstream fix).

Refs: Part A #2179 · umbrella #1899 · upstream istio/ztunnel#1679.

Assisted-By: Claude Code

### DIFF
--- a/charts/rossoctl-deps/files/mesh-selfheal-wrapper.sh
+++ b/charts/rossoctl-deps/files/mesh-selfheal-wrapper.sh
@@ -1,0 +1,100 @@
+#!/usr/bin/env bash
+#
+# mesh-selfheal-wrapper.sh — Part B scheduling+safety wrapper around mesh-recover.sh
+# (rossoctl/rossoctl#1899). Run by the mesh-selfheal CronJob on each tick:
+#   1. detect: mesh-recover.sh --json   (exit 0 healthy / 2 degraded / 3 inconclusive)
+#   2. on degraded, apply cooldown + max-restart(window) gating from a state ConfigMap
+#   3. if allowed: mesh-recover.sh --fix   (rollout restart ztunnel + waypoints + gateway)
+#   4. emit k8s Events throughout; persist restart state back to the ConfigMap
+#
+# Env (set by the CronJob):
+#   GATEWAY_URL, PROBE_HOST (optional), INCLUDE_SPIRE (true/false),
+#   COOLDOWN_SECONDS, MAX_RESTARTS, WINDOW_SECONDS, STATE_NS, STATE_CM, CRONJOB_NAME
+set -uo pipefail
+
+NS="${STATE_NS:?STATE_NS required}"
+CM="${STATE_CM:?STATE_CM required}"
+COOLDOWN="${COOLDOWN_SECONDS:-900}"
+MAX_RESTARTS="${MAX_RESTARTS:-5}"
+WINDOW="${WINDOW_SECONDS:-3600}"
+CRONJOB="${CRONJOB_NAME:-mesh-selfheal}"
+now=$(date +%s)
+
+log() { echo "[mesh-selfheal] $*" >&2; }
+
+# Best-effort k8s Event against the CronJob so `kubectl describe cronjob mesh-selfheal` shows history.
+emit_event() {  # $1=type(Normal|Warning) $2=reason $3=message
+  local ts; ts=$(date -u +%Y-%m-%dT%H:%M:%SZ)
+  kubectl -n "$NS" create -f - >/dev/null 2>&1 <<EOF || log "event emit failed (non-fatal): $2"
+apiVersion: v1
+kind: Event
+metadata:
+  generateName: mesh-selfheal-
+  namespace: ${NS}
+type: ${1}
+reason: ${2}
+message: ${3}
+firstTimestamp: ${ts}
+lastTimestamp: ${ts}
+count: 1
+source:
+  component: mesh-selfheal
+involvedObject:
+  apiVersion: batch/v1
+  kind: CronJob
+  name: ${CRONJOB}
+  namespace: ${NS}
+EOF
+}
+
+read_state() {  # $1=key ; echoes value or 0
+  local v; v=$(kubectl -n "$NS" get cm "$CM" -o "jsonpath={.data.$1}" 2>/dev/null || true)
+  echo "${v:-0}"
+}
+
+# --- 1. detect (never mutates) ---
+args=(--json --gateway-url "$GATEWAY_URL")
+[ -n "${PROBE_HOST:-}" ] && args+=(--probe-host "$PROBE_HOST")
+[ "${INCLUDE_SPIRE:-false}" = "true" ] && args+=(--include-spire)
+
+summary=$(bash /scripts/mesh-recover.sh "${args[@]}" 2>/dev/null); rc=$?
+log "detect rc=$rc summary=$summary"
+
+case "$rc" in
+  0) emit_event Normal MeshHealthy "gateway reachable, mesh certs OK"; exit 0 ;;
+  3) emit_event Warning MeshInconclusive "cannot confirm mesh health (restart would not help): ${summary}"; exit 0 ;;
+  2) : ;;  # degraded/recoverable — fall through to gating
+  *) log "unexpected rc=$rc; treating as inconclusive"; emit_event Warning MeshInconclusive "mesh-recover.sh rc=$rc"; exit 0 ;;
+esac
+
+# --- 2. gating (cooldown + max-restarts within a rolling window) ---
+last=$(read_state last_restart)
+count=$(read_state restart_count)
+wstart=$(read_state window_start)
+[ "$wstart" = "0" ] && wstart=$now
+# reset the rolling window if it has elapsed
+if [ $((now - wstart)) -ge "$WINDOW" ]; then count=0; wstart=$now; fi
+
+if [ $((now - last)) -lt "$COOLDOWN" ]; then
+  emit_event Warning MeshCooldown "degraded but within ${COOLDOWN}s cooldown since last restart — skipping: ${summary}"
+  log "in cooldown ($((now - last))s < ${COOLDOWN}s); skip"
+  exit 0
+fi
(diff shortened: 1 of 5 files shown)


charts/rossoctl-deps/files/mesh-selfheal-wrapper.sh
```diff
@@ -0,0 +1,100 @@
...
+  local ts; ts=$(date -u +%Y-%m-%dT%H:%M:%SZ)
+  kubectl -n "$NS" create -f - >/dev/null 2>&1 <<EOF || log "event emit failed (non-fatal): $2"
+apiVersion: v1
+kind: Event
+metadata:
+  generateName: mesh-selfheal-
+  namespace: ${NS}
+type: ${1}
+reason: ${2}
+message: ${3}
```
pdettori: **suggestion** — The Event heredoc embeds the raw `--json` summary into an unquoted YAML scalar (`message: ${3}`), and several call sites pass `${summary}` (MeshInconclusive rc=3, MeshCooldown, MeshRestartCapHit, MeshDegraded). Because the JSON contains `:`, `kubectl create -f -` fails to parse it — I reproduced this locally and it errors with *"mapping values are not allowed here"*. That means those events silently hit the non-fatal `|| log "event emit failed"` path and never get created — i.e. exactly the "loud Events" the design promises for `kubectl describe cronjob` are the ones that go missing. (MeshHealthy and MeshRecoverAttempted carry no summary, so they still work.)

Suggested fix: emit the message as a literal block scalar so the JSON can't break YAML:
```yaml
message: |-
  ${3}
```
(indented under the heredoc), or strip/escape the JSON before interpolating.

charts/rossoctl-deps/values.yaml
```diff
@@ -441,3 +441,33 @@ mlflow:
...
+# probes the gateway and, on a recoverable outage, runs scripts/k8s/mesh-recover.sh --fix with
+# cooldown / max-restart safety and loud Events. OFF by default; only renders on non-OpenShift
+# (the ambient/ztunnel cert-expiry failure mode is Kind/dev-specific — see the issue). The durable
+# fix is upstream (istio/ztunnel#1679); this is a stop-gap remediator.
+meshSelfHeal:
+  enabled: false
+  # Cron schedule for the probe+remediate tick.
+  schedule: "*/5 * * * *"
+  # Stock image with bash + kubectl + curl (no custom image built for this).
+  image: "registry.k8s.io/kubectl:v1.31.0"
```
pdettori: **suggestion** — `registry.k8s.io/kubectl` is a minimal (debian-base) image that very likely ships **no `curl`**, but `mesh-recover.sh` relies on curl to probe the gateway (`check_mesh_reachability`, curl calls at `mesh-recover.sh:150,153`). With no curl the probe returns `000` → INCONCLUSIVE, so the self-heal never fires — a silent no-op rather than a visible failure. `bash` may also be absent even though the CronJob runs `command: ["bash", ...]`.

Since the live Kind test is a follow-up, this path is currently unverified. Please confirm the image actually has `bash` + `curl` at runtime, or switch to an image that bundles them (e.g. one with a shell + curl alongside kubectl).

charts/rossoctl-deps/templates/mesh-selfheal.yaml
```diff
@@ -0,0 +1,179 @@
...
+          labels:
+            app: mesh-selfheal
+            # Keep this pod OUT of ambient — its kubectl/curl must not traverse the (possibly
+            # broken) ztunnel it is trying to repair. Same mechanism the gateway pods use.
+            istio.io/dataplane-mode: none
+        spec:
+          serviceAccountName: mesh-selfheal
+          restartPolicy: Never
+          securityContext:
+            runAsNonRoot: true
```
pdettori, line 138: **suggestion** — `runAsNonRoot: true` is set without a `runAsUser`. If the chosen image's default user is root (UID 0), the kubelet rejects the pod at admission (*"container has runAsNonRoot and image will run as root"*) and the CronJob never runs. Recommend pinning an explicit non-root `runAsUser` (and verifying the image supports running as that UID) so startup doesn't depend on the image's default user — this pairs with the image-choice question on `values.yaml`.

pdettori (approved):
Well-structured Part B of the mesh self-heal work. Verified: the chart copy of `mesh-recover.sh` is byte-identical to `scripts/k8s/mesh-recover.sh` (sync guard correct), the `rossoctl.labels` helper and `.Values.openshift` gate exist and match existing conventions, and the RBAC is tight and least-privilege (read + patch for rollout restart, named state ConfigMap, events create). Feature is off by default and Kind-only, so nothing here blocks merge.

The three inline suggestions are runtime-viability concerns on a path that hasn't been live-tested yet (live Kind test is a listed follow-up). None block merge, but all three should be resolved before `meshSelfHeal.enabled` is turned on:

1. Event YAML breaks when it embeds the JSON `--json` summary (verified locally — `kubectl create -f -` rejects it), silently dropping the key Warning events.
2. `registry.k8s.io/kubectl` very likely has no `curl` (and maybe no `bash`), which would make the probe a silent no-op.
3. `runAsNonRoot: true` with no `runAsUser` may block pod admission if the image defaults to root.

**Author**: cwiklik (MEMBER — maintainer) · **Commits**: 1, signed-off · **CI**: passing (E2E HyperShift pending maintainer `/run-e2e`)

Assisted-By: Claude Code

charts/rossoctl-deps/files/mesh-selfheal-wrapper.sh
cwiklik: Fixed in d0881397. emit_event now flattens the message to one line and emits it as a literal block scalar (\`message: |-\`), so the \`:\`-laden JSON summary can't break the Event YAML. Verified live on kind-rossoctl: the \`MeshDegraded\` Event was created carrying the full \`{"degraded":true,...}\` summary. Confirmed the old plain-scalar form reproduces your exact error (\`kubectl create\` → "mapping values are not allowed in this context"), and the new form parses.

charts/rossoctl-deps/values.yaml
cwiklik: Fixed in d0881397. You were right, and it's worse than "likely no curl": \`registry.k8s.io/kubectl\` has *no shell at all* (User=0, entrypoint \`/bin/kubectl\`, \`sh\` not found), so \`command: ["bash", ...]\` couldn't even exec. Switched to \`docker.io/alpine/k8s:1.31.0\`, which bundles bash + curl + kubectl (pinned, non-floating). Verified live: the wrapper exec'd, curl probed the gateway (503 → restart → 200).

charts/rossoctl-deps/templates/mesh-selfheal.yaml
cwiklik: Fixed in d0881397. Pinned \`runAsUser/runAsGroup/fsGroup: 65532\` (the image defaults to root, so \`runAsNonRoot\` alone would fail admission). Also added a writable \`HOME\` emptyDir so kubectl's discovery cache works under \`readOnlyRootFilesystem: true\`. Verified live: the pod was admitted and Succeeded as UID 65532 with no runAsNonRoot rejection and no cache warnings.

PR #2453, opened 2026-08-21
fix: constrain agent route path params to RFC-1123

### LINKED ISSUE #2395: Constrain agent route path params (name/namespace) to RFC-1123 — resolves CodeQL py/log-injection
## Summary
CodeQL flags **`py/log-injection`** (medium, ~7 instances) where a **user-provided path param** (`name`, and by extension `namespace`) is logged unsanitized. These are **pre-existing** findings that #2388 (the move-only `agents.py` split) *relocated* into new files — `main`'s `agents.py` already logs `{name}` the same way. #2388 was merged as a pure refactor with a CodeQL override referencing this issue; this issue tracks the actual fix.

## Affected routes
- `rossoctl/backend/app/routers/agents_migration.py` → `migrate_agent(namespace, name, …)` (lines ~240, 254, 267, 276, 295, 298)
- `rossoctl/backend/app/routers/agents_shipwright.py` → `trigger_shipwright_buildrun`, `get_shipwright_build_info` (line ~427)

In each, `name`/`namespace` are **bare `str` path params** interpolated into `logger.info(f"…{name}…")`.

## Risk
Low but real. An authenticated `ROLE_OPERATOR` caller could pass a URL-encoded newline (`%0A`) in the path segment → forged/split log lines (log-integrity, not RCE). k8s rejects invalid names on the actual operation, but some log statements fire before that.

## Fix (boundary validation — resolves the alert legitimately, not a suppression)
Constrain the path params to an RFC-1123 DNS label so injection chars can't reach the logger; invalid input returns 422.

```python
from fastapi import APIRouter, Depends, HTTPException, Path, Query   # add Path

# RFC-1123 DNS label (k8s resource/namespace name)
_K8S_NAME = r"^[a-z0-9]([-a-z0-9]*[a-z0-9])?$"

async def migrate_agent(
    namespace: str = Path(..., pattern=_K8S_NAME, max_length=63),
    name: str = Path(..., pattern=_K8S_NAME, max_length=63),
    ...
):
```
Apply the same to the two shipwright handlers. (Pydantic v2 → `pattern=`, not `regex=`.) Optionally put `_K8S_NAME` in the shared `app/utils/naming.py` and import it to keep it DRY.

## Acceptance
- [ ] `name`/`namespace` path params on the 3 handlers constrained to RFC-1123 (`Path(pattern=…, max_length=63)`).
- [ ] CodeQL `py/log-injection` clears on `agents_migration.py` / `agents_shipwright.py`.
- [ ] Test: a `name` containing a newline → **422** (not logged).
- [ ] Confirm no route legitimately passes dotted/long (>63) names.

Refs #2388 (relocated these findings).

Assisted-By: Claude Code

cwiklik:
## Summary

Resolves the CodeQL **`py/log-injection`** findings on the agent migration + shipwright routes: user-provided path params (`name`, `namespace`) were logged unsanitized, so an authenticated `ROLE_OPERATOR` caller could pass an encoded newline (`%0A`) in a path segment and forge/split log lines (log-integrity, not RCE).

**Fix = boundary validation**, not a suppression: constrain the path params to an RFC-1123 DNS label with FastAPI `Path(pattern=…, max_length=63)`, so injection characters are rejected with **422** before any handler code or logger runs.

Closes #2395. Refs #2388 (the move-only split that relocated these pre-existing findings).

## Changes

- `app/utils/naming.py` — shared `K8S_NAME_PATTERN` (`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`) + `K8S_NAME_MAX_LENGTH` (63).
- `app/routers/agents_migration.py` — `migrate_agent` `name`/`namespace` → `Path(..., pattern=K8S_NAME_PATTERN, max_length=K8S_NAME_MAX_LENGTH)`.
- `app/routers/agents_shipwright.py` — same for all four path-param handlers: `get_shipwright_build_status`, `get_shipwright_buildrun_status`, `trigger_shipwright_buildrun`, `get_shipwright_build_info`.
- `tests/test_route_name_validation.py` — new.

## Why this is safe

Agent/namespace names are already RFC-1123 labels (lowercase alphanumeric + `-`, ≤63) — Deployments/Services derived from them can't be dotted or longer — so no legitimate request is affected. k8s would reject invalid names on the actual op anyway; this just moves the rejection to the boundary so the logger never sees the raw value.

## Testing

- New `tests/test_route_name_validation.py`: newline (`%0A`), uppercase, dotted, underscore, leading-hyphen, and >63-char names all return **422** on both `name` and `namespace` across the migration + shipwright routes; a valid RFC-1123 name is **not** rejected. **19 passed.**
- `ruff check` + `ruff format --check` clean.
- No regressions: shipwright route tests **155 passed**.
- Expect CodeQL `py/log-injection` to clear on `agents_migration.py` / `agents_shipwright.py`.

Assisted-By: Claude Code

### DIFF
--- a/rossoctl/backend/app/utils/naming.py
+++ b/rossoctl/backend/app/utils/naming.py
@@ -5,6 +5,15 @@
 Helpers for deriving valid Kubernetes resource names from user-supplied text.
 """
 
+# RFC-1123 DNS label — the shape of a Kubernetes resource / namespace name:
+# lowercase alphanumeric plus '-', starting and ending with an alphanumeric.
+# Use with FastAPI ``Path(pattern=K8S_NAME_PATTERN, max_length=K8S_NAME_MAX_LENGTH)``
+# to constrain user-supplied path params at the boundary, so injection characters
+# (e.g. an encoded newline) can't reach loggers or downstream calls — invalid input
+# is rejected with 422 before any handler code runs. (rossoctl/rossoctl#2395)
+K8S_NAME_PATTERN = r"^[a-z0-9]([-a-z0-9]*[a-z0-9])?$"
+K8S_NAME_MAX_LENGTH = 63
+
 
 def sanitize_k8s_name(name: str) -> str:
     """Sanitize a name to be valid for Kubernetes resource names.
--- a/rossoctl/backend/app/utils/naming.py
+++ b/rossoctl/backend/app/utils/naming.py
@@ -5,14 +5,22 @@
 Helpers for deriving valid Kubernetes resource names from user-supplied text.
 """
 
+import re
+
 # RFC-1123 DNS label — the shape of a Kubernetes resource / namespace name:
-# lowercase alphanumeric plus '-', starting and ending with an alphanumeric.
-# Use with FastAPI ``Path(pattern=K8S_NAME_PATTERN, max_length=K8S_NAME_MAX_LENGTH)``
-# to constrain user-supplied path params at the boundary, so injection characters
-# (e.g. an encoded newline) can't reach loggers or downstream calls — invalid input
-# is rejected with 422 before any handler code runs. (rossoctl/rossoctl#2395)
-K8S_NAME_PATTERN = r"^[a-z0-9]([-a-z0-9]*[a-z0-9])?$"
+# lowercase alphanumeric plus '-', starting and ending with an alphanumeric, ≤63 chars.
+# Single source of truth for this pattern across the backend (routers previously each
+# defined their own identical ``re.compile(...)`` — chat.py, acp.py, agents_authbridge.py).
+#
+# Two forms:
+#   * K8S_NAME_PATTERN — regex string, for FastAPI ``Path(pattern=..., max_length=...)`` so
+#     malformed path params are rejected with 422 before any handler/logger runs (#2395).
+#   * K8S_NAME_RE — compiled, for direct ``.fullmatch()`` / ``.match()`` validation.
+# The length bound (``{0,61}`` between the anchors) makes the regex self-sufficient, so it
+# holds standalone as well as behind Path's max_length.
+K8S_NAME_PATTERN = r"^[a-z0-9]([-a-z0-9]{0,61}[a-z0-9])?$"
 K8S_NAME_MAX_LENGTH = 63
+K8S_NAME_RE = re.compile(K8S_NAME_PATTERN)
 
 
 def sanitize_k8s_name(name: str) -> str:
--- a/rossoctl/backend/app/routers/acp.py
+++ b/rossoctl/backend/app/routers/acp.py
@@ -18,6 +18,7 @@ from fastapi import APIRouter, Query, WebSocket, WebSocketDisconnect
 
 from app.core.config import settings
 from app.services.acp_bridge import ACPBridge
+from app.utils.naming import K8S_NAME_PATTERN
 
 logger = logging.getLogger(__name__)
 router = APIRouter(prefix="/acp", tags=["acp"])
@@ -29,7 +30,7 @@ JSON_RPC_METHOD_NOT_FOUND = -32601
 JSON_RPC_INTERNAL_ERROR = -32603
 JSON_RPC_PARSE_ERROR = -32700
 
-_K8S_NAME = re.compile(r"^[a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?$")
+_K8S_NAME = re.compile(K8S_NAME_PATTERN)
 
 
 @router.websocket("/ws/{namespace}/{agent_name}")
--- a/rossoctl/backend/app/routers/agents_authbridge.py
+++ b/rossoctl/backend/app/routers/agents_authbridge.py
@@ -34,13 +34,14 @@ from app.core.constants import (
 )
 from app.routers.agents_models import OutboundRoute
 from app.services.kubernetes import KubernetesService, get_kubernetes_service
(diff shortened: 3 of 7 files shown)


rossoctl/backend/app/utils/naming.py
```diff
@@ -5,6 +5,15 @@
...
 """
 
+# RFC-1123 DNS label — the shape of a Kubernetes resource / namespace name:
+# lowercase alphanumeric plus '-', starting and ending with an alphanumeric.
+# Use with FastAPI ``Path(pattern=K8S_NAME_PATTERN, max_length=K8S_NAME_MAX_LENGTH)``
+# to constrain user-supplied path params at the boundary, so injection characters
+# (e.g. an encoded newline) can't reach loggers or downstream calls — invalid input
+# is rejected with 422 before any handler code runs. (rossoctl/rossoctl#2395)
+K8S_NAME_PATTERN = r"^[a-z0-9]([-a-z0-9]*[a-z0-9])?$"
+K8S_NAME_MAX_LENGTH = 63
```
esnible: We have very similar code in three other places with a slightly different name, e.g. in _chat.py_:

```Python
_K8S_NAME_RE = re.compile(r"^[a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?$")
```

Consider using the same var or varname.

rossoctl/backend/app/utils/naming.py
```diff
@@ -5,6 +5,15 @@
 Helpers for deriving valid Kubernetes resource names from user-supplied text.
 """
 
+# RFC-1123 DNS label — the shape of a Kubernetes resource / namespace name:
+# lowercase alphanumeric plus '-', starting and ending with an alphanumeric.
+# Use with FastAPI ``Path(pattern=K8S_NAME_PATTERN, max_length=K8S_NAME_MAX_LENGTH)``
+# to constrain user-supplied path params at the boundary, so injection characters
+# (e.g. an encoded newline) can't reach loggers or downstream calls — invalid input
+# is rejected with 422 before any handler code runs. (rossoctl/rossoctl#2395)
+K8S_NAME_PATTERN = r"^[a-z0-9]([-a-z0-9]*[a-z0-9])?$"
```
esnible: **suggestion** — the same log-injection sink exists in 11 other routers that still take bare `namespace: str` / `name: str`: **94 unconstrained params** across the backend, vs. the 10 fixed here.

`tools.py` alone has 7 f-string logger calls on `name` (e.g. `logger.warning(f"Failed to delete Deployment '{name}': {e}")`) — structurally identical to what this PR just fixed. Same shape in `agents.py`, `agents_authbridge.py`, `skills.py`, `simulation.py`, `chat.py`, `acp.py`, `dream.py`, `agents_manifests.py`, `agents_skills.py`, and `agents_finalize.py`. I checked `main.py` — there's no global validation middleware that would catch them, and these constants are new in this PR, so nothing else was already covering that surface.

Not blocking, and the shared constants here make the rest mechanical. But since this closes #2395, worth either widening the PR or filing a follow-up so the CodeQL `py/log-injection` class doesn't read as resolved while 11 routers are still exposed.

esnible (approved):
Tight, well-scoped security fix. Hoisting the RFC-1123 pattern into `app/utils/naming.py` and applying it via FastAPI `Path(...)` means malformed input is rejected with 422 before any handler code runs — and before the f-string loggers CodeQL flagged.

I verified the fix rather than taking it on faith:

- Ran the new tests on the PR head — 19 passed.
- **Mutation-tested them**: reverting only the production `Path(...)` constraints (keeping the tests) makes **18 of 19 fail**. These tests are load-bearing, not decorative.
- **Confirmed the actual vector**: without the fix, `bad%0Aname` returns **409** — it reaches handler code and its loggers. With the fix, 422 at validation. That is the log-injection path genuinely closed.
- Checked the newline case isn't passing incidentally: the pattern rejects both the encoded (`bad%0Aname`) and decoded (`bad\nname`) forms, so it holds regardless of where Starlette percent-decodes.
- Verified the test route paths against the real router definitions (neither router carries its own prefix, so the `/api/v1/...` paths are correct).

One non-blocking scope note inline. Approving — the fix is correct as far as it goes.

**Areas reviewed**: Python (routers, utils, tests)
**Commits**: 1, signed off (DCO passing)
**CI**: 22 checks passing; `E2E HyperShift` pending a maintainer `/run-e2e` (backend-only change with unit coverage, so likely not required here)

Assisted-By: Claude Code

cwiklik:
Thanks for the thorough review (and the mutation test — good to know the tests are load-bearing). Both comments addressed in `46c11166`:

- ✅ **Consolidate the pattern (naming.py:15 / the 3 duplicate `_K8S_NAME_RE`).** Hoisted the canonical pattern into `app/utils/naming.py` as the single source of truth — `K8S_NAME_PATTERN` (now length-bounded `{0,61}` so it's self-sufficient standalone) plus a compiled `K8S_NAME_RE` — and routed `chat.py`, `acp.py`, and `agents_authbridge.py` (which had two copies) through it. Behavior is identical (same pattern; char-class order is irrelevant), so no functional change — just removes the 4-way duplication. 26 existing chat/authbridge tests + the 19 new ones still pass; ruff clean.

- ✅ **Wider surface (naming.py:14 / ~11 routers, ~94 params).** Agreed it shouldn't read as "resolved" while `tools.py` et al. are still exposed. Filed **#2457** to constrain the remaining routers — now mechanical thanks to the shared constants here. Kept this PR scoped to the migration+shipwright routes it fixes.

Assisted-By: Claude Code

PR #2454, opened 2026-08-21
fix(e2e): repair SPIFFE trust chain in tx-e2e token-exchange leg

cwiklik:
## Summary

Repairs the **SPIFFE trust chain** in the community-KC token-exchange E2E leg, which had been red since ~2026-07-16 with `invalid_client`. **Part of #2342.**

Root-caused live on a Kind cluster (not blind CI iteration). The issue's two leads — a secret-rotation race and a port-forward timeout — were **both wrong**. The real cause: the tx-e2e SPIFFE setup uses **production-shaped URLs that don't match Kind's actual services**. Three layered mismatches, each confirmed by watching Keycloak's error change:

| Layer | Was | Fix |
|---|---|---|
| IdP `issuer` | `spiffe://<domain>` (trust domain) | the SPIRE OIDC issuer (`https://oidc-discovery.<domain>`) — KC matches the SVID's `iss`, which is the OIDC URL, not the trust domain |
| IdP `bundleEndpoint` (JWKS) | `https://…svc/keys` | `http://…svc/keys` on non-OCP — the SPIRE OIDC Service is plain http:80 on Kind, so `https` was unreachable and KC couldn't verify the SVID signature |
| SVID `aud` / issuer | `https://<host>/realms/<realm>` (hardcoded) | derived from KC's **live** realm issuer (`http://keycloak.localtest.me:8080/realms/<realm>` on Kind), patched into `authbridge-config` (the operator-injected source, **not** `spiffe-helper-config`), with a workload restart to re-mint |

### Changes
- **`55-setup-spiffe-idp.sh`**: IdP `issuer` → SPIRE OIDC issuer (override via `SPIRE_OIDC_ISSUER`); `bundleEndpoint` http on non-OCP / https on OCP.
- **`56-enable-spiffe.sh`**: derive audience/issuer from the live realm `.well-known` and patch `authbridge-config` (+ per-workload) + restart workloads.
- **`conftest.py`**: urllib3 Retry adapter (kills the transient port-forward `ReadTimeout` — the "Bug B" symptom); skip secret-based `client_credentials`/exchange tests when the workload clients are `federated-jwt` (SPIFFE, no secret — covered by the JWT-SVID tests); SPIFFE grant timeout 10s→30s.
- Reverted an earlier wrong workflow block (a settle+`rollout restart --all` — `--all` is not a valid flag; the correct restart lives in `56`).

## Validation (live, `kind-rossoctl`)
tx-e2e suite: **14 failed → 12 passed, 13 skipped**. `test_spiffe_client_credentials` passes; secret-based tests correctly skip.

## Not in this PR (distinct follow-up)
`test_spiffe_token_exchange` still fails on **token-exchange authorization** (`"Requested audience not available"`) — a separate Keycloak *standard token-exchange* audience/permission concern, not the trust-chain `invalid_client` this PR fixes. Tracked separately.

## Caveats
- The OCP leg keeps its `https`/service-ca path (branched on `PLATFORM`); needs CI validation on the OCP legs.
- The community-KC leg stays `continue-on-error` until a green release-validation run confirms end to end.

Assisted-By: Claude Code


.github/scripts/token-exchange/56-enable-spiffe.sh
```diff
@@ -63,18 +63,28 @@ print('\n'.join(result))
...
-      -p "{\"data\":{\"helper.conf\":$(echo "$UPDATED_CONF" | jq -Rs .)}}"
-  fi
+  log_info "Keycloak realm issuer: $KC_ISSUER"
+  for CM in authbridge-config $(kubectl get cm -n "$TX_NAMESPACE" -o name 2>/dev/null | sed 's|configmap/||' | grep -E '^authbridge-config-'); do
+    kubectl get cm "$CM" -n "$TX_NAMESPACE" >/dev/null 2>&1 || continue
+    kubectl patch configmap "$CM" -n "$TX_NAMESPACE" --type=merge \
+      -p "{\"data\":{\"JWT_AUDIENCE\":\"$KC_ISSUER\",\"ISSUER\":\"$KC_ISSUER\"}}" 2>/dev/null || true
+  done
+  # Restart the workloads so spiffe-helper re-mints the SVID with the aligned audience.
+  kubectl rollout restart deployment -n "$TX_NAMESPACE" 2>/dev/null || true
```
pdettori, line 83: nit (non-blocking): `kubectl rollout restart deployment` restarts **every** deployment in `$TX_NAMESPACE`, not just the SPIFFE workloads, and the following `rollout status --timeout=180s` then waits on all of them. In a dedicated tx-e2e namespace this is harmless, but scoping the restart with a label selector (the same workload label used elsewhere) would narrow the blast radius and shorten the wait. Optional.

pdettori (approved):
## Review Summary

Well-scoped, thoroughly root-caused fix for the SPIFFE trust chain in the tx-e2e token-exchange leg. The three IdP/SVID mismatches (OIDC issuer vs trust domain, `http` vs `https` JWKS bundle endpoint on Kind, and sourcing the SVID audience from `authbridge-config` rather than `spiffe-helper-config`) are each explained and confirmed live, and the Python-side retry adapter + SPIFFE skip logic is a clean way to absorb the port-forward flakiness. All changes are confined to E2E test infrastructure — no production code.

Verified locally: both scripts run under `set -euo pipefail`; `PLATFORM` is defined (line 11, `detect_platform` default) so the new `$PLATFORM` branch is safe under `set -u`; `KC_URL` comes from `lib.sh`. Supply-chain gate clean (no `.claude/`/`.vscode/` changes).

**Areas reviewed**: Shell (`.github/scripts/token-exchange/`), Python (E2E fixtures/tests)
**Commits**: 2 real commits, both signed-off (DCO green)
**CI**: all required checks passing; E2E HyperShift pending maintainer `/run-e2e` (normal)

One non-blocking nit inline. Approving.

PR #2456, opened 2026-08-21
CI: gate floating image tags in rendered subcharts (#2389)

### LINKED ISSUE #2389: check-release-pins doesn't render subcharts — can't catch :latest in dependencies
## Problem
`scripts/check-release-pins.sh` only inspects **rossoctl's own** `charts/rossoctl/values.yaml`. It does **not render subcharts**, so it cannot see image tags inside a dependency chart (e.g. `operator-chart`).

This is why the operator-chart's `:latest` AuthBridge images (rossoctl/operator companion issue) were **invisible at rc.5** — the pin check passed while the transitive dependency shipped `:latest`.

## Fix
Add a full-render gate to the release check (this repo and ideally the operator repo):
```
helm template <chart> --set <required values> | grep -E 'image:.*(:latest|:main|:master)$' && exit 1 || true
```
i.e. fail the release check if any **rendered** manifest references a floating tag, covering subcharts.

## Interim workaround (in use)
At the v0.7.0 cut this full-render `grep :latest` was run **manually** and confirmed clean. Until this issue lands, that manual gate is applied at every cut.

Assisted-By: Claude Code

cwiklik:
## Summary

`check-release-pins.sh` only inspected rossoctl's own `charts/rossoctl/values.yaml` and `templates/`, so a floating image tag inside a **dependency subchart** (e.g. `operator-chart`'s AuthBridge images) was **invisible** — how a `:latest` shipped at an earlier rc while the pin check stayed green.

Adds an opt-in **`--render`** gate:
- `helm dependency build` + `helm template rossoctl charts/rossoctl --set openshift=false` (subcharts included; `openshift=false` clears the OpenShift-only required-value guards so it renders in CI),
- greps the **rendered** output for any container-registry ref ending in `:latest`/`:main`/`:master` — matches both `image:` fields and subchart ConfigMap-embedded refs (the AuthBridge case); ignores version-pinned (`:vX.Y.Z`) and digest (`@sha256:`) refs,
- fails **loudly** if helm is missing or the render fails (no silent skip).

Wired into `ci-release-pins.yaml`: a `Set up Helm` step + `--render` on the release-gate run (release/* PRs + manual dispatch). `helm dependency build` fetches the OCI subchart **without auth** (public), so it runs in CI.

## Validation

- Clean run → render check **passes** (0 floating tags today — operator #512 pinned operator-chart's images).
- **Negative test**: injecting a floating tag is caught end-to-end — `floating image tag in rendered output: …/backend:latest`, exit 1 — then reverted.
- Grep pattern verified positive/negative: pinned, digest, and bare `tag: latest` (non-registry) refs are correctly ignored.

This is defense-in-depth complementing operator #512 (which fixed the source): even if a *future* dependency ships a floating tag, the release check now catches it.

Closes #2389.

_Assisted-By: Claude Code_

### DIFF
--- a/scripts/check-release-pins.sh
+++ b/scripts/check-release-pins.sh
@@ -29,6 +29,7 @@ DEPS_VALUES_FILE="$DEPS_CHARTS_DIR/values.yaml"
 
 JSON_MODE=false
 VERIFY_IMAGES=false
+RENDER=false
 
 usage() {
     cat <<EOF
@@ -39,6 +40,10 @@ Validate that all image tags and chart dependencies are pinned for release.
 Options:
   --json            Output results as JSON (for CI summary consumption)
   --verify-images   Check that pinned GHCR images exist via docker manifest inspect
+  --render          Render the chart (helm template, incl. subcharts) and fail on any
+                    floating image tag (:latest/:main/:master). Catches tags inside
+                    dependency subcharts that the values/template greps cannot see
+                    (rossoctl/rossoctl#2389). Requires helm.
   -h, --help        Show this help message
 EOF
     exit 0
@@ -48,6 +53,7 @@ while [[ $# -gt 0 ]]; do
     case "$1" in
         --json)           JSON_MODE=true;     shift ;;
         --verify-images)  VERIFY_IMAGES=true; shift ;;
+        --render)         RENDER=true;        shift ;;
         -h|--help)        usage ;;
         *)                echo "Unknown option: $1"; usage ;;
     esac
@@ -451,6 +457,54 @@ if [[ -f "$DEPS_VALUES_FILE" ]]; then
             else
                 add_ok "charts/rossoctl-deps/values.yaml" \
                     "spiffeIdp.image.tag ($spiffe_tag) consistent with platform"
+            fi
+        fi
+    fi
+fi
+
+# ---------------------------------------------------------------------------
+# Check 6c (opt-in, --render): No floating image tags in the RENDERED chart
+#
+# Checks 1-2 only see rossoctl's own values.yaml/templates. They cannot see a
+# floating tag inside a dependency subchart (e.g. operator-chart's AuthBridge
+# images), which is how a :latest shipped invisibly at an earlier rc
+# (rossoctl/rossoctl#2389). Rendering the full chart (subcharts included) and
+# grepping the output for any registry ref ending in :latest/:main/:master
+# closes that gap. Opt-in because it needs helm + fetching OCI dependencies.
+# ---------------------------------------------------------------------------
+
+if [[ "$RENDER" == "true" ]]; then
+    render_label="charts/rossoctl (rendered)"
+    if ! command -v helm >/dev/null 2>&1; then
+        add_error "$render_label" 0 "--render requires helm, which is not installed"
+    else
+        # Build dependencies so subcharts are rendered too. Public OCI, no auth.
+        if ! helm dependency build "$CHARTS_DIR" >/tmp/crp-helm-dep.log 2>&1; then
+            add_error "$render_label" 0 "helm dependency build failed (cannot render subcharts) — see output"
+            cat /tmp/crp-helm-dep.log >&2 || true
+        else
+            # openshift=false avoids the OpenShift-only required-value guards;
+            # image references (incl. subcharts) render regardless.
+            if ! helm template rossoctl "$CHARTS_DIR" --set openshift=false \
+                    > /tmp/crp-render.yaml 2>/tmp/crp-render.err; then
+                add_error "$render_label" 0 "helm template failed — see output"
+                cat /tmp/crp-render.err >&2 || true
+            else
+                # Match a container-registry ref ending in a floating tag, in any
+                # field (image:, or a subchart ConfigMap value like authbridge:).
+                # Excludes version-pinned (:vX.Y.Z) and digest (@sha256:) refs.
+                float_re='(ghcr\.io|quay\.io|docker\.io|registry\.[a-z.]+)[^[:space:]"'\'']*:(latest|main|master)([^0-9A-Za-z._-]|$)'
+                matches=$(grep -nE "$float_re" /tmp/crp-render.yaml || true)
+                if [[ -z "$matches" ]]; then
+                    add_ok "$render_label" "no floating image tags in rendered manifests (subcharts included)"
+                else
+                    while IFS= read -r m; do
+                        [[ -z "$m" ]] && continue
+                        ln="${m%%:*}"
+                        ref=$(printf '%s' "$m" | grep -oE "$float_re" | head -1)
+                        add_error "$render_label" "$ln" "floating image tag in rendered output: ${ref}"
+                    done <<< "$matches"
+                fi
             fi
         fi
     fi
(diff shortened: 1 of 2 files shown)


scripts/check-release-pins.sh
```diff
@@ -456,6 +462,54 @@ if [[ -f "$DEPS_VALUES_FILE" ]]; then
...
+        add_error "$render_label" 0 "--render requires helm, which is not installed"
+    else
+        # Build dependencies so subcharts are rendered too. Public OCI, no auth.
+        if ! helm dependency build "$CHARTS_DIR" >/tmp/crp-helm-dep.log 2>&1; then
+            add_error "$render_label" 0 "helm dependency build failed (cannot render subcharts) — see output"
+            cat /tmp/crp-helm-dep.log >&2 || true
+        else
+            # openshift=false avoids the OpenShift-only required-value guards;
+            # image references (incl. subcharts) render regardless.
+            if ! helm template rossoctl "$CHARTS_DIR" --set openshift=false \
```
esnible, line 493: **suggestion** — the registry allowlist (`ghcr.io|quay.io|docker.io|registry.[a-z.]+`) covers what's in the tree today, but a subchart pulling from `gcr.io`, ECR (`*.dkr.ecr.*.amazonaws.com`), or `mcr.microsoft.com` would render a floating tag that this gate reports as clean. I verified `gcr.io/x/y:latest` and an ECR ref both pass through unmatched.

That's the same blind-spot class #2389 was — an image source the check couldn't see — just one registry over. Since the whole point here is defense-in-depth against a *future* dependency, consider widening to a generic host pattern and excluding known-good tags instead of enumerating registries:

```bash
# any ref that looks like host/path:floating-tag
float_re='[a-z0-9.-]+\.[a-z]{2,}(:[0-9]+)?/[^[:space:]"'\'']*:(latest|main|master)([^0-9A-Za-z._-]|$)'
```

Not blocking — today's chart set is fully covered.

scripts/check-release-pins.sh
```diff
@@ -456,6 +462,54 @@ if [[ -f "$DEPS_VALUES_FILE" ]]; then
...
+                    > /tmp/crp-render.yaml 2>/tmp/crp-render.err; then
```
esnible: **suggestion** — the temp paths (`/tmp/crp-helm-dep.log`, `/tmp/crp-render.yaml`, `/tmp/crp-render.err`) are fixed names with no `trap` cleanup. In CI on an ephemeral runner this is harmless, but the script is also runnable locally (and `--render` is documented in `--help`), where two concurrent invocations clobber each other's render and a multi-user box hits ownership errors on the pre-existing file.

```bash
tmpdir=$(mktemp -d)
trap 'rm -rf "$tmpdir"' EXIT
# ... > "$tmpdir/render.yaml" 2> "$tmpdir/render.err"
```

A rendered manifest also lingering in `/tmp` after the run is mild information leakage on a shared host.

scripts/check-release-pins.sh
esnible, line 493: **nit** — two subtleties in the tag alternation worth a comment, since both are easy to misread later:

1. Case-sensitive: I confirmed `ghcr.io/x/y:LATEST` is **not** matched. Unusual, but `grep -iE` would close it for free.
2. The trailing `([^0-9A-Za-z._-]|$)` intentionally excludes suffixed tags like `:latest-arm64` and `:main-abc123`. That's the right call (`:main-abc123` is effectively pinned), but it isn't obvious from reading the class — one line noting it would save the next reader the reasoning.

Also cosmetic: `ref` keeps the trailing quote on quoted refs, so the message reads `floating image tag in rendered output: ghcr.io/rossoctl/cortex/authbridge:main"`. Affects only the error text, not detection.

esnible (approved):
Verified the gate end-to-end on a worktree of this PR's head rather than reading it only — including the exact #2389 regression shape.

**Verification results**

| Claim | Result |
|---|---|
| Clean run passes | pass (exit 0, `[OK] charts/rossoctl (rendered)`) |
| Negative test catches floating tag | pass (exit 1, `floating image tag in rendered output: .../ui-v2:latest`) |
| **Catches ConfigMap-embedded subchart refs (the #2389 case)** | pass — caught in `rossoctl-platform-config` -> `config.yaml` |
| Legacy checks 1-2 miss that same case | confirmed silent — the gap is real, not hypothetical |
| Pinned / digest / bare `tag: latest` ignored | pass (all three correctly skipped) |
| `helm dependency build` works without auth | pass (anonymous OCI pull of `operator-chart` 0.3.1-alpha.1) |
| `--set openshift=false` required to render | confirmed — `openshift=true` fails on required `mcpGateway.openshiftDomain` |
| `azure/setup-helm` SHA is genuine v5.0.1 | pass (`9bc31f4e...` matches the upstream tag) |
| shellcheck | clean, 0 findings |
| `--json --render` composes | pass (valid JSON, no stderr leakage into stdout) |

The important part is row 3: I injected a floating tag into the subchart's AuthBridge ref and confirmed it surfaces in the rendered `rossoctl-platform-config` ConfigMap's `config.yaml` string — not as an `image:` field — and that the new regex catches it while checks 1-2 stay silent. That is precisely the invisible-`:latest` class from #2389, so this gate closes a real gap. It also integrates cleanly with the existing `add_error`/`add_ok` machinery (line numbers stay numeric for `jq --argjson`) and fails loudly rather than silently skipping.

Approving. Three non-blocking notes inline — all about future-proofing the pattern, none affecting the fix.

**Areas reviewed**: Shell, CI/GitHub Actions, Helm (render behavior)
**Commits**: 2 commits (1 feature + 1 merge); DCO check passing
**CI status**: all required checks passing (Deploy & Test pending, E2E HyperShift awaiting maintainer `/run-e2e`)

_Assisted-By: Claude Code_

scripts/check-release-pins.sh
cwiklik: Done in 0317585d. Widened to a generic host pattern (\`[a-z0-9.-]+\.[a-z]{2,}(:[0-9]+)?/...\`) instead of the registry allowlist, so gcr.io, ECR (\`*.dkr.ecr.*.amazonaws.com\`), and mcr.microsoft.com are now covered — verified all three MATCH while ghcr/quay/docker still do. That closes the one-registry-over blind spot rather than enumerating.

scripts/check-release-pins.sh
cwiklik: Done in 0317585d. Switched to \`crp_tmp=$(mktemp -d)\` with \`trap 'rm -rf "$crp_tmp"' EXIT\`; the three render artifacts now live under it. Verified a re-run leaves zero \`/tmp/crp-*\` and zero \`tmp.*\` dirs behind.

scripts/check-release-pins.sh
cwiklik: Done in 0317585d, all three: (1) switched the greps to \`-iE\` so \`:LATEST\` is caught (verified). (2) Added a comment explaining the trailing \`[^0-9A-Za-z._-]\` boundary intentionally excludes effectively-pinned suffixed tags (\`:latest-arm64\`, \`:main-abc123\`) — behavior unchanged, just documented. (3) Strip the captured trailing delimiter with \`ref="${ref%[!0-9A-Za-z._-]}"\`, so the message now reads \`...:main\` without the stray closing quote.

PR #2464, opened 2026-08-25
fix(mesh): proactive SVID near-expiry detection + SPIRE TTL right-sizing + 503 docs (Part of #1899)

cwiklik:
## Summary

Advances #1899 (mesh-wide 503 from expired SPIRE SVIDs after a long host suspend) beyond the merged reactive self-heal (Part A #2179 / Part B #2452) with three additions. **This is "Part of #1899", not a close** — the root cause (SPIRE re-attest deadlock / ztunnel not re-fetching) is upstream [istio/ztunnel#1679](https://github.com/istio/ztunnel/issues/1679).

## What & why

1. **Proactive near-expiry detection** (`scripts/k8s/mesh-recover.sh`) — a new `check_cert_expiry()` reads the **soonest-expiring ztunnel workload SVID** from ztunnel's admin `config_dump` (the data `istioctl ztunnel-config certificates` surfaces), via `kubectl exec` + `jq`. It emits a new advisory `WARN` status → **exit code 4** → the CronJob wrapper emits a `MeshNearExpiry` Warning Event. **Advisory only — never restarts.** Degrades to a non-fatal `SKIP` when `jq` is absent or `kubectl exec` is denied (the CronJob ServiceAccount has no `pods/exec`), so proactive warning is a manual / `k8s:health`-path capability and the CronJob stays reactive-only — **no new RBAC**.

2. **SPIRE TTL right-sizing (Kind-dev-only)** (`scripts/kind/setup-rossoctl.sh`) — CA TTL `24h → 168h (7d)`, x509 SVID TTL `→ 24h`, so a normal overnight suspend no longer expires every mesh credential at once. OpenShift path untouched.

3. **Troubleshooting docs** — a "503 after host suspend" section in `docs/users-guides/troubleshooting.md` and a proactive-detection note in the `k8s:health` skill.

## Exit-code semantics

`DEGRADED(2) > INCONCLUSIVE(3) > WARN/near-expiry(4) > OK(0)`. A skipped near-expiry check is **neutral** (does not flip a healthy mesh off `OK`).

## Validation (live, `kind-rossoctl`)

- Detection WARN reports the **real ~24h SVID** expiry → exit 4; OK → exit 0; healthy CronJob run (exec denied) → `cert-expiry: SKIP` → `MeshHealthy` (not a false `MeshInconclusive`).
- `rc=4` → `MeshNearExpiry` Warning Event with the full JSON summary, **no rollout restart** (ztunnel restart count unchanged).
- shellcheck clean on all edited scripts; `mesh-recover.sh` chart-copy sync guard passes.

## Caveats / notes for reviewers

- **TTL change is render-verified, not live-reinstalled.** `helm template` confirms `ca_ttl: 168h` and ClusterSPIFFEID `ttl: 24h` render correctly; a live SPIRE reinstall was intentionally skipped to avoid disrupting a running cluster, so it takes effect on the next fresh Kind install.
- **Design pivot vs. the committed spec.** The spec doc (`docs/superpowers/specs/2026-08-24-…`) describes an `openssl`-on-trust-bundle-CA approach. Live testing showed that only ever reads the 10-year Istio root CA (the SPIRE bundle is JWKS, not PEM), so it would never warn before a real outage; the implementation was reworked to read the actual short-lived workload SVID instead. The spec is left as the historical record.
- **Pre-existing follow-up (not fixed here):** `mesh-recover.sh check_spire_agent` uses `-l app=spire-agent`, which matches no pods (real label `app.kubernetes.io/name=agent`); it only runs under `--include-spire`. Worth a separate fix.

Assisted-By: Claude Code

### DIFF
--- a/charts/rossoctl-deps/files/mesh-selfheal-wrapper.sh
+++ b/charts/rossoctl-deps/files/mesh-selfheal-wrapper.sh
@@ -72,6 +72,7 @@ case "$rc" in
   0) emit_event Normal MeshHealthy "gateway reachable, mesh certs OK"; exit 0 ;;
   3) emit_event Warning MeshInconclusive "cannot confirm mesh health (restart would not help): ${summary}"; exit 0 ;;
   2) : ;;  # degraded/recoverable — fall through to gating
+  4) emit_event Warning MeshNearExpiry "trust-bundle CA nearing expiry (advisory; no restart): ${summary}"; exit 0 ;;
   *) log "unexpected rc=$rc; treating as inconclusive"; emit_event Warning MeshInconclusive "mesh-recover.sh rc=$rc"; exit 0 ;;
 esac
 
--- a/.claude/skills/k8s:health/SKILL.md
+++ b/.claude/skills/k8s:health/SKILL.md
@@ -247,10 +247,13 @@ kubectl get destinationrule -A
 > (`scripts/k8s/mesh-recover.sh`) to detect and recover.
 >
 > **Proactively** (before the 503): run `scripts/k8s/mesh-recover.sh` in detect
-> mode from a machine with `openssl`. Exit code **4** / a `cert-expiry: WARN`
-> finding means the SPIRE trust-bundle CA is nearing expiry — plan a data-plane
-> restart or cluster recreate before the outage. Tune the window with
+> mode from a machine with `kubectl` exec access and `jq`. It reads the soonest-
+> expiring ztunnel **workload SVID** (via ztunnel's admin `config_dump`); exit code
+> **4** / a `cert-expiry: WARN` finding means an SVID is nearing expiry — plan a
+> data-plane restart or cluster recreate before the outage. Tune the window with
 > `CERT_WARN_SECONDS` (default 6h). This is advisory: it never restarts anything.
+> (The automated CronJob's ServiceAccount has no `pods/exec`, so it skips this
+> check — proactive warning is a manual / `k8s:health` capability.)
 
 #### SPIRE (Workload Identity)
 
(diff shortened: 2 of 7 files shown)


charts/rossoctl-deps/files/mesh-selfheal-wrapper.sh
```diff
@@ -72,6 +72,7 @@ case "$rc" in
   0) emit_event Normal MeshHealthy "gateway reachable, mesh certs OK"; exit 0 ;;
   3) emit_event Warning MeshInconclusive "cannot confirm mesh health (restart would not help): ${summary}"; exit 0 ;;
   2) : ;;  # degraded/recoverable — fall through to gating
+  4) emit_event Warning MeshNearExpiry "trust-bundle CA nearing expiry (advisory; no restart): ${summary}"; exit 0 ;;
```
pdettori, line 75: **suggestion (non-blocking):** stale wording from the design pivot. This `MeshNearExpiry` Event message says *"trust-bundle CA nearing expiry"*, but per the PR's design-pivot note (and commit `d730b9b`) `check_cert_expiry` now reads the **workload SVID** expiry, not the trust-bundle CA. If this branch ever fires, the emitted Kubernetes Warning Event would be misleading. Suggest something like *"workload SVID nearing expiry"*.

(The `docs/superpowers/specs/...` doc is fine to leave as-is — it's intentionally kept as the historical CA-approach record.)

pdettori (approved):
Clean, well-documented maintainer PR advancing #1899 with proactive SVID near-expiry detection, Kind-only SPIRE TTL right-sizing, and troubleshooting docs. All CI checks pass; no blocking issues.

**Reviewed:** Shell (`mesh-recover.sh` + chart copy, `mesh-selfheal-wrapper.sh`), Helm/Kind setup (`caTTL`/ClusterSPIFFEID `ttl`), docs, and the `.claude/skills/k8s:health/SKILL.md` change (inspected in full — benign doc-only blockquote, no hooks/commands/executables).

**Verified good:**
- `scripts/k8s/mesh-recover.sh` and `charts/rossoctl-deps/files/mesh-recover.sh` are byte-identical (sync guard intact).
- Exit-code precedence `DEGRADED(2) > INCONCLUSIVE(3) > WARN(4) > OK(0)` is correct, and `OK`/`SKIP` correctly set no flag.
- All `check_cert_expiry` degradation paths (no `jq`, no `pods/exec`, empty/unparseable `config_dump`) are non-fatal `SKIP`s — a healthy mesh never flips off `OK`.

One minor accuracy suggestion inline (design-pivot leftover); non-blocking.

Assisted-By: Claude Code

PR #2479, opened 2026-08-26
Docs: Clarify LLM provider selection in weather agent demo

mrsabath:
## Problem

In the **Import New Agent** step of the weather agent demo, the environment-variable instructions gave a single Ollama `.env` URL followed by:

> If using OpenAI LLMs, skip this step

This is ambiguous. It is unclear whether "skip this step" means skip the entire Import New Agent step or just the URL import, and OpenAI users are never told what to use instead. (Raised by a user following the published demo.)

## Fix

- Give an explicit `.env` URL **per provider** so there is nothing to "skip":
  - Local model (Ollama): `.env.ollama`
  - OpenAI: `.env.openai`
- Clarify the note: the OpenAI set uses the `openaiApiKey` value configured in `deployments/envs/.secret_values.yaml` at install time, and no local model is needed. Links to the quickstart install page.
- Fix incidental typos in the same section: unclosed backtick on `Import from File/URL`, "Image Agent"/"Image Tool" -> "Weather", and the malformed `**Select Tool*`.

## Notes

The `.env.openai` file already exists in [rossoctl/examples](https://github.com/rossoctl/examples/tree/main/a2a/weather_service), so the OpenAI URL resolves. Docs-only change; no code affected.

### DIFF
--- a/docs/demos/demo-weather-agent.md
+++ b/docs/demos/demo-weather-agent.md
@@ -26,28 +26,29 @@ Here's a breakdown of the sections:
 
 #### Import New Agent
 
-To deploy the Image Agent:
+To deploy the Weather Agent:
 
 1. Navigate to [Import New Agent](http://rossoctl-ui.localtest.me:8080/Import_New_Agent#import-new-agent) in the Rossoctl UI.
 2. Under **Select Agent**, choose `Weather Service Agent`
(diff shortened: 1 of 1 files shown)


pdettori (approved):
LGTM

PR #2484, opened 2026-08-27
fix: Explain when Context Service is disabled

moonlight16:
## What & why

Context Service is optional, but its routes were previously registered only when CONTEXT_SERVICE_URL was configured. On installations without it, context commands therefore returned an ambiguous 404 that looked like an outdated Rosso server.

Always register the context routes and return an actionable 503 when the integration is disabled. The response explains that a cluster administrator must configure Context Service and links to the existing setup documentation. Agent imports with context attachments use the same response.

## Acceptance tier

### Pillar 1 — Code quality

- [x] Focused Ruff checks pass
- [x] Tests added/updated and passing
- [x] DCO sign-off on all commits

### Pillar 2 — Documentation

- [x] Existing setup documentation is linked from the error response

## Testing

- `uv run pytest tests/test_contexts.py -q` — 24 passed
- `uv run ruff check app/main.py app/routers/contexts.py app/routers/agents_manifests.py tests/test_contexts.py`
- HTTP-level coverage verifies `GET /api/v1/contexts/team1` returns 503, actionable guidance, and the setup URL when Context Service is unset.

Assisted-By: Codex

### DIFF
--- a/rossoctl/backend/app/main.py
+++ b/rossoctl/backend/app/main.py
@@ -54,12 +54,13 @@ class NoCacheMiddleware:
 from app.core.config import settings  # pylint: disable=wrong-import-position
 from app.routers import (  # pylint: disable=wrong-import-position
     agents,
-    tools,
-    namespaces,
-    config,
     auth,
     chat,
+    config,
+    contexts,
+    namespaces,
     shipwright,
+    tools,
 )
 
 # Conditionally import feature-flagged modules.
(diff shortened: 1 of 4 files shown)


evaline-ju (commented):
generally lgtm but will need conflicts resolved

evaline-ju (approved):
agent nit: 503 means "temporarily unavailable, try again later" — but a missing
CONTEXT_SERVICE_URL is a permanent administrative configuration gap, not a transient outage. 501 Not Implemented would be semantically cleaner: "this feature isn't enabled on this installation."

PR #2486, opened 2026-08-28
fix(chart): only manage openai-secret when secrets.openaiApiKey is set

### LINKED ISSUE #2485: Chart re-templates openai-secret to empty on every helm upgrade when secrets.openaiApiKey is unset
### Summary

`charts/rossoctl/templates/agent-namespace-resources.yaml` renders the agent-namespace `openai-secret` Secret **unconditionally**:

```yaml
stringData:
  apikey: {{ $.Values.secrets.openaiApiKey | quote }}
```

When `secrets.openaiApiKey` is left empty — the intended setup when the LLM API key is provisioned out-of-band (secret manager / External Secrets / manual `kubectl`) — **every `helm upgrade` re-applies an empty `apikey`**, silently overwriting a working, operator-managed key with `""`.

### Impact

Agents and tools read the key from `openai-secret` (key `apikey`) via `secretKeyRef` at pod startup. After an upgrade zeroes the key, every LLM call fails (401/timeout) and stays broken until the Secret is restored **and** the workload pods are `rollout restart`ed (running pods cache the env). This was hit during a routine `rossoctl` upgrade — it wiped `team1/openai-secret.apikey` mid-run and every benchmark task errored until the key was manually restored.

### Root cause

The Secret block is not gated on the value being present. The GitHub secrets directly above it in the same template are already conditional (`{{- if and $.Values.secrets.githubUser $.Values.secrets.githubToken }}`); `openai-secret` is the outlier.

### Proposed fix

Guard the block with `{{- if $.Values.secrets.openaiApiKey }}`, matching the existing GitHub-secret idiom. When the value is empty the chart no longer creates or overwrites the Secret, so an out-of-band `openai-secret` survives upgrades. Add a `NOTES.txt` WARNING when the value is unset so operators know they must provision the Secret out-of-band before deploying workloads.

To adopt on an existing release without losing the live key: annotate the Secret `helm.sh/resource-policy=keep` (and disown the Helm ownership metadata) **before** upgrading, so the upgrade that removes it from the release manifest does not prune it.

webchang:
Fixes #2485.

## Problem

The agent-namespace `openai-secret` Secret in `charts/rossoctl/templates/agent-namespace-resources.yaml` was rendered unconditionally:

```yaml
stringData:
  apikey: {{ $.Values.secrets.openaiApiKey | quote }}
```

With `secrets.openaiApiKey` empty (the intended state when the LLM key is provisioned out-of-band), **every `helm upgrade` re-writes an empty `apikey`**, silently zeroing an operator-managed key. Workloads read it via `secretKeyRef` at pod start, so LLM calls fail until the Secret is restored and pods are restarted.

## Change

- **Guard the Secret** with `{{- if $.Values.secrets.openaiApiKey }}`, mirroring the conditional GitHub secrets already in the same template. Empty value ⇒ the chart no longer creates or overwrites `openai-secret`, so an out-of-band Secret survives upgrades. Non-empty value ⇒ unchanged behavior (backward compatible).
- **`NOTES.txt` WARNING** when `secrets.openaiApiKey` is unset, telling operators to provision `openai-secret` out-of-band before deploying workloads.

## Validation

`helm template` (chart 0.7.0):
- key empty → `openai-secret` **not** rendered (0 objects)
- key set → rendered exactly once

Enacted on a live cluster (chart 0.7.0, `--reuse-values`):
- `helm diff upgrade` showed the **only** change was `openai-secret` dropped from the release manifest — no other churn.
- After annotating the live Secret `helm.sh/resource-policy=keep` and disowning Helm ownership metadata, the `helm upgrade` **preserved** the real key (previously this exact operation wiped it).
- NOTES `WARNING` rendered on the live upgrade.
- Post-upgrade: platform pods healthy; a gsm8k benchmark run succeeded (pass_rate 1.0, 0 errored), confirming the out-of-band Secret is consumed correctly.

## Adoption note (existing releases)

To adopt without losing a live key, **before** upgrading annotate the Secret `helm.sh/resource-policy=keep` (and strip its Helm ownership label/annotations) so the upgrade that removes it from the release manifest does not prune it.

Assisted-By: Claude Code

### DIFF
--- a/charts/rossoctl/templates/NOTES.txt
+++ b/charts/rossoctl/templates/NOTES.txt
@@ -18,6 +18,24 @@ Show all service URLs and credentials:
 ```
   .github/scripts/local-setup/show-services.sh
 ```
+{{- if not .Values.secrets.openaiApiKey }}
+
+============================================================================
+  WARNING: secrets.openaiApiKey is NOT set
+============================================================================
+The chart did NOT create/manage the `openai-secret` Secret in the agent
+namespace(s). Agents and tools that call an LLM read their API key from
+`openai-secret` (key: `apikey`) at pod startup and will fail to run until it
+exists. Provision it out-of-band BEFORE deploying workloads, e.g.:
+
+  kubectl -n <agent-namespace> create secret generic openai-secret \
+    --from-literal=apikey="$YOUR_LLM_API_KEY"
+
+Leaving secrets.openaiApiKey empty is the supported way to keep the key out
+of Helm values so that a later `helm upgrade` cannot overwrite an
+operator-managed secret with an empty value.
+============================================================================
+{{- end }}
 
 
 
--- a/charts/rossoctl/templates/agent-namespace-resources.yaml
+++ b/charts/rossoctl/templates/agent-namespace-resources.yaml
@@ -52,6 +52,11 @@ data:
 ---
 {{ end }}
 # 4. OpenAI API Key Secret
+#    Only managed by the chart when secrets.openaiApiKey is set. When left empty the Secret is
+#    expected to be provisioned out-of-band (e.g. `kubectl create secret generic openai-secret
+#    --from-literal=apikey=...`) so that `helm upgrade` does not overwrite an operator-managed
+#    key with an empty value. Mirrors the conditional GitHub secrets above.
+{{- if $.Values.secrets.openaiApiKey }}
 apiVersion: v1
 kind: Secret
 metadata:
@@ -63,6 +68,7 @@ type: Opaque
 stringData:
   apikey: {{ $.Values.secrets.openaiApiKey | quote }}
 ---
+{{- end }}
 # 5. Slack Bot Tokens Secret
 apiVersion: v1
 kind: Secret


cwiklik (changes requested):
Clean, focused fix — conditionally renders `openai-secret` only when `secrets.openaiApiKey` is non-empty, preventing `helm upgrade` from zeroing an out-of-band-managed key. Template logic is correct (validated by Helm Chart Lint, chart-tests, and Deploy & Test CI). NOTES.txt warning is a nice operator UX touch.

**Must-fix:**
- Commit `bedb32f` is missing a `Signed-off-by:` trailer — DCO CI check is failing. Amend with `git commit --amend --signoff` and force-push.

**Nit:**
- PR body footer says "Generated with Claude Code" — repo convention (CLAUDE.md) is `Assisted-By: Claude Code` for PR metadata.

cwiklik (approved):
Re-review after my prior CHANGES_REQUESTED. Both flagged items are resolved (commit re-signed → DCO passes; footer now `Assisted-By: Claude Code`).

The conditional guard on `openai-secret` is correct in all github/openai combinations — each namespace iteration emits its own leading `---`, so document separators stay valid when the block is omitted, and the pattern mirrors the existing conditional GitHub secrets in the same template. Strictly safer behavior (no more empty-value overwrite of an out-of-band-managed key). NOTES.txt operator warning is a nice touch.

**Areas reviewed**: Helm chart templates (`NOTES.txt`, `agent-namespace-resources.yaml`)
**CI**: passing (20/20, incl. Helm Chart Lint, chart-tests, Deploy & Test)

LGTM.

PR #2488, opened 2026-08-28
feat(backend): accept AuthBridge plugin preset and pass onto AgentRuntime.spec

### LINKED ISSUE #2487: AuthBridge plugin presets are not enactable via the agent-create API
## Problem

AuthBridge supports layer-3 plugin **presets** (`auth-only` / `ibac-only` / `full`), per-plugin
policy (`enforce` / `observe` / `off`), and a chain-default `on_error`. Today none of this is
reachable through the agent-create API:

- `CreateAgentRequest` has no field to select a preset or per-plugin policy, so a caller can only
  toggle the coarse `authBridgeEnabled` (which yields the operator's default two-plugin synthesis,
  `jwt-validation` + `token-exchange`).
- As a result, comparative runs that need `ibac-only`, `full`, or `full` with `ibac:observe` cannot
  be provisioned over HTTP at all — the preset must be hand-edited into the per-agent
  `authbridge-config-<agent>` ConfigMap with `kubectl`.

## Proposal

Accept the preset selection on the agent-create request and thread it onto the
`AgentRuntime.spec`, mirroring the existing `authBridgeMode` / `mtlsMode` passthrough. The operator
then renders the full canonical pipeline from those spec fields (see the companion operator issue/PR).

New optional request fields:
- `pluginPreset`: `auth-only` | `ibac-only` | `full`
- `plugins`: list of `name:policy` override tokens (e.g. `"ibac:observe"`, `"mcp-parser:off"`)
- `onError`: `enforce` | `observe` | `off` (chain-default policy for selected plugins)

All three are optional and default to `None`, so existing callers are unaffected.

## Companion

Requires the operator to render the preset into the per-agent pipeline:
rossoctl/operator (see linked PR). Validated end-to-end cross-cluster (Service → backend →
operator webhook → per-agent ConfigMap → AuthBridge sidecar), including the `enforce` vs `observe`
IBAC behavior contrast.

webchang:
Closes #2487.

## What

Adds three optional fields to the agent-create request and threads them onto the `AgentRuntime.spec`
so the operator can render the full AuthBridge layer-3 pipeline (companion: rossoctl/operator#523):

- `pluginPreset`: `auth-only` | `ibac-only` | `full`
- `plugins`: list of `name:policy` override tokens (e.g. `"ibac:observe"`)
- `onError`: `enforce` | `observe` | `off` (chain-default policy)

Mirrors the existing `authBridgeMode` passthrough. The fields flow through both the image path
(`agents_create.py`) and the build-from-source Shipwright round-trip (`agents_skills.py` →
`agents_finalize.py`), and are emitted into the manifest by `_build_agentruntime_manifest`.
All three default to `None`; existing callers are unaffected.

## Changed

- `agents_models.py` — new request fields (+ the duplicated finalize request).
- `agents_manifests.py` — set `spec.pluginPreset` / `plugins` / `onError` when present.
- `agents_create.py`, `agents_finalize.py`, `agents_skills.py`, `models/shipwright.py` — thread the
  values through the create + build-from-source paths.

## Test plan

Validated end-to-end cross-cluster (Benchmarking Service → this backend → operator webhook →
per-agent `authbridge-config-<agent>` ConfigMap → AuthBridge sidecar). Preset deploys that
previously returned 422 now provision the intended pipeline; a small IBAC-comparison benchmark
showed the expected `enforce` (denies non-essential outbound calls) vs `observe` (passes) behavior.

### DIFF
--- a/rossoctl/backend/app/routers/agents_manifests.py
+++ b/rossoctl/backend/app/routers/agents_manifests.py
@@ -238,6 +238,9 @@ def _build_agentruntime_manifest(
     auth_bridge_mode: Optional[str] = None,
     mtls_mode: Optional[str] = None,
     tls_bridge_enabled: bool = False,
+    plugin_preset: Optional[str] = None,
+    plugins: Optional[List[str]] = None,
+    on_error: Optional[str] = None,
 ) -> dict:
     """Build an AgentRuntime CR manifest for the given workload."""
     kind_map = {
@@ -267,6 +270,16 @@ def _build_agentruntime_manifest(
     # CRD field off envoy-sidecar agents so the validating webhook doesn't reject).
     if tls_bridge_enabled:
         spec["tlsBridgeMode"] = "enabled"
+    # AuthBridge layer-3 plugin composition. Only set when a preset is chosen;
+    # unset → operator keeps its default 2-plugin pipeline (jwt-validation +
+    # token-exchange), so existing agents are unaffected. plugins/onError are
+    # only meaningful alongside a preset (they refine the selected pipeline).
+    if plugin_preset:
+        spec["pluginPreset"] = plugin_preset
+        if plugins:
+            spec["plugins"] = plugins
+        if on_error:
+            spec["onError"] = on_error
     return {
         "apiVersion": f"{CRD_GROUP}/{CRD_VERSION}",
         "kind": "AgentRuntime",
@@ -291,6 +304,9 @@ def _ensure_agentruntime(
     auth_bridge_mode: Optional[str] = None,
     mtls_mode: Optional[str] = None,
     tls_bridge_enabled: bool = False,
+    plugin_preset: Optional[str] = None,
+    plugins: Optional[List[str]] = None,
+    on_error: Optional[str] = None,
 ) -> None:
     """Create an AgentRuntime CR for the workload. Skip if it already exists."""
     manifest = _build_agentruntime_manifest(
@@ -301,6 +317,9 @@ def _ensure_agentruntime(
         auth_bridge_mode,
         mtls_mode,
         tls_bridge_enabled,
+        plugin_preset,
+        plugins,
+        on_error,
     )
     try:
         kube.create_custom_resource(
--- a/rossoctl/backend/app/models/shipwright.py
+++ b/rossoctl/backend/app/models/shipwright.py
@@ -139,6 +139,12 @@ class ResourceConfigFromBuild(BaseModel):
     # that read it back via model_dump() during Shipwright finalize.
     k8sResourceLimits: Optional[Dict[str, str]] = None
     k8sResourceRequests: Optional[Dict[str, str]] = None
+    # AuthBridge layer-3 plugin composition stashed at form-submit time. Declared
+    # here for the same reason as k8sResourceLimits above — undeclared annotation
+    # keys are dropped by this model's default extra="ignore".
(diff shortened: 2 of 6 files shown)


cwiklik (changes requested):
Well-structured feature — threads three new optional AuthBridge plugin-composition fields (`pluginPreset`, `plugins`, `onError`) through both the direct-image and build-from-source paths, mirroring the existing `authBridgeMode` pattern. Proper `Literal` type constraints at the API boundary, conditional nesting in `_build_agentruntime_manifest` (plugins/onError only set alongside a preset) is clean, and the Shipwright stash/read-back round-trip is consistent with the other fields.

**Must-fix:**
- Commit `9c01e94` is missing a `Signed-off-by:` trailer — DCO CI check is failing. Amend with `git commit --amend --signoff` and force-push.

Code itself LGTM — just needs the sign-off.

rossoctl/backend/app/routers/agents_manifests.py
```diff
@@ -267,6 +270,16 @@ def _build_agentruntime_manifest(
     # CRD field off envoy-sidecar agents so the validating webhook doesn't reject).
     if tls_bridge_enabled:
         spec["tlsBridgeMode"] = "enabled"
+    # AuthBridge layer-3 plugin composition. Only set when a preset is chosen;
+    # unset → operator keeps its default 2-plugin pipeline (jwt-validation +
+    # token-exchange), so existing agents are unaffected. plugins/onError are
+    # only meaningful alongside a preset (they refine the selected pipeline).
+    if plugin_preset:
```
cwiklik, line 277: suggestion (non-blocking): `plugins`/`onError` are only emitted alongside a `pluginPreset` (nested under this `if plugin_preset:`). A caller that sends `plugins`/`onError` **without** a preset gets a silent no-op. The field docstrings say these "refine the selected pipeline," so if the operator contract genuinely requires a preset, consider rejecting that combination at the API boundary (422) rather than silently dropping it — otherwise it's an easy-to-miss surprise for API consumers. Fine to leave as-is if the operator treats them as preset-only refinements.

cwiklik (approved):
Re-review after my prior CHANGES_REQUESTED — the DCO must-fix is resolved (commit re-signed, DCO passes).

Threads three optional AuthBridge plugin-composition fields (`pluginPreset`/`plugins`/`onError`) onto `AgentRuntime.spec`, mirroring the existing `authBridgeMode`/`mtlsMode` passthrough. **Verified full field parity** across both deploy paths — the direct-image path (`agents_create` → manifest) and the build-from-source Shipwright stash→readback round-trip (`agents_skills` stash → `ResourceConfigFromBuild` → `agents_finalize` readback → model reconstruct + `_ensure_agentruntime`). All three fields thread through every stage; no drops. `Literal` constraints at the boundary, `extra=ignore` readback handled by declaring the fields on `ResourceConfigFromBuild`, backward compatible (all default `None`).

One non-blocking inline suggestion on preset-gating. Nit: the PR body footer says "🤖 Generated with Claude Code" — repo convention (CLAUDE.md) is `Assisted-By: Claude Code`. Optional: a unit test asserting the fields reach the manifest (and the preset-gating) would lock the behavior, though this matches the existing untested passthrough pattern and you validated it end-to-end.

**Areas reviewed**: Python backend (models, routers, manifest builder) + multi-path parity
**CI**: passing (E2E HyperShift gated on maintainer /run-e2e)

LGTM.

PR #2507, opened 2026-08-31
fix(backend): don't stamp rossoctl.io/type on Shipwright-finalized workloads (#2489)

### LINKED ISSUE #2489: Agent import with Workload Type 'Deployment' fails: 403 from agent-label-protection policy, no workload created
## Summary

When importing an **agent** via the Rossoctl UI and selecting **Workload Type: `Deployment`** (instead of the default `Sandbox`), the Shipwright source build succeeds but **no workload and no AgentRuntime are created**. The agent silently never deploys.

## Root cause

The backend's `finalize_shipwright_build` (`rossoctl/backend/app/routers/agents.py:4652`) attempts to create a `Deployment` with the `rossoctl.io/type` label already set. This is rejected by a ValidatingAdmissionPolicy:

```
deployments.apps "weather-service-advanced" is forbidden:
ValidatingAdmissionPolicy 'agent-label-protection' with binding 'agent-label-protection' denied request:
The rossoctl.io/type label on team1/weather-service-advanced can only be applied by the
rossoctl-operator via an AgentRuntime CR. Create an AgentRuntime targeting this workload
instead of manually setting the label.
code: 403
```

So the Deployment code path sets a label that only the operator (via an AgentRuntime CR) is allowed to set. The `Sandbox` path goes through the operator and works correctly.

## Impact

- The build completes (and re-triggers), consuming resources, but the user gets no pod, no AgentRuntime, and no error surfaced in the UI.
- Following any doc that says 'Workload Type: Deployment' (e.g. the weather-agent demos) leads straight into this.

## Repro

1. Import an agent via the UI (build from source).
2. Set **Workload Type** to `Deployment` (override the `Sandbox` default).
3. Deploy. Observe the build pod completes but no `agentruntime`/pod for the agent appears.
4. Backend logs show the 403 above.

## Expected

Either: the backend should create the agent via an AgentRuntime CR (not a raw labeled Deployment) for the Deployment workload type too; or the UI should not offer 'Deployment' for agents if it cannot work; or the failure should surface clearly in the UI.

## Environment

- Kind on macOS, operator v0.7.0.
- Found while running `authbridge/demos/weather-agent/demo-ui-advanced.md`.

mrsabath:
## Summary

Fixes #2489 — importing an agent with **Workload Type: `Deployment`** succeeded at build time but silently created no Deployment, pod, or AgentRuntime. E2E on Kind uncovered **two** independent bugs on the Deployment finalize path; both are fixed here.

### Bug 1 — `rossoctl.io/type` label 403 (the reported symptom)
`finalize_shipwright_build` re-applied *all* `rossoctl.io/*` labels from the Shipwright Build onto the raw workload, including `rossoctl.io/type`. The `agent-label-protection` ValidatingAdmissionPolicy only lets the operator SA set that label (via `AgentRuntime` reconciliation), so the create was rejected with 403.

**Fix:** new pure helper `_carryover_workload_labels()` (`agents_manifests.py`) carries every `rossoctl.io/*` label **except** `rossoctl.io/type` — mirroring `_build_common_labels`, which already omits it. Used in the Deployment / StatefulSet / Job branches of `finalize_shipwright_build`. The operator stamps `rossoctl.io/type` afterwards from the AgentRuntime CR. Sandbox and Service paths unchanged.

### Bug 2 — `AttributeError: contexts` → 500 (found during E2E)
With bug 1 fixed, finalize still 500'd: `finalize_shipwright_build` reads `request.contexts` unconditionally, but the router-split refactor added the `k8sResourceLimits`/`k8sResourceRequests` mirror fields to `FinalizeShipwrightBuildRequest` and **missed `contexts`**. Every Deployment/StatefulSet/Job finalize raised `AttributeError` before any workload was created (and looped in the build reconciler).

**Fix:** mirror `CreateAgentRequest.contexts` on `FinalizeShipwrightBuildRequest` (`Optional`, default `None`).

### UI — silent-failure UX
The build-progress auto-finalize `useEffect` re-fired on every poll tick after a finalize error (phase stays `Succeeded`), hammering the backend and preventing the error Alert from stabilizing — matching the issue's "build re-triggers, consuming resources." New `shouldAutoFinalize()` helper + a `hasAutoFinalized` latch make it fire at most once per build; the user re-runs via the existing **Retry** button. Applied to `BuildProgressPage.tsx` and `ToolBuildProgressPage.tsx`.

## Testing

- **Backend:** `test_shipwright_finalize_labels.py` (label-strip + `contexts` field regression); full suite 782 passed; ruff clean. (Pre-existing unrelated `test_migration.py` collection error not touched.)
- **UI:** `buildFinalize.test.ts` (fire-once / no-retrigger); all 93 vitest pass; `tsc -b` + build pass; eslint clean.
- **Kind E2E (verified):** agent imported as Workload Type `Deployment` finalizes cleanly. Backend log:
  ```
  Created ServiceAccount 'weather-deploy-test2'
  Created Deployment 'weather-deploy-test2' with image '...:v0.0.1'
  Created Service 'weather-deploy-test2'
  Created AgentRuntime 'weather-deploy-test2'
  Successfully finalized orphaned build 'team1/weather-deploy-test2'
  ```
  Result: Deployment 1/1, AgentRuntime `type=agent` READY, pod Running 1/1, no `agent-label-protection` 403, no `AttributeError`.

Assisted-By: Claude Code


esnible (approved):
### Summary
Fixes two independent, verified bugs blocking the Deployment/StatefulSet/Job finalize path (#2489): a 403 from carrying `rossoctl.io/type` onto raw workloads (violates the `agent-label-protection` admission policy), and a missing `contexts` field on `FinalizeShipwrightBuildRequest` causing an `AttributeError` → 500. Also fixes a UI retrigger loop where auto-finalize hammered the backend after a finalize error. Both backend claims were independently verified against `origin/main` (the label-omission pattern already used in `_build_common_labels`, and the unconditional `request.contexts` access in `finalize_shipwright_build`) and hold up. New tests cover the regression on both backend (pytest) and frontend (vitest).

**Author**: mrsabath (MEMBER — maintainer)
**Areas reviewed**: Python (backend routers/models/tests), Frontend (TSX pages, new util + test)
**Commits**: 4 commits (3 substantive + 1 merge), all substantive commits signed-off: yes
**CI status**: passing (E2E HyperShift and Deploy & Test are pending manual `/run-e2e` trigger — not blocking; all required checks green)

No blocking issues. Clean, well-scoped fix with good regression coverage.

PR #2510, opened 2026-09-01
fix(ci): Save two minutes per test -- fix weather agent deploy to detect secrets correctly

esnible:
## Summary

`74-deploy-weather-agent.sh` waited on `secret/rossoctl-keycloak-client-credentials`, but the operator creates the secret with a generated suffix (`rossoctl-keycloak-client-credentials-<suffix>`). The exact-name `kubectl wait --for=create` could never match, so the step always burned its full 120s and fell through to the warning path.

Replaced with the poll-and-grep pattern already used by `token-exchange/70-deploy-test-workloads.sh`.

## Changes

- Poll `kubectl get secrets -n team1 -o name | grep -c rossoctl-keycloak-client-credentials`, 24 attempts × 5s (same 120s budget).
- Threshold `-ge 1` (one workload here; the token-exchange script uses `-ge 2` for its two deployments).
- Timeout stays non-fatal: same warning plus the three diagnostic dumps.
- Kept `|| echo "0"` so `grep -c` exiting 1 on zero matches doesn't trip `set -e`.

## Testing

`bash -n` passes. Exercised by the operator E2E path that runs this script.

Assisted-By: Claude Code

### DIFF
--- a/.github/scripts/operator/74-deploy-weather-agent.sh
+++ b/.github/scripts/operator/74-deploy-weather-agent.sh
@@ -174,13 +174,21 @@ log_success "Weather-service deployed via Deployment + Service (operator-indepen
 # Keycloak and creates the Secret.  Wait here to avoid a flaky timeout later.
 # ============================================================================
 log_info "Waiting for operator to create client credentials secret..."
-if ! kubectl -n team1 wait --for=create secret/rossoctl-keycloak-client-credentials --timeout=120s 2>/dev/null; then
+CRED_FOUND=false
+for _ in $(seq 1 24); do
+    CRED_COUNT=$(kubectl get secrets -n team1 -o name 2>/dev/null | grep -c rossoctl-keycloak-client-credentials || echo "0")
+    if [[ "$CRED_COUNT" -ge 1 ]]; then
+        log_success "Found $CRED_COUNT client credentials secret(s) created by operator"
+        CRED_FOUND=true
+        break
+    fi
+    sleep 5
+done
+if [[ "$CRED_FOUND" != "true" ]]; then
     log_warn "Credentials secret not found after 120s — pod may be stuck in ContainerCreating"
     kubectl get pods -n rossoctl-system 2>&1 || true
     kubectl get secrets -n team1 2>&1 || true
     kubectl logs -n rossoctl-system -l app.kubernetes.io/instance=rossoctl --tail=20 2>&1 || true
-else
-    log_success "Client credentials secret created by operator"
 fi
 
 # WORKAROUND: Fix Service targetPort mismatch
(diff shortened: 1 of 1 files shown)


clawgenti (commented):
Fixes a real bug where `kubectl wait --for=create` on the exact secret name silently burned the full 120s timeout because the operator appends a generated suffix. The poll-and-grep pattern is already established in the repo and this is a correct, consistent application of it.

---

.github/scripts/operator/74-deploy-weather-agent.sh
```diff
@@ -174,13 +174,21 @@ log_success "Weather-service deployed via Deployment + Service (operator-indepen
 # Keycloak and creates the Secret.  Wait here to avoid a flaky timeout later.
 # ============================================================================
 log_info "Waiting for operator to create client credentials secret..."
-if ! kubectl -n team1 wait --for=create secret/rossoctl-keycloak-client-credentials --timeout=120s 2>/dev/null; then
+CRED_FOUND=false
+for _ in $(seq 1 24); do
+    CRED_COUNT=$(kubectl get secrets -n team1 -o name 2>/dev/null | grep -c rossoctl-keycloak-client-credentials || echo "0")
```
cwiklik: suggestion (non-blocking): `grep -c` prints `0` **and** exits 1 on zero matches, so `... | grep -c ... || echo "0"` captures `$'0\n0'` (a double value). On every poll iteration where the secret isn't there yet — i.e. the normal wait path — the next line `[[ "$CRED_COUNT" -ge 1 ]]` then logs `bash: [[: 0\n0: arithmetic syntax error in expression` to stderr. It's functionally harmless (the `[[ ]]` is an if-condition so `set -e` doesn't abort, and it correctly doesn't break), but it's misleading noise in the CI log. Fix: use `|| true` instead of `|| echo "0"` — `grep -c` already emits its own `0` to stdout, so you get a clean single `0`. Verified under bash.

cwiklik (approved):
Correctly fixes the exact-name `kubectl wait --for=create` that could never match the operator's suffixed secret (`rossoctl-keycloak-client-credentials-<suffix>`), replacing it with the poll-and-grep pattern already used by `token-exchange/70-deploy-test-workloads.sh`. Logic is sound, the 120s budget is preserved (24×5s), the `-ge 1` threshold is right for the single workload here, and the timeout stays non-fatal with the diagnostic dumps. DCO signed, no `.claude`/`.vscode`.

One non-blocking suggestion inline on the `grep -c` guard (double-value on the zero-match path → a bash arithmetic-error line each poll iteration). LGTM.

**Areas reviewed**: Shell (CI script)
**CI**: passing (Deploy & Test running; E2E HyperShift gated on maintainer /run-e2e)

Issue #2527, opened 2026-09-08
docs: guard against Ollama model-name drift in demo docs (source of truth in rossoctl/examples)


mrsabath:
## Problem

The Ollama **default model name** is documented as a `**Note:**` block that is duplicated verbatim (or near-verbatim) across five demo docs:

- `docs/demos/demo-weather-agent.md` — `llama3.2:3b-instruct-fp16`
- `docs/demos/demo-image-agent.md` — `llama3.2:3b-instruct-fp16`
- `docs/demos/demo-generic-agent.md` — `llama3.2:3b-instruct-fp16`
- `docs/demos/demo-file-organizer-agent.md` — `llama3.2:3b-instruct-fp16`
- `docs/demos/demo-slack-research-agent.md` — `granite3.3:8b`

The **source of truth** for each model is the `.env.ollama` file in a *different repo*
(`rossoctl/examples`, e.g. `a2a/weather_service/.env.ollama`). Nothing in this repo's CI
can detect the docs drifting out of sync with those files — which is exactly the drift
that PR #2516 (C3) fixes by hand.

## Proposed guard (from @huang195's review on #2516)

A CI step in `.github/workflows/docs-ci.yaml` that fetches the raw `.env.ollama` from
`rossoctl/examples` and greps that the model string appears in the corresponding demo doc.

## Tradeoffs to weigh before implementing

- Catches drift automatically, fork-safe (read-only, no secrets)
- **Couples this repo's CI to another repo's contents** — a rename/move in `rossoctl/examples` breaks CI here
- **Network flake** — external raw.githubusercontent fetch on every docs PR

## Alternatives worth considering

1. **Cross-repo grep guard** (as proposed) — simplest, but the coupling/flake above.
2. **Vendor the model name** into this repo (a small checked-in fixture the docs and CI both reference) — no network, but a second thing to keep in sync.
3. **Single shared include / snippet** for the Note block so the string lives in one place per model — removes the *duplication* even if it doesn't solve cross-repo drift.

## Scope

Docs/CI only. Not blocking any demo. C3 (#2516) lands independently.

_Assisted-By: Claude Code_

mrsabath:
### Investigation update — the source-of-truth situation is worse than a name mismatch

Checked `rossoctl/examples` for each demo's `.env.ollama`:

| Demo doc | Doc says | `.env.ollama` in examples repo? |
|---|---|---|
[...]

PR #2516, opened 2026-09-04
docs: fix Weather Agent Ollama model name (C3)

mrsabath:
## Summary

Both Weather Agent demo docs told readers the default Ollama model is `gpt-oss:latest`, but the agent's env file (`rossoctl/examples` → `a2a/weather_service/.env.ollama`) sets:

```
LLM_MODEL=llama3.2:3b-instruct-fp16
```

So a reader following the note runs `ollama pull gpt-oss:latest` — the wrong model for what the demo actually deploys. Corrected the note in both:

- `docs/demos/demo-weather-agent.md`
- `docs/demos/cli-demo-weather-agent.md`

## Verification

- Source of truth confirmed: `.env.ollama` = `llama3.2:3b-instruct-fp16`.
- Checked the rest of both docs while here — the linked targets (`getting-started/install.md`, `overview/5-quickstart.md`, `users-guides/troubleshooting.md`) all resolve, and no other stale references were found in these files.
- `docs/getting-started/llms/local-models.md` uses `qwen2.5:3b` as a generic example and lists `gpt-oss:latest` in a "Tested Models" table for *other* agents — neither is what the weather demo's env sets, so the demo docs are matched to the env file rather than to that table.

This is correction **C3** from the weather-agent demo verification effort; it was intended for #2479 but merged without it. Pairs with the merged demo-fix PRs rossoctl/cortex#818 and rossoctl/cortex#819.

Fixes #2515

Assisted-By: Claude (Anthropic AI) <noreply@anthropic.com>

PR #2529, opened 2026-09-08
ci(docs): guard weather demo against Ollama model-name drift

mrsabath:
## What

Adds a **Docs CI** job (`ollama-model-drift`) that fetches the source-of-truth
`.env.ollama` from `rossoctl/examples` and asserts the weather demo doc
references the same `LLM_MODEL`. This catches the exact class of drift that was
fixed by hand in #2516 (C3).

Implements @huang195's suggestion from the #2516 review, scoped down to what's
actually buildable today.

## Why scoped to the weather demo only

While implementing the guard I found the drift is deeper than a stale string:
**only the weather service ships an `.env.ollama`** in `rossoctl/examples`. The
image / generic / file-organizer / slack demos all instruct users to import an
`.env.ollama` that **does not exist** in the examples repo, so there is nothing
to check them against yet.

| Demo doc | `.env.ollama` in examples repo? |
|---|---|
| weather | ✅ `a2a/weather_service/.env.ollama` → `LLM_MODEL=llama3.2:3b-instruct-fp16` |
| image / generic / file-organizer | ❌ only `.env.openai` |
| slack | ❌ no `.env.ollama`; `.env.template` has `TASK_MODEL_ID` (different var) |

Extending the guard to the other four is blocked on the examples repo
publishing those files (or the docs dropping the reference). Full breakdown and
tracking in #2527.

## Design notes

- Follows the existing `docs-ci.yaml` gating/advisory convention.
- **Real drift** (script exit 1) **fails** the job.
- A **source-fetch failure** (exit 2, e.g. transient `raw.githubusercontent.com`
  flake) is **advisory** — it warns but does not fail, so this repo's CI does not
  go red on another repo's/CDN's momentary unavailability. This directly answers
  the cross-repo-coupling / network-flake tradeoff @huang195 flagged.
- Fetch has a 3-try retry with backoff.

## Testing

- Pre-C3 doc (`gpt-oss:latest`) → script exits 1, drift detected ✅
- Post-C3 doc (`llama3.2:3b-instruct-fp16`) → script exits 0, in sync ✅
- `bash -n` clean; YAML validated.

Layers cleanly on top of the now-merged C3 (#2516).

Refs: #2527
Assisted-By: Claude Code

### DIFF
--- a/.github/scripts/docs/check-ollama-model-drift.sh
+++ b/.github/scripts/docs/check-ollama-model-drift.sh
@@ -1,0 +1,78 @@
+#!/usr/bin/env bash
+#
+# check-ollama-model-drift.sh
+#
+# Guards against the demo docs drifting out of sync with the Ollama model name
+# that is the source of truth in the rossoctl/examples repo.
+#
+# Currently scoped to the weather demo ONLY, because it is the sole agent that
+# actually publishes an `.env.ollama` in rossoctl/examples. The image, generic,
+# file-organizer, and slack demos reference an `.env.ollama` that does not yet
+# exist in the examples repo -- see rossoctl/rossoctl#2527. Extend the CHECKS
+# array once those files are published.
+#
+# Exit codes: 0 = in sync, 1 = drift detected, 2 = could not fetch source.
+
+set -euo pipefail
+
+# doc_path <TAB> raw_env_url
+CHECKS=(
+  "docs/demos/demo-weather-agent.md	https://raw.githubusercontent.com/rossoctl/examples/refs/heads/main/a2a/weather_service/.env.ollama"
+)
+
+fetch_with_retry() {
+  # $1 = url. Retries to absorb transient network flake.
+  local url="$1" attempt
+  for attempt in 1 2 3; do
+    if curl -fsSL --max-time 20 "$url"; then
+      return 0
+    fi
+    echo "  fetch attempt ${attempt} failed for ${url}" >&2
+    sleep $((attempt * 2))
+  done
+  return 1
+}
+
+status=0
+for entry in "${CHECKS[@]}"; do
+  doc="${entry%%$'\t'*}"
+  url="${entry##*$'\t'}"
+
+  echo "Checking ${doc} against ${url}"
+
+  if [[ ! -f "${doc}" ]]; then
+    echo "::error file=${doc}::doc not found" >&2
+    status=1
+    continue
+  fi
+
+  env_contents="$(fetch_with_retry "${url}")" || {
+    echo "::error::could not fetch ${url} after retries (network flake?)" >&2
+    status=2
+    continue
+  }
+
+  # Extract LLM_MODEL=... (tolerate optional quotes/whitespace).
+  model="$(printf '%s\n' "${env_contents}" \
+    | grep -iE '^[[:space:]]*LLM_MODEL[[:space:]]*=' \
+    | head -1 \
+    | sed -E 's/^[^=]*=[[:space:]]*//; s/^["'\'']//; s/["'\'']$//; s/[[:space:]]*$//')"
+
+  if [[ -z "${model}" ]]; then
+    echo "::error::no LLM_MODEL found in ${url}" >&2
+    status=2
+    continue
+  fi
+
+  if grep -qF -- "${model}" "${doc}"; then
+    echo "  OK: '${model}' present in ${doc}"
+  else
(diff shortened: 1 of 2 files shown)


.github/scripts/docs/check-ollama-model-drift.sh
```diff
@@ -0,0 +1,78 @@
...
+  else
+    echo "::error file=${doc}::model '${model}' from ${url} not found in ${doc} (docs drifted?)" >&2
+    status=1
+  fi
+done
+
+if [[ "${status}" -eq 0 ]]; then
+  echo "All Ollama model references are in sync."
+fi
+exit "${status}"
```
esnible, line 78: **suggestion** — `status` is overwritten rather than accumulated, so a later exit-2 clobbers an earlier exit-1.

Since `docs-ci.yaml` converts exit 2 into `exit 0` with only a warning, the sequence "entry A drifts, entry B fails to fetch" produces a **green check while real drift is present**. I reproduced this locally with a two-entry array: final exit was 2, and the workflow would pass it.

This is latent today because `CHECKS` has a single entry — but both this script's header comment and the workflow comment tell maintainers to extend the array, so the bug lands the moment someone follows those instructions.

Suggested fix — track the two conditions separately and let drift win:

```bash
drift=0; fetch_fail=0
# on drift detected:  drift=1
# on fetch failure:   fetch_fail=1

if [[ "$drift" -eq 1 ]]; then exit 1; fi
if [[ "$fetch_fail" -eq 1 ]]; then exit 2; fi
echo "All Ollama model references are in sync."
exit 0
```

That way a genuine drift always fails the job even if an unrelated entry had a network flake in the same run.

.github/scripts/docs/check-ollama-model-drift.sh
```diff
@@ -0,0 +1,78 @@
...
+# doc_path <TAB> raw_env_url
+CHECKS=(
+  "docs/demos/demo-weather-agent.md	https://raw.githubusercontent.com/rossoctl/examples/refs/heads/main/a2a/weather_service/.env.ollama"
+)
+
+fetch_with_retry() {
+  # $1 = url. Retries to absorb transient network flake.
+  local url="$1" attempt
+  for attempt in 1 2 3; do
+    if curl -fsSL --max-time 20 "$url"; then
```
esnible, line 27: **suggestion** — `docs/demos/cli-demo-weather-agent.md` is checkable today and worth adding now.

That doc hardcodes this exact URL at line 47 (`--envVarsURL https://raw.githubusercontent.com/rossoctl/examples/refs/heads/main/a2a/weather_service/.env.ollama`) and restates the model name at line 66. Unlike the image/generic/file-organizer/slack docs, it is **not** blocked on #2527 — the source of truth already exists.

I confirmed it currently contains `llama3.2:3b-instruct-fp16`, so adding this entry passes immediately:

```bash
"docs/demos/cli-demo-weather-agent.md	https://raw.githubusercontent.com/rossoctl/examples/refs/heads/main/a2a/weather_service/.env.ollama"
```

Small aside while you're in here: the header comment says weather is "the sole agent that actually publishes an `.env.ollama`" in the examples repo. There are actually three — `weather_service`, `simple_generalist` (`LLM_MODEL=gpt-oss:latest`), and `git_issue_agent` (uses `TASK_MODEL_ID`, so genuinely out of scope). The PR body's table is accurate about the demos it lists; it's just this broader "sole agent" phrasing that could mislead whoever extends the array next.

esnible (approved):
Well-scoped CI guard that does what it claims. Verified end to end:

- Source of truth fetched from `rossoctl/examples`: `LLM_MODEL=llama3.2:3b-instruct-fp16` ✅
- `docs/demos/demo-weather-agent.md` contains that string, so the check passes on `main` ✅
- `shellcheck` clean on the new script ✅
- Exercised the fetch-failure path locally: exits 2, workflow converts to warning ✅

The advisory-on-fetch-failure design is a good answer to the cross-repo-coupling concern raised in the #2516 review.

Two non-blocking suggestions inline. The first is the one I'd actually act on before extending this: the exit-code precedence means the advisory path can swallow genuine drift once `CHECKS` grows beyond one entry — which the code comments actively invite maintainers to do.

**Areas reviewed**: Shell, CI/GitHub Actions, Docs
**CI status**: all lint/security checks passing; `Deploy & Test` pending, `E2E HyperShift` awaiting `/run-e2e`

Issue #963, opened 2026-03-16
feature:  Create a first MVP which run an Exgentic benchmark (e.g. appworld, tau2) against existing Kagenti generalist agent


yoavkatz:
### Feature Description

We want to demonstrate that [Exgentic](https://github.com/Exgentic/exgentic/) benchmarks can be seemlessly integrated into Kagenti and become a platform for generating realistic and scalable agent work loads  for platform validation (see Epic #962)

Exgentic takes leading benchmarks like Tau-Bench , Appworld, BrowseComp, SWE bench verified) and exposes them through standard protocols (e.g. MCP) so any agent can access them, without modifying the agent code.

### Proposed Solution

1. Add a mcp server wrapper to Exgentic benchmark (part of Exgentic code base) which allows creating an arbitrary number of sessions , each exposing a task in the  benchmark.
2. Create test harness for Exgentic in  https://github.com/kagenti/workload-harness
3. Run the test harness against the generalist model in Kagenti with manual deployment
4. Add automatic deployment of agent and mcp benchmarks as part of the harness

## High-Level Architecture
The system comprises:
- **Exgentic MCP Server** exposing:
  - *Admin tools:* `create_session`, `close_session`, `evaluate_session`  
  - *Agent tools:* `action(session_id, ...)`   
- **Kagenti MCP wrapper** to integrate the MCP server with Kagenti (dockerize it)  
- A **Kagenti/Exgentic test harness** for end-to-end benchmarking.   

## Execution Model
1. For each required load unit (parallel thread):
   - `(session_id, task) = benchmark_mcp.call("create_session")`   
   - `startTime = time()`   
   - `agent.invoke_agent("{task}. Use session id {session_id} in all accesses")`   
   - `completionTime = time() – startTime`   
   - `success = benchmark_mcp.call("evaluate_session", {"session_id": session_id})`  
   - `stats[session_id] = (completion_time, success)`; l
2. Aggregate into overall statistics.   

- Initially we will support benchmarks that do not require using docker within them.

### Want to contribute?

- [x] I would like to work on this issue.

yoavkatz:
This task was completed .  See 

https://github.com/kagenti/workload-harness/blob/main/exgentic_a2a_runner/README.md

Issue #960, opened 2026-03-16
fix: E2E CI weather-service fails to connect to weather-tool due to missing serviceAccountName


pdettori:
## Summary

The E2E Kind CI test `test_agent_simple_query` fails with:

```
Error: Cannot connect to MCP weather service at http://weather-tool-mcp.team1.svc.cluster.local:8000/mcp
```

This works in local installs where the kagenti-webhook is not injecting sidecars, but fails in CI where the webhook IS active.

## Root Cause

The static deployment YAMLs used by CI (`weather_service_deployment.yaml`, `weather_service_deployment_ocp.yaml`) and the weather-tool deployment script do not set `serviceAccountName`. The kagenti-webhook's `deriveWorkloadName()` falls back to the ReplicaSet name (which includes a hash suffix), producing a wrong SPIFFE ID:

```
spiffe://localtest.me/ns/team1/sa/weather-service-554ff44969   ← hash suffix!
```

This triggers a failure chain:

1. `kagenti-client-registration` sidecar finds an expected env var missing (keyed off the correct workload name) and **skips registration**
2. The gRPC ext-proc server on `:9090` is never started
3. Envoy's `ext_proc` filter cannot connect to `127.0.0.1:9090` → `Connection refused`
4. All outbound traffic from the weather-service pod (routed through envoy via `proxy-init` iptables rules) fails
5. The weather-service agent cannot reach `weather-tool-mcp:8000/mcp`

CI log evidence:
```
spiffe_id="spiffe://localtest.me/ns/team1/sa/weather-service-554ff44969"
Expected environment variable missing. Skipping client registration of spiffe://localtest.me/ns/team1/sa/weather-service-554ff44969.
delayed connect error: Connection refused   ← envoy ext_proc to 127.0.0.1:9090
upstream reset: reset reason: remote connection failure
```

## Why It Works Locally

In a typical local install the kagenti-webhook is either not deployed or the `team1` namespace is not labeled for injection, so no sidecars are injected and traffic flows directly to `weather-tool-mcp`.

## Fix

Add `serviceAccountName` to the static deployment manifests and the weather-tool deployment script so the webhook derives the correct workload name:

- `kagenti/examples/agents/weather_service_deployment.yaml` → `serviceAccountName: weather-service`
- `kagenti/examples/agents/weather_service_deployment_ocp.yaml` → `serviceAccountName: weather-service`
[...]

PR #2551, opened 2026-09-11
docs: Cortex laptop observability — read the numbers, first-run path, troubleshooting

mrsabath:
## What & why

Documentation for the **Cortex v0.9.0** laptop tool, the observability release ([rossoctl/cortex#962](https://github.com/rossoctl/cortex/issues/962)). It gives a first-time user, on their own workstation, everything needed to install Cortex, watch a session, and understand what the numbers mean and why the tool is worth leaving running. This backs the epic's **exit criterion 5** — a new user self-serves from the docs alone.

Written against the v0.9.0 target UX (`install.sh`, `abctl observe`, `abctl service`, tool pruning). Where a detail is not yet verifiable against a shipped binary, an HTML `VERIFY v0.9.0` comment marks it for the outside-reader check.

> **Draft.** #963 depends on the metrics work landing ([#950](https://github.com/rossoctl/cortex/issues/950), [#951](https://github.com/rossoctl/cortex/issues/951), [#952](https://github.com/rossoctl/cortex/issues/952)). This PR is ready for review of structure and voice now; the 5 `VERIFY v0.9.0` markers and the worked-example figures should be confirmed against a v0.9.0 binary before it leaves draft.

## Sub-issues addressed

| Issue | Where |
|---|---|
| [#963](https://github.com/rossoctl/cortex/issues/963) interpret metrics + value | New `get-started/reading-the-numbers.md` |
| [#959](https://github.com/rossoctl/cortex/issues/959) first-run path | `get-started/laptop.md` spine: install → observe → read → stop |
| [#960](https://github.com/rossoctl/cortex/issues/960) laptop troubleshooting | New "On a laptop" section in `operate/troubleshooting.md` |
| [#961](https://github.com/rossoctl/cortex/issues/961) separate laptop from Kubernetes | Laptop reader stays in `get-started/`; troubleshooting split laptop / Kubernetes |
| [#867](https://github.com/rossoctl/cortex/issues/867) explain the `/` filter | Documented inline in the abctl keys (substring match on METHOD, not a query language) |
| [#844](https://github.com/rossoctl/cortex/issues/844) Context Guru value | "What you see for it" + the pruning-savings section |

## The new page

`get-started/reading-the-numbers.md` covers:

- The value case, readable **before** installing, and a "the data stays on your computer" note — Cortex captures the traffic locally and sends it nowhere.
- The `abctl observe` views and keys, including the `/` filter.
- The five token categories, and **why cache-read and cache-write are priced apart**.
- How to read cost (estimated vs. authoritative), latency (TTFT vs. total, percentiles), and the tool-pruning savings, **including what a zero means**.
- A worked example on a representative session.

## Concepts extension

The new page is wired into `concepts/index.md` (a new Core row), the `RossoCortex` page, the `context-compaction` page, `observability.md`, and both landing pages. The get-started sidebar is renumbered to seat the new page after the laptop quickstart.

## Notes for reviewers

- The 5 `VERIFY v0.9.0` comments name the exact things to confirm against a shipped binary (install command, `abctl observe`/`service`, CA path, log paths, rate source, and the local-only claim relative to the opt-in central-collector feature [#898](https://github.com/rossoctl/cortex/issues/898)).
- The worked-example table uses representative figures from the real captured fields; it carries a note to replace it with a captured session before release.
- [#867](https://github.com/rossoctl/cortex/issues/867) also proposes renaming the TUI filter prompt to `METHOD name filter…`. That is a code change in `cortex`, out of scope here; the docs describe the current behaviour.

## Acceptance tier

- [x] Tier 0 — Maintenance (docs only)

## Checklist

- [x] Follows repo conventions; STE100 throughout
- [x] DCO sign-off; `Assisted-By` trailer, no `Co-Authored-By`
- [x] All internal links and heading anchors resolve
- [ ] `VERIFY v0.9.0` markers confirmed against a shipped binary (blocks leaving draft)

---

Assisted-By: Claude Code

### DIFF
--- a/docs/get-started/reading-the-numbers.md
+++ b/docs/get-started/reading-the-numbers.md
@@ -199,6 +199,13 @@ One session is useful. A week of sessions is more useful. The value is in the ch
 You see these only if Cortex runs while you work. It runs as a background service and adds no step to
 your day. See [Manage the service](laptop.md#manage-the-service).
 
+## Give feedback
+
+Did the numbers tell you what you needed, or not? Was a figure confusing? Tell us. Open the **Laptop
+feedback** form on [rossoctl/cortex](https://github.com/rossoctl/cortex/issues/new/choose), or write a
+message in [Slack](https://ibm.biz/rossoctl-slack). Feedback on the local tool is what makes the next
+release clearer.
+
 ## Related pages
 
 - [Quickstart on a laptop](laptop.md) installs Cortex and starts `abctl observe`.
--- a/docs/get-started/reading-the-numbers.md
+++ b/docs/get-started/reading-the-numbers.md
@@ -1,0 +1,208 @@
+---
+title: Read the numbers
+sidebar_label: Read the numbers
+description: What Cortex shows for each session, what each number means, and how to act on it.
+sidebar_position: 3
+---
+
+You installed Cortex and started `abctl observe`. Now you see traffic. This page explains what the
+numbers mean and what to do with them.
+
+Cortex shows you what your coding agent sent and what it cost. It shows the model calls, the tool
+calls and the token counts for each session, on your own computer, as they happen. Your agent does
+not report this. Your model provider reports a monthly total, not one session while you work. Cortex
+is the view between the two.
+
+:::note The data stays on your computer
+Cortex captures this traffic on your computer and keeps it there. It does not collect the data and it
+does not send the data to Rossoctl or to any other service. The data is for you to read and to inspect.
+When you stop the service, the data goes with it. See [Manage the service](laptop.md#manage-the-service).
+:::
+
+<!-- VERIFY v0.9.0: confirm the local-only claim once persistence (#901, sqlite) lands — the store is
+     on-disk and local, and no telemetry is sent by default. Central reporting is a separate, opt-in
+     feature (#898), not part of the laptop tool. -->
+
+
+<!-- VERIFY v0.9.0: the metrics view, the token/cost/latency figures and the pruning figure depend
+     on #950, #951 and #952. Confirm the field names, the labels and the layout against `abctl
+     observe` from a v0.9.0 binary before release, and replace the worked example below with a
+     captured session. -->
+
+## What the numbers tell you
+
+The numbers support four decisions:
+
+- **Find an expensive prompt.** One session that costs much more than the others shows you where the
+  cost is.
+- **See whether caching operates.** A high cache-read count means the model reuses your prompt. A low
+  count means it does not, and you pay the full rate each turn.
+- **Catch a session that does not end.** A token count that grows without a result is an agent in a
+  loop. You stop it before it costs more.
+- **Decide whether to prune tool definitions.** The pruning figure shows the tokens that Cortex
+  removed, so you know if the feature is worth enabling. See [Cost control](../concepts/experiments/cost-control.md).
+
+You can read this page before you install Cortex. It describes what Cortex shows you that you cannot
+otherwise see.
+
+## Watch a session
+
+`abctl observe` opens a terminal interface. It has three views.
+
+- **Sessions.** A list of the sessions, with the most recent one first. Each row shows the identifier,
+  the time of the last event, the event count and the token total.
+- **Events.** The calls in one session. Each row shows the time, the direction, the protocol, the
+  model or the method, the status, the duration and the host. Press `Enter` on a session to open it.
+- **Detail.** The full content of one event, as formatted JSON. Press `Enter` on an event to open it.
+
+### Keys
+
+| Key | Action |
+| --- | --- |
+| `↑` `↓` or `k` `j` | Move between rows |
+| `Enter` | Open the selected row |
+| `Esc` | Return to the previous view |
+| `/` | Filter the events |
+| `p` | Pause and resume the stream |
+| `y` | Write the event to a file in `/tmp` |
+| `g` `G` | Move to the top or the bottom |
+| `q` or `Ctrl+C` | Quit |
+
+The `/` key filters the events by a **substring match on the method**. For example, `messages` shows
+only the events whose method contains that text. It is a text match on the method, not a query
+language. You cannot filter on a condition such as a duration.
+
+## Read the tokens
+
+A model call has a cost in tokens. Cortex separates the tokens into five categories, because each
+category has a different price.
+
+| Category | What it is | Why it is separate |
+| --- | --- | --- |
+| **Input** | The prompt tokens that the model read for the first time | You pay the full rate for these. |
+| **Cache read** | The prompt tokens that the model read from its cache | Much less than the input rate. This is caching that operates. |
+| **Cache write** | The prompt tokens that the model wrote to its cache | More than the input rate. This is the one-time cost to store the prompt. |
+| **Output** | The tokens that the model generated | |
+| **Reasoning** | The output tokens that the model used to reason | A part of the output on some models. |
+
+### Why cache read and cache write are two numbers
+
+A model can cache a part of your prompt. The next call that sends the same part reads it from the
+cache. This is useful for a coding agent, because the agent sends the same system prompt and the same
+files on each turn.
+
+The two cache numbers have different prices, so Cortex keeps them apart:
+
+- **A cache write costs more than an input token.** You pay a premium once, to store the prompt.
+- **A cache read costs a fraction of an input token.** You save on each later turn that reuses the
+  prompt.
+
+The split tells you whether caching earns its cost:
+
+- A large **cache-read** count means the cache operates. You pay the low rate for most of the prompt.
+- A **cache-write** count that repeats, with little cache-read, means the cache does not hold. You pay
+  the premium again and again and get no saving. This happens when the prompt changes on each turn.
+
+## Read the cost
+
+Cortex applies a price to each token category and adds the results. The figure is the cost of the
+session.
+
+- **The cost is an estimate, unless the provider returns an exact figure.** Cortex computes the cost
+  from published rates for each model. When the provider returns an exact cost (for example, the
+  `x-litellm-response-cost` header from LiteLLM), Cortex uses that figure instead.
+- **The rates come from a table for each model.** A model that Cortex does not recognize has no rate,
+  so its cost is zero even when its token counts are correct.
+
+<!-- VERIFY v0.9.0: confirm the rate source (built-in table vs. config), and the exact provider
+     headers Cortex reads for an authoritative cost, once #950/#952 land. -->
+
+## Read the latency
+
+Cortex records two times for each model call.
+
+- **Time to first token.** The time from the request to the first part of the reply. For a reply that
+  streams, this is the time until you see the first word. It is the number that decides how fast the
+  agent feels.
+- **Total response time.** The time from the request to the last part of the reply. It depends on the
+  length of the reply.
+
+For a set of calls, Cortex reports percentiles.
+
+- **p50** is the middle value. Half of the calls are faster. It is the typical experience.
+- **p95** and **p99** are the slow calls. Ninety-five or ninety-nine percent of the calls are faster.
+  These are the calls that a user notices. A p50 that is good with a p99 that is bad means that most
+  calls are fast, but the slow calls are very slow.
+
+## Read the pruning savings
+
+If you enable tool pruning, Cortex removes the tool definitions that the agent does not use, before
+the request goes to the model. The savings figure is the tokens and the cost that the pruning
+removed. See [Cost control](../concepts/experiments/cost-control.md).
+
+- **A figure above zero** is the saving for that session. You paid less because Cortex sent a smaller
+  request.
+- **A figure of zero** means that Cortex removed nothing. The usual cause is an agent with no tools.
+  An agent that sends no tool definitions has nothing to prune, so zero is the correct figure and not
+  a failure.
+
+This figure is the answer to one question: what does pruning do for me? You enable pruning, you read
+the figure, and you decide if the saving is worth the feature.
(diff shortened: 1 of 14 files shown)


docs/get-started/reading-the-numbers.md
```diff
@@ -0,0 +1,215 @@
...
+### Keys
+
+| Key | Action |
+| --- | --- |
+| `↑` `↓` or `k` `j` | Move between rows |
+| `Enter` | Open the selected row |
+| `Esc` | Return to the previous view |
+| `/` | Filter the events |
+| `p` | Pause and resume the stream |
+| `y` | Write the event to a file in `/tmp` |
```
huang195: **must-fix** — two rows of this reference table send people the wrong way, and a reference table is the one thing readers will not second-guess.

**`y`** writes to `~/.cortex/abctl-events/<timestamp>-<rand>.json`, not `/tmp`. cortex's own abctl README says so in three separate places, including its deferred-work list ("Native clipboard (currently writes a file under `~/.cortex/abctl-events`)").

**`/`** (line 71) is described in a way that is actively wrong rather than merely incomplete. `events_pane.go:matchEventRow` builds a haystack from many event fields and matches your substring against all of them, so "a substring match on the method" is not what happens. And there are exactly two query forms that the paragraph tells people do not exist:

- `deny` — matches denied events
- `plugin:<name>` — matches events carrying that plugin's record

Those are the two filters an operator actually reaches for, so they are worth documenting rather than denying. The same method-only sentence appears in `laptop.md:57`, so both need the same edit.

docs/get-started/reading-the-numbers.md
```diff
@@ -0,0 +1,215 @@
...
+| `q` or `Ctrl+C` | Quit |
+
+The `/` key filters the events by a **substring match on the method**. For example, `messages` shows
+only the events whose method contains that text. It is a text match on the method, not a query
+language. You cannot filter on a condition such as a duration.
+
+## Read the tokens
+
+A model call has a cost in tokens. Cortex separates the tokens into five categories, because each
+category has a different price.
```
huang195: **suggestion** — "five categories, because each category has a different price" is four categories. `pricing/rates.go` has `numTiers = 4`: `TierInput`, `TierCacheWrite`, `TierCacheRead`, `TierOutput`. There is no reasoning tier.

Your own table row on line 86 already says as much — "a part of the output on some models" — which contradicts the premise one paragraph above it. The worked example then lists "Reasoning tokens 3,100" as its own row, implying Cortex reports it separately.

Either fold it into a note under Output, or say plainly that reasoning tokens are billed as output and Cortex does not price them apart. The second reads better and is true.

docs/get-started/reading-the-numbers.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Read the cost
+
+Cortex applies a price to each token category and adds the results. The figure is the cost of the
+session.
+
+- **The cost is an estimate, unless the provider returns an exact figure.** Cortex computes the cost
+  from published rates for each model. When the provider returns an exact cost (for example, the
+  `x-litellm-response-cost` header from LiteLLM), Cortex uses that figure instead.
+- **The rates come from a table for each model.** A model that Cortex does not recognize has no rate,
```
huang195: **suggestion** — "A model that Cortex does not recognize has no rate, so its cost is zero even when its token counts are correct" is backwards, and it happens to describe the precise behaviour the code exists to prevent.

`costevent.Decode` refuses to treat a non-positive cost as priced, with the reasoning in the comment: "free" and "unknown" must not be conflated, because a consumer that treated 0 as priced would render `$0.00` for traffic it simply cannot price. So an unpriced model shows **no cost**, not a cost of zero.

That is also the better story: the tool admits it does not know, rather than quietly claiming the call was free.

docs/operate/troubleshooting.md
```diff
@@ -4,13 +4,132 @@ description: The failures that occur most often, and the recovery for each one.
...
+   the proxy variable and the certificate variable. See
+   [Other agents](../get-started/laptop.md#other-agents).
+3. **The agent sends no traffic yet.** Send a message to the agent, and watch for the events.
+
+### The numbers are wrong or absent
+
+- **The token count is zero.** RossoCortex reads the token counts from the reply of the model. A
+  reply that Cortex cannot parse gives no counts. Confirm that the model is an OpenAI-compatible or an
+  Anthropic endpoint.
+- **The cost is zero, but the tokens are correct.** Cortex has no rate for that model. See
```
huang195: **suggestion** — same inversion as the one on `reading-the-numbers.md:114`: "The cost is zero, but the tokens are correct." What a user actually sees is a *missing* cost alongside correct tokens.

Worth fixing here especially, because this is the page someone lands on when the number confuses them — and "no cost shown" versus "a cost of zero" is exactly the distinction they arrived to resolve.

huang195 (changes requested):
*Dobra* — the shape of this is genuinely good, and I want that on the record before the red ink. Splitting troubleshooting into laptop-then-Kubernetes with the heading levels pushed down correctly, the four-decisions framing at the top of the new page (find an expensive prompt / see whether caching operates / catch a session that does not end / decide whether to prune), the cache-write-repeating-with-no-cache-read diagnosis, and the `VERIFY v0.9.0` markers sitting on exactly the claims that depend on unlanded issues — that is someone writing for a reader rather than for a checklist. The `sidebar_position` renumbering is also complete and collision-free (clean 1→8 across all eight pages, I checked), which is the thing that usually goes wrong in a PR like this.

Then I fact-checked the checkable claims against `rossoctl/cortex@main`, and four of them do not survive:

| Claim | Reality |
|---|---|
| `y` writes "to a file in `/tmp`" | `~/.cortex/abctl-events/<timestamp>-<rand>.json` |
| `/` is "a substring match on the method… not a query language" | `matchEventRow` matches a haystack of many event fields, and supports `deny` and `plugin:<name>` |
| "five categories, because each has a different price" | `numTiers = 4` — input, cache-write, cache-read, output |
| unrecognised model → "its cost is zero" | inverted; a non-positive cost is deliberately **not** priced |

The first two are in a keys table and a filter paragraph — the parts a reader trusts most and verifies least. Someone presses `y`, goes rummaging in `/tmp`, and concludes the feature is broken; someone else reads that `/` only matches the method and never finds `deny` or `plugin:<name>`. Cheap to fix, expensive to ship, hence the block. Details inline.

Things I checked that are **right**, so you do not have to re-verify them: `~/.cortex/ca/ca.crt` (I was suspicious because `ca/bundle.crt` also exists in the tree — it is the outlier), the `x-litellm-response-cost` header name, and the 47600/47601/47602 port split.

**On the red X, which is not yours.** `Ollama Model Drift` fails with `doc not found`: `.github/scripts/docs/check-ollama-model-drift.sh` hardcodes `docs/demos/demo-weather-agent.md`, and `docs/demos/` does not exist on `main` at all any more. `docs-ci.yaml` is path-filtered to `**.md`/`docs/**`, which is why #2548–2550 show n/a for that check and your docs PR is the one holding the bag. Nothing to change here — it wants a one-line path fix (or a deletion) in its own PR, otherwise every docs PR from now on is red.

docs/get-started/reading-the-numbers.md
```diff
@@ -0,0 +1,225 @@
...
+| `q` or `Ctrl+C` | Quit |
+
+The `/` key filters the events by a **text match**. For example, `messages` shows only the events
+that contain that text. The match covers many fields of an event, not the method alone — the host,
+the method, the plugin name, the reason, the path, and the message content all count.
+
+Two searches are special:
+
+- `deny` shows only the events that a plugin denied.
+- `plugin:<name>` shows only the events that the named plugin acted on. For example, `plugin:jwt-validation`.
```
huang195, line 78: **fix before leaving draft** — you fixed my filter complaint by documenting `plugin:<name>`, and then picked the one plugin in the entire tree that cannot be found that way. I laughed, then checked twice, because I did not believe it either.

`matchEventRow` resolves `plugin:` as an **exact lookup** in `r.event.Plugins`, and `SessionEvent.Plugins` is populated only from `Extensions.Custom` entries written with `PluginEventSuffix` (`"/event"`) — an opt-in for plugins that publish a plugin-public record. `jwtvalidation/plugin.go` never writes `Extensions.Custom`; it has four `pctx.Record(pipeline.Invocation{...})` calls, and `Record` populates `Invocations`, a different field entirely. So `Plugins["jwt-validation"]` is never set and this filter returns an empty list, every time.

The extra sting: jwt-validation events *are* findable — by typing the bare `jwt-validation`, because `eventHaystack` includes `iv.Plugin`. So for your example plugin the plain search works and the special syntax does not, which is exactly backwards from how the bullet reads.

Swap in a plugin that publishes a record. Both verified:

- `plugin:tool-prune` — `toolprune/event.go` writes `Custom[p.Name()+PluginEventSuffix]`, `Name()` returns `"tool-prune"`, and the listener strips the suffix, so the key really is `tool-prune`.
- `plugin:cost` — `costevent.Key = "cost"`. Note this one is a *record* name rather than a plugin name, so "the named plugin" is doing a little work in that sentence if you want it strictly true.

Same reason as the last round: this is a copy-pasteable instruction in a reference section. The reader types it, gets nothing, and concludes the filter is broken rather than that the docs picked a bad example.

docs/get-started/reading-the-numbers.md
```diff
@@ -0,0 +1,225 @@
...
+
+| Field | Value |
+| --- | --- |
+| Model | `claude-sonnet-4` |
+| Events | 34 |
+| Input tokens | 12,400 |
+| Cache read | 486,000 |
+| Cache write | 61,200 |
+| Output tokens | 8,900 |
+| Reasoning tokens | 3,100 |
```
huang195, line 181: **suggestion** — the prose fix above is exactly right ("They are a part of the output count, not a fifth priced category" matches `pricing.Usage`'s own comment almost word for word), but the worked example did not get the memo:

```md
| Output tokens | 8,900 |
| Reasoning tokens | 3,100 |
```

Sibling rows, with no hint that one is inside the other. A reader who sums the token rows double-counts 3,100, and nothing on the table says whether 8,900 already contains it. `| Reasoning tokens (part of the output) | 3,100 |` and you are done — last loose thread from the four-versus-five fix.

docs/get-started/reading-the-numbers.md
```diff
@@ -0,0 +1,225 @@
...
+| `Enter` | Open the selected row |
+| `Esc` | Return to the previous view |
+| `/` | Filter the events |
+| `p` | Pause and resume the stream |
+| `y` | Write the event to a file in `~/.cortex/abctl-events/` |
+| `g` `G` | Move to the top or the bottom |
+| `q` or `Ctrl+C` | Quit |
+
+The `/` key filters the events by a **text match**. For example, `messages` shows only the events
+that contain that text. The match covers many fields of an event, not the method alone — the host,
```
huang195, line 72: **nit** — the six-field list is accurate, but `eventHaystack` also covers `Identity.Subject`, `Identity.ClientID`, and every invocation `Details` **key and value** — the source comment there specifically calls out searching `target_audience` and its value.

"Every event from this caller" and "every event where this plugin wrote this detail" are the two searches an operator reaches for on day two, and they are the two the list omits. A trailing "…and the caller identity, and any diagnostic detail a plugin recorded" would cover it.

huang195 (approved):
*No dobra* — all four blockers from my last pass are in, and I re-verified every one against `rossoctl/cortex@main` rather than taking the commit message's word for it:

| Claim | Verified |
|---|---|
| `y` → `~/.cortex/abctl-events/` | `tui/app.go`: `const yankDirRel = ".cortex/abctl-events"`, with its own "Not `os.TempDir()`" comment explaining why |
| `/` matches many fields; `deny` and `plugin:<name>` are the two special forms | `matchEventRow` — exactly those two, then substring over `eventHaystack`. Structure is right; see the inline note on the example |
| `deny` = events a plugin denied | `eventMatchesDeny`: `Phase == SessionDenied` or any `Action == ActionDeny`. Accurate |
| Four priced categories, reasoning folded in | `numTiers = 4`, and the wording now matches `pricing.Usage`'s own comment ("Reasoning is omitted (a subset of Output, already counted)") |
| Unpriced model → blank cost, not zero | Both sites fixed. "A blank cost with correct tokens means the model is unpriced, not free" is exactly the `costevent.Decode` distinction |

Two unasked-for improvements worth naming, because they show the fix was understood rather than pattern-matched: the token table reordered to input → cache-write → cache-read → output, which now matches the `Tier` iota order in `rates.go`; and `laptop.md` got a cross-link to the filter explanation instead of a duplicated half-sentence, so there is one place to keep correct instead of two.

The new feedback commit checks out too, including the claim most likely to have been invented: `.github/ISSUE_TEMPLATE/laptop-feedback.yml` really exists in cortex with `name: Laptop feedback`, so `/issues/new/choose` will show it. And "the form asks for each item in the list above" holds — all five map to a field (`abctl --version` → Cortex version, `abctl service status` and the log → Service status and logs, `uname -sm` → Operating system and architecture, agent+model → Agent and model).

**Approving**, since this is still a draft and the one real problem left is a single identifier. But it does need fixing before this leaves draft, alongside the `VERIFY v0.9.0` markers you already have queued: `plugin:jwt-validation` names the one plugin in the tree that publishes no plugin-public record, so the only example of that filter returns an empty list. Details inline.

**On the red X, still not yours.** `Ollama Model Drift` fails identically to three days ago — `check-ollama-model-drift.sh` hardcodes `docs/demos/demo-weather-agent.md`, and I re-confirmed `docs/demos/` is a 404 on `main`. Nothing to change here, but it has been red for three days and every docs PR inherits it, so someone should land the one-line path fix or delete the check.

Assisted-By: Claude Code

PR #2553, opened 2026-09-14
CI: repoint the Ollama model-drift guard to the restructured docs

mrsabath:
## What & why

The `Ollama Model Drift` check fails on **every docs PR** since the documentation restructure ([#2530](https://github.com/rossoctl/rossoctl/pull/2530)) merged. This **repoints** it so it works again.

The guard (`check-ollama-model-drift.sh`, run by the `ollama-model-drift` job) verified that a docs page contained the `LLM_MODEL` value from the weather demo's `.env.ollama` in `rossoctl/examples`. It pointed at `docs/demos/demo-weather-agent.md`, which the restructure removed — so the script exits with `doc not found`, and because the job is path-filtered to `docs/**` and `**.md`, every documentation PR inherits the red check ([#2551](https://github.com/rossoctl/rossoctl/pull/2551) is the one currently holding the bag).

## Repoint, not retire

> **This PR changed direction after review.** It originally deleted the guard, on the stated grounds that no published doc names the model any more. @esnible fact-checked that and it was wrong: `docs/get-started/configure-a-model.md:55` names `llama3.2:3b-instruct-fp16` verbatim and asserts "The default in the `ollama` preset" — the exact relationship the guard exists to protect — and the deleted script's own `grep -F` matches it against the live `.env.ollama`. So the guard can be kept, green and meaningful, rather than removed. Thanks Ed.

The model configuration moved to `docs/get-started/configure-a-model.md` in the restructure. This PR points `CHECKS` at that file. The check stays exactly as it was — fetch `.env.ollama`, read `LLM_MODEL`, confirm the doc contains it — now against the page that actually names the model.

The page is a **menu** of tested models, so the guarded value can move rows without breaking the check, as long as the current `.env.ollama` model still appears somewhere on the page. That is looser than pinning one exact row, and deliberately so: a contributor reordering the table or rewording the Notes column should not go red, but the examples repo repinning to a model the page never mentions should.

## Change

- Repoint `CHECKS` in `check-ollama-model-drift.sh` from the deleted `docs/demos/demo-weather-agent.md` to `docs/get-started/configure-a-model.md`.
- Update the script header and the job comment to describe the new target; drop the stale [#2527](https://github.com/rossoctl/rossoctl/issues/2527) "extend the CHECKS array" note (moot — see below).
- No change to the check logic, the job structure, or the `paths:` filter.

## Loose ends Ed flagged

- **#2527** referenced the guard's old "extend CHECKS once the other demos publish `.env.ollama`" follow-up. That framing is moot now; the guard covers the model page rather than per-demo docs. Worth closing or retargeting so it does not send someone looking for the old per-demo scope.

## Refs

Refs [#2527](https://github.com/rossoctl/rossoctl/issues/2527). Unblocks the red check on every docs PR, including #2551.

Assisted-By: Claude Code


.github/workflows/docs-ci.yaml
```diff
@@ -85,33 +85,3 @@ jobs:
             --config .lychee.toml
             "**/*.md"
           fail: false
```
esnible, line 87: **suggestion** — the retire-vs-repoint rationale rests on a claim that doesn't hold up:

> No published doc restates the env file's model verbatim any more, so the guard has nothing to protect.

`docs/get-started/configure-a-model.md:55` does restate it verbatim:

```
| `llama3.2:3b-instruct-fp16` | 3B | The default in the `ollama` preset. |
```

That is character-for-character the value currently in `rossoctl/examples`' `a2a/weather_service/.env.ollama` (`LLM_MODEL=llama3.2:3b-instruct-fp16`, fetched live from `main` while reviewing). I ran the deleted script's exact comparison — `grep -qF -- "${model}" "${doc}"` — against that file, and **it matches**. Repointing `CHECKS` to `docs/get-started/configure-a-model.md` would have yielded a green, meaningful guard rather than a deletion.

What makes this more than pedantry: that row doesn't merely happen to contain the string, it *asserts the very relationship the guard existed to protect* — "The default in the `ollama` preset." That's exactly the kind of claim that rots silently when the examples repo repins its model, and this PR removes the only automated detection of it. The drift risk described in the script's header comment is still live; it just no longer has a tripwire.

I'm not blocking on this, because the decision to retire is defensible on grounds the PR states correctly and this comment doesn't dispute: the page is a **menu** of three tested models, and pinning CI to one row of it is genuinely brittle — a contributor reordering the table, rewording the Notes column, or dropping that row would go red for no real reason. And a guard that fails on every docs PR is worse than no guard. My objection is narrowly to the *stated reason*, not the outcome.

So, either of these would resolve it:

1. **Repoint** `CHECKS` at `docs/get-started/configure-a-model.md` and keep the guard, accepting the table-brittleness tradeoff; or
2. **Amend the rationale** to the accurate version — the model *is* still named, but only inside a menu, so pinning CI to it would be brittle and low-value. Same conclusion, honestly grounded, and it tells the next person what would actually have to change for the guard to come back.

Two loose ends while you're here, neither worth its own thread:

- **#2527** is referenced as "the pre-existing fragility of this guard," and its actionable part — extend `CHECKS` once the other demos publish `.env.ollama` — is now moot. Worth closing or retargeting so it doesn't send someone looking for a script that no longer exists.
- The `.github/scripts/docs/**` entry in the `paths:` filter is now dead, since the deleted script was the only file under that directory. Agreed with leaving it out of this diff; just flagging that it's fully dead rather than merely unused, in case a follow-up sweeps it.

esnible (approved):
## Summary

Verified the premise rather than taking it on trust, and it holds: `docs/demos/` is gone from `main` (0 paths in the tree), so the script's `doc not found` exit is real; and `Ollama Model Drift` is the **only** failing job on #2551 — Link Check and Markdown Lint both pass, across three consecutive runs on that branch. The deleted script is also genuinely the sole file under `.github/scripts/docs/`, so the directory does go with it as described. Retiring a guard that turns every docs PR red is the right instinct, and the diff is clean and single-purpose.

One correction to the rationale, inline. The "Retire rather than repoint" section rests on a claim that isn't accurate, and it happens to be the load-bearing one. It doesn't change my verdict — the retire decision is defensible on the *other* grounds the PR states correctly — but the stated reason should be the true one, since it's what a future contributor will read when deciding whether to bring the guard back.

Two minor notes folded into that comment rather than filed separately: the stale `.github/scripts/docs/**` entry in the `paths:` filter (correctly called out as a deliberate no-op — agreed, leaving it keeps this diff to one purpose), and #2527's "extend the CHECKS array" follow-up now being moot.

**Author**: mrsabath (MEMBER — maintainer)
**Areas reviewed**: CI/GitHub Actions, Shell (deleted script), Docs (cross-referenced), security
**CI status**: passing (19 checks green; `Deploy & Test` pending at review time) — docs-ci is green on this branch

---
Assisted-By: Claude Code

PR #2558, opened 2026-09-14
Fix: remove unimplemented sandbox/triggers/integrations feature flags

### LINKED ISSUE #1764: :bug: SANDBOX flag enabled but sandbox modules not installed — skipping
### Describe the bug

If Kagenti is installed with 

```YAML
featureFlags:
  sandbox: true
```

The _kagenti-backend_ Deployment does not serve the sandbox API (http://kagenti-ui.localtest.me:8080/api/v1/sandbox/team1/agents returns 404) and logs

```
SANDBOX flag enabled but sandbox modules not installed — skipping
```

### Steps To Reproduce

Install Kagenti with

```bash
cat <<EOF > /tmp/enable-flags-v4.yaml
featureFlags:
  sandbox: true
  authbridgeAPI: true
  admin: true
EOF

scripts/kind/setup-kagenti.sh --with-istio --with-spire --with-ui --with-otel --with-agent-sandbox --preload-images --kagenti-values /tmp/enable-flags-v4.yaml && echo done || echo failed
```

Browse to http://kagenti-ui.localtest.me:8080/sandboxes

### Expected Behavior

- The modules should be installed in the container build.  I installed from main, and the image installed is `ghcr.io/kagenti/kagenti/backend:v0.7.0-alpha.2`, which should have the module
- Even if the module isn't installed, if the feature flag is enabled the Kagenti backend `http://kagenti-ui.localtest.me:8080/api/v1/sandbox/<team>/agents` endpoint should return 500 if the feature flag is set
- The backend should log the ImportError that caused the failure, to help troubleshooters know what is missing.

### Additional Context

The actual error is

```
cannot import name 'sandbox' from 'app.routers' (/app/app/routers/__init__.py)
```

cc @cwiklik

cwiklik:
## What & why

Three feature flags — `sandbox`, `integrations`, `triggers` — gated UI whose backend routers were never merged. Enabling any of them produced a navigable UI where every API call returned 404, and the failure was silent by construction: each flag's router imports sat in a `try/except ImportError` that logged a warning and continued, so `_*_modules_loaded` stayed `False`, `include_router` was skipped, and `/api/v1/config/features` kept reporting the flag as `true`.

This was not hypothetical. `deployments/envs/ocp_values.yaml` and `deployments/envs/ocp_ci_values.yaml` both set all three flags to `true`, so those deployments have been serving a broken Sessions/Sandboxes UI for as long as those files have existed.

Finishing the backends was not on the table — the routers live in an unmerged PR (#996) of unknown currency and the UI has drifted for months. A flag that must not be enabled is worse than no flag, because `values.yaml` presents it as a supported option.

This PR deletes all three flags and every artifact that exists only to serve them.

Closes #1764

### This is a privilege reduction, not just a cleanup

Removing the `sandbox` RBAC block permanently drops from the backend ClusterRole:

- `pods` `create`
- `pods/exec` `get` + `create` — container-escape-grade, and no shipping code path uses it
- `pods/log`
- `secrets` `create`/`patch`/`delete` (`KubernetesService.create_secret` has no callers)
- `configmaps` `create`/`patch`/`delete`

**This is not inert.** The two OCP env values files above were the only place in the repo where that RBAC block actually rendered, so this materially changes the OCP and OCP-CI backend ClusterRole. **An OCP-CI smoke test is worth running before merge.**

One caveat, called out honestly: the live `skills` router deletes ConfigMaps (`app/routers/skills.py:737,882,894`, `app/services/skill_autosync.py:225`), and after this change the rendered ClusterRole grants `configmaps` only `get,list,create,update,patch` — no `delete` from any rule. This is **not a regression** (`sandbox` defaulted to `false`, and the OCP envs don't enable `skills`), but the sandbox block had been accidentally masking a pre-existing chart gap, and this PR makes that permanent. Filed as follow-up 3 below.

### The one edit that is not a deletion

`charts/rossoctl/templates/ui.yaml` gated PersistentVolumeClaim RBAC on `{{- if or .Values.featureFlags.sandbox .Values.featureFlags.simulatedTools }}`. That block is **narrowed, not deleted** — `simulatedTools` is a live shipping feature (epic #2151) whose delete path enumerates and tears down PVCs. Deleting it with the rest of the sandbox RBAC would have silently stripped PVC permissions from a working feature. Two new `helm unittest` cases pin the behaviour in both directions.

### Two things worth a reviewer's eye

1. **`SkillWhisperer.tsx` is deleted.** It lands in the `sandbox` set because only sandbox pages import it — despite the name's proximity to the live `skills` flag. The import graph is correct (it has no other importer), but the name invites a double-take.
2. **Four files named `Sandbox*` survive on purpose.** `SandboxWizard.tsx`, `SandboxConfig.tsx`, `SandboxAgentsPanel.tsx` and `utils/workloadType.ts` belong to the live `agentSandbox` feature. The deletion set was computed by import-graph reachability, not filename matching, precisely because a name-matching sweep would have broken working code. Likewise `sandboxService` and `modelsService` in `api.ts` are kept: both hit dead endpoints, but both are called by the live wizard.

## Scope

37 UI files deleted, plus `charts/rossoctl/templates/integration-crd.yaml` (its first line wrapped the whole file in the flag conditional). Backend: 3 Settings fields, 3 guard blocks, 3 `include_router` branches, a dead shutdown branch, and 3 response-model fields. Chart: env vars, RBAC, values entries. Also `scripts/openshell/deploy-shared.sh` and the `CLAUDE.md` flags table.

Commits are vertical slices — one per flag, each spanning backend + chart + UI + scripts — so no commit leaves a flag half-removed. The first commit is a no-op safety improvement: it adds `exc_info=True` to the **four surviving** guards (`skills`, `acp`, `simulated_tools`, `dreaming`), whose modules do exist. Today a genuine `ImportError` inside one of those is reported as "modules not installed" with the traceback discarded — the same information loss that let this issue reach production.

## Known follow-ups (pre-existing, deliberately not fixed here)

1. **The live `agentSandbox` wizard calls a route that was never merged.** `SandboxWizard.tsx:43` imports `services/eventService.ts` and calls `eventService.getDefaults()` at `:267`, targeting the `events` router — one of the eight modules the dead `sandbox` flag was importing. `eventService.ts` is therefore kept and untrimmed (`AuthContext.tsx:16` also imports from it).
2. **The sidecar manager has no shutdown hook.** The removed `if _sandbox_modules_loaded:` branch held the only one ever written, but it was never reachable. No runtime behaviour changes.
3. **`configmaps delete` is missing from the chart's ClusterRole** while the live `skills` router needs it — see the caveat above.

## Acceptance tier

Tier 0: no net-new user-facing behaviour. The removed UI was non-functional by construction and both flags defaulted to `false`.

### Pillar 1 — Code quality (all tiers)

- [x] `make lint` and pre-commit pass — exit 0, pylint 9.94/10, unchanged from the previous run
- [x] Tests added/updated and passing — new `tests/test_removed_feature_flags.py` asserts each removed flag is absent from **both** `FeatureFlagsResponse` and `Settings` while the live flags remain, so a bad merge that reintroduces one fails; plus 2 new `helm unittest` cases pinning the PVC RBAC narrowing
- [x] DCO sign-off on all commits (`git commit -s`)
- [ ] (Tier 1+) Behind a feature flag, off by default — N/A for Tier 0

### Pillar 2 — Documentation (all tiers; scope scales)

- [x] Docs affected by this change are updated — `CLAUDE.md`'s "Current flags" table loses all three rows, one per vertical commit, leaving `rossoctl_feature_flag_admin`. `docs/reference/install-options.md` and `docs/operate/install-helm.md` never mentioned these flags.

Pillars 3 and 4 are Tier 1+ and do not apply. One environment note regardless of tier: **OpenShift** is the environment materially affected, via the ClusterRole change described above.

## Verification

Run on the final tree:

| Check | Result |
|---|---|
| `uv run pytest -q --ignore=tests/test_migration.py` | 794 passed |
| `npm run typecheck` / `build` | exit 0 |
| `npm run test:unit` | 93 passed |
| `helm unittest charts/rossoctl` | 22/22 passed |
| `helm lint charts/rossoctl` | 0 failed |
| `helm template --set featureFlags.simulatedTools=true \| grep -c persistentvolumeclaims` | 1 (the §5 regression guard) |
| `make lint` | exit 0, 9.94/10 |

Repo-wide sweep for `featureFlags.{sandbox,triggers,integrations}` / `FEATURE_FLAG_*` / `feature_flag_*`: zero matches, excluding only the regression test (which must name them as literals) and `docs/`, which holds a historical plan document deliberately left alone. The sweep's known blind spot — the bare YAML key form, which the template-shaped regex cannot see — was checked separately and repo-wide.

Live `agentSandbox` / `agent_sandbox` references were confirmed still present afterwards, so the sweep was specific rather than a blanket deletion.

`npm run lint` reports 18 errors / 7 warnings; all pre-exist on `main` in files this PR does not touch, and the count is unchanged.

## Note on the removed keys

The chart has no `values.schema.json`, so a stale `featureFlags.sandbox: true` in someone's values becomes a silently ignored key rather than a Helm error. For a flag that never worked and defaulted to `false`, degrading from "silently broken UI" to "silently ignored key" is a strict improvement; adding a schema is out of scope. Verified: setting all three removed keys on this branch produces a byte-identical render.

Assisted-By: Claude Code


charts/rossoctl/templates/ui.yaml
```diff
@@ -499,9 +470,9 @@ rules:
   - apiGroups: ["rbac.authorization.k8s.io"]
     resources: ["rolebindings"]
     verbs: ["get", "create"]
-  {{- if or .Values.featureFlags.sandbox .Values.featureFlags.simulatedTools }}
-  # PersistentVolumeClaims — sandbox agent workspace storage, and simulated-tool
-  # StatefulSet volumes that the delete path enumerates and tears down (#2151).
+  {{- if .Values.featureFlags.simulatedTools }}
```
esnible, line 473: The narrowing here is correct — verified `simulatedTools` keeps `get,list,create,patch,delete` on PVCs, and gating this on the live flag instead of deleting it with the rest of the sandbox block is the right call. Using this line to raise something about a **different** RBAC rule in this same file, since it has no changed line to anchor to.

**suggestion** — the `configmaps delete` caveat and follow-up #3 are overstated. The PR body says:

> after this change the rendered ClusterRole grants `configmaps` only `get,list,create,update,patch` — no `delete` from any rule

But `ui.yaml:440-445` grants it:

```yaml
{{- if .Values.featureFlags.skills }}
# Skills-specific: delete required for skill deletion via the Skills API.
- apiGroups: [""]
  resources: ["configmaps"]
  verbs: ["delete"]
{{- end }}
```

I diffed that against `main` (`ui.yaml:454-459` there) — **byte-identical and untouched by this PR**. So the chart already couples `configmaps delete` to the same flag that gates the `skills` router needing it, which is the correct coupling: with `skills=true` the permission is granted by the skills rule; with `skills=false` the router isn't loaded and nothing needs it.

That also undercuts the "the sandbox block had been accidentally masking a pre-existing chart gap" framing — there's no gap being masked, because the permission is already conditioned on the feature that requires it.

Why this is worth correcting rather than leaving as a harmless wording slip: follow-up #3 as written points someone at adding `delete` to the **unconditional** configmaps rule at `:438-439`. Doing that would grant cluster-wide configmap deletion regardless of any flag — widening privilege in a PR whose headline contribution is reducing it (`pods/exec`, `secrets` mutation, `pods create`). Given how carefully the rest of this PR reasons about exactly that tradeoff, I doubt that's the intended outcome.

Suggested resolution: drop follow-up #3, or restate it as "verified: `configmaps delete` is correctly gated on `featureFlags.skills`; no action needed."

One transparency note on my own check — I could not confirm by rendering, because `helm template` fails on this tree with `missing in charts/ directory: operator-chart`, which is a local dependency issue unrelated to your change. The conclusion rests on the template source plus the `main` comparison, and your `helm unittest` 22/22 is consistent with it. If you read the rules differently, say so and I'll recheck.

esnible (approved):
## Summary

Verified the deletion set independently rather than relying on the (unusually thorough) PR body, since the risk in a −16k diff is a deletion that catches live code:

- **No dangling imports.** I extracted the 36 deleted UI modules from the diff and grepped all **68** surviving `.ts`/`.tsx` files on the branch for `from '…/<module>'` — zero hits. The import-graph reachability claim holds.
- **The four surviving `Sandbox*` files match the claim** exactly (`SandboxWizard`, `SandboxConfig`, `SandboxAgentsPanel`, `utils/workloadType`, plus `types/sandbox.ts`). Computing the set by reachability rather than filename was the right call, and it shows.
- **The PVC RBAC narrowing is correct** — the single riskiest edit here, and it does what it says.
- **Vertical-slice commits confirmed** — one per flag, so no commit leaves a flag half-removed. That structure made this reviewable at all.

One correction inline: follow-up #3's `configmaps delete` claim doesn't hold up. It's a documentation-accuracy issue in the PR body, not a defect in the chart — but it's worth fixing because acting on it as written would widen privilege in a PR whose purpose is reducing it.

The `pods/exec` removal is a real privilege reduction and this is right to foreground it rather than bury it as cleanup.

**On the pre-merge gap**: the PR correctly identifies that an OCP-CI smoke test is warranted, since the two OCP env values files were the only place the sandbox RBAC block rendered. `Deploy & Test` and `OpenShell PoC (Kind)` both pass, but neither exercises OpenShift, and `E2E HyperShift` is still waiting on a maintainer `/run-e2e`. That's the gap to close before merging — approving the code, not waiving that run.

**Areas reviewed**: Python (backend), Helm/K8s (RBAC, chart), Frontend (import graph), YAML, Shell, Docs, security/RBAC
**Commits**: 7 commits, all signed-off: yes
**CI status**: passing (24 checks incl. Deploy & Test 27m, OpenShell PoC 34m); E2E HyperShift pending `/run-e2e`

---
Assisted-By: Claude Code

PR #2561, opened 2026-09-15
Chore(release): pin v0.8.0-alpha.2, bump operator-chart to 0.4.0-alpha.1

cwiklik:
## Summary

Prepares the `v0.8.0-alpha.2` tag. The `release/*` branch name makes `ci-release-pins.yaml` a hard gate on this PR rather than an advisory summary.

## Changes

| Change | How |
|---|---|
| Nine platform image tags → `v0.8.0-alpha.2` | `scripts/pin-release-tags.sh v0.8.0-alpha.2` (both `charts/rossoctl/values.yaml` and `charts/rossoctl-deps/values.yaml`) |
| `Chart.yaml` `version` / `appVersion` → `0.8.0-alpha.2` | same script |
| `operator-chart` `0.3.1-alpha.1` → `0.4.0-alpha.1` | by hand — `dependencies[].version` in `Chart.yaml` plus the `values.yaml` override block. `pin-release-tags.sh` does not touch dependency versions |
| `Chart.lock` | `helm dependency update charts/rossoctl/` |

## What the operator bump contains

**The CRD change is additive**: `AgentRuntime` gained three optional fields — `pluginPreset`, `plugins`, `onError` — for the AuthBridge layer-3 plugin-pipeline presets. No new `required` entries, so it is backward-compatible and the minor bump is justified.

**The chart is not additive**, and this is the part of the bump that changes runtime behavior (thanks @mrsabath for catching that an earlier revision of this description said "additive only", which steered attention away from it). Across `v0.3.1-alpha.1...v0.4.0-alpha.1`:

- `charts/operator/templates/manager/configmap-spiffe-helper.yaml` is **removed** (−36)
- `charts/operator/templates/manager/manager.yaml` (+3/−46) strips the `spiffe-helper` sidecar, the `jwt-svid` volume, and the `--jwt-svid-path` flag

JWT-SVID delivery moves from a sidecar-written file to an in-process `workloadapi.Client`. Safe for this PR specifically — the rossoctl charts never passed `--jwt-svid-path`, so the flag removal breaks no wiring here — but it is a sidecar removal, not an addition, and deserves a look on that basis.

## Sidecar tags come from the subchart

Per rossoctl/rossoctl#2399 this chart deliberately does **not** override the AuthBridge sidecar tags. Rendering confirms they arrive via `operator-chart` and land on cortex `v0.7.0-alpha.10`:

```
ghcr.io/rossoctl/operator/rossoctl-operator:0.4.0-alpha.1
ghcr.io/rossoctl/cortex/authbridge:v0.7.0-alpha.10
ghcr.io/rossoctl/cortex/authbridge-envoy:v0.7.0-alpha.10
ghcr.io/rossoctl/cortex/authbridge-lite:v0.7.0-alpha.10
ghcr.io/rossoctl/cortex/proxy-init:v0.7.0-alpha.10
```

That range carries an OpenShift fix: cortex `alpha.8` broke in-cluster DNS on OCP (a `CLUSTER_CIDRS` default of `10.0.0.0/8` excluded the 172.30/172.31 service network, so UDP/53 hit the terminal DROP). `alpha.10` removes `CLUSTER_CIDRS` and derives the DNS exemption from the pod's `/etc/resolv.conf`.

## Validation

`bash scripts/check-release-pins.sh` — all 8 checks pass. `--render` adds a 9th (subchart floating-tag scan of the rendered manifests), also passing.

**E2E HyperShift should run before the tag lands.** The OCP DNS fix above is the substantive reason this cortex range matters, and Kind's service network (10.96/16) sits inside the old 10/8 default — which is exactly why the original bug survived CI. Passing Kind-based CI is weak evidence about the thing being fixed.

## Known gaps, not addressed here

- `charts/openshell/values.yaml:16` carries a floating `tag: main-d7d8306` for the Keycloak credentials driver, outside `check-release-pins.sh`, which hardcodes `charts/rossoctl` and `charts/rossoctl-deps`.
- `rossoctl/examples` references `ghcr.io/rossoctl/examples/weather_service:latest` from demo manifests and docs outside `charts/`, invisible to the validator.
- **Doc drift from the sidecar removal** (follow-up): `docs/security/authentication-modes.md`, `docs/diagrams/02-rossoctl-client-registration.mmd`, `docs/diagrams/03-rossoctl-client-registration-final.mmd`, and `docs/diagrams/README.md` still describe the operator's spiffe-helper container and `/opt/jwt_svid.token`. AuthBridge's bundled spiffe-helper is a different thing and those references remain correct.


charts/rossoctl/Chart.yaml
```diff
@@ -2,11 +2,11 @@ apiVersion: v2
...
 type: application
-version: 0.8.0-alpha.1
-appVersion: "0.8.0-alpha.1"
+version: 0.8.0-alpha.2
+appVersion: "0.8.0-alpha.2"
 
 dependencies:
 - name: operator-chart
-  version: 0.3.1-alpha.1
+  version: 0.4.0-alpha.1
```
mrsabath, line 10: The description calls this bump "Additive only (94 insertions, 0 deletions)" — accurate for the CRD files, but the chart itself is not additive.

`v0.3.1-alpha.1...v0.4.0-alpha.1` removes `charts/operator/templates/manager/configmap-spiffe-helper.yaml` (−36) and strips the `spiffe-helper` sidecar, the `jwt-svid` volume, and the `--jwt-svid-path` flag from `manager.yaml` (+3/−46). JWT-SVID delivery moves from a sidecar-written file to an in-process `workloadapi.Client`.

No action needed on the pin — the rossoctl charts never passed `--jwt-svid-path`, so nothing here breaks. Worth restating in the description though, since "additive only" steers reviewers away from the one part of this bump that changes runtime behavior.

mrsabath (approved):
Verified the pins and every checkable claim in the description. The release itself is sound.

**What I verified**

- **Pin completeness** — all nine platform tags are at `v0.8.0-alpha.2` at HEAD, no stragglers on `alpha.1`. The only remaining `alpha.1` is the intended `operator-chart` `0.4.0-alpha.1`.
- **`operator v0.4.0-alpha.1` exists** as a real tag in `rossoctl/operator`.
- **The CRD change is genuinely additive** — `pluginPreset`, `plugins`, and `onError` are all `+optional`, and the diff adds no new `required` entries. Backward-compatible, so the minor bump is justified.
- **Sidecar tags** land on cortex `v0.7.0-alpha.10` via the subchart exactly as described (`charts/operator/values.yaml:247-250`).
- **The cortex `CLUSTER_CIDRS` fix** is real — removed at `alpha.10`, with a regression test asserting no 10/8 remnant in the nat ruleset.
- **Both stated known gaps are accurate** — `charts/openshell/values.yaml:16` does carry a floating `tag: main-d7d8306`, and `check-release-pins.sh:18-24` does hardcode only the two chart dirs.

**One correction to the description**

The operator bump is described as "Additive only (94 insertions, 0 deletions)." That holds for the CRD files, but the chart is not additive. Comparing `v0.3.1-alpha.1...v0.4.0-alpha.1`:

- `charts/operator/templates/manager/configmap-spiffe-helper.yaml` is **removed** (−36)
- `manager.yaml` (+3/−46) strips the `spiffe-helper` sidecar, the `jwt-svid` volume, and the `--jwt-svid-path` flag

JWT-SVID delivery moved from a file written by a sidecar to an in-process `workloadapi.Client` in `ClientRegistrationReconciler.fetchJWTSVID`.

This is safe for this PR — the rossoctl charts never passed `--jwt-svid-path`, so the flag removal breaks no wiring, and Deploy & Test passes. The framing is the issue rather than the change: a reviewer trusting "additive only" would skip exactly the sidecar-removal risk that warrants a look.

**Consequent doc drift (follow-up, not a blocker)**

Four docs still describe the operator's spiffe-helper container, including the now-nonexistent `/opt/jwt_svid.token`:

- `docs/security/authentication-modes.md:90`
- `docs/diagrams/02-rossoctl-client-registration.mmd`
- `docs/diagrams/03-rossoctl-client-registration-final.mmd`
- `docs/diagrams/README.md`

These describe the *operator's* sidecar specifically — AuthBridge's bundled spiffe-helper is unrelated and those docs remain correct.

**Before tagging**

E2E HyperShift is still pending a maintainer `/run-e2e`, and the description flags the OCP DNS fix as unvalidated on OpenShift. Since that fix is the substantive reason this cortex range matters — and Kind's service network sits inside the old 10/8 default, which is why the original bug survived CI — passing CI is weak evidence about the thing being fixed. Worth running E2E before the tag lands.

**Areas reviewed**: Helm/K8s, YAML, cross-repo version verification, docs accuracy
**Commits**: 1, signed off, conventional format, imperative, 68-char subject
**CI status**: all passing (E2E HyperShift pending maintainer trigger)

PR #2564, opened 2026-09-16
Docs: lead the RossoCortex page with its value

mrsabath:
## What & why

The [RossoCortex concept page](https://www.rossoctl.dev/docs/dev/concepts/core/cortex) opened with *"RossoCortex is the data plane of Rossoctl. It is a proxy..."* — architecture-first. A reader had to get through the four control points, the plugin chain, and the authentication/authorization split before learning why they would want to run it. For the laptop developer the page is aimed at, that is backwards.

This reorders and rewrites the page to lead with the value, then the install path, then the how-it-works detail.

## What changed

- **New opening — the value.** What you actually get: RossoCortex shows every model call, tool call, and external API request the agent makes, with the token count and cost of each, live on your own machine; and it can reduce token usage by pruning unused tool definitions and compacting tool output. No change to the agent.
- **A privacy assurance up top.** A "Your data stays on your computer" note: collected locally, written to disk, never sent to Rossoctl or any other service, yours to inspect and delete. Initial readers raise this as a concern first; the page now answers it early, in the same wording as `reading-the-numbers.md`.
- **A "What you get" list** — see every call / real token+cost numbers / drill into any request / reduce token usage / no code change — each summarized and linked to the detailed page (laptop quickstart, read the numbers, cost control, context compaction) rather than duplicated here.
- **An "Install it" pointer** to the 5-minute laptop quickstart, right after the value.
- **"How it works"** — the existing architecture content (where it runs, the four control points, the plugin chain, authentication vs authorization, what it does not do) is **unchanged in substance**, moved below and nested as optional depth for the reader who wants it.

## Notes

- **Move, not cut.** The platform/Kubernetes and security material is accurate and still serves the operator reader; it is reordered, not removed. The heading levels shift from `##` to `###` under the new "How it works".
- STE100 throughout. All internal links and heading anchors verified to resolve (laptop, read-the-numbers, cost-control, context-compaction, identity, authbridge, flows, and the `#experimental-features` anchor).
- The frontmatter `description` moves from "The data plane proxy, the four control points and the plugin chain" to a value-led line.

## Acceptance tier

- [x] Tier 0 — Maintenance (docs only)

Assisted-By: Claude Code

PR #2567, opened 2026-09-16
Docs: add a documentation contributor guide, and align the docs skills with it

Ibrahim2595:
## What & why

The repository has no standard for the pages under `docs/`. A contributor who wants to add or change a page has to infer the conventions from the existing pages.

- `CONTRIBUTING.md` covers diagrams only.
- `FEATURE_ACCEPTANCE.md` Pillar 2 requires documentation at every tier, but does not say how to write it.
- The two skills that were supposed to carry the standard disagree with the published pages. `docs:review` refers to `meta:write-docs` as "(planned)", and `meta:write-docs` describes a structure that no published page uses.

Measured against the 44 published pages:

| The `meta:write-docs` skill required | Published pages that do it |
| --- | --- |
| A `#` title in the body | 0 of 44 (the frontmatter `title` provides it) |
| A manual table of contents over 50 lines | 0 of 44 (Docusaurus builds it) |
| `---` rules between sections | 0 of 44 |
| Blockquote callouts, `> **Note:**` | 0 of 44 (31 uses of `:::` callouts: 14 note, 12 warning, 4 info, 1 tip) |

A contributor who followed the skill produced a page that did not match the doc set. An agent that used the skill did the same.

## What changed

**New — `docs/_internal/docs-contributor-guide.md`.** The authoritative standard, and the answer to "how do I contribute a page":

- **Where content lives.** One reader for each of the six sections, and the location for an image, a diagram and a glossary term.
- **What to do if a page or a section is missing.** Four ordered rules: change a page, add a page, add an experiment page with its maturity row, or open an issue. A missing section is an issue, not a `TODO` in a page.
- **The frontmatter contract.** `title`, `description` and `sidebar_position`, and no `#` in the body.
- **How to write.** The ASD-STE100 rules that the pages already keep, with a do-not/do table. The published pages contain 0 contractions, 0 uses of "we", 0 uses of "please" and 0 uses of "simply", so the guide records the practice rather than imposing a new one.
- **A template for each page type.** Task, concept, experiment, reference.
- **Accuracy rules.** A claim must agree with the page it links to; unreleased behaviour is not documented as released (use a `VERIFY` comment); a Beta or Alpha feature carries its maturity and its default.
- **The two checks that Docs CI runs**, and the note that `.lychee.toml` excludes `docs/_internal/` and `.claude/`.

**Rewritten — `.claude/skills/meta:write-docs/SKILL.md`.** It now gives the workflow in five phases and points to the guide, so the rules live in one file and cannot drift again. Phase 2 tells the agent to read every page it will cite before it writes a claim. An "Errors to Avoid" table lists the failures from real reviews. The design-document rules the skill held are correct for `docs/_internal/`, so they are kept and scoped to that tree.

**Aligned — `.claude/skills/docs:review/SKILL.md`.** The structure checklist is split into a product-page list and an internal-document list. The callout convention is corrected to `:::note` and `:::warning`. The "(planned)" references are removed. Four accuracy checks are added, from the review of #2564: each claim agrees with the page it links to; no unreleased behaviour; a Beta or Alpha feature states its maturity and its default; token counts and cost are scoped to a model call.

**`CONTRIBUTING.md`** links the guide from the "Contributing Documentation" section. The diagrams subsection is unchanged.

## Notes

- The guide lives in `docs/_internal/`, next to `dev-guide.md` and `svg-diagram-style-guide.md`, and carries `draft: true`. It is not published to rossoctl.dev.
- The guide is written in ASD-STE100, and it keeps its own rules: 0 contractions, 0 "we", 0 sentences over 26 words.
- `markdownlint-cli2` with the repository configuration: 0 issues. Every relative link and every anchor resolves; the only link-shaped strings that do not are the placeholders inside the template code blocks.
- Nothing under `docs/<section>/` changes. No published page is edited.

## Acceptance tier

- [x] Tier 0 — Maintenance (docs only)

## Review notes

Two decisions worth a maintainer's opinion:

1. **The guide is authoritative and the skills point to it.** The alternative is to keep the rules in the skill. One source stops the drift that this PR corrects, but it means a contributor without an agent and an agent read the same file.
2. **The prose rules are recorded, not invented.** Every rule in "How to write" is a practice the 44 published pages already keep. If a maintainer wants a different rule, the guide is the place to change it.


docs/_internal/docs-contributor-guide.md
```diff
@@ -0,0 +1,306 @@
...
+| This will allow you to reduce tokens. | Two experimental plugins reduce the tokens. |
+| We recommend a sidecar. | Use a sidecar on Kubernetes. |
+
+### Callouts
+
+Use a Docusaurus callout. Do not use a blockquote for a callout.
+
+| Callout | Use |
+| --- | --- |
+| `:::note` | Information that the reader can skip. |
```
mrsabath, line 120: **nit** — the description's evidence table gives "30 uses of `:::note` and `:::warning`". Measured across the 44 sidebar pages on `main`, the counts are:

| Directive | Uses |
|---|---|
| `:::note` | 14 |
| `:::warning` | 12 |
| `:::info` | 4 |
| `:::tip` | 1 |
| **Total** | **31** |

So `:::note` plus `:::warning` is 26, and 31 is the figure for all four directive types — this table's full set. Either number supports the argument; 30 is neither.

Not worth a new commit on its own, and nothing in the guide depends on it. Raising it only because this guide's authority comes from being measured rather than asserted, so a number that does not reproduce is the one kind of error that costs it something. If you touch the branch again, "31 uses of `:::` callouts" would match this table exactly.

mrsabath (approved):
## Summary

This PR argues from counts, so I re-measured every one of them: fetched all 46 `.md` files under `docs/` at `main` and counted with code fences and frontmatter stripped.

| Claim | Measured | Verdict |
|---|---|---|
| `#` title in the body — 0 of 44 | 0 | Confirmed |
| Manual table of contents — 0 of 44 | 0 | Confirmed |
| `---` rules between sections — 0 of 44 | 0 | Confirmed |
| Blockquote callouts — 0 of 44 | 0 | Confirmed |
| `:::note` / `:::warning` — "30 uses" | 26 (31 across all `:::` types) | Off by four — see the inline note |
| 0 contractions, 0 "we", 0 "please", 0 "simply" | 0 / 0 / 0 / 0 | Confirmed |
| Frontmatter contract (`title`, `description`, `sidebar_position`) | all three present on all 44 | Confirmed |
| `.lychee.toml` excludes `docs/_internal/` and `.claude/` | both in `exclude_path` | Confirmed |
| The five "(planned)" references are removed | `grep -c planned` is 0 in both skills | Confirmed |

### On the 44 against 46

Worth recording, because it first looks like an arithmetic error and is not. The repository holds 46 published Markdown files. Exactly two are directory READMEs — `docs/data-governance/README.md` and `docs/diagrams/README.md` — and both do carry a body `#` title. Remove those two and 44 sidebar pages remain, on which every count above holds exactly. The denominator is correct; the description does not say how it was reached, which is the only reason it invites a second look.

### Two things the CI cannot check, checked by hand

**The split checklist is not a contradiction.** `docs:review` line 66 forbids a table of contents and line 75 requires one. Read together with the headings they sit under, the first governs a product page and the second governs a document under `docs/_internal/`. That is the two-sets model the guide defines, applied correctly.

**The guide's own links are outside Link Check.** `.lychee.toml` excludes `docs/_internal/`, so the five relative links the new guide adds are unverified by the pipeline. I resolved each against HEAD: `CONTRIBUTING.md`, `FEATURE_ACCEPTANCE.md`, `PERSONAS_AND_ROLES.md`, `docs/get-started/reading-the-numbers.md` and `docs/_internal/svg-diagram-style-guide.md` all exist. The `../../PERSONAS_AND_ROLES.md` target is the repository-root copy, which does exist — there is also a near-identical copy under `docs/_internal/`, but that is a pre-existing duplication in the repository and not this PR's business.

I also checked the guide against the rules it sets for itself. It carries `draft: true`, it opens with a `#` title as its own internal-document rule requires, and it holds no contraction and no "we" outside the do-not column of the writing table — where they appear precisely because the rule is quoting what to avoid.

### On the two decisions raised for a maintainer

Pointing both skills at one authoritative file is the right call. The drift this PR corrects — a skill that required a structure no published page used — is what two sources of truth produce, and a contributor reading the same file as the agent is a feature rather than a cost.

Recording existing practice instead of inventing rules is what makes the guide defensible. Every prose rule is measurable against the doc set today, so a maintainer who disagrees is changing a documented practice rather than arguing with an assertion.

**Areas reviewed**: Docs, skills (Markdown), CONTRIBUTING, security, conventions
**Commits**: 1, signed off; DCO pass
**CI**: 23/23 green, including Markdown Lint and Link Check
**Security**: no secrets; no published page edited; `draft: true` keeps the guide off rossoctl.dev
**Blocking findings**: none

docs/_internal/docs-contributor-guide.md
Ibrahim2595: Confirmed, and thank you for re-measuring. I reproduced your figures independently on `main` — 44 sidebar pages, `:::note` 14, `:::warning` 12, `:::info` 4, `:::tip` 1. So 26 for note plus warning, 31 for all four. My own earlier measurement produced those same per-type counts, so 30 was a mis-summation on the way into the description, not a different method.

The description is corrected: it now reads "31 uses of `:::` callouts: 14 note, 12 warning, 4 info, 1 tip", which matches your table exactly and no longer needs the reader to work out which set the number covers.

Your reason for raising it is the right one. This guide asks contributors to verify a claim against its source, so a number in its own description that does not reproduce is the one error that costs it standing.

The figure also appears once in the commit message. Nothing in the guide file itself carries a count, so the merged content is unaffected. `dismiss_stale_reviews` is off on `main`, so an amend and force-push would keep your approval — I am happy to do that if you want the history clean, or to leave it and let the squash body be corrected at merge. Your call; I will not touch the branch without a word from you.

Ibrahim2595:
@mrsabath can we merge?

PR #2568, opened 2026-09-16
CI: notify the website when the docs change, so a merge reaches rossoctl.dev in minutes

Ibrahim2595:
## What & why

`www.rossoctl.dev` renders `/docs/dev/*` from **this** repository's `docs/`. The site repository (`rossoctl/.github`) does not commit `docs/` at all — its deploy workflow runs `scripts/sync-docs.sh`, which mirrors `rossoctl/rossoctl:docs/` 1:1 (minus `docs/_internal/`) and regenerates the Contributing page from `CONTRIBUTING.md`.

That deploy workflow already listens for the merge signal:

```yaml
# rossoctl/.github/.github/workflows/deploy.yaml
on:
  repository_dispatch:
    types: [docs-updated]   # "docs changed upstream; sent by
                            #  rossoctl/rossoctl .github/workflows/docs-dispatch.yaml"
```

**That sender was never added to this repository.** So the receiver has no caller, and a merged docs change reaches the public site only on the site's daily 06:17 UTC fallback — up to 24 hours late. #2564 merged today and is still not on the site for that reason.

## What changed

One new file: `.github/workflows/docs-dispatch.yaml`.

- **Trigger:** a push to `main` that touches `docs/**` or `CONTRIBUTING.md` — the two paths `sync-docs.sh` reads. A change to neither does not rebuild the site.
- **Action:** sends the `docs-updated` `repository_dispatch` to `rossoctl/.github`, which starts the existing build-and-deploy. The site is live a few minutes after the merge instead of the next morning.
- **Payload:** the source repo, commit SHA, actor and run URL, so a site build can be traced back to the commit that caused it.
- **`concurrency: docs-dispatch`** — a burst of merges produces one rebuild, not one per commit.
- **`workflow_dispatch`** for a manual re-send.
- No third-party action is used, so there is nothing to pin. `gh api` runs the dispatch.

## Required one-time setup

`GITHUB_TOKEN` is scoped to the repository that issues it, so it cannot dispatch to `rossoctl/.github`. A maintainer needs to add one secret:

1. Create a **fine-grained PAT** (or a GitHub App token) with **Contents: read and write** on `rossoctl/.github` **only**.
2. Store it here:

   ```bash
   gh secret set SITE_DISPATCH_TOKEN --repo rossoctl/rossoctl
   ```

**The workflow degrades safely until then.** With no secret it emits a warning and exits 0 — it never fails a docs merge, and the daily fallback still resyncs the site. Nothing regresses if the secret is never added; the site just stays on its 24-hour cadence.

If maintainers would rather not hold a PAT, the alternative is to tighten the site's own `schedule:` in `deploy.yaml` (for example hourly, gated on the upstream `docs/` tree SHA changing). That needs no secret, but it polls and it is not event-driven. I am happy to send that version instead.

## Verification

I built the site locally from the current `main` of both repositories, with the same steps CI runs (`npm ci && npm run sync-docs && npm run build`):

- `[SUCCESS] Generated static files in "build"`.
- `sync-docs` mirrored **80 files**, and `docs/_internal/` was correctly excluded.
- `build/docs/dev/concepts/core/cortex/index.html` contains the merged #2564 text — "Privacy and data handling", "Two experimental plugins can also reduce" — and none of the disk/deletion wording that #2564 removed.
- The version dropdown renders `dev` and `v0.7 (latest)`.

So the pipeline is sound; only the trigger was missing.

Checks on this PR: `actionlint` clean, and `yamllint` with the config from `security-scans.yaml` reports 0 errors and 0 warnings.

## Acceptance tier

- [x] Tier 0 — Maintenance (CI only)


.github/workflows/docs-dispatch.yaml
```diff
@@ -0,0 +1,101 @@
...
+          echo "Dispatched. The site build runs at:"
+          echo "https://github.com/rossoctl/.github/actions/workflows/deploy.yaml"
+
+      - name: Summary
+        if: always()
+        run: |
+          {
+            echo "### Docs dispatch"
+            echo ""
+            if [ "${{ steps.dispatch.outputs.notified }}" = "true" ]; then
```
mrsabath, line 94: **suggestion** — the summary misreports a *dispatch failure* as "token not set".

`notified` is only ever written as `false` on the missing-token path (line 67). If `gh api` fails for any other reason — PAT expired, wrong scope, `rossoctl/.github` renamed, API 5xx — `set -euo pipefail` aborts the step before `notified=true` is written on line 90, so the output is **empty**, not `false`.

This `if: always()` step then takes its `else` branch and tells the reader:

> `SITE_DISPATCH_TOKEN` is not set, so the site was not notified.

That is actively misleading in exactly the situation where someone is debugging a token that *is* set but broken — it sends them to re-add a secret that already exists. The step itself will be red, so the failure is visible; it is the stated cause that is wrong.

A third state distinguishes the two. Keying off the step conclusion is the smallest change:

```yaml
      - name: Summary
        if: always()
        env:
          NOTIFIED: ${{ steps.dispatch.outputs.notified }}
          OUTCOME: ${{ steps.dispatch.conclusion }}
        run: |
          {
            echo "### Docs dispatch"
            echo ""
            if [ "${NOTIFIED}" = "true" ]; then
              echo "Sent \`docs-updated\` to rossoctl/.github. The site rebuilds"
              echo "\`/docs/dev/*\` from this commit."
            elif [ "${NOTIFIED}" = "false" ]; then
              echo "\`SITE_DISPATCH_TOKEN\` is not set, so the site was not notified."
              echo "It resyncs on its daily 06:17 UTC schedule."
            else
              echo "The dispatch step did not complete (\`${OUTCOME}\`). The token may be"
              echo "expired or wrongly scoped — check the step log above."
              echo "The site still resyncs on its daily 06:17 UTC schedule."
            fi
          } >> "$GITHUB_STEP_SUMMARY"
```

That also picks up the `env:` change from the other nit.

.github/workflows/docs-dispatch.yaml
```diff
@@ -0,0 +1,101 @@
...
+      - "CONTRIBUTING.md"
+  workflow_dispatch: {}
+
+permissions:
+  contents: read
+
+# One dispatch at a time. A burst of merges then produces one rebuild of the
+# site rather than one for each commit.
+concurrency:
+  group: docs-dispatch
```
mrsabath, line 41: **nit** — repo convention here is top-level `permissions: {}` with per-job grants.

`remove-safe-to-test.yaml` and `security-post-merge.yaml` both do this, the former with the comment "Explicit permissions - jobs request only what they need".

This workflow's job needs no `GITHUB_TOKEN` permissions at all — it never checks out the repository and authenticates to the target with `SITE_DISPATCH_TOKEN` instead. So `permissions: {}` is both stricter than `contents: read` and more consistent with its neighbours.

.github/workflows/docs-dispatch.yaml
```diff
@@ -0,0 +1,101 @@
...
+          {
+            echo "### Docs dispatch"
+            echo ""
+            if [ "${{ steps.dispatch.outputs.notified }}" = "true" ]; then
+              echo "Sent \`docs-updated\` to rossoctl/.github. The site rebuilds"
+              echo "\`/docs/dev/*\` from this commit."
+            else
+              echo "\`SITE_DISPATCH_TOKEN\` is not set, so the site was not notified."
+              echo "It resyncs on its daily 06:17 UTC schedule."
+            fi
```
mrsabath, line 100: **nit** — `${{ }}` interpolated directly into a `run:` block rather than passed through `env:`.

Not a live vulnerability: `steps.dispatch.outputs.notified` is workflow-controlled, not attacker-controlled. But this repository pins every action to a SHA and gates E2E behind `safe-to-test` for TOCTOU reasons, so the `env:`-over-interpolation habit matches the prevailing posture — and the dispatch step above already does it correctly for `SHA`, `ACTOR` and `RUN_URL`.

The suggested rewrite in the other comment on this step covers this.

mrsabath (approved):
Well-scoped Tier 0 fix for a real, verified gap. Approving.

## Verification

I checked every central claim against live upstream sources via the API rather than a local clone:

| Claim | Result |
|---|---|
| `rossoctl/.github` `deploy.yaml` listens for `repository_dispatch: types: [docs-updated]` | Confirmed — and its header comment literally names `rossoctl/rossoctl .github/workflows/docs-dispatch.yaml` as the sender |
| That sender does not exist in this repository | Confirmed — 25 workflows, no `docs-dispatch.yaml` |
| Daily fallback is 06:17 UTC | Confirmed — `cron: "17 6 * * *"` |
| `sync-docs.sh` reads exactly `docs/` and `CONTRIBUTING.md` | Confirmed — and it excludes `docs/_internal/` |
| A fine-grained PAT needs **Contents: write** on the target for `POST /repos/{owner}/{repo}/dispatches` | Confirmed against GitHub's fine-grained-PAT permission table — the setup note is correctly and minimally scoped |

So the diagnosis is right, the fix closes the right gap, and the degrade-to-warning path genuinely cannot fail a docs merge. Nice touch that the receiver's own comment already documented this sender — the contract was written down before either side existed, which is exactly why the gap went unnoticed.

## Findings

One suggestion about failure-path reporting and two consistency nits, all inline. Nothing blocking.

**Areas reviewed**: CI/GitHub Actions, shell in embedded `run:` blocks, YAML, security (token scope, secret handling, action pinning), cross-repo contract
**Commits**: 2 (one substantive plus a merge), all signed-off, `ci:` prefix consistent with repo convention
**CI status**: 20/24 passing, 0 failing, 2 pending (E2E awaits a maintainer `/run-e2e`; Deploy & Test), 2 skipping

PR #2570, opened 2026-09-18
fix(backend): read Task artifacts in chat /send response extraction

### LINKED ISSUE #2569: chat /send returns "No response from agent" for agents that answer with a Task artifact
## Problem

`POST /api/v1/chat/{namespace}/{name}/send` returns HTTP **200** with
`{"content": "No response from agent"}` for any agent that answers with a completed A2A
**Task**, even though the agent produced the answer correctly. `GET .../agent-card` and
`POST .../stream` on the same agent both work, so the UI's streaming path hides the bug —
but any client using `/send` (including the UI's non-streaming call and scripted callers)
sees only the placeholder.

Observed on the stock `weather-service` example on OpenShift with v0.8.0-rc.1.

## Root cause

`rossoctl/backend/app/routers/chat.py` extracts content from only two response shapes:

1. `result.status.message.parts[].text`
2. `result.parts[].text` (a bare `Message`)

A completed A2A Task puts its output in **`result.artifacts[].parts[].text`** and commonly
leaves `status.message` unset. A direct in-cluster `message/send` against the weather agent
returns:

```
result keys = [artifacts, contextId, history, id, kind, status]
status.state = completed
status.message = None
artifacts = 1
artifacts[0].parts[0].text = "The weather in Rome is currently: ... 68.5°F ..."
```

Neither branch matches, so `content` stays `""` and the `or "No response from agent"`
fallback fires. The streaming handler in the same module already reads
`result.artifact.parts[]` — only the non-streaming path is missing it.

A related sharp edge: when an agent sends `status.message: null` explicitly, the first
branch's key check (`"message" in result_data["status"]`) is true and the following
`.get("parts", ...)` raises on `None`, surfacing as a 500.

## Proposal

Read `result.artifacts[*].parts[*].text` as an additional shape in the `/send` extraction
block, preferred over `status.message` (a completed Task's answer lives in the artifacts),
and reuse the existing `_extract_text_from_parts` helper so `/send` and `/stream` handle
parts identically — that also gives `/send` the helper's `data`-part handling for free.
Guard against a present-but-null `status.message` while there.

The agent, AuthBridge, and the LLM path are all fine in this scenario; only the backend's
parsing drops the answer, so this is a self-contained backend fix.

webchang:
## What & why

Closes #2569.

`POST /api/v1/chat/{ns}/{name}/send` returned HTTP 200 with `"No response from agent"` for
every agent that answers with a completed A2A **Task** — including the stock
`weather-service` example. The extraction block in `chat.py` handled only
`result.status.message.parts[]` and `result.parts[]`, but a completed Task carries its
output in **`result.artifacts[].parts[]`** and commonly leaves `status.message` unset, so
neither branch matched and the `or "No response from agent"` fallback fired. The streaming
handler in the same module already reads `result.artifact.parts[]`; only the non-streaming
path was missing it.

The fix reads artifacts first (a completed Task's answer lives there), falls back to
`status.message` for agents that answer in-status, and reuses the existing
`_extract_text_from_parts` helper so `/send` and `/stream` handle parts identically — which
also gives `/send` the helper's `data`-part handling. It additionally stops an explicit
`status.message: null` from raising a 500: the old key check (`"message" in status`) was
true for a null value and the following `.get("parts", ...)` blew up on `None`.

## Changed

- `rossoctl/backend/app/routers/chat.py` — artifacts-first extraction in `send_message`,
  via `_extract_text_from_parts`.
- `rossoctl/backend/tests/test_chat_send_artifacts.py` — new: 6 tests covering the
  artifact shape, multi-artifact concatenation, null `status.message`, and regression
  guards for the in-status and bare-`Message` shapes plus the empty-result placeholder.

## Test plan

- `uv run pytest tests/test_chat_send_artifacts.py tests/test_chat_invoke_url.py tests/test_chat_stream_401.py` → 30 passed.
- The 3 new artifact/null-message tests fail on `main` and pass with this change; the 3
  regression tests pass both before and after.
- `make lint` + pre-commit pass (both run as commit hooks here).
- Bug reproduced end-to-end on OpenShift (v0.8.0-rc.1, `team1/weather-service`): `/send`
  → 200 `"No response from agent"` while `/stream` returned the real answer with
  `final:true` and AuthBridge logged `inbound authorized`. A direct in-cluster
  `message/send` confirmed the answer sits in `artifacts[0].parts[0].text` with
  `status.message` unset.
- **Verified live on that same cluster** by mounting this `chat.py` over the released
  `backend:v0.8.0-rc.1` image (ConfigMap `subPath` overlay) and restarting the backend.
  Same request, same agent, before → after:
  - `/send` → 200 `"No response from agent"` **→** 200 with the real answer
    ("The weather in Rome is currently: … 68.3°F, wind 2.3 mph, 39° northeast …").
  - Multi-turn follow-up reusing `session_id` → 200 with the Paris answer, `session_id`
    preserved (contextId handling unchanged).
  - `/stream` (2551 bytes, 1 artifact event, `final:true`) and `/agent-card` (200)
    unaffected; zero `ERROR`/`Traceback` lines in the backend log across the run.
  - The overlay was removed and the deployment returned to the stock image afterwards.

## Acceptance tier

### Pillar 1 — Code quality (all tiers)

- [x] `make lint` and pre-commit pass
- [x] Tests added/updated and passing
- [x] DCO sign-off on all commits (`git commit -s`)
- [ ] (Tier 1+) Behind a feature flag, off by default — N/A, Tier 0 bugfix restoring
      documented A2A behavior; a flag would leave the broken path as the default.

### Pillar 2 — Documentation (all tiers; scope scales)

- [x] Docs affected by this change are updated — none affected; no user-facing API,
      config, or flag change.

### Pillars 3 & 4

Tier 1+ only; not applicable. For reference, the defect was found on OpenShift and the
affected code path is environment-independent (pure response parsing).

Assisted-By: Claude Code


rossoctl/backend/app/routers/chat.py
```diff
@@ -312,19 +312,21 @@ async def send_message(
...
-                            content += part["text"]
-                        elif hasattr(part, "text"):
-                            content += part.text
+                # Handle Task response. A completed Task carries its output in
+                # result.artifacts[].parts[] and commonly leaves status.message
+                # unset, so read artifacts first and fall back to the status
+                # message for agents that answer in-status.
+                artifacts = result_data.get("artifacts") or []
+                status_message = (result_data.get("status") or {}).get("message") or {}
+                if artifacts:
```
esnible, line 321: **suggestion**: this chain branches on artifacts being *present*, not on artifacts *yielding text*. When an artifact exists but `_extract_text_from_parts` returns `""`, the `elif`s are skipped and the response falls through to `"No response from agent"` — a narrow regression against `main`, which would have used the status message or the bare parts.

It is reachable rather than theoretical, because the helper deliberately yields nothing for some real part kinds:

- an image data-part hits `elif not content_type.startswith("image/")` and contributes `""`;
- an A2A `FilePart` (`{"kind": "file", "file": {...}}`) has neither `text` nor `data`, so no branch matches.

So an agent that returns an image artifact plus an in-status text explanation goes from showing that text (on `main`) to showing the placeholder. Same for a textless artifact alongside a bare `parts` reply.

I differential-tested old vs. new across seven shapes; these two are the only ones that regress. Gating the fallback on accumulated content instead of the `elif` fixes both while preserving every shape this PR intends to fix (artifacts, in-status, bare message, null `status.message`, empty result):

```python
artifacts = result_data.get("artifacts") or []
for artifact in artifacts:
    if isinstance(artifact, dict):
        content += _extract_text_from_parts(artifact.get("parts", []))
if not content:
    status_message = (result_data.get("status") or {}).get("message") or {}
    if status_message:
        content += _extract_text_from_parts(status_message.get("parts", []))
    elif "parts" in result_data:
        content += _extract_text_from_parts(result_data["parts"])
```

Worth a seventh test pinning it — e.g. an artifact whose only part is an image data-part, together with a `status.message` carrying the text.

Not blocking: it fails to the same placeholder rather than to a wrong answer or a 500, and it is narrower than the bug being fixed.

esnible (approved):
Correct diagnosis, and both halves of the fix verified rather than taken on faith. I differential-tested the old and new extraction logic across seven A2A response shapes:

| Shape | `main` | this PR |
|---|---|---|
| Completed Task w/ text artifacts | `""` -> placeholder | **fixed** |
| In-status message | works | works |
| Bare `Message` | works | works |
| `status.message: null` | **500 (AttributeError)** | **fixed** |
| `status` present, no `message`, + bare parts | works | works |
| Artifact present but textless + in-status text | `"STATUS-FALLBACK"` | `""` -> placeholder |
| Artifact present but textless + bare parts | `"BARE"` | `""` -> placeholder |

The primary fix holds, and so does the null-message claim — I reproduced the `AttributeError` against `main`'s logic directly, confirming `"message" in status` is indeed true for an explicit null.

The last two rows are the one finding, left inline: the `elif` chain conditions on artifact presence rather than extracted content. Non-blocking (it degrades to the same placeholder, not to a wrong answer), and a one-line change fixes it.

For consistency, the streaming handler at `chat.py:503` already uses a truthiness guard (`if "message" in status and status["message"]`), which is why `/stream` never had the null crash — this brings `/send` in line with it.

Other checks:

- **Helper reuse is a strict superset** for dict parts, adding `kind=text` and `data` handling. The dropped `elif hasattr(part, "text")` branch was dead code, since `result_data` comes from `response.json()` and only yields dicts.
- **Tests** — 6 new, assertive, no skips or `xfail`; they pin the artifact shape, multi-artifact concatenation, null `status.message`, and three regression guards. The only gap is the textless-artifact case above.
- **Security** — no secrets, no new dependencies, no auth/authz change; Bandit and CodeQL green.
- **Feature flag** — CLAUDE.md requires flags for new features, but this is a Tier 0 bugfix restoring documented A2A behavior; the author's reasoning that a flag would leave the broken path as the default is right.

The live-cluster before/after on `v0.8.0-rc.1` — with the overlay removed and the deployment returned to the stock image afterwards — is unusually thorough validation for a parsing bugfix.

**Areas reviewed**: Python, Tests, Security, Docs
**Agent/IDE config (.claude/.vscode)**: none — supply-chain gate clean
**CI status**: passing (22 checks; `E2E HyperShift` pending a maintainer `/run-e2e`, `Scorecard` skipped)

pdettori (approved):
### Summary
Clean, well-scoped bugfix. `send_message` now reads a completed A2A Task's answer from `result.artifacts[].parts[]` first, falling back to `status.message.parts[]`, reusing the existing `_extract_text_from_parts` helper for parity with the streaming path (and picking up its `data`-part handling as a disclosed side effect). It also fixes the latent crash where an explicit `status.message: null` used to pass the `"message" in status` check and then blow up on `.get("parts", ...)` against `None`.

Verified the control flow against `_stream_from_response` and the new/old extraction branches by hand — the if/elif/elif chain (artifacts → status.message → bare parts) is correct, including the empty-list/falsy-dict edge cases the null-message fix depends on. The 6 new tests cover artifact extraction, multi-artifact concatenation, the null-message regression, and both pre-existing shapes (in-status message, bare Message) plus the empty-result placeholder — good coverage of the branches touched.

**Author**: webchang (MEMBER — maintainer)
**Areas reviewed**: Python (backend/app/routers/chat.py, new test file)
**CI status**: all checks passing (E2E HyperShift pending, gated behind maintainer `/run-e2e` command — not required for merge)

No must-fix, suggestion, or nit findings.

